PR #1865, opened 2020-04-08
Add ExternalSource to "C API"

klecki:
#### Why we need this PR?
We want access for external source through simplified C[-like] API. 

 - What solution was applied:
     Added new API calls for setting external source, corresponding to setting a TensorList or vector of Tensors.
    C API
     The relevant changes
 - Validation and testing:
     Added a GTest test, that compares the C API created from serialized pipeline with the same Pipeline. Checks pipeline with regular file reader and two variants of setting external source.
     Docstrings for C API. 

**JIRA TASK**: *[DALI-1368]*

### DIFF
--- a/dali/c_api/c_api.cc
+++ b/dali/c_api/c_api.cc
@@ -74,12 +74,12 @@ void daliPrefetchSeparate(daliPipelineHandle* pipe_handle,
 }
 
 void daliSetExternalInput(daliPipelineHandle* pipe_handle, const char* name, device_type_t device,
-                          const void* data_ptr, dali_data_type_t data_type, const int64_t* shape,
+                          const void* data_ptr, dali_data_type_t data_type, const int64_t* shapes,
                           int sample_dim, const char* layout_str) {
   DALI_ENFORCE(device == CPU, "GPU data cannot be passed as external source");
   dali::Pipeline* pipeline = reinterpret_cast<dali::Pipeline*>(pipe_handle->pipe);
-  std::vector<int64_t> shapes_tmp(shape, shape + sample_dim * pipeline->batch_size());
-  dali::TensorListShape<> shapes(std::move(shapes_tmp), pipeline->batch_size(), sample_dim);
+  std::vector<int64_t> shapes_tmp(shapes, shapes + sample_dim * pipeline->batch_size());
+  dali::TensorListShape<> tl_shape(std::move(shapes_tmp), pipeline->batch_size(), sample_dim);
   dali::TensorLayout layout{};
   if (layout_str != nullptr) {
     layout = dali::TensorLayout(layout_str);
@@ -90,8 +90,8 @@ void daliSetExternalInput(daliPipelineHandle* pipe_handle, const char* name, dev
   // We cast away the const from data_ptr, as there is no other way of passing it to the
   // TensorList, as we must also set the shape and type metadata.
   // It is passed further as const TensorList, so it's data cannot be modified.
-  data.ShareData(const_cast<void*>(data_ptr), shapes.num_elements() * elem_sizeof);
-  data.Resize(shapes, type_info);
+  data.ShareData(const_cast<void*>(data_ptr), tl_shape.num_elements() * elem_sizeof);
+  data.Resize(tl_shape, type_info);
   data.SetLayout(layout);
   const auto& cdata = data;
   pipeline->SetExternalInput(name, cdata);
@@ -99,12 +99,12 @@ void daliSetExternalInput(daliPipelineHandle* pipe_handle, const char* name, dev
 
 void daliSetExternalInputTensors(daliPipelineHandle* pipe_handle, const char* name,
                                  device_type_t device, const void* const* data_ptr,
-                                 dali_data_type_t data_type, const int64_t* shape,
+                                 dali_data_type_t data_type, const int64_t* shapes,
                                  int64_t sample_dim, const char* layout_str) {
   DALI_ENFORCE(device == CPU, "GPU data cannot be passed as external source");
   dali::Pipeline* pipeline = reinterpret_cast<dali::Pipeline*>(pipe_handle->pipe);
-  std::vector<int64_t> shapes_tmp(shape, shape + sample_dim * pipeline->batch_size());
-  dali::TensorListShape<> shapes(std::move(shapes_tmp), pipeline->batch_size(), sample_dim);
+  std::vector<int64_t> shapes_tmp(shapes, shapes + sample_dim * pipeline->batch_size());
+  dali::TensorListShape<> tl_shape(std::move(shapes_tmp), pipeline->batch_size(), sample_dim);
   dali::TensorLayout layout{};
   if (layout_str != nullptr) {
     layout = dali::TensorLayout(layout_str);
@@ -116,8 +116,8 @@ void daliSetExternalInputTensors(daliPipelineHandle* pipe_handle, const char* na
     // We cast away the const from data_ptr, as there is no other way of passing it to the
     // Tensor as we must also set the shape and type metadata.
     // The vector that we pass to pipeline is const.
-    data[i].ShareData(const_cast<void*>(data_ptr[i]), shapes[i].num_elements() * elem_sizeof);
-    data[i].Resize(shapes[i], type_info);
+    data[i].ShareData(const_cast<void*>(data_ptr[i]), tl_shape[i].num_elements() * elem_sizeof);
+    data[i].Resize(tl_shape[i], type_info);
     data[i].SetLayout(layout);
   }
   const auto& cdata = data;
--- a/dali/c_api/c_api.cc
+++ b/dali/c_api/c_api.cc
@@ -80,7 +80,6 @@ void daliSetExternalInput(daliPipelineHandle* pipe_handle, const char* name, dev
   dali::Pipeline* pipeline = reinterpret_cast<dali::Pipeline*>(pipe_handle->pipe);
   std::vector<int64_t> shapes_tmp(shape, shape + sample_dim * pipeline->batch_size());
   dali::TensorListShape<> shapes(std::move(shapes_tmp), pipeline->batch_size(), sample_dim);
-  std::string source_name(name);
   dali::TensorLayout layout{};
   if (layout_str != nullptr) {
     layout = dali::TensorLayout(layout_str);
@@ -95,7 +94,7 @@ void daliSetExternalInput(daliPipelineHandle* pipe_handle, const char* name, dev
   data.Resize(shapes, type_info);
   data.SetLayout(layout);
   const auto& cdata = data;
-  pipeline->SetExternalInput(source_name, cdata);
+  pipeline->SetExternalInput(name, cdata);
 }
 
 void daliSetExternalInputTensors(daliPipelineHandle* pipe_handle, const char* name,
@@ -106,7 +105,6 @@ void daliSetExternalInputTensors(daliPipelineHandle* pipe_handle, const char* na
   dali::Pipeline* pipeline = reinterpret_cast<dali::Pipeline*>(pipe_handle->pipe);
   std::vector<int64_t> shapes_tmp(shape, shape + sample_dim * pipeline->batch_size());
   dali::TensorListShape<> shapes(std::move(shapes_tmp), pipeline->batch_size(), sample_dim);
-  std::string source_name(name);
   dali::TensorLayout layout{};
   if (layout_str != nullptr) {
     layout = dali::TensorLayout(layout_str);
@@ -123,7 +121,7 @@ void daliSetExternalInputTensors(daliPipelineHandle* pipe_handle, const char* na
     data[i].SetLayout(layout);
   }
   const auto& cdata = data;
-  pipeline->SetExternalInput(source_name, cdata);
+  pipeline->SetExternalInput(name, cdata);
 }
 
 void daliRun(daliPipelineHandle* pipe_handle) {
--- a/dali/c_api/c_api.cc
+++ b/dali/c_api/c_api.cc
@@ -73,12 +73,10 @@ void daliPrefetchSeparate(daliPipelineHandle* pipe_handle,
   }
 }
 
-dali_result_t daliSetExternalInput(daliPipelineHandle* pipe_handle, const char* name,
-    device_type_t device, const void* data_ptr, dali_data_type_t data_type, int64_t* shape,
-    int sample_dim, const char* layout_str) {
-  if (device != CPU) {
-    return dali_result_t::DALI_FAILURE;  // "GPU data cannot be passed as external source"
-  }
+void daliSetExternalInput(daliPipelineHandle* pipe_handle, const char* name, device_type_t device,
+                          const void* data_ptr, dali_data_type_t data_type, const int64_t* shape,
+                          int sample_dim, const char* layout_str) {
+  DALI_ENFORCE(device == CPU, "GPU data cannot be passed as external source");
   dali::Pipeline* pipeline = reinterpret_cast<dali::Pipeline*>(pipe_handle->pipe);
   std::vector<int64_t> shapes_tmp(shape, shape + sample_dim * pipeline->batch_size());
   dali::TensorListShape<> shapes(std::move(shapes_tmp), pipeline->batch_size(), sample_dim);
@@ -88,7 +86,7 @@ dali_result_t daliSetExternalInput(daliPipelineHandle* pipe_handle, const char* 
     layout = dali::TensorLayout(layout_str);
   }
   dali::TensorList<dali::CPUBackend> data;
-  const auto &type_info = dali::TypeTable::GetTypeInfo(static_cast<dali::DALIDataType>(data_type));
+  const auto& type_info = dali::TypeTable::GetTypeInfo(static_cast<dali::DALIDataType>(data_type));
   auto elem_sizeof = type_info.size();
   // We cast away the const from data_ptr, as there is no other way of passing it to the
   // TensorList, as we must also set the shape and type metadata.
@@ -96,17 +94,15 @@ dali_result_t daliSetExternalInput(daliPipelineHandle* pipe_handle, const char* 
   data.ShareData(const_cast<void*>(data_ptr), shapes.num_elements() * elem_sizeof);
   data.Resize(shapes, type_info);
   data.SetLayout(layout);
-  const auto &cdata = data;
+  const auto& cdata = data;
   pipeline->SetExternalInput(source_name, cdata);
-  return dali_result_t::DALI_SUCCESS;
 }
 
-dali_result_t daliSetExternalInputTensors(daliPipelineHandle* pipe_handle, const char* name,
-    device_type_t device, const void* const* data_ptr, dali_data_type_t data_type, int64_t* shape,
-    int64_t sample_dim, const char* layout_str) {
-  if (device != CPU) {
-    return dali_result_t::DALI_FAILURE;  // "GPU data cannot be passed as external source"
-  }
+void daliSetExternalInputTensors(daliPipelineHandle* pipe_handle, const char* name,
+                                 device_type_t device, const void* const* data_ptr,
+                                 dali_data_type_t data_type, const int64_t* shape,
+                                 int64_t sample_dim, const char* layout_str) {
+  DALI_ENFORCE(device == CPU, "GPU data cannot be passed as external source");
   dali::Pipeline* pipeline = reinterpret_cast<dali::Pipeline*>(pipe_handle->pipe);
   std::vector<int64_t> shapes_tmp(shape, shape + sample_dim * pipeline->batch_size());
   dali::TensorListShape<> shapes(std::move(shapes_tmp), pipeline->batch_size(), sample_dim);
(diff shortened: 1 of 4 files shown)


klecki:
!build

dali/c_api/c_api.cc
```diff
@@ -70,6 +73,64 @@ void daliPrefetchSeparate(daliPipelineHandle* pipe_handle,
   }
 }
 
+dali_result_t daliSetExternalInput(daliPipelineHandle* pipe_handle, const char* name,
+    device_type_t device, const void* data_ptr, dali_data_type_t data_type, int64_t* shape,
```
mzient, line 77:
```suggestion
    device_type_t device, const void* data_ptr, dali_data_type_t data_type, const int64_t* shape,
```

dali/c_api/c_api.cc
```diff
@@ -70,6 +73,64 @@ void daliPrefetchSeparate(daliPipelineHandle* pipe_handle,
...
+  const auto &type_info = dali::TypeTable::GetTypeInfo(static_cast<dali::DALIDataType>(data_type));
+  auto elem_sizeof = type_info.size();
+  // We cast away the const from data_ptr, as there is no other way of passing it to the
+  // TensorList, as we must also set the shape and type metadata.
+  // It is passed further as const TensorList, so it's data cannot be modified.
+  data.ShareData(const_cast<void*>(data_ptr), shapes.num_elements() * elem_sizeof);
+  data.Resize(shapes, type_info);
+  data.SetLayout(layout);
+  const auto &cdata = data;
+  pipeline->SetExternalInput(source_name, cdata);
```
mzient, line 100: can't it throw?

dali/c_api/c_api.cc
```diff
@@ -70,6 +73,64 @@ void daliPrefetchSeparate(daliPipelineHandle* pipe_handle,
...
+  for (int i = 0; i < pipeline->batch_size(); i++) {
+    // We cast away the const from data_ptr, as there is no other way of passing it to the
+    // Tensor as we must also set the shape and type metadata.
+    // The vector that we pass to pipeline is const.
+    data[i].ShareData(const_cast<void*>(data_ptr[i]), shapes[i].num_elements() * elem_sizeof);
+    data[i].Resize(shapes[i], type_info);
+    data[i].SetLayout(layout);
+  }
+  const auto &cdata = data;
+  pipeline->SetExternalInput(source_name, cdata);
```
mzient, line 130: likewise - can't it throw?

dali/c_api/c_api.h
```diff
@@ -69,6 +80,42 @@ extern "C" {
       int cpu_prefetch_queue_depth,
       int gpu_prefetch_queue_depth);
 
+  /**
+   * @brief Feed the data to ExternalSource as contiguous allocation.
+   *
+   * @param pipe_handle Pointer to pipeline handle
+   * @param name Pointer to null terminated string with External Source name to be fed
```
mzient, line 87:
```suggestion
   * @param name Pointer to a null terminated string with the name of the External Source to be fed
```

dali/c_api/c_api.h
```diff
@@ -23,17 +23,23 @@
...
 extern "C" {
 #endif
-  struct daliPipelineHandle {
+  /**
+   * @brief Handle for DALI C-like API.
+   *
+   * @note Beware, the C API is just C-like API for handling some mangling issues and
+   * it can throw exceptions.
+   */
+  typedef struct daliPipelineHandle {
```
mzient, line 32:
```suggestion
  typedef struct {
```
I don't think anyone's been using it as `struct daliPipelineHandle` - and this repetition is just asking for some compiler to start to complain.

dali/c_api/c_api.h
```diff
@@ -23,17 +23,23 @@
...
     void* pipe;
     void* ws;
-  };
+  } daliPipelineHandle;
 
-  enum device_type_t {
+  typedef enum device_type_t {
```
mzient, line 37:
```suggestion
  typedef enum {
```

dali/c_api/c_api.h
```diff
@@ -23,17 +23,23 @@
...
     CPU = 0,
     GPU = 1
-  };
+  } device_type_t;
 
-  typedef enum {
+  typedef enum dali_data_type_t {
```
mzient, line 42:
```suggestion
  typedef enum {
```

dali/c_api/c_api.h
```diff
@@ -49,6 +55,11 @@ extern "C" {
     DALI_BOOL            = 11
   } dali_data_type_t;
 
+  typedef enum dali_result_t {
```
mzient, line 58:
```suggestion
  typedef enum {
```

dali/c_api/c_api_test.cc
```diff
@@ -0,0 +1,245 @@
...
+    pipe_ptr->RunGPU();
+  }
+  daliPrefetchUniform(&handle, prefetch_queue_depth);
+
+  dali::DeviceWorkspace ws;
+  for (int i = 0; i < prefetch_queue_depth; i++) {
+    ComparePipelinesOutputs(handle, *pipe_ptr);
+  }
+
+  SequentialFill(view<uint8_t>(input), 42 * prefetch_queue_depth);
```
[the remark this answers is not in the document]
mzient: To see that setting the input actually works _each time_?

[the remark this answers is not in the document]
klecki: Unless you're asking about ExternalSourceMultipleAllocPipe and ExternalSourceSingleAllocPipe it checks passing the external source data in two different ways. I would keep it separate for that reason.

dali/c_api/c_api.h
klecki: There's a slight difference with typedefing an anonymous and named struct (and some people discourage the practice totally), but I don't think we rely on those properties anywhere.

dali/c_api/c_api.cc
klecki: Yes, like almost every call to pipeline in this file.

dali/c_api/c_api.h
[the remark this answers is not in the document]
klecki: Yes.

dali/c_api/c_api.cc
klecki: Done

dali/c_api/c_api.h
klecki: Done

dali/c_api/c_api.h
klecki: Done

dali/c_api/c_api.h
klecki: Done

dali/c_api/c_api.h
klecki: Done

dali/c_api/c_api.h
```diff
@@ -69,6 +80,42 @@ extern "C" {
       int cpu_prefetch_queue_depth,
       int gpu_prefetch_queue_depth);
 
+  /**
+   * @brief Feed the data to ExternalSource as contiguous allocation.
```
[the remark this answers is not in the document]
klecki: Done

dali/c_api/c_api.h
klecki: Done

dali/c_api/c_api.h
```diff
@@ -69,6 +80,42 @@ extern "C" {
...
+   * @param sample_dim The dimensionality of the data.
+   * @param layout_str Optional layout provided as a pointer null terminated string.
+   *                   Can be set to NULL.
+   */
+  DLL_PUBLIC dali_result_t daliSetExternalInput(daliPipelineHandle *pipe_handle, const char *name,
+     device_type_t device, const void *data_ptr, dali_data_type_t data_type, int64_t *shape,
+     int sample_dim, const char *layout_str);
+
+  /**
+   * @brief Feed the data to ExternalSource as set of separate buffers.
```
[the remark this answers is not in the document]
klecki: Done

klecki:
!build

dali/c_api/c_api.cc
```diff
@@ -70,6 +73,59 @@ void daliPrefetchSeparate(daliPipelineHandle* pipe_handle,
...
+  const auto& type_info = dali::TypeTable::GetTypeInfo(static_cast<dali::DALIDataType>(data_type));
+  auto elem_sizeof = type_info.size();
+  // We cast away the const from data_ptr, as there is no other way of passing it to the
+  // TensorList, as we must also set the shape and type metadata.
+  // It is passed further as const TensorList, so it's data cannot be modified.
+  data.ShareData(const_cast<void*>(data_ptr), shapes.num_elements() * elem_sizeof);
+  data.Resize(shapes, type_info);
+  data.SetLayout(layout);
+  const auto& cdata = data;
+  pipeline->SetExternalInput(source_name, cdata);
```
mzient, line 98: Nitpick: why not just pass `name`?

dali/c_api/c_api.cc
```diff
@@ -70,6 +73,59 @@ void daliPrefetchSeparate(daliPipelineHandle* pipe_handle,
...
+  for (int i = 0; i < pipeline->batch_size(); i++) {
+    // We cast away the const from data_ptr, as there is no other way of passing it to the
+    // Tensor as we must also set the shape and type metadata.
+    // The vector that we pass to pipeline is const.
+    data[i].ShareData(const_cast<void*>(data_ptr[i]), shapes[i].num_elements() * elem_sizeof);
+    data[i].Resize(shapes[i], type_info);
+    data[i].SetLayout(layout);
+  }
+  const auto& cdata = data;
+  pipeline->SetExternalInput(source_name, cdata);
```
mzient, line 126: likewise - why have the extra `source_name` variable?

dali/c_api/c_api_test.cc
```diff
@@ -0,0 +1,245 @@
...
+  pipe_ptr->SetExternalInput(input_name, input);
+  std::string layout = "HWC";
```
mzient, line 233: I know it's just a test, but... `std::string` when all you need is `c_str()`?

dali/c_api/c_api.h
```diff
@@ -69,6 +75,44 @@ extern "C" {
...
+   * @param pipe_handle Pointer to pipeline handle
+   * @param name Pointer to a null terminated string with the name of the External Source to be fed
+   * @param device Device of the supplied memory. Only CPU is supported.
+   * @param data_ptr Pointer to contiguous buffer containing all samples
+   * @param data_type Type of the provided data
+   * @param shape Pointer to an array containing shape of all samples concatenated one after
+   *              another. Should contain batch_size * sample_dim elements.
+   * @param sample_dim The dimensionality of the data.
+   * @param layout_str Optional layout provided as a pointer null terminated string.
+   *                   Can be set to NULL.
```
szalpal, line 90: Maybe setting it to `NULL` as default argument might be a good idea?

dali/c_api/c_api.h
```diff
@@ -69,6 +75,44 @@ extern "C" {
...
+   *
+   * @param pipe_handle Pointer to pipeline handle
+   * @param name Pointer to a null terminated string with the name of the External Source to be fed
+   * @param device Device of the supplied memory. Only CPU is supported.
+   * @param data_ptr Pointer to an array containing batch_size pointers to separate Tensors.
+   * @param data_type Type of the provided data
+   * @param shape Pointer to an array containing shape of all samples concatenated one after
+   *              another. Should contain batch_size * sample_dim elements.
+   * @param sample_dim The dimensionality of the data.
+   * @param layout_str Optional layout provided as a pointer null terminated string.
```
szalpal, line 108:
```suggestion
   * @param layout_str Optional layout provided as a null terminated string.
```

dali/c_api/c_api.h
```diff
@@ -69,6 +75,44 @@ extern "C" {
...
+   * @brief Feed the data to ExternalSource as a set of separate buffers.
+   *
+   * @param pipe_handle Pointer to pipeline handle
+   * @param name Pointer to a null terminated string with the name of the External Source to be fed
+   * @param device Device of the supplied memory. Only CPU is supported.
+   * @param data_ptr Pointer to an array containing batch_size pointers to separate Tensors.
+   * @param data_type Type of the provided data
+   * @param shape Pointer to an array containing shape of all samples concatenated one after
+   *              another. Should contain batch_size * sample_dim elements.
+   * @param sample_dim The dimensionality of the data.
```
szalpal, line 107:
```suggestion
   * @param sample_dim The dimensionality of a single sample.
```
It's to be 100% clear, whether this should include number of samples or not

dali/c_api/c_api.h
```diff
@@ -69,6 +75,44 @@ extern "C" {
       int cpu_prefetch_queue_depth,
       int gpu_prefetch_queue_depth);
 
+  /**
+   * @brief Feed the data to ExternalSource as contiguous memory.
```
szalpal, line 79: I'd propose to put here a remark about performence differences between these two functions, or no perf diffs, if there are none. E.g. it might be possible that user can choose, if he want to put the samples into contiguous mem himself, if he can do it efficiently

dali/c_api/c_api.h
```diff
@@ -69,6 +75,44 @@ extern "C" {
...
+   * @param shape Pointer to an array containing shape of all samples concatenated one after
+   *              another. Should contain batch_size * sample_dim elements.
+   * @param sample_dim The dimensionality of the data.
+   * @param layout_str Optional layout provided as a pointer null terminated string.
+   *                   Can be set to NULL.
+   */
+  DLL_PUBLIC void daliSetExternalInputTensors(daliPipelineHandle* pipe_handle, const char* name,
+                                              device_type_t device, const void* const* data_ptr,
+                                              dali_data_type_t data_type, const int64_t* shape,
+                                              int64_t sample_dim, const char* layout_str);
```
szalpal, line 114: I'm not an expert in C APIs, but I think that together with const char* you should require length of that string, in case there are nulls inside. Or it's common to require only one null always at the end?

dali/c_api/c_api.h
klecki: If we want to stick with this being C API (at least pretend that it is such), there is no such thing as default value in C.

dali/c_api/c_api.h
klecki: What about?
```suggestion
   * @param layout_str Optional layout provided as a pointer to null terminated string.
```

dali/c_api/c_api.h
klecki: We can go the way of `snprintf`, but I'm not sure if we really want to pile more length related arguments here. NULL terminated strings are quite common.

dali/c_api/c_api.cc
klecki: Does it make any difference? It will still create a temporary.

dali/c_api/c_api.h
klecki: I think most we should do is point which C++ Pipeline API call are we using underneath. This API is just thin wrapper over that.

dali/c_api/c_api.h
szalpal: If we want to be precise here, looking at most probable usage: `daliSetExternalInputTensors(handl, "input", ...)`, it's actually a string, not really a pointer ;) But yeah, pick whichever suits you better

dali/c_api/c_api.h
szalpal: Fair enough ;)

dali/c_api/c_api.h
szalpal: Whatever we do here, I'm just pointing out, that from user's perspective this is quite important information, what's the actual difference (or if there's no difference) apart from convinency

dali/c_api/c_api.h
klecki: I'll stick to what is done at cppreference: https://en.cppreference.com/w/cpp/string/byte/strcmp
```pointers to the null-terminated byte strings to compare```

File: dali/c_api/c_api.h (hunk 17 above)
Comment (klecki) [reply dali/c_api/c_api.h:107]: Done

File: dali/c_api/c_api.h (hunk 18 above)
Comment (klecki) [reply dali/c_api/c_api.h:79]: TBH, the answer to that is a little bit complicated and it would also change additionally if we allow the GPU inputs. The Pipeline doc doesn't provide such information.

File: dali/c_api/c_api_test.cc [hunk 20]
```diff
@@ -0,0 +1,245 @@
...
+  constexpr int prefetch_queue_depth = 2;
+  constexpr bool async = true;
+  constexpr float output_size = 20.f;
+  std::string input_name = "inputs"s;
+
+  std::unique_ptr<Pipeline> GetTestPipeline(bool is_file_reader, const std::string &output_device) {
+    auto pipe_ptr = std::make_unique<Pipeline>(batch_size, num_thread, device_id, seed, pipelined,
+                                               prefetch_queue_depth, async);
+    auto &pipe = *pipe_ptr;
+      TensorList<CPUBackend> data;
```
jantonguirao, line 47: nitpick: wrong indentation

klecki: Done

dali/c_api/c_api.h
```diff
@@ -69,6 +75,46 @@ extern "C" {
...
+   * @param shape Pointer to an array containing shape of all samples concatenated one after
+   *              another. Should contain batch_size * sample_dim elements.
+   * @param sample_dim The dimensionality of a single sample.
+   * @param layout_str Optional layout provided as a pointer to null-terminated byte string.
+   *                   Can be set to NULL.
+   */
+  DLL_PUBLIC void daliSetExternalInput(daliPipelineHandle* pipe_handle, const char* name,
+                                       device_type_t device, const void* data_ptr,
+                                       dali_data_type_t data_type, const int64_t* shape,
+                                       int sample_dim, const char* layout_str);
```
szalpal, line 96: What I don't like about this API here, is that it doesn't forsee strided memory support. Let's say (with GPU external input) user has input data in strided memory. This way he'd need to make one additional copy. And since we actually do a copy there, it might be useful.

I know, strided mem is not supported there yet, but it doesn't look like big  deal, to modify the entry-point copy in Dali

mzient: It sounds like a great idea for a follow-up task.

dali/c_api/c_api.h
```diff
@@ -69,6 +75,46 @@ extern "C" {
...
+   * @param data_type Type of the provided data
+   * @param shape Pointer to an array containing shape of all samples concatenated one after
+   *              another. Should contain batch_size * sample_dim elements.
+   * @param sample_dim The dimensionality of a single sample.
+   * @param layout_str Optional layout provided as a pointer to null-terminated byte string.
+   *                   Can be set to NULL.
+   */
+  DLL_PUBLIC void daliSetExternalInput(daliPipelineHandle* pipe_handle, const char* name,
+                                       device_type_t device, const void* data_ptr,
+                                       dali_data_type_t data_type, const int64_t* shape,
```
mzient, line 95:
```suggestion
                                       dali_data_type_t data_type, const int64_t* shapes,
```
rationale: there's a chance the user might miss the docs and just put there what he thinks the name says - and those used to dealing with batches as tensors, would be suriprised to see wrong results or even segfaults.

dali/c_api/c_api.h
```diff
@@ -69,6 +75,46 @@ extern "C" {
...
+   * @param data_type Type of the provided data
+   * @param shape Pointer to an array containing shape of all samples concatenated one after
+   *              another. Should contain batch_size * sample_dim elements.
+   * @param sample_dim The dimensionality of a single sample.
+   * @param layout_str Optional layout provided as a pointer to null-terminated byte string.
+   *                   Can be set to NULL.
+   */
+  DLL_PUBLIC void daliSetExternalInputTensors(daliPipelineHandle* pipe_handle, const char* name,
+                                              device_type_t device, const void* const* data_ptr,
+                                              dali_data_type_t data_type, const int64_t* shape,
```
mzient, line 115:
```suggestion
                                              dali_data_type_t data_type, const int64_t* shapes,
```

dali/c_api/c_api.h
JanuszL: The choice is rather not about the performance but how your data looks like - if it is contiguous or not. I don't think anyone would convert continuous<-> noncontinuous just to use one of these functions.

dali/c_api/c_api.h
klecki: The support for strides is out of scope for this task (and this PR), we would need to adjust the copying in ExternalSource operator probably to make it work.

klecki:
!build

dali/c_api/c_api.h
mzient: I think so far we're the only user of this API.
We know it's broken.
The function returns void and should return an error code.
There are more issues - this adds up to one thing: this signature is not stable anyway, so we can add the paramters in the next iteration, when the signature changes anyway.

dali/c_api/c_api.h
JanuszL: We would need to add this for C++ API as the Numpy array we accept can have strides as well. So it is definitely bigger than this task PR.

PR #1870, opened 2020-04-10
Add "no exec check" to SmallVector

mzient:
Add "no exec check" to SmallVector to prevent warnings in host-only functions.

#### Why we need this PR?
*Pick one, remove the rest*
- It prevents NVCC from spitting out nonsensical warnings in host-only code.

 - What solution was applied:
     * Add DALI_NO_EXEC_CHECK to functions that call potentially host-only code, but are marked as host/device
     * SmallVector - it's just adding pragmas; the code generatio is not affected, so the output should not change
     * N/A
 - Validation and testing:
     * SmallVector tests suffice
     * N/A

**JIRA TASK**: N/A


mzient:
!build

PR #1882, opened 2020-04-18
Reduce GPU - kernel frontend

mzient:
#### Why we need this PR?
*Pick one, remove the rest*
- It adds new feature needed because of we want fast, selectable axis reductions.

 - What solution was applied:
     * CRTP base class
     * Stage setup and switch by stage kind
     * kernels/reduce module
     * design / distribution of content across files
 - Validation and testing:
     * Unit tests
     * Doxygen, comments

**JIRA TASK**: DALI-1250

### DIFF
--- a/dali/kernels/reduce/reduce_gpu.h
+++ b/dali/kernels/reduce/reduce_gpu.h
@@ -1,0 +1,41 @@
+// Copyright (c) 2020, NVIDIA CORPORATION. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef DALI_KERNELS_REDUCE_REDUCE_GPU_H_
+#define DALI_KERNELS_REDUCE_REDUCE_GPU_H_
+
+#include <memory>
+#include "dali/kernels/kernel.h"
+#include "dali/core/tensor_view.h"
+
+namespace dali {
+namespace kernels {
+
+template <typename Out, typename In>
+class SumGPU {
+ public:
+  KernelRequirements Setup(KernelContext &ctx,
+                           const TensorListShape<> &in_shape,
+                           span<const int> axes);
+
+  void Run(KernelContext &ctx, const OutListGPU<Out> &out, const InListGPU<In> &in);
+ private:
+  class Impl;
+  std::unique_ptr<Impl> impl_;
+};
+
+}  // namespace kernels
+}  // namespace dali
+
+#endif  // DALI_KERNELS_REDUCE_REDUCE_GPU_H_
--- a/dali/kernels/reduce/reduce_gpu.h
+++ b/dali/kernels/reduce/reduce_gpu.h
@@ -23,17 +23,52 @@ namespace dali {
 namespace kernels {
 
 template <typename Out, typename In>
-class SumGPU {
+class DLL_PUBLIC SumGPU {
  public:
+  SumGPU();
+  ~SumGPU();
+
+  /**
+   * @brief Sets up the reduction
+   *
+   * Sets up the reduction according to the parameters. The indices of dimensions to be reduced
+   * are provided in `axes` parameter.
+   * For a successful batch reduction, the reduced shape of all samples must be equal (but the
+   * input may have non-uniform shape, as long as the non-uniform dimensions are reduced).
+   *
+   * @param ctx          the execution environment
+   * @param in_shape     shape of the input tensor list
+   * @param axes         indices of axes to reduce along
+   * @param keep_dims    if true, the reduced dimensions are kept in the output shape, with the
+   *                     extent of 1
+   * @param reduce_batch if true, reduces respective output values of all samples in the batch
+   *                     and outputs a single tensor
+   */
   KernelRequirements Setup(KernelContext &ctx,
                            const TensorListShape<> &in_shape,
-                           span<const int> axes);
+                           span<const int> axes, bool keep_dims, bool reduce_batch);
 
+  /**
+   * @brief Performs the reduction, according to the parameters specified in Setup.
+   */
   void Run(KernelContext &ctx, const OutListGPU<Out> &out, const InListGPU<In> &in);
+
  private:
   class Impl;
   std::unique_ptr<Impl> impl_;
 };
+
+extern template class SumGPU<uint64_t, uint8_t>;
+extern template class SumGPU<float, uint8_t>;
+extern template class SumGPU<int64_t, int8_t>;
+extern template class SumGPU<float, int8_t>;
+extern template class SumGPU<uint64_t, uint16_t>;
+extern template class SumGPU<float, uint16_t>;
+extern template class SumGPU<int64_t, int16_t>;
+extern template class SumGPU<float, int16_t>;
+extern template class SumGPU<int64_t, int32_t>;
+extern template class SumGPU<float, int32_t>;
+extern template class SumGPU<float, float>;
 
 }  // namespace kernels
 }  // namespace dali
--- a/dali/kernels/reduce/reduce_gpu_impl.cuh
+++ b/dali/kernels/reduce/reduce_gpu_impl.cuh
@@ -246,11 +246,11 @@ auto GetPreprocessorsHelper(std::true_type, const ReduceImpl *impl, WorkArea *wa
   return impl->GetPreprocessorsImpl(*wa);
 }
 
-template <int non_reduced_dim, typename ReduceImpl>
+template <int non_reduced_dims, typename ReduceImpl>
 auto GetPreprocessorBanksHelper(
       std::true_type, const ReduceImpl *impl, WorkArea *wa, int reduced_axis)
-      ->decltype(impl->GetPreprocessorBanksImpl(*wa, reduced_axis)) {
-  return impl->GetPreprocessorBanksImpl(*wa, reduced_axis);
+      ->decltype(impl->template GetPreprocessorBanksImpl<non_reduced_dims>(*wa, reduced_axis)) {
+  return impl->template GetPreprocessorBanksImpl<non_reduced_dims>(*wa, reduced_axis);
 }
 
 template <typename ReduceImpl>
@@ -909,6 +909,20 @@ class ReduceImplGPU {
     CUDA_CALL(cudaGetLastError());
   }
 
+  template <typename Fn>
+  int GetMaxBlockSize(Fn *f) {
+    static int max_block_size[1024] = {};
+    int device = 0;
+    cudaGetDevice(&device);
+    if (!max_block_size[device]) {
+      cudaFuncAttributes attr = {};
+      CUDA_CALL(cudaFuncGetAttributes(&attr, f));
+      max_block_size[device] = attr.maxThreadsPerBlock;
+    }
+    return max_block_size[device];
+  }
+
+
   template <typename StageOut, typename StageIn, bool is_first, bool is_last>
   void LaunchStage(Context &ctx, const ReductionStage &stage,
                    ReductionKindTag<ReductionKind::Inner>) {
@@ -919,8 +933,15 @@ class ReduceImplGPU {
     auto *pre = GetPreprocessorBanks<is_first, 1>(wa, stage.axis);
     auto *post = GetPostprocessors<is_last>(wa);
 
+    using pre_bank_t = std::remove_cv_t<std::remove_reference_t<decltype(*pre)>>;;
+    using post_t = std::remove_cv_t<std::remove_reference_t<decltype(*post)>>;;
+    using red_t = std::remove_reference_t<decltype(This().GetReduction())>;
+
+    int max_block_size = GetMaxBlockSize(
+      ReduceInnerKernel<StageOut, StageIn, red_t, pre_bank_t, post_t>);
+
     wa.CopyParamsToDevice(ctx.stream);
-    dim3 block(32, 24);
+    dim3 block(32, max_block_size / 32);
     int gridx = std::max(32, 512/stage.num_samples());
     dim3 grid(gridx, stage.num_samples());
 
@@ -934,7 +955,6 @@ class ReduceImplGPU {
     CUDA_CALL(cudaGetLastError());
   }
 
-
   template <typename StageOut, typename StageIn, bool is_first, bool is_last>
   void LaunchStage(Context &ctx, const ReductionStage &stage,
                    ReductionKindTag<ReductionKind::Middle>) {
@@ -945,8 +965,15 @@ class ReduceImplGPU {
     auto *pre = GetPreprocessorBanks<is_first, 2>(wa, stage.axis);
     auto *post = GetPostprocessors<is_last>(wa);
 
+    using pre_bank_t = std::remove_cv_t<std::remove_reference_t<decltype(*pre)>>;;
+    using post_t = std::remove_cv_t<std::remove_reference_t<decltype(*post)>>;;
+    using red_t = std::remove_reference_t<decltype(This().GetReduction())>;
+
+    int max_block_size = GetMaxBlockSize(
+      ReduceMiddleKernel<StageOut, StageIn, red_t, pre_bank_t, post_t>);
+
     wa.CopyParamsToDevice(ctx.stream);
-    dim3 block(32, 24);
+    dim3 block(32, max_block_size / 32);
     int gridx = std::max(32, 512/stage.num_samples());
     dim3 grid(gridx, stage.num_samples());
     const int shm_size = 0x8000;  // 32 kB shared mem
--- a/dali/kernels/reduce/reduce_gpu_impl.cuh
+++ b/dali/kernels/reduce/reduce_gpu_impl.cuh
@@ -1,0 +1,1066 @@
+// Copyright (c) 2020, NVIDIA CORPORATION. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef DALI_KERNELS_REDUCE_REDUCE_GPU_IMPL_CUH_
+#define DALI_KERNELS_REDUCE_REDUCE_GPU_IMPL_CUH_
+
+#include <cassert>
+#include <memory>
+#include <utility>
+#include <vector>
+#include "dali/kernels/kernel.h"
+#include "dali/kernels/reduce/reduce_axes_gpu_impl.cuh"
+#include "dali/kernels/reduce/reductions.h"
+#include "dali/kernels/reduce/reduce_setup_utils.h"
+#include "dali/core/convert.h"
+#include "dali/core/format.h"
+#include "dali/core/small_vector.h"
+#include "dali/core/span.h"
+#include "dali/core/static_switch.h"
+#include "dali/core/tensor_view.h"
+#include "dali/core/traits.h"
+
+namespace dali {
+namespace kernels {
+namespace reduce_impl {
+
+enum class ReductionKind {
+  All,     ///< Reduce contiguous memory
+  Sample,  ///< Reduce samples separately
+  Block,   ///< Reduce contiguous samples separately
+  Inner,   ///< Reduce inner dimension
+  Middle,  ///< Reduce middle or outer dimension
+  None,    ///< No reduction - can happen when reduced dimension is degenerate
+  Fold,    ///< Reduce samples, keep spatial extents
+};
+
+struct ReductionShape {
+  int64_t outer, inner, reduced_in, reduced_out;
+
+  int64_t input_elements() const {
+    return outer * inner * reduced_in;
+  }
+
+  int64_t output_elements() const {
+    return outer * inner * reduced_out;
+  }
+};
+
+struct ReductionStage {
+  ReductionKind kind;
+  int axis;
+  int index;
+  bool is_last = false;
+
+  int num_samples() const {
+    return shape.size();
+  }
+
+  int64_t input_elements() const {
+    int64_t n = 0;
+    for (auto &rs : shape)
+      n += rs.input_elements();
+    return n;
+  }
+
+  int64_t output_elements() const {
+    int64_t n = 0;
+    for (auto &rs : shape)
+      n += rs.output_elements();
+    return n;
+  }
+
+  void calculate_offsets() {
+    int64_t ofs = 0;
+    int n_in = num_samples();
+    input_offsets.resize(n_in);
+    for (int i = 0; i < n_in; i++) {
+      input_offsets[i] = ofs;
+      ofs += shape[i].input_elements();
+    }
+
+    // if the stage produces contiguous output or just one sample, we only
+    // need one output pointer - otherwise we need a pointer for each sample separately
+    int n_out = kind == ReductionKind::All ||     // only one sample in output
+                kind == ReductionKind::Block ||   // contiguous output
+                kind == ReductionKind::Sample ||  // contiguous output
+                kind == ReductionKind::Fold       // only one sample in output
+                ? 1 : num_samples();
+
+    output_offsets.resize(n_out);
+    ofs = 0;
+    for (int i = 0; i < n_out; i++) {
+      output_offsets[i] = ofs;
+      ofs += shape[i].output_elements();
+    }
+  }
+
+  vector<ReductionShape> shape;
+  vector<int64_t> input_offsets, output_offsets;
+};
+
+
+struct TempBufferSizes {
+  int64_t io_buffers = 0;
+  int64_t param_buffers = 0;
+
+  void Reserve(TempBufferSizes sz) {
+    if (sz.io_buffers > io_buffers)
+      io_buffers = sz.io_buffers;
+    if (sz.param_buffers > param_buffers)
+      param_buffers = sz.param_buffers;
+  }
+
+  template <typename T>
+  void AddParam(int nsamples, size_t alignment = alignof(T)) {
+    param_buffers = align_up(param_buffers, alignment) + nsamples * sizeof(T);
+  }
+
+  template <typename T>
+  void AddIO(int nsamples, size_t alignment = alignof(T)) {
+    io_buffers = align_up(io_buffers, alignment) + nsamples * sizeof(T);
+  }
+};
+
+
+/**
+ * @brief Manages working memory for multi-stage operations
+ *
+ * The layout is:
+ * Host:  parameters
+ * GPU:   parameters  I/O ......... O/I
+ *
+ * Every other stage the roles of inputs and outputs are switched, so the outputs of the
+ * previous stage become the inputs of current one.
+ */
+struct WorkArea {
+  char *host_memory = nullptr, *gpu_memory = nullptr;
+
+  int64_t next_param_offset = 0;
+  int stage_index = 0;
+
+  void BeginStage(int stage_idx) {
+    next_param_offset = 0;
+    stage_index = stage_idx;
+  }
+
+  template <typename T>
+  T *ParamBuffer(int64_t n, size_t alignment = alignof(T)) {
+    next_param_offset = align_up(next_param_offset, alignment);
+    T *out = reinterpret_cast<T *>(host_memory + next_param_offset);
+    next_param_offset += n * sizeof(T);
+    assert(next_param_offset <= buffer_sizes.param_buffers);
+    return out;
+  }
+
+  template <typename T>
+  const T *InputBuffer(int64_t elements) {
+    return IOBuffer<const T>(elements, (stage_index & 1) == 1);
+  }
+
+  template <typename T>
+  T *OutputBuffer(int64_t elements) {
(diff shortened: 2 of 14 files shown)


dali/kernels/reduce/reduce_setup_utils.h
```diff
@@ -0,0 +1,194 @@
...
+      DALI_ENFORCE(sample_shape[a] == first_sample_shape[a], make_string(
+        "Reduce: batch reduction requires that all samples have the same extent in non-reduced "
+        "dimensions.\nError at sample ", i, " axis ", a, ": the extent is ", sample_shape[a],
+        " != ", first_sample_shape[a], " in the first sample in the batch."));
+    }
+  }
+}
+
+
+inline void CheckAxes(span<const int> axes, int dim) {
```
szalpal, line 132:
```suggestion
inline void CheckAxes(span<const int> axes, int ndims) {
```
`dim` IMHO suggests, it's an index of some specific dimension

dali/kernels/reduce/reduce_setup_utils.h
```diff
@@ -0,0 +1,194 @@
...
+  uint64_t mask = 0;
+  for (auto a : axes) {
+    if (a < 0 || a >= dim)
+      throw std::out_of_range(make_string("Axis index out of range: ", a, " not in 0..", dim-1));
+    uint64_t amask = 1ul << a;
+    if (mask & amask)
+      throw std::invalid_argument(make_string("Duplicate axis index ", a));
```
szalpal, line 139: This will only work, when `ndims<64`, right? Maybe `assert(ndims < 64)` somewhere at the beginning would be good, to document this?

dali/kernels/reduce/reduce_setup_utils.h
```diff
@@ -0,0 +1,194 @@
...
+    mask |= amask;
+  }
+}
```
szalpal, line 142: I'm not opposed to throwing otherwise than `DALI_ENFORCE`, but since it's our current workflow, I just want to put extra attention by asking the question. Could it have negative side effects? If not, then I guess we can agree that it's ok to throw not only in `DALI_ENFORCE`

dali/kernels/reduce/reduce_setup_utils.h
```diff
@@ -0,0 +1,194 @@
...
+
+
+inline void CheckBatchReduce(const TensorListShape<> &tls, span<const int> axes) {
+  if (tls.num_samples() == 0)
+    return;
+
+  uint64_t mask = to_bit_mask(axes);
+  SmallVector<int, 6> non_reduced;
```
szalpal, line 112: Why the `6` here?

dali/kernels/reduce/reduce_setup_utils.h
```diff
@@ -0,0 +1,194 @@
...
+  return mask;
+}
+
+/**
+ * @param out_axes a collection of integers, supporting `clear` and `push_back`.
+ * @param dim_groups a collection of integer pairs, supporting `clear` and `push_back`
+ */
+template <typename Axes, typename DimGroups, int ndim>
+void SimplifyReduction(Axes &out_axes, DimGroups &dim_groups,
+                       const TensorListShape<ndim> &in_shape, span<const int> axes) {
```
szalpal, line 56: It's not clear, from the documentation, what exactly is a `dim_group`. Could you put a description? (especially, because it's an output parameter)

dali/kernels/reduce/reduce_setup_utils.h
```diff
@@ -0,0 +1,194 @@
...
+
+template <typename ArrayLike, int ndim>
+inline void CalculateReductionFactors(ArrayLike &out, const TensorListShape<ndim> &in_shape,
+                                      span<const int> axes) {
+  for (int i = 0; i < in_shape.num_samples(); i++) {
+    auto sample_shape = in_shape.tensor_shape_span(i);
+    if (volume(sample_shape) == 0) {
+      out[i] = 0;  // degenerate sample - no reduction possible
+      continue;
+    }
```
szalpal, line 182: Hmmm, I'm lost in the 0-dim specialization. Isn't, in case of 0-dim, reduction also impossible, but `volume==1`?

mzient: It's not 0-dim. It's 0-volume, regardless of number of dimensions. Still, thanks for getting my attention to this, because it's actually a bug - it should be `max<decltype(Out[0])>` not zero - the reduction is infinite.

dali/kernels/reduce/reduce_setup_utils.h
mzient: I think we already throw other exceptions in kernels (but not in operators).

dali/kernels/reduce/reduce_setup_utils.h
mzient: Actually, I have more stringent limitations in other places, depending on the kind of preprocessing used. I'll document them and add appropriate validation.

dali/kernels/reduce/reduce_setup_utils.h
mzient: It's what we use in TensorShape.

dali/kernels/reduce/reduce_setup_utils.h
```diff
@@ -0,0 +1,194 @@
...
+constexpr enable_if_t<std::is_integral<T>::value, int> ilog2(T x) {
+  int n = 0;
+  while (x >>= 1)
+    n++;
+  return n;
+}
+
+
+template <typename OutType = uint64_t, typename BitIndices>
+OutType to_bit_mask(const BitIndices &bit_indices) {
```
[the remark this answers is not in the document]
mzient: Yes, it will output incorrect mask.
I though about moving this to core/util.h - in that case the thing I'd expect from such a function is _speed_. For the sake of if, I'll not add verification here - especially if I mark this function as `__host__ __device__` - I've had quite painful encounter with `assert` in device code eating up inordinate number of registers.

dali/kernels/reduce/reduce_setup_utils.h
```diff
@@ -0,0 +1,194 @@
...
+namespace kernels {
+namespace reduce_impl {
+
+template <typename T>
+constexpr enable_if_t<std::is_integral<T>::value, int> ilog2(T x) {
+  int n = 0;
+  while (x >>= 1)
+    n++;
+  return n;
+}
```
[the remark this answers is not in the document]
mzient: Kidding, yes?

dali/kernels/reduce/reduce_gpu.h
```diff
@@ -0,0 +1,41 @@
...
+namespace kernels {
+
+template <typename Out, typename In>
+class SumGPU {
+ public:
+  KernelRequirements Setup(KernelContext &ctx,
+                           const TensorListShape<> &in_shape,
+                           span<const int> axes);
+
+  void Run(KernelContext &ctx, const OutListGPU<Out> &out, const InListGPU<In> &in);
```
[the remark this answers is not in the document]
mzient: Yes.... I'll remove that sentence from there... in a separate PR ;)

dali/kernels/reduce/reduce_gpu.h
```diff
@@ -0,0 +1,41 @@
...
+#include <memory>
+#include "dali/kernels/kernel.h"
+#include "dali/core/tensor_view.h"
+
+namespace dali {
+namespace kernels {
+
+template <typename Out, typename In>
+class SumGPU {
+ public:
```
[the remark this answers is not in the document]
mzient: Oh, this file is a stub, really. There will be more reduction types here.

dali/kernels/reduce/reduce_setup_utils.h
szalpal: How about this then?
```
  SmallVector<int, MAX_NDIMS> non_reduced;
  SmallVector<int, 6 /* max number of dims */> non_reduced;
```

dali/kernels/reduce/reduce_setup_utils.h
szalpal: I know it's 0-volume. But when we have 0-dim (a scalar), reduction is also impossible, isn't it? So I suppose this condition should look something like
```
if (volume(sample_shape) == 0 || sample_shape.ndims() == 0)
```

mzient: It's not like it. When reducing, you force a reduced dimension to 1. If the tensor is empty (0 volume) you can't conjure up the data to be reduced.
A scalar, on the other hand, can't be reduced due to a totally different condition - there's no valid axis index.
I wonder what I should do with these 0-sized tensors - when not all axes are reduced and the output is produced with `keep_dims=true`, I can still set the reduced extents to 1 and non-reduced ones to 0, so the output is empty.

mzient:
!build

dali/kernels/reduce/reduce_setup_utils.h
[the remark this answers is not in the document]
szalpal: I just wanted to point out, that it's not obvious and the incorrect mask will be an error that's hard to spot later on. How much speed would we sacrifice by verifying value here? In per-cents?

mzient:
!build

mzient:
!build

dali/kernels/reduce/reduce_setup_utils.h
```diff
@@ -0,0 +1,228 @@
...
+  int d = in_shape.sample_dim();
+
+  uint64_t mask = to_bit_mask(axes);
+
+  int i;
+
+  // skip leading degenerate dimensions
+  for (i = 0; i < d && is_degenerate_dim(in_shape, i); i++) {}
+
+  if (i < d) {
```
klecki, line 86: Can you maybe reverse the condition, do early exit in if case, and do the rest without else?

dali/kernels/reduce/reduce_setup_utils.h
```diff
@@ -0,0 +1,228 @@
...
+    n++;
+  return n;
+}
+
+/**
+ * @brief Returns an integer where bits at indicies in `bit_indices` are set to 1.
+ * @remarks Indices that are outside the bit-width of OutType are ignored.
+ */
+template <typename OutType = uint64_t, typename BitIndices>
+OutType to_bit_mask(const BitIndices &bit_indices) {
```
klecki, line 49: I also prefer snake_case, but shouldn't this be `ToBitMask`?

dali/kernels/reduce/reduce_gpu_impl.cuh
```diff
@@ -0,0 +1,1092 @@
...
+      }
+    } else {
+      buf_sizes.AddIO<Out>(stage.output_elements());
+    }
+  }
+
+  /**
+   * @brief Initializes processing where no per-sample dimension is reduced
+   *
+   * The reduction is either pre- and post-processing only or a of repsective elements
```
klecki, line 428: respective

klecki: also what's up with: `or a of ...`

dali/kernels/reduce/reduce_gpu_impl.cuh
```diff
@@ -0,0 +1,1092 @@
...
+  Sample,  ///< Reduce samples separately
+  Block,   ///< Reduce contiguous samples separately
+  Inner,   ///< Reduce inner dimension
+  Middle,  ///< Reduce middle or outer dimension
+  None,    ///< No reduction - can happen when reduced dimension is degenerate
+  Fold,    ///< Reduce samples, keep spatial extents
+};
+
+struct ReductionShape {
+  int64_t outer, inner, reduced_in, reduced_out;
```
klecki, line 49: TBH, I have no idea what those variables are supposed to describe.

dali/kernels/reduce/reduce_gpu_impl.cuh
```diff
@@ -0,0 +1,1092 @@
...
+#include "dali/core/traits.h"
+
+namespace dali {
+namespace kernels {
+namespace reduce_impl {
+
+enum class ReductionKind {
+  All,     ///< Reduce contiguous memory
+  Sample,  ///< Reduce samples separately
+  Block,   ///< Reduce contiguous samples separately
```
JanuszL, line 41: So we assume that samples can be not continuous?

dali/kernels/reduce/reduce_gpu_impl.cuh
```diff
@@ -0,0 +1,1092 @@
...
+    input_offsets.resize(n_in);
+    for (int i = 0; i < n_in; i++) {
+      input_offsets[i] = ofs;
+      ofs += shape[i].input_elements();
+    }
+
+    // if the stage produces contiguous output or just one sample, we only
+    // need one output pointer - otherwise we need a pointer for each sample separately
+    int n_out = kind == ReductionKind::All ||     // only one sample in output
+                kind == ReductionKind::Block ||   // contiguous output
```
JanuszL, line 96: Nitpick, you can keep the same order as this values appears in the enum.

dali/kernels/reduce/reduce_gpu_impl.cuh
```diff
@@ -0,0 +1,1092 @@
...
+  vector<ReductionShape> shape;
+  vector<int64_t> input_offsets, output_offsets;
+};
+
+
+struct TempBufferSizes {
+  int64_t io_buffers = 0;
+  int64_t param_buffers = 0;
+
+  void Reserve(TempBufferSizes sz) {
```
JanuszL, line 118: It doesn't reserve anything, I don't know if this is a fortunate name.

dali/kernels/reduce/reduce_gpu_impl.cuh
```diff
@@ -0,0 +1,1092 @@
...
+  KernelRequirements Setup(KernelContext &ctx,
+                           const TensorListShape<> &in_shape,
+                           span<const int> axes,
+                           bool keep_dims,
+                           bool reduce_batch) {
+    if (in_shape.sample_dim() > 64)
+      throw std::range_error("Reduce supports up to 64 dimensions");
+    reduce_batch_ = reduce_batch;
+    CheckAxes(axes, in_shape.sample_dim());
+    if (reduce_batch)
```
JanuszL, line 224:
```suggestion
    if (reduce_batch_)
```

dali/kernels/reduce/reduce_gpu_impl.cuh
```diff
@@ -0,0 +1,1092 @@
...
+
+  /// DO NOT redefine in CRTP derived class
+  identity GetPostprocessor(std::false_type is_last) const {
+    return {};
+  }
+
+  void CalculateTempBuffers(ScratchpadEstimator &se) {
+    int nstages = stages_.size();
+    buffer_sizes_ = {};
+    for (int stage_idx = 0; stage_idx < nstages; stage_idx++) {
```
JanuszL, line 327:
```suggestion
    for (auto& stage: stages_) {
```

dali/kernels/reduce/reduce_gpu_impl.cuh
```diff
@@ -0,0 +1,1092 @@
...
+      auto &stage = stages_[stage_idx];
+      TempBufferSizes stage_buffers;
+      CalculateTempBuffers(stage_buffers, stage);
+      buffer_sizes_.Reserve(stage_buffers);
+    }
+
+    // now reserve the scratchpad - overaligned to 64 bytes, just in case
```
JanuszL, line 334: In case of what? It seem like a war for some access beyond the array.

dali/kernels/reduce/reduce_gpu_impl_test.cu
```diff
@@ -0,0 +1,657 @@
...
+#include "dali/kernels/scratch.h"
+#include "dali/test/test_tensors.h"
+#include "dali/test/tensor_test_utils.h"
+#include "dali/core/tensor_shape_print.h"
+
+namespace dali {
+
+
+template <typename T, int ndim>
+std::ostream &operator<<(std::ostream &os, const TensorView<StorageCPU, T, ndim> &t) {
```
JanuszL, line 28: Please move to come common utils.

dali/kernels/reduce/reduce_gpu_impl.cuh
```diff
@@ -0,0 +1,1092 @@
...
+    stage.index = 0;
+    stage.is_last = true;
+    stage.kind = reduce_batch_ ? ReductionKind::Fold : ReductionKind::None;
+    int N = in_shape_.num_samples();
+    stage.shape.resize(N);
+    for (int i = 0; i < N; i++) {
+      ReductionShape &rs = stage.shape[i];
+      rs.inner = volume(in_shape_.tensor_shape_span(i));
+      rs.outer = rs.reduced_in = 1;
+      rs.reduced_out = (i == 0 || !reduce_batch_) ? 1 : 0;
```
klecki, line 444: What's with reduced_in?

dali/kernels/reduce/reduce_gpu_impl.cuh
```diff
@@ -0,0 +1,1092 @@
...
+      ReductionStage &stage = stages_[s];
+      if (stage.kind == ReductionKind::All) {
+        // ReduceAll cannot happen at stage 0
+        assert(s > 0 || nsamples == 1);
+        int64_t r = 0;
+        if (s > 0) {
+          for (auto &rs : stages_[s-1].shape)
+            r += rs.reduced_out;
+        } else {
+          r = max_reduced_extent;
```
klecki, line 529: if ReduceAll cannot happen at `s == 0`, and we even assert this, the whole `if (s > 0) {} else {}` checks are superfluous, also the `r = max_reduced_extent` seems a bit weird.

dali/kernels/reduce/reduce_setup_utils.h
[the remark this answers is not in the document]
mzient: If I put an assert there and it makes me change CUDA block size? Perhaps 50%.

dali/kernels/reduce/reduce_gpu_impl.cuh
```diff
@@ -0,0 +1,1058 @@
...
+        r = CalcReducedExtent(r, remaining_stages);
+        stage.shape[0].reduced_out = r;
+      } else {
+        int64_t max_in = 0;
+        if (s > 0) {
+          for (int i = 0; i < nsamples;  i++) {
+            auto &rs = stage.shape[i];
+            int64_t r = stages_[s-1].shape[i].reduced_out;
+            rs.reduced_in = r;
+            rs.inner = rs.outer = 1;
```
klecki, line 549: Why the reduced_in, inner and outer are set only in the `if (s > 0)` case?

dali/kernels/reduce/reduce_setup_utils.h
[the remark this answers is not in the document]
mzient: Anyway, I think you're looking at it as a high level function. It's a _very low level_ function. It's like a glorified bit shift operator - it certainly doesn't throw when you do 1<<100.

dali/kernels/reduce/reduce_setup_utils.h
```diff
@@ -0,0 +1,194 @@
...
+    n++;
+  return n;
+}
+
+
+template <typename OutType = uint64_t, typename BitIndices>
+OutType to_bit_mask(const BitIndices &bit_indices) {
+  OutType mask = 0;
+  for (int idx : bit_indices)
+    mask |= OutType(1) << idx;
```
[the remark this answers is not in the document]
mzient: It has a big potential for generating stupid warnings, that's certain. As for the "improvement" - I fail to see any.

dali/kernels/reduce/reduce_setup_utils.h
mzient: I'll move to_bit_mask and ilog2 to code/util.h - just not now, because it would cause massive rebuilds when switching branches.

dali/kernels/reduce/reduce_gpu_impl.cuh
```diff
@@ -0,0 +1,1058 @@
...
+
+    for (int s = 0; s < substages; s++) {
+      ReductionStage &stage = stages_[s];
+      stage.index = s;
+      // stage 0 is never ReduceAll, because that would require contiguous input
+      stage.kind = s == 0 ? ReductionKind::Sample
+                          : reduce_batch_ ? ReductionKind::All : ReductionKind::Block;
+      stage.shape.resize(stage.kind == ReductionKind::All ? 1 : nsamples);
+    }
+
```
klecki, line 517: Correct me if I'm wrong, but is the whole processing below just setting up progressively smaller stages with condition stage[i].reduced_in = stage[i-1].reduced_out?
And inner = outer = 1 for all stages?
Isn't the reduce_batch_ step happening just ather the first one, and than you should be able to handle this with two variants of simple loop for stages 1..substages-1, in one you have `nsmaples` shapes in other only 1.

dali/kernels/reduce/reduce_gpu_impl.cuh
```diff
@@ -0,0 +1,1058 @@
...
+  void Run(KernelContext &kctx, const OutListGPU<Out> &out, const InListGPU<In> &in) {
+    assert(!stages_.empty());
+    Context ctx;
+    auto host_mem_size = buffer_sizes_.param_buffers;
+    auto gpu_mem_size = buffer_sizes_.param_buffers + buffer_sizes_.io_buffers;
+
+    ctx.stream = kctx.gpu.stream;
+    ctx.work_area.buffer_sizes = buffer_sizes_;
+    ctx.work_area.host_memory = kctx.scratchpad->Allocate<char>(AllocType::Host, host_mem_size, 64);
+    ctx.work_area.gpu_memory = kctx.scratchpad->Allocate<char>(AllocType::GPU, gpu_mem_size, 64);
```
klecki, line 998: You're not using the same alloc&pad approach to gpu param & io buffers (in setup you configure them in two separate steps). It probably won't matter, but that's probably the only issue I will be able to notice, as the CalculateTempBuffers is incomprehensible for me rn (at least till I understand the memory use of the whole kernel).

dali/kernels/reduce/reduce_gpu_impl.cuh
```diff
@@ -0,0 +1,1058 @@
...
+    return samples;
+  }
+
+  template <typename StageOut, typename StageIn, bool is_first, bool is_last>
+  void LaunchStage(Context &ctx, const ReductionStage &stage,
+                   ReductionKindTag<ReductionKind::All>) {
+    assert(!is_first || ctx.input.is_contiguous());
+    WorkArea &wa = ctx.work_area;
+
+    auto pre = This().GetPreprocessor(bool_const<is_first>());
```
klecki, line 777: Can the `All` reduction be first?

dali/kernels/reduce/reduce_setup_utils.h
mzient: Why not.

dali/kernels/reduce/reduce_gpu_impl.cuh
mzient: Yes.

dali/kernels/reduce/reduce_gpu_impl.cuh
mzient: I can do it, but using the parameter is safer, in case the check is moved above the member assignment.

dali/kernels/reduce/reduce_gpu_impl.cuh
mzient: It's above.

dali/kernels/reduce/reduce_gpu_impl.cuh
mzient: I don't see how this would be simpler.

dali/kernels/reduce/reduce_gpu_impl.cuh
mzient: Stale comment - look at the assert again.

dali/kernels/reduce/reduce_gpu_impl.cuh
mzient: It's missing, but they're not used in ReduceAll/ReduceBlock anyway.

dali/kernels/reduce/reduce_gpu_impl.cuh
mzient: If there's just one sample in the input.

dali/kernels/reduce/reduce_gpu_impl.cuh
mzient: The calculation of buffer size happens elsewhere and here we don't know what types are allocated - so we're very conservative about alignment.

dali/kernels/reduce/reduce_gpu_impl.cuh
mzient: Right you are. Actually, I should align param_buffer's size and allocate - and reserve - all gpu memory in one go.

mzient:
!build

mzient:
!build

mzient:
!build

dali/kernels/reduce/reduce_gpu.h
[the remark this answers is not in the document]
mzient: OK, now it's not a stub, but it's still partial. Mean, RMS and StdDev will come in a followup.

dali/kernels/reduce/reduce_setup_utils.h
```diff
@@ -0,0 +1,229 @@
...
+ *
+ * @param in_shape   shape of the input
+ * @param axes       list of reduced axes; axis indices must be <= 63
+ *
+ * A dimension can be collapsed with its neighbor if:
+ *  - its extent is 1 in all samples
+ *  - both dimensions are reduced or non-reduced.
+ */
+template <typename Axes, typename DimGroups, int ndim>
+void SimplifyReduction(Axes &out_axes, DimGroups &dim_groups,
```
JanuszL, line 73:
```suggestion
void SimplifyReduction(Axes &out_axes, DimGroups &out_dim_groups,
```
?

dali/kernels/reduce/reduce_setup_utils.h
```diff
@@ -0,0 +1,229 @@
...
+    if (!(mask & (1 << a)))
+      non_reduced.push_back(a);
+  }
+
+  auto first_sample_shape = tls.tensor_shape_span(0);
+
+  for (int i = 1; i < tls.num_samples(); i++) {
+    auto sample_shape = tls.tensor_shape_span(i);
+    for (int a : non_reduced) {
+      DALI_ENFORCE(sample_shape[a] == first_sample_shape[a], make_string(
```
JanuszL, line 149: DO you want to use `DALI_ENFORCE` inside kernel? I think the plain exception is the way.

dali/kernels/reduce/reduce_setup_utils.h
```diff
@@ -0,0 +1,229 @@
...
+      throw std::out_of_range(make_string("Axis index out of range: ", a, " not in 0..", ndim-1));
+    uint64_t amask = 1ul << a;
+    if (mask & amask)
+      throw std::invalid_argument(make_string("Duplicate axis index ", a));
+    mask |= amask;
+  }
+}
+
+
+inline void CalculateReducedShape(TensorListShape<> &out_shape,
```
JanuszL, line 178: Any doxygen doc, please.

dali/kernels/reduce/reduce_gpu_impl.cuh
JanuszL: So I would rephrase the comment to explain that.

dali/kernels/reduce/reduce_gpu_impl_test.cu
```diff
@@ -0,0 +1,533 @@
...
+
+  std::mt19937_64 rng(12345);
+
+  TestTensorList<uint8_t> in;
+  in.reshape(in_shape);
+  auto in_cpu = in.cpu();
+  UniformRandomFill(in_cpu, rng, 0, 255);
+
+  TestTensorList<int64_t> out;
+  out.reshape(req.output_shapes[0]);
```
JanuszL, line 373: Any particular reason why you don't call `cudaMemset` lik in the previous test case?

dali/kernels/reduce/reduce_gpu_test.cc
```diff
@@ -0,0 +1,87 @@
...
+  TestTensorList<uint8_t> in;
+  in.reshape(in_shape);
+  auto in_cpu = in.cpu();
+  UniformRandomFill(in_cpu, rng, 0, 255);
+
+  TestTensorList<uint64_t> out;
+  out.reshape(req.output_shapes[0]);
+  sum.Run(ctx, out.gpu(), in.gpu());
+
+  auto out_cpu = out.cpu(ctx.gpu.stream);
```
JanuszL, line 59: Doesn't it call `cudaMemcpyAsync`? It requires to sync to have the memory available on the CPU, right?

mzient: AFAIK copying to non-pinned memory is implicitly synchronized with the host.

dali/kernels/reduce/reduce_setup_utils.h
mzient: will do

dali/kernels/reduce/reduce_setup_utils.h
mzient: I'll replace it with appropriate error (invalid argument or logic error).

dali/kernels/reduce/reduce_gpu_impl.cuh
mzient: Done (param_buffers are aligned in CalculateTempBuffers.

dali/kernels/reduce/reduce_setup_utils.h
mzient: Done.

dali/kernels/reduce/reduce_gpu_impl_test.cu
mzient: I only used cudaMemset in case of degenerate input (zero size) - in that cases, the expected value is zero, which may coincide with how the memory is initialized on allocation. Here, I have some real value to compare against and it's exceedingly unlikely that the memory will happen to contain the expected result.

mzient:
!build

dali/kernels/reduce/reduce_gpu_test.cc
JanuszL: OK

dali/kernels/reduce/reduce_gpu_impl_test.cu
JanuszL: OK

mzient:
!build

PR #1890, opened 2020-04-22
Copy out core* files if the test_body fails

klecki:
Copy them to top DALI directory from test working directory
for unified location

#### Why we need this PR?
To capture core dumps in the CI

 - What solution was applied:
	Run the test body without exiting when the test fails, copy the core* files to predefined location and report the error.
     QA tests
 - Key points relevant for the review: N/A
 - Validation and testing:	CI
 - Documentation (including examples): N/A

### DIFF
--- a/qa/test_template.sh
+++ b/qa/test_template.sh
@@ -62,12 +62,12 @@ do
         set +e
         RV=0
         test_body || RV=$?
-        set -e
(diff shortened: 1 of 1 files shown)


klecki:
#1896 test this thing by forcing abort() which will cause a core dump to be written.

PR #1892, opened 2020-04-22
C API ExternalSource for GPU input

szalpal:
As the title suggests.

What shall deserve a remark is that I introduced additional functions for ExternalSource: `daliSetExternalInputStream` and `daliSetExternalInputTensorsStream`. This is because, in case of GPU source, user might want to specify the stream which is used to copy the data. If a regular function is used, default stream is used (`0`)

Great thanks to @klecki for writing tests for ExternalSource. Thanks to them this went smooothly. I changed them to TYPED, so that it's parameterized by Backend

I know I reformatted whole `c_api.h`. It has been indented, while I believe, that `extern "C" {}` block should follow rules of `namespace {}` block, thus do not impose indentation. In case you can't spot the differences, take a look at lines `80 - 140` in this file. That's the only difference.

In this reformatting frenzy I might have reformatted also other parts of code. Sorry for that, if anyone will have some great complaints, I'll fix this.

EDIT: There were lots of discussions on this topic, so I'll summarize it here.
Generally the problem is how to provide synchronization in 2 points - user API call and ExternalSource RunImpl. I decided for a following solution. User provides a stream, on which the `user storage -> ext src storage` copy will be conducted. He has to synchronize on this stream. After scheduling the copies on this stream, event is also recorded. Using this event, RunImpl will synchronize its own copy. This way we achieve following benefits:
1. API call is used the same way, regardless of operator and data backends
2. ExternalSource doesn't require separate storages for CPU and GPU backends
3. There is a slight "problem" - user will need to provide stream also in 1 case (out of 4), in which it's redundant: CPU op and CPU data. However, I decided, that he'll live with it - the API assumes CUDA is available.

I created also a convenience, blocking function, which will handle the stream for the user, if he doesn't care about it. It's somewhat response for point 3 above

### DIFF
--- a/dali/c_api/c_api.cc
+++ b/dali/c_api/c_api.cc
@@ -18,6 +18,7 @@
 #include <vector>
 
 #include "dali/c_api.h"
+#include "dali/core/format.h"
 #include "dali/core/tensor_shape.h"
 #include "dali/pipeline/pipeline.h"
 #include "dali/plugin/copy.h"
@@ -28,11 +29,11 @@
 namespace {
 
 template<typename Backend>
-void SetExternalInput(daliPipelineHandle* pipe_handle, const char* name, device_type_t device,
-                          const void* data_ptr, dali_data_type_t data_type, const int64_t* shapes,
-                          int sample_dim, const char* layout_str) {
+void SetExternalInput(daliPipelineHandle *pipe_handle, const char *name, device_type_t device,
+                      const void *data_ptr, dali_data_type_t data_type, const int64_t *shapes,
+                      int sample_dim, const char *layout_str, cudaStream_t stream = 0) {
   DALI_ENFORCE(device == CPU, "GPU data cannot be passed as external source");
-  dali::Pipeline* pipeline = reinterpret_cast<dali::Pipeline*>(pipe_handle->pipe);
+  dali::Pipeline *pipeline = reinterpret_cast<dali::Pipeline *>(pipe_handle->pipe);
   std::vector<int64_t> shapes_tmp(shapes, shapes + sample_dim * pipeline->batch_size());
   dali::TensorListShape<> tl_shape(std::move(shapes_tmp), pipeline->batch_size(), sample_dim);
   dali::TensorLayout layout{};
@@ -40,25 +41,26 @@ void SetExternalInput(daliPipelineHandle* pipe_handle, const char* name, device_
     layout = dali::TensorLayout(layout_str);
   }
   dali::TensorList<Backend> data;
-  const auto& type_info = dali::TypeTable::GetTypeInfo(static_cast<dali::DALIDataType>(data_type));
+  const auto &type_info = dali::TypeTable::GetTypeInfo(static_cast<dali::DALIDataType>(data_type));
   auto elem_sizeof = type_info.size();
   // We cast away the const from data_ptr, as there is no other way of passing it to the
   // TensorList, as we must also set the shape and type metadata.
   // It is passed further as const TensorList, so it's data cannot be modified.
-  data.ShareData(const_cast<void*>(data_ptr), tl_shape.num_elements() * elem_sizeof);
+  data.ShareData(const_cast<void *>(data_ptr), tl_shape.num_elements() * elem_sizeof);
   data.Resize(tl_shape, type_info);
   data.SetLayout(layout);
-  const auto& cdata = data;
-  pipeline->SetExternalInput(name, cdata);
+  const auto &cdata = data;
+  pipeline->SetExternalInput(name, cdata, stream);
 }
 
+
 template<typename Backend>
-void SetExternalInputTensors(daliPipelineHandle* pipe_handle, const char* name,
-                                 device_type_t device, const void* const* data_ptr,
-                                 dali_data_type_t data_type, const int64_t* shapes,
-                                 int64_t sample_dim, const char* layout_str) {
+void SetExternalInputTensors(daliPipelineHandle *pipe_handle, const char *name,
+                             device_type_t device, const void *const *data_ptr,
+                             dali_data_type_t data_type, const int64_t *shapes, int64_t sample_dim,
+                             const char *layout_str, cudaStream_t stream = 0) {
   DALI_ENFORCE(device == CPU, "GPU data cannot be passed as external source");
-  dali::Pipeline* pipeline = reinterpret_cast<dali::Pipeline*>(pipe_handle->pipe);
+  dali::Pipeline *pipeline = reinterpret_cast<dali::Pipeline *>(pipe_handle->pipe);
   std::vector<int64_t> shapes_tmp(shapes, shapes + sample_dim * pipeline->batch_size());
   dali::TensorListShape<> tl_shape(std::move(shapes_tmp), pipeline->batch_size(), sample_dim);
   dali::TensorLayout layout{};
@@ -66,18 +68,18 @@ void SetExternalInputTensors(daliPipelineHandle* pipe_handle, const char* name,
     layout = dali::TensorLayout(layout_str);
   }
   std::vector<dali::Tensor<Backend>> data(pipeline->batch_size());
-  const auto& type_info = dali::TypeTable::GetTypeInfo(static_cast<dali::DALIDataType>(data_type));
+  const auto &type_info = dali::TypeTable::GetTypeInfo(static_cast<dali::DALIDataType>(data_type));
   auto elem_sizeof = type_info.size();
   for (int i = 0; i < pipeline->batch_size(); i++) {
     // We cast away the const from data_ptr, as there is no other way of passing it to the
     // Tensor as we must also set the shape and type metadata.
     // The vector that we pass to pipeline is const.
-    data[i].ShareData(const_cast<void*>(data_ptr[i]), tl_shape[i].num_elements() * elem_sizeof);
+    data[i].ShareData(const_cast<void *>(data_ptr[i]), tl_shape[i].num_elements() * elem_sizeof);
     data[i].Resize(tl_shape[i], type_info);
     data[i].SetLayout(layout);
   }
-  const auto& cdata = data;
-  pipeline->SetExternalInput(name, cdata);
+  const auto &cdata = data;
+  pipeline->SetExternalInput(name, cdata, stream);
 }
 
 }  // namespace
@@ -120,7 +122,7 @@ void daliPrefetchUniform(daliPipelineHandle* pipe_handle, int queue_depth) {
 
 void daliPrefetchSeparate(daliPipelineHandle* pipe_handle,
                           int cpu_queue_depth, int gpu_queue_depth) {
-  dali::Pipeline* pipeline = reinterpret_cast<dali::Pipeline*>(pipe_handle->pipe);
+  dali::Pipeline *pipeline = reinterpret_cast<dali::Pipeline *>(pipe_handle->pipe);
   for (int i = 0; i < gpu_queue_depth; ++i) {
     pipeline->RunCPU();
     pipeline->RunGPU();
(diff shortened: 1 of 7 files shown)


szalpal:
!build

dali/c_api/c_api.cc
```diff
@@ -12,19 +12,77 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+#include "dali/c_api.h"  // NOLINT [build/include]
```
JanuszL, line 15: Why moving to the top?

dali/c_api/c_api.cc
```diff
@@ -12,19 +12,77 @@
...
 #include "dali/pipeline/pipeline.h"
 #include "dali/plugin/copy.h"
 #include "dali/plugin/plugin_manager.h"
 #include "dali/pipeline/data/tensor_list.h"
 #include "dali/pipeline/data/backend.h"
 
+namespace {
+
+template<typename Backend>
+void SetExternalInput(daliPipelineHandle *pipe_handle, const char *name, device_type_t device,
```
JanuszL, line 33:
```suggestion
void SetExternalInput(daliPipelineHandle *pipe_handle, const char *name,
```

dali/c_api/c_api.cc
```diff
@@ -12,19 +12,77 @@
...
+  data.Resize(tl_shape, type_info);
+  data.SetLayout(layout);
+  const auto &cdata = data;
+  pipeline->SetExternalInput(name, cdata, stream);
+}
+
+
+template<typename Backend>
+void SetExternalInputTensors(daliPipelineHandle *pipe_handle, const char *name,
+                             device_type_t device, const void *const *data_ptr,
```
JanuszL, line 59:
```suggestion
                             const void *const *data_ptr,
```

include/dali/c_api.h
```diff
@@ -23,214 +23,239 @@
...
+ *             to be fed
+ * @param device Device of the supplied memory. Only CPU is supported.
+ * @param data_ptr Pointer to contiguous buffer containing all samples
+ * @param data_type Type of the provided data
+ * @param shapes Pointer to an array containing shapes of all samples concatenated one after
+ *              another. Should contain batch_size * sample_dim elements.
+ * @param sample_dim The dimensionality of a single sample.
+ * @param layout_str Optional layout provided as a pointer to null-terminated byte string.
+ *                   Can be set to NULL.
+ * @param stream CUDA stream to use when copying the data onto GPU. If used with `device == CPU`,
```
JanuszL, line 97: There is no such arg in this function. Still you should state that for the GPU device it will use a default stream. If the user wants different one `daliSetExternalInputStream` should be used.

include/dali/c_api.h
```diff
@@ -23,214 +23,239 @@
...
+ *               this value will be ignored
+ */
+DLL_PUBLIC void daliSetExternalInput(daliPipelineHandle *pipe_handle, const char *name,
+                                     device_type_t device, const void *data_ptr,
+                                     dali_data_type_t data_type, const int64_t *shapes,
+                                     int sample_dim, const char *layout_str);
+
```
JanuszL, line 104: I think you need a separate docs for this function as well. I don't think doxygen will tak the one from the above.

include/dali/c_api.h
```diff
@@ -23,214 +23,239 @@
...
+ *             to be fed
+ * @param device Device of the supplied memory. Only CPU is supported.
+ * @param data_ptr Pointer to an array containing batch_size pointers to separate Tensors.
+ * @param data_type Type of the provided data
+ * @param shapes Pointer to an array containing shapes of all samples concatenated one after
+ *              another. Should contain batch_size * sample_dim elements.
+ * @param sample_dim The dimensionality of a single sample.
+ * @param layout_str Optional layout provided as a pointer to null-terminated byte string.
+ *                   Can be set to NULL.
+ * @param stream CUDA stream to use when copying the data onto GPU. If used with `device == CPU`,
```
JanuszL, line 129: As above.

dali/pipeline/operator/builtin/external_source.h
```diff
@@ -158,7 +159,7 @@ class ExternalSource : public Operator<Backend> {
       data = tl_data_.GetEmpty();
     }
 
-    data.front()->Copy(tl, 0);
+    data.front()->Copy(tl, stream);
```
JanuszL, line 162: It sucks. No!!!
The external source keeps data on CPU now. With your change you will have round trip - source from GPU, copy to CPU inside SetDataSource, copy to GPU inside RunImpl. You need to extend external_source.h to be able to store data on the GPU as well.

dali/pipeline/operator/builtin/external_source.h
```diff
@@ -185,7 +188,7 @@ class ExternalSource : public Operator<Backend> {
 
     data.front()->resize(t.size());
     for (size_t i = 0; i < t.size(); ++i) {
-      (*(data.front()))[i].Copy(t[i], 0);
+      (*(data.front()))[i].Copy(t[i], stream);
```
JanuszL, line 191: As above.

szalpal:
!build

dali/c_api/c_api_test.cc
```diff
@@ -39,81 +39,112 @@ constexpr bool pipelined = true;
...
       EXPECT_EQ(idx, ref_shape.sample_dim());
       free(shape);
     }
 
-    TensorList<CPUBackend> c_output;
-    auto &regular_output = ws.Output<CPUBackend>(0);
-    c_output.Resize(regular_output.shape(), TypeInfo::Create<uint8_t>());
-    daliCopyTensorListNTo(&handle, c_output.raw_mutable_data(), 0, device_type_t::CPU, 0, false);
-    Check(view<uint8_t>(c_output), view<uint8_t>(regular_output));
+    TensorList<CPUBackend> output1, output2;
```
klecki, line 120: Can you give them less generic names, like c_api_out_cpu, regular_our_cpu?

dali/pipeline/operator/builtin/external_source.h
```diff
@@ -167,13 +168,15 @@ class ExternalSource : public Operator<Backend> {
...
    */
-  inline void SetDataSource(const vector<Tensor<CPUBackend>> &t) {
+  template<typename SrcBackend>
+  inline void SetDataSource(const vector<Tensor<SrcBackend>> &t, cudaStream_t stream = 0) {
     DALI_ENFORCE(OperatorBase::batch_size_ == static_cast<int>(t.size()),
-      "Data list provided to ExternalSource needs to have batch_size length.");
+                 "Data list provided to ExternalSource needs to have batch_size length.");
     // Note: If we create a GPU source, we will need to figure
     // out what stream we want to do this copy in. CPU we can
     // pass anything as it is ignored.
```
klecki, line 182: If we do that copy on some external stream, and than in RunImpl do another copy out of those buffers on yet another stream, are we sure, that there is no data race?

Don't we need some cross-stream sync here? Can you check it?

I'm afraid of data races.
With CPU and stream 0 there should be no such issues AFAIK, as it is quite synchronous, but when you start using other streams there may be problems.

klecki:
I have another general remark. How efficient are the cases where we supply data on different device than the op is placed?

If the user is not familiar enough with internals, he can do
ExternalSource(device="cpu"), supply GPU data and use it only on GPU.
That will do internal copy to CPU buffers of the Operator and next put MakeContiguous that will do CPU->GPU copy.
If the user already has the contiguous data (or even separate buffers), it would be just device->device copy instead of this device->host->device roundtrip.

I was wondering if we should somehow limit GPU inputs to only ExternalSource(device="gpu"), but that limits the scope of operators that can be used.

dali/pipeline/operator/builtin/external_source.h
JanuszL: I think the user should sync on the stream he uses to copy. I no stream is provided and the default is used we should sync on our own.
The current assumption is that after you exit from `SetDataSource` data should be safe inside the DALI, so you can do whatever you want with the data source provided to it.
I can imagine you can ask DALI for its stream, work on it and put some even synchronization on it to make sure that `SetDataSource` consumed it. But I don't know how error-prone it could be.

JanuszL:
> I was wondering if we should somehow limit GPU inputs to only ExternalSource(device="gpu"), but that limits the scope of operators that can be used.

Maybe just DALI warn in such case?

klecki:
> > I was wondering if we should somehow limit GPU inputs to only ExternalSource(device="gpu"), but that limits the scope of operators that can be used.
> 
> Maybe just DALI warn in such case?

Good idea, I was thinking about reporting such case, but you would need to do a not so simple diagnostic to warn about that. It would be to check if the ExternalSource is consumed directly by MakeContiguous and SetExteranlSource with GPU data was used.

dali/c_api/c_api_test.cc
szalpal: Done

include/dali/c_api.h
szalpal: Introduced doxygen's member groups (`///@{`)

include/dali/c_api.h
szalpal: Introduced doxygen's member groups (`///@{`)

include/dali/c_api.h
szalpal: Introduced doxygen's member groups (`///@{`). Also, I changed the name to `daliSetExternalInputCudaStream`, cause with the previous one it sounded like the `input stream` would be set using this function.

dali/c_api/c_api.cc
szalpal: Done

dali/c_api/c_api.cc
szalpal: Done

dali/c_api/c_api.cc
szalpal: I thought it's our way, to include the header file corresponding to the cc file on top of the include list?

dali/pipeline/operator/builtin/external_source.h
szalpal: Fixed

dali/pipeline/operator/builtin/external_source.h
szalpal: Fixed

dali/pipeline/operator/builtin/external_source.h
JanuszL: I'm not sure if this is what we wants. I can imagine a GPU ExternalSource that accepts CPU input and the opposite. @klecki  what do you think?

szalpal: With @klecki we figured out 4 scenarios:
```
Lp.    DataLocation    OpDevice
1.          CPU          CPU
2.          GPU          GPU
3.          CPU          GPU
4.          GPU          CPU
```
ad 1. Legit
ad 2. Legit
ad 3. Legit - data will be copied onto GPU by ExternalSource.
ad 4. Nope. Since in DALI once you go GPU, you never go back, I feel that this should be illegal. So if you have GPU data, don't expect to work with it on CPU op.

Remaining question is, if right now the ExternalSource will work for the 3rd scenario, I can write a test for it

JanuszL: 3 - it currently does and we have a test for that https://github.com/NVIDIA/DALI/blob/master/dali/test/python/test_pipeline.py#L1215
4 - it is true for the pipeline, but data on the GPU is outside the pipeline so I think we should also consider supporting it for some post-processing after inference.

szalpal:
!build

dali/pipeline/operator/builtin/external_source.h
szalpal: About the `4` - I believe that it's what happens now: user specifies the location of data in function invocation, it get piped to `SetExternalInputHelper` and then `SetDataSource`, where it gets copied into proper memory. The downside is, that it's a little bit obfuscated around this `SetExternalInputHelper` (we have 2 different `Backends` - storage's and operator's). I've put proper comment there.

JanuszL: In the invocation, you specify where you want to get your data after the operator run. The internal storage should be the same as the input to make the copy as fast as possible as `SetDataSource` call is synchronous from the user point of view and we don't want it to last more than it is essential.

szalpal: Hmm, that's not what docs indicate:
> @param device Device of the supplied memory.

As I understood this, user specified, where the data exists and the info about where the data should be at the beginning of the pipeline is determined by operator's device in the pipeline. Am I missing something here?

klecki: Yes, @szalpal is correct. The `device` param does indicate what kind of memory user supplies to us, so we now if we copy from CPU or GPU.
It doesn't indicate where to copy the memory, as this is the task for the Pipeline definition and the setup of ExternalSource operator which is out of scope of C API.

As you noted, the GPU (DataLocation) -> CPU (OpDevcie) may hit the inefficient scenario. It also enables users to place the ExternalSource on CPU, supply it with GPU data, and use the outputs only in GPU operators. We currently don't have any sensible way of diagnosing that, and I would like to prohibit such inefficient scenario (additional GPU->CPU->GPU copies).

If the user wants to use GPU data in CPU, unless there is a valid use-case for that, I would make them do an explicit copy on their own. API shouldn't allow to easily do bad things.

JanuszL: I'm talking about Python API, as the C one should is still a secondary thing for us. So I want the Python side to be flexible and not broken in the first place.
So one thing is the device where the ExternalSource is located and the second what kind of data you supply. I think we should allow all combinations and when the data is on the GPU and ExternalSource is located on the CPU we should warn, but not break.

JanuszL: So extend the ExternalSource  to house CPU and GPU storage, add some holder which will tell what kind of data should be consumed in the next iteration - CPU or GPU (maybe you can extend `data_in_tl_` and store some enum instead of bool there). Then extend the implementation with the selection of proper place holder. I don't think it it would cost much work.

klecki: The same reasoning would apply for Python API, that GPU->CPU->GPU will be bad.

JanuszL: "GPU->CPU->GPU" is bad, but "GPU->CPU" maybe useful.

dali/c_api/c_api.cc
```diff
@@ -73,83 +131,99 @@ void daliPrefetchSeparate(daliPipelineHandle* pipe_handle,
...
-  pipeline->SetExternalInput(name, cdata);
+
+void daliSetExternalInput(daliPipelineHandle *pipe_handle, const char *name, device_type_t device,
+                          const void *data_ptr, dali_data_type_t data_type, const int64_t *shapes,
+                          int sample_dim, const char *layout_str) {
+  cudaStream_t stream;
+  CUDA_CALL(cudaStreamCreate(&stream));
+  daliSetExternalInputAsync(pipe_handle, name, device, data_ptr, data_type, shapes, sample_dim,
+                                 layout_str, stream);
+  CUDA_CALL(cudaStreamSynchronize(stream));
```
JanuszL, line 142: You should call `cudaStreamDestroy`, or CUDAStream.
I don't know if creating a stream for every call isn't to time consuming operation.

JanuszL: Maybe you should query DALI for a stream or create a static stream in this fucntion.

dali/c_api/c_api.cc
```diff
@@ -73,83 +131,99 @@ void daliPrefetchSeparate(daliPipelineHandle* pipe_handle,
...
+
+void daliSetExternalInputTensors(daliPipelineHandle *pipe_handle, const char *name,
+                                 device_type_t device, const void *const *data_ptr,
+                                 dali_data_type_t data_type, const int64_t *shapes,
+                                 int64_t sample_dim, const char *layout_str) {
+  cudaStream_t stream;
+  CUDA_CALL(cudaStreamCreate(&stream));
+  daliSetExternalInputTensorsAsync(pipe_handle, name, device, data_ptr, data_type, shapes,
+                                        sample_dim, layout_str, stream);
+  CUDA_CALL(cudaStreamSynchronize(stream));
```
JanuszL, line 173: As above.

szalpal: Changed to create stream per pipeline

dali/c_api/c_api.cc
szalpal: Changed to create stream per pipeline

dali/pipeline/operator/builtin/external_source.cu
```diff
@@ -27,41 +27,47 @@ struct ExternalSource<GPUBackend>::RecycleFunctor {
...
     cv_.wait(busy_lock, [&data = data_in_tl_]{return !data.empty();});
     auto is_data_in_tl = data_in_tl_.front();
     data_in_tl_.pop_front();
     DALI_ENFORCE(is_data_in_tl, "Cannot feed non-contiguous data to GPU op.");
     data = tl_data_.PopFront();
+    internal_copy_to_gpu = copy_to_gpu_events_.PopFront();
     cuda_event = cuda_events_.GetEmpty();
   }
 
+  cudaEventSynchronize(*internal_copy_to_gpu.front());
```
klecki, line 63: Why not something like `cudaStreamWaitEvent`? Do we need this busy wait here?

szalpal:
!build

dali/pipeline/operator/builtin/external_source.cu
szalpal: Done

szalpal:
!build

include/dali/c_api.h
```diff
@@ -16,221 +16,268 @@
...
+ *               provided stream.
+ */
+DLL_PUBLIC void
+daliSetExternalInputAsync(daliPipelineHandle *pipe_handle, const char *name,
+                          device_type_t device, const void *data_ptr,
+                          dali_data_type_t data_type, const int64_t *shapes,
+                          int sample_dim, const char *layout_str,
+                          cudaStream_t stream);
+
+DLL_PUBLIC void
```
JanuszL, line 116: I would put some remark that if the GPU memory is provided it need to be available on the GPU already as this variant uses internal, non default stream to copy it to DALI.

szalpal:
> Is not handled efficiently as in both cases we have H2D or D2H copy inside SetDataSource.

In these cases D2H and H2D has to happen somewhere... How do you know that how it's done right know is ineffective?

JanuszL:
> In these cases D2H and H2D has to happen somewhere... How do you know that how it's done right > know is ineffective?

I we can do this inside DALI thread that is good, if in the user one that is bad. `SetDataSource` should be as fast as possible.

dali/pipeline/operator/builtin/external_source.h
mzient: It looks OK now. The validation to (possibly) reject GPU->CPU copy should happen in Pipeline if anywhere.
I need to track the stream usage outside this function to spot potential issues at call site.

dali/c_api/c_api.cc
```diff
@@ -12,19 +12,77 @@
...
+  auto elem_sizeof = type_info.size();
+  for (int i = 0; i < pipeline->batch_size(); i++) {
+    // We cast away the const from data_ptr, as there is no other way of passing it to the
+    // Tensor as we must also set the shape and type metadata.
+    // The vector that we pass to pipeline is const.
+    data[i].ShareData(const_cast<void *>(data_ptr[i]), tl_shape[i].num_elements() * elem_sizeof);
+    data[i].Resize(tl_shape[i], type_info);
+    data[i].SetLayout(layout);
+  }
+  const auto &cdata = data;
```
mzient, line 80: Is it necessary?

include/dali/c_api.h
```diff
@@ -16,221 +16,268 @@
 #define DALI_C_API_H_
 
 #include <cuda_runtime_api.h>
-#include <inttypes.h>
+#include <cinttypes>
```
mzient, line 19:
```suggestion
#include <inttypes.h>
```
This is supposed to be C API!

mzient: Add a file `c_api_build_test.c` which would include c_api.h just to see if it compiles in C.

dali/pipeline/operator/builtin/external_source.h
```diff
@@ -39,13 +39,21 @@ struct CudaEventWrapper {
...
+
   ~CudaEventWrapper() {
     auto res = cudaEventDestroy(event);
     if (res != cudaSuccess) {
       DALI_ERROR("Fatal error:: " + to_string(res) + " in ~CudaEventWrapper()");
       std::terminate();
     }
   }
+
+  DEFAULT_COPY_MOVE_ASSIGN(CudaEventWrapper);
```
mzient, line 51: Remove this entire class.
Use `CUDAEvent` from dali/core instead.

JanuszL: It is a legacy code, created before CUDAEvent was avaialble.

include/dali/c_api.h
klecki: I wanted to do that, but we decided to hold our horses on reworking this to be actually a C api, not a "C-like" api.

dali/c_api/c_api.cc
szalpal: Removed

include/dali/c_api.h
szalpal: "Supposed" ;)
I agree and will do it but in a followup PR: right now there are `bool`s used in this API, so let's make it into the next one

This error fixed.

szalpal:
!build

szalpal:
!build

PR #1897, opened 2020-04-23
Add an ability to strip DALI whl binary from debug symbols

JanuszL:
- adds an ability to run strip debug symbols step for DALI wheel
- it is useful for Debug and RelWithDebInfo builds
- when used it will produce two wheels *_debug.whl and *.whl

#### Why we need this PR?
*Pick one, remove the rest*
- It adds an ability to strip DALI whl binary from debug symbols

 - What solution was applied:
     adds an ability to run strip debug symbols step for DALI wheel
     it is useful for Debug and RelWithDebInfo builds
     when used it will produce two wheels *_debug.whl and *.whl
     build system
     NA
 - Validation and testing:
     CI
     compilation guide updated

**JIRA TASK**: *[NA]*


klecki:
I wonder if it wouldn't be better to have a wheel + just separate debug symbols for it, you can create them by:
`strip --only-keep-debug lib_with_debug_symbols.so`

JanuszL:
> I wonder if it wouldn't be better to have a wheel + just separate debug symbols for it, you can create them by:
> `strip --only-keep-debug lib_with_debug_symbols.so`

You mean:
`objcopy --only-keep-debug "${tostripfile}" "${debugdir}/${debugfile}"`
and store them somewhere, right?

PR #1902, opened 2020-04-24
"Fix" (remove) Optical flow test bug

szalpal:
#### Why we need this PR?
*Pick one, remove the rest*
- It fixes a bug - test failing for OF

 - What solution was applied:
     There was a test for OF, that used pregenerated data to verify the correctness of API implementation. The problem with this test was, that it, well, used pregenerated data. Therefore, when driver got updated, result has changed (for better or for worse, hard to tell), pregenerated data got outdated and the test was failing. This test should be fixed not to use pregenerated data, or removed. I propose to remove this test. This behaviour is actually already tested in OF's Nosetest. To refactor the test, I'd need to mimic the testing approach from Nosetest, however IMHO it would be too much effort for the effect we achieve, without all the numpy's utilities. Alternatively, I'd need to figure out some other testing scenario, but again, I feel that Nosetest handles everything that should be handled

How do I know it was a problem with test, not with OF algorithm?
Because GTest failed, while Nosetest succeeded, and they basically test the same thing (Nosetest is even broader)

As the result, I removed the test. Also, there was one function, that might come in handy in future (copying `cv::Mat` onto GPU), I put it in `cv_mat_utils` file, where it belongs 
     NA
     NA
 - Validation and testing:
     L0 and L1
     NA

**JIRA TASK**: *[Use DALI-1394 or NA]*

### DIFF
--- a/dali/test/cv_mat_utils.h
+++ b/dali/test/cv_mat_utils.h
@@ -1,4 +1,4 @@
-// Copyright (c) 2018, NVIDIA CORPORATION. All rights reserved.
+// Copyright (c) 2018-2020, NVIDIA CORPORATION. All rights reserved.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -16,6 +16,8 @@
 #define DALI_TEST_CV_MAT_UTILS_H_
 
 #include <opencv2/opencv.hpp>
+#include <utility>
+#include <tuple>
 #include "dali/core/geom/box.h"
 #include "dali/test/mat2tensor.h"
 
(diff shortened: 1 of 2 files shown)


szalpal:
!build

dali/test/cv_mat_utils.h
```diff
@@ -46,6 +48,41 @@ cv::Mat_<cv::Vec<T, nchannels>> copy_to_mat(Box<2, int> roi, T *base, int rows,
   return out_copy;
 }
 
+
+/**
```
JanuszL, line 52: What is a rationale behind that change. You missed it in the PR description.

szalpal: Sorry, missed that. Updated the description:
>Also, there was one function, that might come in handy in future (copying cv::Mat onto GPU), I put it in cv_mat_utils file, where it belongs

PR #1903, opened 2020-04-24
Add Erase GPU Kernel

klecki:

#### Why we need this PR?
Add Erase GPU kernel implementation

 - What solution was applied:
     Kernel was implemented, there is a lengthy comment describing the approach. 
     There was also fix for CPU Kernel that appeared when I used it for tests.
     Dali Kernels
     GPU Kernel
 - Validation and testing:
     Validation using CPU kernel which was fixed for that purpose.
     Docstring with explanation.

**JIRA TASK**: *[Use DALI-1367]*

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -176,7 +176,8 @@ CUDA_get_gencode_args(CUDA_gencode_flags ${CUDA_targeted_archs})
 message(STATUS "Generated gencode flags: ${CUDA_gencode_flags}")
 
 # Add ptx & bin flags for cuda
-set(CMAKE_CUDA_FLAGS "${CMAKE_CUDA_FLAGS} ${CUDA_gencode_flags} --compiler-options -fvisibility=hidden")
+set(CMAKE_CUDA_FLAGS "${CMAKE_CUDA_FLAGS} ${CUDA_gencode_flags} --compiler-options -fvisibility=hidden -Xptxas -v")
+
 
 # Include directories
 include_directories(
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -176,7 +176,7 @@ CUDA_get_gencode_args(CUDA_gencode_flags ${CUDA_targeted_archs})
 message(STATUS "Generated gencode flags: ${CUDA_gencode_flags}")
 
 # Add ptx & bin flags for cuda
-set(CMAKE_CUDA_FLAGS "${CMAKE_CUDA_FLAGS} ${CUDA_gencode_flags} --compiler-options -fvisibility=hidden -Xptxas -v")
+set(CMAKE_CUDA_FLAGS "${CMAKE_CUDA_FLAGS} ${CUDA_gencode_flags} --compiler-options -fvisibility=hidden")
 
 
 # Include directories
--- a/dali/kernels/erase/erase_gpu.h
+++ b/dali/kernels/erase/erase_gpu.h
@@ -383,11 +383,10 @@ struct EraseGpu {
     const int num_samples = in.num_samples();
 
     if (channel_dim == -1) {
-      DALI_ENFORCE(
-          fill_values.size() == 1 || fill_values.size() == 0,
-          make_string(
-              "Kernel is unaware of channel dimension, exactly 0 or 1 fill value is expected, got: ",
-              fill_values.size(), "."));
+      DALI_ENFORCE(fill_values.size() == 1 || fill_values.size() == 0,
+                   make_string("Kernel is unaware of channel dimension, exactly 0 or 1 fill value "
+                               "is expected, got: ",
+                               fill_values.size(), "."));
     } else {
       // ensure that dim `channel_dim` in every sample matches fill_values.size();
       for (int i = 0; i < num_samples; i++) {
@@ -522,7 +521,6 @@ struct EraseGpu {
 
     erase_gpu_impl<channel_dim><<<grid_dim, block_dim, 0, stream>>>(
         sample_desc_gpu, region_dim, erase_regions_count, fill_values_span);
-
   }
 };
 
--- a/dali/kernels/erase/erase_gpu.h
+++ b/dali/kernels/erase/erase_gpu.h
@@ -17,15 +17,13 @@
 
 #include <vector>
 
-#include "dali/core/tensor_view.h"
-#include "dali/kernels/kernel.h"
-
-#include "dali/core/geom/vec.h"
-#include "dali/core/geom/box.h"
-#include "dali/kernels/common/utils.h"
-
-#include "dali/core/cuda_event.h"
+#include "dali/core/dev_array.h"
 #include "dali/core/format.h"
+#include "dali/core/geom/box.h"
+#include "dali/core/geom/vec.h"
+#include "dali/core/tensor_view.h"
+#include "dali/kernels/common/utils.h"
+#include "dali/kernels/kernel.h"
 
 namespace dali {
 namespace kernels {
@@ -78,18 +76,10 @@ struct erase_sample_desc {
 };
 
 template <int ndim>
-DALI_HOST_DEV ivec<ndim> get_region_cover(ivec<ndim> region_dims, ivec<ndim> sample_dims) {
+DALI_HOST_DEV ivec<ndim> get_region_start(int region_idx, ivec<ndim> region_shape,
+                                          ivec<ndim> sample_shape) {
   // let's calculate how many regions cover given axes:
-  auto region_cover_shape = (sample_dims + region_dims - 1) / region_dims;
-  return region_cover_shape;
-}
-
-// Hope that SROA is strong in this one
-template <int ndim>
-DALI_HOST_DEV ivec<ndim> get_region_start(int region_idx, ivec<ndim> region_dims,
-                                          ivec<ndim> sample_dims) {
-  // let's calculate how many regions cover given axes:
-  auto region_cover_shape = (sample_dims + region_dims - 1) / region_dims;
+  auto region_cover_shape = div_ceil(sample_shape, region_shape);
   // this introduces the same ordering of regions as data layout
   auto region_cover_strides = GetStrides(region_cover_shape);
   ivec<ndim> region_position{};  // n-dim index of region among other regions
@@ -98,13 +88,13 @@ DALI_HOST_DEV ivec<ndim> get_region_start(int region_idx, ivec<ndim> region_dims
     region_idx -= region_position[i - 1] * region_cover_strides[i - 1];
     region_position[i] = region_idx / region_cover_strides[i];
   }
-  return region_position * region_dims;
+  return region_position * region_shape;
 }
 
 template <int ndim>
-__device__ ibox<ndim> get_region(int region_idx, ivec<ndim> region_dims, ivec<ndim> sample_dims) {
-  auto anchor = get_region_start(region_idx, region_dims, sample_dims);
-  return {anchor, anchor + region_dims};
+__device__ ibox<ndim> get_region(int region_idx, ivec<ndim> region_shape, ivec<ndim> sample_shape) {
+  auto anchor = get_region_start(region_idx, region_shape, sample_shape);
+  return {anchor, anchor + region_shape};
 }
 
 constexpr unsigned FULL_MASK = 0xffffffffu;
@@ -127,7 +117,7 @@ constexpr bool select_outer_loops(int ndim, int current_dim, int channel_dim) {
  */
 template <typename Worker, int channel_dim, int current_dim = 0, typename T, int ndim>
 __device__ std::enable_if_t<select_outer_loops(ndim, current_dim, channel_dim)>
-erase_generic(erase_sample_desc<T, ndim> sample, ibox<ndim> region, span<T> fill_values = {},
+erase_generic(erase_sample_desc<T, ndim> sample, ibox<ndim> region, span<T> fill_values,
               span<ibox<ndim>> erase_regions = {}, ivec<ndim> coordinate = {},
               int64_t offset_base = 0) {
   constexpr int d = current_dim;
@@ -150,7 +140,7 @@ erase_generic(erase_sample_desc<T, ndim> sample, ibox<ndim> region, span<T> fill
  */
 template <typename Worker, int channel_dim, int current_dim = 0, typename T, int ndim>
 __device__ std::enable_if_t<ndim - current_dim == 2 && channel_dim < ndim - 2>
-erase_generic(erase_sample_desc<T, ndim> sample, ibox<ndim> region, span<T> fill_values = {},
+erase_generic(erase_sample_desc<T, ndim> sample, ibox<ndim> region, span<T> fill_values,
               span<ibox<ndim>> erase_regions = {}, ivec<ndim> coordinate = {},
               int64_t offset_base = 0) {
   constexpr int d = current_dim;
@@ -185,7 +175,7 @@ erase_generic(erase_sample_desc<T, ndim> sample, ibox<ndim> region, span<T> fill
  */
 template <typename Worker, int channel_dim, int current_dim = 0, typename T, int ndim>
 __device__ std::enable_if_t<ndim - current_dim == 3 && channel_dim >= ndim - 2>
-erase_generic(erase_sample_desc<T, ndim> sample, ibox<ndim> region, span<T> fill_values = {},
+erase_generic(erase_sample_desc<T, ndim> sample, ibox<ndim> region, span<T> fill_values,
               span<ibox<ndim>> erase_regions = {}, ivec<ndim> coordinate = {},
               int64_t offset_base = 0) {
   constexpr int d = current_dim;
@@ -196,17 +186,26 @@ erase_generic(erase_sample_desc<T, ndim> sample, ibox<ndim> region, span<T> fill
   int boundary_x = ::min(region.hi[dX], sample.sample_shape[dX]);
   int boundary_c = ::min(region.hi[dC], sample.sample_shape[dC]);
 
+  int flattened_start = region.lo[dX] * sample.sample_stride[dX] + region.lo[dC] + threadIdx.x;
+  int flattened_end = region.hi[dX] * sample.sample_stride[dX] + region.hi[dC];
+  int start_x = flattened_start / sample.sample_stride[dX];
+  int start_c = flattened_start % sample.sample_stride[dX];
+  int step_x = blockDim.x / sample.sample_stride[dX];
+  int step_c = blockDim.x % sample.sample_stride[dX];
   for (int y = region.lo[dY] + threadIdx.y; y < boundary_y; y += blockDim.y) {
     coordinate[dY] = y;
     int64_t offset_y = offset_base + y * sample.sample_stride[dY];
-    int flattened_start = region.lo[dX] * sample.sample_stride[dX] + region.lo[dC] + threadIdx.x;
-    int flattened_end = region.hi[dX] * sample.sample_stride[dX] + region.hi[dC];
-    for (int x = flattened_start; x < flattened_end; x += blockDim.x) {
-      coordinate[dX] = x / sample.sample_stride[dX];
-      coordinate[dC] = x % sample.sample_stride[dX];
+    for (int x = start_x, c = start_c, flat = flattened_start; flat < flattened_end;
+         x += step_x, c += step_c, flat += blockDim.x) {
+      if (c >= sample.sample_stride[dX]) {
+        x += 1;
+        c -= sample.sample_stride[dX];
+      }
+      coordinate[dX] = x;
+      coordinate[dC] = c;
       if (region.lo[dX] <= coordinate[dX] && coordinate[dX] < boundary_x &&
           region.lo[dC] <= coordinate[dC] && coordinate[dC] < boundary_c) {
-        int64_t offset_x = offset_y + x;
+        int64_t offset_x = offset_y + flat;
         Worker::template copy_or_erase<channel_dim>(sample, erase_regions, coordinate,
                                                              offset_x, fill_values);
       }
@@ -219,7 +218,7 @@ erase_generic(erase_sample_desc<T, ndim> sample, ibox<ndim> region, span<T> fill
  */
 template <typename Worker, int channel_dim, int current_dim = 0, typename T, int ndim>
 __device__ std::enable_if_t<ndim == 1 && current_dim == 0> erase_generic(
-    erase_sample_desc<T, ndim> sample, ibox<ndim> region, span<T> fill_values = {},
+    erase_sample_desc<T, ndim> sample, ibox<ndim> region, span<T> fill_values,
     span<ibox<ndim>> erase_regions = {}, ivec<ndim> coordinate = {}, int64_t offset_base = 0) {
   constexpr int d = current_dim;
   int boundary = ::min(region.hi[d], sample.sample_shape[d]);
@@ -239,7 +238,7 @@ __device__ std::enable_if_t<ndim == 1 && current_dim == 0> erase_generic(
  */
 template <typename Worker, int channel_dim, int current_dim = 0, typename T, int ndim>
 __device__ std::enable_if_t<(ndim == 2 && current_dim == 0 && channel_dim >= 0)> erase_generic(
-    erase_sample_desc<T, ndim> sample, ibox<ndim> region, span<T> fill_values = {},
+    erase_sample_desc<T, ndim> sample, ibox<ndim> region, span<T> fill_values,
     span<ibox<ndim>> erase_regions = {}, ivec<ndim> coordinate = {}, int64_t offset_base = 0) {
   constexpr int d = current_dim;
   constexpr int dC = d + 1;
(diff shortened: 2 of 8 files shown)


dali/kernels/erase/erase_gpu.h
```diff
@@ -0,0 +1,526 @@
...
+struct erase_sample_desc {
+  const T *in = nullptr;
+  T* out = nullptr;
+  const ibox<ndim> *erase_regions = nullptr;
+  ivec<ndim> sample_shape;
+  ivec<ndim> sample_stride;
+};
+
+template <int ndim>
+DALI_HOST_DEV ivec<ndim> get_region_cover(ivec<ndim> region_dims, ivec<ndim> sample_dims) {
```
mzient, line 81: This is confusing... we use `dims` more as axes than actual values.
```suggestion
DALI_HOST_DEV ivec<ndim> get_region_cover(ivec<ndim> region_sizes, ivec<ndim> sample_sizes) {
```
?

dali/kernels/erase/erase_gpu.h
```diff
@@ -0,0 +1,526 @@
...
+  // let's calculate how many regions cover given axes:
```
mzient, line 82: Does it calculate how many regions we must have, in each axis, to cover the sample? If so, i think the function name is not very informative - in fact, for me the line that actually implements the function is much more self-explanatory than the cryptic name. I'd suggest something with tiles... or maybe just overload `div_ceil` to handle vectors.

dali/kernels/erase/erase_gpu.h
```diff
@@ -0,0 +1,526 @@
...
+    // CUDA_CALL(cudaGetLastError());
+    // float t = 0;
+    // cudaEventElapsedTime(&t, start, end);
+    // int64_t read = in.num_elements() * sizeof(T)
+    //                + erased_regions.num_elements() * sizeof(ibox<ndim>);
+    // int64_t written = out.num_elements() * sizeof(T);
+    // int64_t total_transferred = read + written;
+
+    // std::cerr << "Transferred " << total_transferred << " in " << t << "[ms] with "
+    //           << total_transferred / t * 1e-6 << " GB/s" << endl;
```
mzient, line 519: remove?

dali/kernels/erase/erase_gpu_test.cu
```diff
@@ -0,0 +1,390 @@
...
+          *regions_tv(j) = region_box;
+        }
+      }
+    }
+  }
+
+  int num_erase_regions_;
+  int region_generation_;
+  bool full_fill_;
+  std::vector<T> fill_values;
```
JanuszL, line 245:
```suggestion
  std::vector<T> fill_values_;
```

JanuszL: You should do this for `test_shape`, `batch_size` , input`, `output`, `baseline` and `regions` as well.

dali/kernels/erase/erase_gpu_test.cu
```diff
@@ -0,0 +1,390 @@
...
+
+    auto out_view = output.gpu();
+
+     // AND THIS IS NO BETTER:
+    ScratchpadAllocator scratch_alloc;
+    scratch_alloc.Reserve(req.scratch_sizes);
+    auto scratchpad = scratch_alloc.GetScratchpad();
+    ctx.scratchpad = &scratchpad;
+
+    // auto start = CUDAEvent::CreateWithFlags(0);
```
JanuszL, line 152: ?

dali/kernels/erase/erase_gpu_test.cu
```diff
@@ -0,0 +1,390 @@
...
+    // auto end =   CUDAEvent::CreateWithFlags(0);
+    // cudaEventRecord(start);
+
+    kernel.Run(ctx, out_view, in_view, regions_gpu, fill_values);
+
+    // cudaEventRecord(end);
+    CUDA_CALL(cudaDeviceSynchronize());
+    CUDA_CALL(cudaGetLastError());
+
+    // float t = 0;
```
JanuszL, line 162: Maybe print this in debug or always (as we do in some reduce tests)?

dali/kernels/erase/erase_gpu.h
```diff
@@ -0,0 +1,526 @@
...
+  auto region_cover_shape = (sample_dims + region_dims - 1) / region_dims;
+  return region_cover_shape;
+}
+
+// Hope that SROA is strong in this one
+template <int ndim>
+DALI_HOST_DEV ivec<ndim> get_region_start(int region_idx, ivec<ndim> region_dims,
+                                          ivec<ndim> sample_dims) {
+  // let's calculate how many regions cover given axes:
+  auto region_cover_shape = (sample_dims + region_dims - 1) / region_dims;
```
JanuszL, line 92: Why not just call `get_region_cover`?

dali/kernels/erase/erase_gpu.h
```diff
@@ -0,0 +1,526 @@
...
+    memcpy(fill_values_cpu, fill_values.data(), num_fill_values * sizeof(T));
+
+    CUDA_CALL(
+      cudaMemcpyAsync(sample_desc_gpu, sample_desc_cpu, num_samples * sizeof(sample_t),
+                      cudaMemcpyHostToDevice, stream));
+    CUDA_CALL(
+      cudaMemcpyAsync(fill_values_gpu, fill_values_cpu, num_fill_values * sizeof(T),
+                      cudaMemcpyHostToDevice, stream));
+
+    // auto start = CUDAEvent::CreateWithFlags(0);
```
JanuszL, line 500: ?

dali/kernels/erase/erase_gpu.h
```diff
@@ -0,0 +1,526 @@
...
+    // auto end =   CUDAEvent::CreateWithFlags(0);
+    // cudaEventRecord(start);
+
+    erase_gpu_impl<channel_dim><<<grid_dim, block_dim, 0, stream>>>(sample_desc_gpu, region_dim,
+                                                       erase_regions_count,
+                                                       make_span(fill_values_gpu, num_fill_values));
+
+    // cudaEventRecord(end);
```
JanuszL, line 508: ?

dali/kernels/erase/erase_gpu.h
```diff
@@ -0,0 +1,526 @@
...
+        full_cover = true;
+        break;
+      }
+    }
+  }
+  __syncthreads();
+
+  //  Check if we have full or no cover at all
+  int total_erase_regions = filtered_region_idx;
+  __syncthreads();
```
JanuszL, line 350: Why we need this additional sync here?

dali/kernels/erase/erase_gpu.h
klecki: Yes, I intended to remove it, but I don't know if we want something like this in test/under some debug conditions etc.
If we put it around DALI kernel the result is not measuring the CUDA kernel call only.

dali/kernels/erase/erase_gpu.h
```diff
@@ -0,0 +1,526 @@
...
+      sample_desc_cpu[i].erase_regions = erased_regions.data[i];
+      for (int dim = 0; dim < ndim; dim++) {
+        sample_desc_cpu[i].sample_shape[dim] = in.tensor_shape_span(i)[dim];
+      }
+      sample_desc_cpu[i].sample_stride = GetStrides(sample_desc_cpu[i].sample_shape);
+    }
+    memcpy(fill_values_cpu, fill_values.data(), num_fill_values * sizeof(T));
+
+    CUDA_CALL(
+      cudaMemcpyAsync(sample_desc_gpu, sample_desc_cpu, num_samples * sizeof(sample_t),
```
jantonguirao, line 494: you can get away with a single `cudaMemcpyAsync`, since the memory from the scratchpad is contiguous

dali/kernels/erase/erase_gpu.h
```diff
@@ -0,0 +1,526 @@
...
+    }
+    dim3 grid_dim = {(uint32_t)max_regions, (uint32_t)num_samples, 1};
+    dim3 block_dim = {32, 32, 1};  // fixed block dim
+    auto* sample_desc_cpu = ctx.scratchpad->Allocate<sample_t>(AllocType::Host, num_samples);
+    auto* fill_values_cpu = ctx.scratchpad->Allocate<T>(AllocType::Host, num_fill_values);
+    auto* sample_desc_gpu = ctx.scratchpad->Allocate<sample_t>(AllocType::GPU, num_samples);
+    auto* fill_values_gpu = ctx.scratchpad->Allocate<T>(AllocType::GPU, num_fill_values);
+
+    for (int i = 0; i < num_samples; i++) {
+      sample_desc_cpu[i].in = in.data[i];
```
jantonguirao, line 483: nitpick,
```
auto &sample = sample_desc_cpu[i]
```
then reuse

dali/kernels/erase/erase_gpu.h
```diff
@@ -0,0 +1,526 @@
...
+  static_assert(
+      -1 <= channel_dim && channel_dim < ndim,
+      "Channel dim can either be ignored (for channel_dim = -1) or must be in [0, ndim) interval.");
+  using sample_t = erase_sample_desc<T, ndim>;
+
+  KernelRequirements Setup(KernelContext &context,
+                          //  const TensorListView<StorageGPU, const T, ndim> &in
+                           const InListGPU<T, ndim> &in,
+                           const InListGPU<ibox<ndim>, 1> &erased_regions,
+                           const std::vector<T> &fill_values = {0}) {
```
jantonguirao, line 388: span maybe?

dali/kernels/erase/erase_gpu.h
```diff
@@ -0,0 +1,526 @@
...
+          fill_values.size() == 1,
+          make_string(
+              "Kernel is unaware of channel dimension, exactly 1 fill value is expected, got: ",
+              fill_values.size(), "."));
+    } else {
+      // ensure that dim `channel_dim` in every sample matches fill_values.size();
+      for (int i = 0; i < num_samples; i++) {
+        DALI_ENFORCE(in.shape.tensor_shape_span(i)[channel_dim] == fill_values.size() ||
+                         fill_values.size() == 1,
+                     make_string("Channel dimension in all samples must match the number of number "
```
jantonguirao, line 402: the number of number of elements ? :)

dali/kernels/erase/erase_gpu.h
```diff
@@ -0,0 +1,526 @@
...
+ * Possible optimization/obfuscation - factor out checks over dimension to specific level of
+ * the loop nest (for the mixed cover case).
+ *
+ */
+
+template <typename T, int ndim>
+using box = Box<ndim, T>;
+
+template <int ndim>
+using ibox = box<int32_t, ndim>;
```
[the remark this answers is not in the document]
mzient: because ivec?

dali/kernels/erase/erase_gpu.h
```diff
@@ -0,0 +1,526 @@
...
+/**
+ * @brief Outer layer of recursive loop over current region
+ */
+template <typename Worker, int channel_dim, int current_dim = 0, typename T, int ndim>
+__device__ std::enable_if_t<select_outer_loops(ndim, current_dim, channel_dim)>
+erase_generic(erase_sample_desc<T, ndim> sample, ibox<ndim> region, span<T> fill_values = {},
+              span<ibox<ndim>> erase_regions = {}, ivec<ndim> coordinate = {},
+              int64_t offset_base = 0) {
+  constexpr int d = current_dim;
+  int boundary = ::min(region.hi[d], sample.sample_shape[d]);
```
mzient, line 134: I think you should trim both sides, not just `hi`.

dali/kernels/erase/erase_gpu.h
```diff
@@ -0,0 +1,526 @@
...
+  const auto region_idx = blockIdx.x;
+  const auto sample_idx = blockIdx.y;
+
+  const auto &sample = samples[sample_idx];
+  ivec<ndim> sample_dims = sample.sample_shape;
+
+  auto region_box = get_region(region_idx, region_dims, sample_dims);
+
+  const auto *erase_regions_ptrs = sample.erase_regions;
+  constexpr int max_regions = 1024;
```
mzient, line 327: This could be calculated as, say, `40kB / sizeof(ibox<ndim>)`

dali/kernels/erase/erase_gpu.h
```diff
@@ -0,0 +1,526 @@
...
+    bool copy = true;
+    for (auto &region : erase_regions) {
+      if (region.contains(coordinate)) {
+        copy = false;
+      }
+    }
+    if (copy) {
+      sample.out[offset] = sample.in[offset];
+    } else {
+      sample.out[offset] = fill_value;
```
mzient, line 310: I don't know how smart NVCC is, but it'd say that this:
```c++
sample.out[offset] = copy ? sample.in[offset] : fill_value;
```
has fewer opportunities for the compiler to screw up optimization and write coalescing.

dali/kernels/erase/erase_gpu.h
mzient: either that, or `InTensorCPU<T, 1>`

mzient: On top of that - I think we can safely say that the requirement of passing everything to Setup _and_ Run was proven unnecessary - so you can just leave out if it's not used before Run.

dali/kernels/erase/erase_gpu.h
mzient: Scratchpad::ToGPU to the rescue!

dali/kernels/erase/erase_gpu.h
```diff
@@ -0,0 +1,526 @@
...
+    CUDA_CALL(
+      cudaMemcpyAsync(fill_values_gpu, fill_values_cpu, num_fill_values * sizeof(T),
+                      cudaMemcpyHostToDevice, stream));
+
+    // auto start = CUDAEvent::CreateWithFlags(0);
+    // auto end =   CUDAEvent::CreateWithFlags(0);
+    // cudaEventRecord(start);
+
+    erase_gpu_impl<channel_dim><<<grid_dim, block_dim, 0, stream>>>(sample_desc_gpu, region_dim,
+                                                       erase_regions_count,
```
mzient, line 505: nitpick: weird (and large) indent - perhaps break at `(` and use just a few indents for arguments?

dali/kernels/erase/erase_gpu.h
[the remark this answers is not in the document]
klecki: yeah, for the consistent look & feel.

dali/kernels/erase/erase_gpu_test.cu
klecki: Done

dali/kernels/erase/erase_gpu.h
klecki: I'd rather go with shape, size feels like 1D thing.

CMakeLists.txt
```diff
@@ -176,7 +176,8 @@ CUDA_get_gencode_args(CUDA_gencode_flags ${CUDA_targeted_archs})
 message(STATUS "Generated gencode flags: ${CUDA_gencode_flags}")
 
 # Add ptx & bin flags for cuda
-set(CMAKE_CUDA_FLAGS "${CMAKE_CUDA_FLAGS} ${CUDA_gencode_flags} --compiler-options -fvisibility=hidden")
+set(CMAKE_CUDA_FLAGS "${CMAKE_CUDA_FLAGS} ${CUDA_gencode_flags} --compiler-options -fvisibility=hidden -Xptxas -v")
```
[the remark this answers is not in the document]
klecki: I will post this as a separate PR, I wanted to check register usage in the compiled kernels

dali/kernels/erase/erase_gpu.h
klecki: After accessing the shared memory, but it's probably not needed here.
@mzient?

JanuszL: It is read so I think ti should be safe.

dali/kernels/erase/erase_gpu.h
klecki: Done

dali/kernels/erase/erase_gpu.h
klecki: Replaced by div_ceil everywhere.

dali/kernels/erase/erase_gpu_test.cu
```diff
@@ -0,0 +1,390 @@
...
+  {
+    ivec<4> region_dims = {2, 2, 32, 32};
+    ivec<4> sample_dims = {16, 8, 32, 32};
+    ivec<4> expected_cover = {8, 4, 1, 1};
+    verify_regions(region_dims, sample_dims, expected_cover);
+  }
+}
+
+template <typename T, int ndim, int channel_dim = -1>
+struct EraseGpuKernelTest :
```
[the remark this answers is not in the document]
klecki: That's just inverts the workaround, instead of abusing the value-parametrized test and instantiating it several times with different types for appropriate sets of values, we would abuse the type-parametrized type instances and would need to loop over sets of values inside of them.

Not much of a difference, but I would need to rework a lot of this boilerplate and I don't think it's worth the effort.

dali/kernels/erase/erase_gpu_test.cu
```diff
@@ -0,0 +1,390 @@
...
+    public testing::TestWithParam<std::tuple<
+      int,  // num erase regions
+      int,  // region generation scheme
+      int,  // default fill or full channel fill
+      TensorShape<ndim>>> {  // shape of input
```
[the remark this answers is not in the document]
klecki: Done. I got those tuples ingrained in me from writing OpenCV tests.

dali/kernels/erase/erase_gpu.h
klecki: As you can see, I'm doing error checking in setup based on the size of fill_values.

klecki: And I don't think there is a nice way of providing default value in case of span, so unless someone suggests something, I would opt for vector. We cannot create span from rvalue, the view also wouldn't work without some additional backing memory unless I'm wrong.

dali/kernels/erase/erase_gpu.h
klecki: Done

dali/kernels/erase/erase_gpu.h
klecki: Looking how well it handles vec & Box, I would say it's quite good.

dali/kernels/erase/erase_gpu.h
klecki: Done. Some leftover probably.

dali/kernels/erase/erase_gpu.h
klecki: Done.
I wonder how we can do an early check if we won't go over that. It would depend on how are the regions spread out.

dali/kernels/erase/erase_gpu.h
klecki: done

dali/kernels/erase/erase_gpu.h
klecki: Didn't we have a magic one that was supposed to do that in one go?
Still I think the cudaMemcpy is a bit clearer than ToGPU that's a bit cryptic and doesn't have any docstring. 
Similar thing to the div_ceil vs get_region_cover above

dali/kernels/erase/erase_gpu.h
klecki: Done by specialcasing empty span

dali/kernels/erase/erase_gpu_test.cu
```diff
@@ -0,0 +1,390 @@
...
+
+INSTANTIATE_ERASE_SUITE(EraseGpuKernel1fTest, values_1);
+INSTANTIATE_ERASE_SUITE(EraseGpuKernel2fTest, values_2);
+INSTANTIATE_ERASE_SUITE(EraseGpuKernel2NCfTest, values_2NC);
+INSTANTIATE_ERASE_SUITE(EraseGpuKernel3fTest, values_3);
+INSTANTIATE_ERASE_SUITE(EraseGpuKernel3fHWCTest, values_3HWC);
+INSTANTIATE_ERASE_SUITE(EraseGpuKernel3fCHWTest, values_3CHW);
+INSTANTIATE_ERASE_SUITE(EraseGpuKernel4fDHCWTest, values_4DHCW);
+INSTANTIATE_ERASE_SUITE(EraseGpuKernel4fDHWCTest, values_4DHWC);
+INSTANTIATE_ERASE_SUITE(EraseGpuKernel5fTest, values_5);
```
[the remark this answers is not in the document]
klecki: Kinda done

dali/kernels/erase/erase_gpu.h
klecki: Removed

dali/kernels/erase/erase_gpu_test.cu
klecki: No strong desire to keep it, so removed everywhere.

klecki:
!build

dali/kernels/erase/erase_gpu.h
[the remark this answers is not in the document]
szalpal: So maybe `Box` should be changed?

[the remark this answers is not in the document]
klecki: We can change it, but maybe not in this PR

dali/kernels/erase/erase_gpu.h
klecki: Done regardless

dali/kernels/erase/erase_gpu.h
JanuszL: I 👍 for ToGPU, if we have it we should use it and be consistent. Or remove `ToGPU` at all.

dali/kernels/erase/erase_gpu_test.cu
```diff
@@ -0,0 +1,399 @@
...
+    ivec<4> expected_cover = {8, 4, 1, 1};
+    verify_regions(region_shape, sample_shape, expected_cover);
+  }
+}
+
+template <int ndim>
+struct EraseTestParams {
+  int num_erase_regions;
+  int region_generation;
+  int fill_type;
```
mzient, line 95: Shouldn't those two be enums? Or at least you could compare them against enum values instead of explaining the meaning in comments.

dali/kernels/erase/erase_gpu.h
```diff
@@ -0,0 +1,530 @@
...
+/**
+ * We have 2 full most inner loops, channel_dim is somewhere on the way here, or doesn't exist
+ *
+ * We handle this by covering the innermost dimensions by:
+ * for (y : blockDim.y)
+ *   for (x : blockDim.x)
+ *     process {coordinate[0], coordinate[1], ..., coordinate[current_dim - 1], y, x}
+ */
+template <typename Worker, int channel_dim, int current_dim = 0, typename T, int ndim>
+__device__ std::enable_if_t<ndim - current_dim == 2 && channel_dim < ndim - 2>
```
mzient, line 142:
```suggestion
__device__ std::enable_if_t<ndim - current_dim == 2 && (channel_dim < ndim - 2)>
```

dali/kernels/erase/erase_gpu.h
```diff
@@ -0,0 +1,530 @@
...
+      }
+    }
+  }
+}
+
+/**
+ * @brief Edge case kernel, 1D case
+ */
+template <typename Worker, int channel_dim, int current_dim = 0, typename T, int ndim>
+__device__ std::enable_if_t<ndim == 1 && current_dim == 0> erase_generic(
```
mzient, line 220: Is it possible that ndim == 1 and current_dim != 0? If not, then 
```suggestion
__device__ std::enable_if_t<ndim == 1> erase_generic(
```

dali/kernels/erase/erase_gpu.h
```diff
@@ -0,0 +1,530 @@
...
+  auto anchor = get_region_start(region_idx, region_shape, sample_shape);
+  return {anchor, anchor + region_shape};
+}
+
+constexpr unsigned FULL_MASK = 0xffffffffu;
+
+/**
+ * @brief Returns true if that parameter configuration is just outer layer of recursive loop call
+ */
+constexpr bool select_outer_loops(int ndim, int current_dim, int channel_dim) {
```
mzient, line 105: Nitpick (?): the name doesn't explain much... I think that something with recurse in the name would be more appropriate - after all, don't you always loop over the outer dimension?

dali/kernels/erase/erase_gpu.h
klecki: Why this and not 
```suggestion
__device__ std::enable_if_t<(ndim - current_dim == 2) && (channel_dim < ndim - 2)>
```
or
```suggestion
__device__ std::enable_if_t<(ndim - current_dim == 2 && channel_dim < ndim - 2)>
```

dali/kernels/erase/erase_gpu.h
klecki: How about
```suggestion
constexpr bool is_outer_loop(int ndim, int current_dim, int channel_dim) {
```

dali/kernels/erase/erase_gpu.h
klecki: Yes, you're right here.

PR #1911, opened 2020-04-27
Fix support for dynamic per-sample shape in Warp operators

mzient:
- fix detection of the presence of size argument
- suppport multiple input types for size argument.

#### Why we need this PR?
*Pick one, remove the rest*
- It fixes a bug: invalid handling of tensor `size` argument

 - What solution was applied:
     * Use `ArgumentDefined` instead of `HasArgument` to detect ArgumentInput
     * Use TYPE_SWITCH to support multiple input types for `size`
     * WarpAffine, Rotate
     * N/A
 - Validation and testing:
     * Python test
     * N/A

**JIRA TASK**: N/A

### DIFF
--- a/dali/operators/image/remap/warp_param_provider.h
+++ b/dali/operators/image/remap/warp_param_provider.h
@@ -19,6 +19,7 @@
 #include <vector>
 #include <string>
 
+#include "dali/core/static_switch.h"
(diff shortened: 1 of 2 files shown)


mzient:
!build

Issue #966, opened 2019-06-12
Illegal Memory access


qorbanpour:
When in my  dali_tf.DALIIterator() I set  device_id = d ( d is the device_id )  it throws:
[/opt/dali/dali/util/cuda_utils.h:69] CUDA runtime api error "an illegal memory access was encountered" . it only works if   device_id is set to = 0 .
I have two gpus on my system and another system with 8 gpus same issue .

     for d in  range(DEVICES):
        with tf.device('/gpu:%i' % d):
            image = daliop(serialized_pipeline = serialized_pipes[d], 
                shapes = [(batch_size  ,sequence_length , 3 , 224, 224 )],
                dtypes = [tf.int32],   
                device_id=0 )#setting device_id to d throws illegal memory access. only 0 works.

My whole code :

    class VideoReaderPipeline(Pipeline):
    def __init__(self, batch_size, sequence_length, num_threads, device_id, files, crop_size,num_gpus):
        super(VideoReaderPipeline, self).__init__(batch_size, num_threads, device_id, seed=12)
        self.reader = ops.VideoReader(device="gpu", filenames=files, sequence_length=sequence_length, normalized=False,
                                     random_shuffle=False, image_type=types.RGB, dtype=types.UINT8, initial_fill=16, shard_id = device_id, num_shards = num_gpus)

        self.crop = ops.CropCastPermute(device="gpu", crop=crop_size, output_layout=types.NHWC, output_dtype=types.FLOAT)
        self.uniform = ops.Uniform(range=(0.0, 1.0))
        self.transpose = ops.Transpose(device="gpu", perm=[ 0 ,3 , 1, 2])

    def define_graph(self):
        input = self.reader(name="Reader")
        cropped = self.crop(input, crop_pos_x=self.uniform(), crop_pos_y=self.uniform())
        output = self.transpose(cropped)
        return output

    def get_batch_test_dali(args, ds_type):
    batch_size = args.batchsize
    file_root = '/home/dl/base-app/DALI_old/docs/examples/video/superres_pytorch/data_dir/720p/scenes/val'
    sequence_length = 2#args.frames,
    crop_size = args.crop_size
    DEVICES= args.DEVICES

    container_files = os.listdir(file_root)
    container_files = [file_root + '/' + f for f in container_files]
    pipelines = [VideoReaderPipeline(batch_size=batch_size,
                                        sequence_length=sequence_length,
                                        num_threads=2,
                                        device_id= device_id,
                                        files=container_files,
[...]

PR #1914, opened 2020-04-28
Fix jitter operator illegal memory access

JanuszL:
- JitterAugment accesses random generator on the GPU beyond allowed range
- added a fix to make sure that access to random generator is inside allowed range

#### Why we need this PR?
*Pick one, remove the rest*
- It fixes a bug of jitter operator illegal memory access

 - What solution was applied:
     added a fix to make sure that access to random generator is inside allowed range
     jitter
     NA
 - Validation and testing:
     CI
     NA

In response to https://github.com/NVIDIA/DALI/issues/966

**JIRA TASK**: *[NA]*

### DIFF
--- a/dali/operators/util/randomizer_impl_gpu.cuh
+++ b/dali/operators/util/randomizer_impl_gpu.cuh
@@ -17,6 +17,7 @@
 
 #include "dali/operators/util/randomizer.h"
 
+#include <math.h>
 #include <curand.h>
 #include <curand_kernel.h>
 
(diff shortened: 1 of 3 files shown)


dali/operators/util/randomizer_impl_gpu.cuh
```diff
@@ -38,7 +39,8 @@ Randomizer<GPUBackend>::Randomizer(int seed, size_t len) {
   len_ = len;
   cudaGetDevice(&device_);
   states_ = GPUBackend::New(sizeof(curandState) * len, true);
-  initializeStates<<<128, 256>>>(len_, seed, reinterpret_cast<curandState*>(states_));
+  initializeStates<<<std::ceil(len_/static_cast<float>(block_size_)), block_size_>>>
```
mzient, line 42:
```suggestion
  initializeStates<<<div_ceil(len_, block_size_), block_size_>>>
```
using floating point arithmetic may lead to incorrect results when len_ > 256*2^23

JanuszL: done

PR #1919, opened 2020-04-29
Mean and Standard Deviation GPU kernels

mzient:
#### Why we need this PR?
*Pick one, remove the rest*
- It adds new feature needed because we want to have (directional) normalization.

 - What solution was applied:
     * Added classes which implement pre- and post-processors
     * Added pre- and post-processing for mean, variance, root mean square and inverse, regularized root mean square
     * Added pImpl facades and instantiations thereof.
     * ReduceGPU
     * the contents of `mean_stddev_gpu_impl.cuh`
 - Validation and testing:
     * Unit tests
     * Comments, Doxygen

**JIRA TASK**: DALI-1250

### DIFF
--- a/dali/kernels/reduce/mean_gpu.cu
+++ b/dali/kernels/reduce/mean_gpu.cu
@@ -40,7 +40,8 @@ KernelRequirements MeanGPU<Out, In>::Setup(
 }
 
 template <typename Out, typename In>
-void MeanGPU<Out, In>::Run(KernelContext &ctx, const OutListGPU<Out> &out, const InListGPU<In> &in) {
+void MeanGPU<Out, In>::Run(
+    KernelContext &ctx, const OutListGPU<Out> &out, const InListGPU<In> &in) {
   assert(impl_ != nullptr);
   impl_->Run(ctx, out, in);
 }
@@ -82,7 +83,8 @@ KernelRequirements RootMeanSquareGPU<Out, In>::Setup(
 }
 
 template <typename Out, typename In>
-void RootMeanSquareGPU<Out, In>::Run(KernelContext &ctx, const OutListGPU<Out> &out, const InListGPU<In> &in) {
+void RootMeanSquareGPU<Out, In>::Run(
+    KernelContext &ctx, const OutListGPU<Out> &out, const InListGPU<In> &in) {
   assert(impl_ != nullptr);
   impl_->Run(ctx, out, in);
 }
--- a/dali/kernels/reduce/mean_gpu.cu
+++ b/dali/kernels/reduce/mean_gpu.cu
@@ -1,0 +1,105 @@
+// Copyright (c) 2020, NVIDIA CORPORATION. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include <memory>
+#include "dali/kernels/reduce/reduce_gpu.h"
+#include "dali/kernels/reduce/mean_stddev_gpu_impl.cuh"
+
+namespace dali {
+namespace kernels {
+
+template <typename Out, typename In>
+class MeanGPU<Out, In>::Impl : public reduce_impl::MeanImplGPU<Out, In> {
+};
+
+template <typename Out, typename In>
+MeanGPU<Out, In>::MeanGPU() = default;
+
+template <typename Out, typename In>
+MeanGPU<Out, In>::~MeanGPU() = default;
+
+template <typename Out, typename In>
+KernelRequirements MeanGPU<Out, In>::Setup(
+    KernelContext &ctx,
+    const TensorListShape<> &in_shape, span<const int> axes, bool keep_dims, bool reduce_batch) {
+  if (!impl_) {
+    impl_ = std::make_unique<Impl>();
+  }
+  return impl_->Setup(ctx, in_shape, axes, keep_dims, reduce_batch);
+}
+
+template <typename Out, typename In>
+void MeanGPU<Out, In>::Run(KernelContext &ctx, const OutListGPU<Out> &out, const InListGPU<In> &in) {
+  assert(impl_ != nullptr);
+  impl_->Run(ctx, out, in);
+}
+
+template class MeanGPU<uint8_t, uint8_t>;
+template class MeanGPU<float, uint8_t>;
+template class MeanGPU<int8_t, int8_t>;
+template class MeanGPU<float, int8_t>;
+template class MeanGPU<uint16_t, uint16_t>;
+template class MeanGPU<float, uint16_t>;
+template class MeanGPU<int16_t, int16_t>;
+template class MeanGPU<float, int16_t>;
+template class MeanGPU<int32_t, int32_t>;
+template class MeanGPU<float, int32_t>;
+template class MeanGPU<uint32_t, uint32_t>;
+template class MeanGPU<float, uint32_t>;
+template class MeanGPU<float, float>;
+
+
+
+template <typename Out, typename In>
+class RootMeanSquareGPU<Out, In>::Impl : public reduce_impl::RootMeanSquareImplGPU<Out, In> {
+};
+
+template <typename Out, typename In>
+RootMeanSquareGPU<Out, In>::RootMeanSquareGPU() = default;
+
+template <typename Out, typename In>
+RootMeanSquareGPU<Out, In>::~RootMeanSquareGPU() = default;
+
+template <typename Out, typename In>
+KernelRequirements RootMeanSquareGPU<Out, In>::Setup(
+    KernelContext &ctx,
+    const TensorListShape<> &in_shape, span<const int> axes, bool keep_dims, bool reduce_batch) {
+  if (!impl_) {
+    impl_ = std::make_unique<Impl>();
+  }
+  return impl_->Setup(ctx, in_shape, axes, keep_dims, reduce_batch);
+}
+
+template <typename Out, typename In>
+void RootMeanSquareGPU<Out, In>::Run(KernelContext &ctx, const OutListGPU<Out> &out, const InListGPU<In> &in) {
+  assert(impl_ != nullptr);
+  impl_->Run(ctx, out, in);
+}
+
+template class RootMeanSquareGPU<uint8_t, uint8_t>;
+template class RootMeanSquareGPU<float, uint8_t>;
+template class RootMeanSquareGPU<int8_t, int8_t>;
+template class RootMeanSquareGPU<float, int8_t>;
+template class RootMeanSquareGPU<uint16_t, uint16_t>;
+template class RootMeanSquareGPU<float, uint16_t>;
+template class RootMeanSquareGPU<int16_t, int16_t>;
+template class RootMeanSquareGPU<float, int16_t>;
+template class RootMeanSquareGPU<int32_t, int32_t>;
+template class RootMeanSquareGPU<float, int32_t>;
+template class RootMeanSquareGPU<uint32_t, uint32_t>;
+template class RootMeanSquareGPU<float, uint32_t>;
+template class RootMeanSquareGPU<float, float>;
+
+}  // namespace kernels
+}  // namespace dali
--- a/dali/kernels/reduce/mean_stddev_gpu_impl.cuh
+++ b/dali/kernels/reduce/mean_stddev_gpu_impl.cuh
@@ -310,7 +310,7 @@ class RegularizedInvRMS {
 };
 
 /**
- * @brief Implements regularize inverse standard  deviation reduction with externally provided mean
+ * @brief Implements regularized inverse standard  deviation reduction with externally provided mean
  */
 template <typename Out, typename In, typename Mean = Out, typename Acc = Out>
 class InvStdDevImplGPU :
--- a/dali/kernels/reduce/mean_stddev_gpu_impl.cuh
+++ b/dali/kernels/reduce/mean_stddev_gpu_impl.cuh
@@ -311,7 +311,7 @@ struct RegularizedInvSqrt {
 
   template <typename T>
   DALI_HOST_DEV Out operator()(T x) const {
-    float s = scale * (x + reg);
+    float s = scale * x + reg;
     return s ? ConvertSat<Out>(rsqrt(s)) : Out(0);
   }
 };
--- a/dali/kernels/reduce/mean_stddev_gpu_impl.cuh
+++ b/dali/kernels/reduce/mean_stddev_gpu_impl.cuh
@@ -1,0 +1,227 @@
+// Copyright (c) 2020, NVIDIA CORPORATION. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef DALI_KERNELS_REDUCE_MEAN_STDDEV_GPU_IMPL_CUH_
+#define DALI_KERNELS_REDUCE_MEAN_STDDEV_GPU_IMPL_CUH_
+
+#include "dali/kernels/reduce/reduce_gpu_impl.cuh"
+#include "dali/kernels/reduce/reduce_drop_dims.h"
+
+namespace dali {
+namespace kernels {
+namespace reduce_impl {
+
+template <typename Out, typename In, typename Actual>
+class MeanImplBase {
+ public:
+  Actual &This() { return static_cast<Actual&>(*this); }
+  const Actual &This() const { return static_cast<const Actual&>(*this); }
+
+  struct Postprocessor {
+    std::conditional_t<std::is_same<Out, double>::value, double, float> inv_div = 1;
+
+    template <typename T>
+    DALI_HOST_DEV Out operator()(T x) const {
+      return ConvertSat<Out>(x * inv_div);
+    }
+  };
+
+  Postprocessor *GetPostprocessorsImpl(WorkArea &wa) const {
+    assert(!This().ReduceBatch());
+    int n = This().SimplifiedOutputShape().num_samples();
+    Postprocessor *pp = wa.ParamBuffer<Postprocessor>(n);
+    for (int i = 0; i < n; i++) {
+      DALI_ENFORCE(This().ReducedElements(i) > 0, "Cannot calculate a mean from 0 elements");
+      pp[i].inv_div = 1.0 / This().ReducedElements(i);
+    }
+    return pp;
+  }
+
+  Postprocessor GetPostprocessorImpl() const {
+    assert(This().ReduceBatch());
+    DALI_ENFORCE(This().TotalReducedElements() > 0, "Cannot calculate a mean from 0 elements");
+    return { static_cast<float>(1.0 / This().TotalReducedElements()) };
+  }
+};
+
+template <typename Out, typename In, typename Acc = Out>
+class MeanImplGPU : public ReduceImplGPU<Out, In, Acc, MeanImplGPU<Out, In, Acc>>,
+                    public MeanImplBase<Out, In, MeanImplGPU<Out, In, Acc>> {
+ public:
+  using MeanBase = MeanImplBase<Out, In, MeanImplGPU<Out, In, Acc>>;
+  using typename MeanBase::Postprocessor;
+  reductions::sum GetReduction() const { return {}; }
+};
+
+template <typename Out, typename In, typename Mean, typename Actual>
+class VarianceImplBase {
+ public:
+  Actual &This() { return static_cast<Actual&>(*this); }
+  const Actual &This() const { return static_cast<const Actual&>(*this); }
+
+  void SetMean(const InListGPU<Mean> &mean, cudaStream_t stream) {
+    mean_ = mean;
+    mean_.reshape(This().SimplifiedOutputShape());
+  }
+
+  InListGPU<Mean> mean_;
+
+  struct Preprocessor {
+    const Mean *__restrict__ mean = nullptr;
+    template <typename T>
+    DALI_HOST_DEV DALI_FORCEINLINE
+    auto operator()(const T &x) const noexcept {
+      #ifdef __CUDA_ARCH__
+        auto d = x - __ldg(mean);
+      #else
+        auto d = x - *mean;
+      #endif
+        return d * d;
+    }
+  };
+
+  static_assert(sizeof(Preprocessor) == sizeof(Mean*),
+    "A variance functor must carry only a pointer to the mean");
+
+  template <int non_reduced_dims>
+  using PreprocessorBank = VariancePreprocessor<non_reduced_dims, Mean>;
+
+  void InitMean(const InListGPU<Mean> &mean) {
+    mean_ = reshape(mean, This().SimplifiedOutputShape());
+  }
+
+  Preprocessor GetPreprocessorImpl() const {
+    assert(This().SimplifiedOutputShape().num_elements() == 1);
+    return Preprocessor { mean_.data[0] };
+  }
+
+  Preprocessor *GetPreprocessorsImpl(WorkArea &wa) const {
+    int n = This().SimplifiedOutputShape().num_samples();
+    assert(This().SimplifiedOutputShape().num_elements() == n);
+    Preprocessor *pp = wa.ParamBuffer<Preprocessor>(n);
+    for (int i = 0; i < n; i++) {
+      pp[i] = { mean_.data[i] };
+    }
+    return pp;
+  }
+
+  PreprocessorBank<1> *
+  GetPreprocessorBanks(WorkArea &wa, int axis, std::integral_constant<int, 1>) const {
+    using Bank = PreprocessorBank<1>;
+    int n = This().SimplifiedOutputShape().num_samples();
+    Bank *banks = wa.ParamBuffer<Bank>(n);
+
+    for (int i = 0; i < n; i++) {
+      auto shape = This().SimplifiedOutputShape().tensor_shape_span(i);
+      auto &bank = banks[i];
+      bank.mean = mean_.data[i];
+      bank.stride[0] = volume(shape.begin() + axis, shape.end());  // outer stride
+    }
+    return banks;
+  }
+
+  PreprocessorBank<2> *
+  GetPreprocessorBanks(WorkArea &wa, int axis, std::integral_constant<int, 2>) const {
+    using Bank = PreprocessorBank<2>;
+    int n = This().SimplifiedOutputShape().num_samples();
+    Bank *banks = wa.ParamBuffer<Bank>(n);
+
+    SmallVector<int, 6> remaining_axes;
+    for (int a : This().SimplifiedAxes())
+      if (a > axis)
+        remaining_axes.push_back(a - axis - 1);
+    int mask = to_bit_mask(remaining_axes);
+
+    for (int i = 0; i < n; i++) {
+      auto &bank = banks[i];
+      auto shape = This().SimplifiedInputShape().tensor_shape_span(i);
+      auto inner_shape = span<const int64_t>(shape.begin() + axis + 1, shape.end());
+      bank.mean = mean_.data[i];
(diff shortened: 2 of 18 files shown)


mzient:
!build

mzient:
!build

dali/kernels/reduce/reduce_gpu_impl.cuh
```diff
@@ -258,8 +261,8 @@ auto GetPreprocessorsHelper(std::true_type, const ReduceImpl *impl, WorkArea *wa
 template <int non_reduced_dims, typename ReduceImpl>
 auto GetPreprocessorBanksHelper(
       std::true_type, const ReduceImpl *impl, WorkArea *wa, int reduced_axis)
-      ->decltype(impl->template GetPreprocessorBanksImpl<non_reduced_dims>(*wa, reduced_axis)) {
-  return impl->template GetPreprocessorBanksImpl<non_reduced_dims>(*wa, reduced_axis);
+      ->decltype(impl->GetPreprocessorBanksImpl(*wa, reduced_axis, int_const<non_reduced_dims>())) {
```
mzient, line 264: That solves the CUDA 9 bug.

dali/kernels/reduce/mean_gpu.cu
```diff
@@ -0,0 +1,107 @@
...
+#include "dali/kernels/reduce/mean_stddev_gpu_impl.cuh"
+
+namespace dali {
+namespace kernels {
+
+template <typename Out, typename In>
+class MeanGPU<Out, In>::Impl : public reduce_impl::MeanImplGPU<Out, In> {
+};
+
+template <typename Out, typename In>
```
klecki, line 26: Do you really need to `default` constructors and destructors in the .cu file instead of inline in class definition?

dali/kernels/reduce/reduce_setup_utils.h
```diff
@@ -204,11 +204,11 @@ inline void CalculateReducedShape(TensorListShape<> &out_shape,
     int out_d = 0;
     for (int d = 0; d < in_dim; d++) {
       if (mask & (1ul << d)) {
-      assert(out_d < out_dim);
         if (keep_dims)
           out_sample_shape[out_d++] = 1;
         continue;  // skip reduced axes
       }
+      assert(out_d < out_dim);
```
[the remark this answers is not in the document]
mzient: It wasn't.

dali/kernels/reduce/mean_stddev_gpu_impl.cuh
```diff
@@ -0,0 +1,385 @@
...
+  using scale_t = std::conditional_t<std::is_same<Out, double>::value, double, float>;
+  using Postprocessor = ScaleAndConvert<Out, scale_t>;
+
+  Postprocessor *GetPostprocessorsImpl(WorkArea &wa) const {
+    assert(!This().ReduceBatch());
+    int n = This().SimplifiedOutputShape().num_samples();
+    Postprocessor *pp = wa.ParamBuffer<Postprocessor>(n);
+    for (int i = 0; i < n; i++) {
+      DALI_ENFORCE(This().ReducedElements(i) > 0, "Cannot calculate a mean from 0 elements");
+      pp[i].scale = 1.0 / This().ReducedElements(i);
```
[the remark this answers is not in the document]
mzient: I'd rather have the division in double and then round - I don't know how much it really changes, but performancewise the cost is negligible (it's once per sample).

dali/kernels/reduce/mean_stddev_gpu_impl.cuh
```diff
@@ -0,0 +1,385 @@
...
+  using MeanBase = MeanImplBase<Out, In, MeanImplGPU<Out, In, Acc>>;
+  using typename MeanBase::Postprocessor;
+  reductions::sum GetReduction() const { return {}; }
+};
+
+/**
+ * @brief Subtracts a mean value stored in specified memory location
+ */
+template <class Mean>
+struct SubtractMeanIndirect {
```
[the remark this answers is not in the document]
mzient: I'll rename.

dali/kernels/reduce/mean_stddev_gpu_impl.cuh
```diff
@@ -0,0 +1,385 @@
...
+      auto d = x - __ldg(mean);
+    #else
+      auto d = x - *mean;
+    #endif
+      return d * d;
+  }
+};
+
+/**
+ * @brief Calculates variance, given a tensor of means.
```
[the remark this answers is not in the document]
mzient: Will change.

dali/kernels/reduce/mean_stddev_gpu_impl.cuh
```diff
@@ -0,0 +1,385 @@
...
+
+  Preprocessor GetPreprocessorImpl() const {
+    assert(This().SimplifiedOutputShape().num_elements() == 1);
+    return Preprocessor { mean_.data[0] };
+  }
+
+  Preprocessor *GetPreprocessorsImpl(WorkArea &wa) const {
+    int n = This().SimplifiedInputShape().num_samples();
+    assert(This().SimplifiedOutputShape().num_elements() ==
+           This().SimplifiedOutputShape().num_samples());
```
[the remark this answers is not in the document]
mzient: :+1:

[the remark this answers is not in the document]
mzient: Nope - n is number of _input_ samples, here I compare number of _output_ samples and it may differ for batch reduction.

dali/kernels/reduce/mean_stddev_gpu_impl_test.cu
```diff
@@ -0,0 +1,448 @@
...
+  TestTensorList<uint8_t> in;
+  in.reshape(in_shape);
+  auto in_cpu = in.cpu();
+  UniformRandomFill(in_cpu, rng, 0, 255);
+
+  TestTensorList<float> out;
+  out.reshape(req.output_shapes[0]);
+  mean.Run(ctx, out.gpu(), in.gpu());
+
+  auto out_cpu = out.cpu(ctx.gpu.stream);
```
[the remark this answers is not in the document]
mzient: No, it's not pinned.

dali/kernels/reduce/mean_gpu.cu
mzient: Yes, I do - otherwise unique_ptr complains about incomplete type.

dali/kernels/reduce/mean_gpu.cu
```diff
@@ -0,0 +1,107 @@
...
+template class MeanGPU<int32_t, int32_t>;
+template class MeanGPU<float, int32_t>;
+template class MeanGPU<uint32_t, uint32_t>;
+template class MeanGPU<float, uint32_t>;
+template class MeanGPU<float, float>;
+
+
+
+template <typename Out, typename In>
+class RootMeanSquareGPU<Out, In>::Impl : public reduce_impl::RootMeanSquareImplGPU<Out, In> {
```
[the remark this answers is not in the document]
mzient: It's a trade-off. Mean and RootMeanSquare share a good deal of functions, so splitting them into multiple translation units will most certainly increase the total CPU time (even if real time for rebuilding just these files could be shorter).

dali/kernels/reduce/mean_stddev_gpu_impl.cuh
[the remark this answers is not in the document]
jantonguirao: Oh, I missed that

dali/kernels/reduce/mean_stddev_gpu_impl.cuh
```diff
@@ -0,0 +1,386 @@
...
+    int n = This().SimplifiedOutputShape().num_samples();
+    Postprocessor *pp = wa.ParamBuffer<Postprocessor>(n);
+    for (int i = 0; i < n; i++) {
+      DALI_ENFORCE(This().ReducedElements(i) > 0, "Cannot calculate a mean from 0 elements");
+      pp[i].scale = 1.0 / This().ReducedElements(i);
+    }
+    return pp;
+  }
+
+  Postprocessor GetPostprocessorImpl() const {
```
klecki, line 54: When reading the comment above GetPostprocessor, there is:
```
   * Note the is_first/is_last compile-time parameters. Overload resolution will pick the
   * right function with distinct return type.
```
but actually there are no such arguments (in reduce_gpu_impl.cuh), instead there are `do_pre/postprocess`

dali/kernels/reduce/mean_stddev_gpu_impl.cuh
```diff
@@ -0,0 +1,386 @@
...
+
+#include "dali/kernels/reduce/reduce_gpu_impl.cuh"
+#include "dali/kernels/reduce/reduce_drop_dims.h"
+
+namespace dali {
+namespace kernels {
+namespace reduce_impl {
+
+template <typename Out, typename Scale>
+struct ScaleAndConvert {
```
klecki, line 26: I feel like this belongs into the MeanImplBase, it's used only there, and those two are completely tied together with how they are implemented.

dali/kernels/reduce/mean_stddev_gpu_impl.cuh
klecki: Also, the comment there deserves the mention, that one is used for per-sample reductions and the other for global/batch reduction.

At first I was about to comment, that they do the same thing twice in a slightly different way (and a loop). 
Still, you need to look at the implementation to know (or maybe I didn't read enough of reduce_gpu_impl).

dali/kernels/reduce/mean_stddev_gpu_impl.cuh
```diff
@@ -0,0 +1,386 @@
...
+  void SetMean(const InListGPU<Mean> &mean, cudaStream_t stream) {
+    mean_ = mean;
+    mean_.reshape(This().SimplifiedOutputShape());
+  }
+
+  InListGPU<Mean> mean_;
+
+  using Preprocessor = VarianceIndirect<Mean>;
+
+  static_assert(sizeof(Preprocessor) == sizeof(Mean*),
```
klecki, line 145: Again you're using one particular type here, but this looks like it should be a customization point with this requirement.

dali/kernels/reduce/mean_stddev_gpu_impl.cuh
```diff
@@ -0,0 +1,386 @@
...
+      bank.stride[0] = volume(out_shape.begin() + axis, out_shape.end());  // outer stride
+      bank.stride[1] = 1;  // inner stride, always 1?
+      bank.inner_dims = DropDims(inner_shape, mask);  // reindexing, if necessary
+    }
+    return banks;
+  }
+
+  template <int non_reduced_dims>
+  PreprocessorBank<non_reduced_dims> *
+  GetPreprocessorBanksImpl(WorkArea &wa, int axis, int_const<non_reduced_dims> nrd) const {
```
klecki, line 216: There is no definition for non_reduced_dims > 2. Just pointing that out, I'm still looking up what `PreprocessorBanks` do compared to `Preprocessors` and `Preprocessor`. 

Some description of what the signatures are supposed to be on such [Pre/Post]processor[Banks] and what the result is expected to be would be nice (so I would be able to verify if the implementation actually does what's intended of it).

dali/kernels/reduce/mean_stddev_gpu_impl.cuh
```diff
@@ -0,0 +1,386 @@
...
+    return { m };
+  }
+};
+
+template <typename Mean>
+struct VariancePreprocessor<2, Mean> {
+  const Mean *mean;
+  i64vec<2> stride;
+  /// Calculates the fully reduced inner offset based on non-reduced `pos[1]`
+  DropDims inner_dims;
```
klecki, line 116: This will be nitpicky, but according to the Google Style Guide's nomenclature that we use, I would say this is going into the `class` area, with the members that must be set to make the `Get()` work (especially with inner_dims). So maybe give this a constructor?

Also, is this really a VariancePreprocessor or more of VariancePreprocessorBank?

dali/kernels/reduce/mean_stddev_gpu_impl.cuh
```diff
@@ -0,0 +1,386 @@
...
+
+  DALI_HOST_DEV DALI_FORCEINLINE
+  reductions::variance<Mean> Get(const i64vec<2> &pos) const {
```
klecki, line 119: Why does preprocessor return reductions instead of doing preprocessing?

dali/kernels/reduce/mean_stddev_gpu_impl.cuh
```diff
@@ -0,0 +1,386 @@
...
+    Mean m = __ldg(mean + offset);
+  #else
+    Mean m = mean[offset];
+  #endif
+    return { m };
+  }
+};
+
+template <typename Out, typename In, typename Mean, typename Actual>
+class VarianceImplBase {
```
klecki, line 131: Is this whole class copying 1 or n pointers from mean_ TLV to the preprocessor objects of corresponding samples?
I just want to make sure if I understand that.

Can you shed some light when the GetPreprocessorBanks for ndim=2 happens?

dali/kernels/reduce/mean_stddev_gpu_impl.cuh
```diff
@@ -0,0 +1,386 @@
...
+  Scale scale = 1;
+
+  template <typename T>
+  DALI_HOST_DEV Out operator()(T x) const {
+    return ConvertSat<Out>(sqrt(x * scale));
+  }
+};
+
+template <typename Out, typename In, typename Actual>
+class RootMeanImplBase {
```
klecki, line 232: This is the same as MeanImplBase, but uses the ScaleSqrtConvert instead of ScaleAndConvert, right?

I'm thinking if taking some common parts into ElementwiseScaleAndSomething would have any benefit for tersness of this file, or is it better to keep separate.
Maybe keep it simpler.

dali/kernels/reduce/mean_gpu.cu
```diff
@@ -0,0 +1,107 @@
...
+};
+
+template <typename Out, typename In>
+MeanGPU<Out, In>::MeanGPU() = default;
+
+template <typename Out, typename In>
+MeanGPU<Out, In>::~MeanGPU() = default;
+
+template <typename Out, typename In>
+KernelRequirements MeanGPU<Out, In>::Setup(
```
JanuszL, line 33: Setup and run seems to be repetitive. Can it be a moved to the base class (at least part)?

dali/kernels/reduce/mean_stddev_gpu_impl.cuh
JanuszL: This sounds pretty generic. It can be moved to some utils as well.

dali/kernels/reduce/mean_gpu.cu
```diff
@@ -0,0 +1,107 @@
...
+// limitations under the License.
+
+#include <memory>
+#include "dali/kernels/reduce/reduce_gpu.h"
+#include "dali/kernels/reduce/mean_stddev_gpu_impl.cuh"
+
+namespace dali {
+namespace kernels {
+
+template <typename Out, typename In>
```
klecki, line 22: General remark: I wouldn't complain if such file had a succinct comment that this is mostly boilerplate for kernels & predefined set of explicit instantiations here. Similar thing for files with Kernel declarations & extern templates. But that's just my opinion.

dali/kernels/reduce/mean_stddev_gpu_impl.cuh
```diff
@@ -0,0 +1,386 @@
...
+class InvStdDevImplGPU :
+      public ReduceImplGPU<Out, In, Acc, InvStdDevImplGPU<Out, In, Mean, Acc>>,
+      public VarianceImplBase<Out, In, Mean, InvStdDevImplGPU<Out, In, Mean, Acc>>,
+      public RegularizedInvRMS<Out, In, InvStdDevImplGPU<Out, In, Mean, Acc>> {
+ public:
+  using ReduceBase = ReduceImplGPU<Out, In, Acc, InvStdDevImplGPU<Out, In, Mean, Acc>>;
+  using VarBase = VarianceImplBase<Out, In, Mean, InvStdDevImplGPU<Out, In, Mean, Acc>>;
+  using MeanBase = RegularizedInvRMS<Out, In, InvStdDevImplGPU<Out, In, Mean, Acc>>;
+
+  using Preprocessor = typename VarBase::Preprocessor;
```
klecki, line 364: When do you need to just use `using Preprocessor = ...` and when do you provide the GetPrerocessorImpl & co? 
Is it beacause one of the ImplBase/Base classes (btw, RegularizedInvRMS doesn't have Base in it's name) already provide them, and you need to bring only the types to scope here?

Maybe going a bit into the design of those classes, I like the customization points for Pre & Post processors, but I wonder, if it wouldn't be easier to read if they were just explicitly defined.

Basically those ImplBase classes represent a non-documented concept. It would be really helpful if there was a doc explaining, that you can provide such class, that needs to provide:
P GetPreprocessor(...) // for batch reduction
P* GetPreprocessor() // for per sample reduction
...
Where `P` is expected to provide call operator.
Such "Concept"/Policy class could be passed to ReduceImplGpu, and it would access it's Pre/Post processors directly without a chain of unholy overloads.

dali/kernels/reduce/mean_stddev_gpu_impl.cuh
```diff
@@ -0,0 +1,386 @@
...
+  using param_t = std::conditional_t<std::is_same<Out, double>::value, double, float>;
+  using Postprocessor = RegularizedInvSqrt<Out, param_t>;
+
+  void SetRegularizationTerm(param_t reg) {
+    regularization_ = reg * reg;
+  }
+
+  param_t regularization_ = 0.0f;
+
+  Postprocessor *GetPostprocessorsImpl(WorkArea &wa) const {
```
klecki, line 335: Just a thought, wouldn't one function like:
GetPostprocessor(int reduced_elements){} work as well. The ReduceImpl could call it in a loop for per-sample reductions, or once with TotalReducedElements() for batch-reduction and this would remove the need to define the same thing here twice, which you did for all of those clases in this file.

dali/kernels/reduce/reduce_all_kernel_gpu.h
```diff
@@ -35,6 +35,7 @@ class DLL_PUBLIC ReduceAllGPU {
   // TODO(janton): remove this when implemented
```
JanuszL, line 35: Is this still a valid comment?

dali/kernels/reduce/reduce_gpu.h
```diff
@@ -62,14 +62,253 @@ extern template class SumGPU<uint64_t, uint8_t>;
...
+
+
+template <typename Out, typename In, typename Mean = Out>
+class DLL_PUBLIC StdDevGPU {
+ public:
+  StdDevGPU();
+  ~StdDevGPU();
+
+  /**
+   * @brief Sets up the reduction
```
JanuszL, line 195: I would add a formula to the doc string as in InvStdDevGPU.

dali/kernels/reduce/reduce_gpu.h
```diff
@@ -62,14 +62,253 @@ extern template class SumGPU<uint64_t, uint8_t>;
...
+
+template <typename Out, typename In>
+class DLL_PUBLIC RootMeanSquareGPU {
+ public:
+  RootMeanSquareGPU();
+  ~RootMeanSquareGPU();
+
+  /**
+   * @brief Sets up the reduction
+   *
```
JanuszL, line 142: I would add a formula to the doc string as in InvStdDevGPU.

dali/kernels/reduce/reduce_gpu_impl.cuh
JanuszL: So yo can add some comment in the code.

dali/kernels/reduce/mean_stddev_gpu_impl.cuh
mzient: It depends - see GetPreprocessonBanks - if batch-reducing, only the first (and presumably the only) sample is used; otherwise each bank gets its own pointer.

dali/kernels/reduce/mean_stddev_gpu_impl.cuh
mzient: Sorry, I don't understand this comment.

dali/kernels/reduce/mean_stddev_gpu_impl.cuh
mzient: You can find a lengthy comment in `reduce_axes_gpu_impl.cuh` explaining what a preprocessor bank is.

dali/kernels/reduce/mean_stddev_gpu_impl.cuh
mzient: It could be sth like `MeanBaseImpl<typename Out, typename In, typename Actual, typename Potprocessor>`.

dali/kernels/reduce/mean_stddev_gpu_impl.cuh
mzient: The comment is stale. I'll update it.

dali/kernels/reduce/mean_gpu.cu
mzient: I know... but I think it the alternative would either require macros or be unnecessarily complicated.

dali/kernels/reduce/mean_stddev_gpu_impl.cuh
mzient: It's in the reduction namespace, but variance is in fact just a functor that returns (x - mean)^2

dali/kernels/reduce/mean_stddev_gpu_impl.cuh
mzient: It's a bank. I'll see how many more line breaks it will cause :\

mzient: Done.

dali/kernels/reduce/mean_stddev_gpu_impl.cuh
mzient: No, it's per-sample and samples can have different reduction factors.

dali/kernels/reduce/mean_stddev_gpu_impl.cuh
mzient: Done.

dali/kernels/reduce/mean_stddev_gpu_impl.cuh
mzient: The problem is, that the base class ReduceImplBase also defines Preprocessor, Postprocessor, etc - we need to shadow them properly here; without this using, there's ambiguity, because the type is defined in multiple base classes.

mzient:
>     * ImplBase classes look like a concept that should be documented, probably setting Pre/Post processors explicitly. This can remove the need for the GetPreprocessor magic.

This would bring even more boilerplate with identity pre-/postprocessors in every single derived class.

>     * GetPreprocessor and GetPrerpocessors can be unified.

They cannot, since a null is a valid argument for postprocessors - it just creates a default one for each sample directly on device.

>     * There are parts with a lot of boilerplate but I think it's better to keep it as it's easier to understand.

dali/kernels/reduce/reduce_drop_dims.h
```diff
@@ -51,7 +51,7 @@ struct DropDims {
   int64_t mod[kMaxDims];
   int start = 2 * kMaxDims;
 
-  DropDims() = default;
+  DALI_HOST_DEV DropDims() {}
```
mzient, line 54: NVCC9 is produces error with =default when there's no host/device and a warning when it is there :(

dali/kernels/reduce/reduce_gpu.h
```diff
@@ -62,14 +62,253 @@ extern template class SumGPU<uint64_t, uint8_t>;
...
+  ~MeanGPU();
+
+  /**
+   * @brief Sets up the reduction
+   *
+   * Sets up the reduction according to the parameters. The indices of dimensions to be reduced
+   * are provided in `axes` parameter.
+   * For a successful batch reduction, the reduced shape of all samples must be equal (but the
+   * input may have non-uniform shape, as long as the non-uniform dimensions are reduced).
+   *
```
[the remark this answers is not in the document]
mzient: I think a mean is quite well defined, nothing fancy there...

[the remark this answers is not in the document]
JanuszL: By the definition, yes, but here you have these additional flavors, and args like `reduced_pos` and `pos` in some kernels. So I'm more asking how exactly it is implemented here.

dali/kernels/reduce/mean_stddev_gpu_impl.cuh
klecki: I was talking about the fact, that you use VarianceIndirect preprocessor as the only thing here. Why this check, it's not like you parametrize over Preprocessors.

dali/kernels/reduce/reduce_gpu_impl.cuh
mzient: It doesn't make much sense to comment it in the code - how would I do it? Show an example of (arguably more complicated) code that _didn't_ work?

JanuszL: State that this is the fix for CUDA 9 compiler bug and give some background. So when we deprecate CUDA 9 we can remove this WAR.

mzient: But I think it's actually better this way! It's not a workaround.

dali/kernels/reduce/mean_stddev_gpu_impl.cuh
```diff
@@ -57,21 +57,11 @@ class MeanImplBase {
...
-  }
-
-  Postprocessor GetPostprocessorImpl() const {
-    assert(This().ReduceBatch());
-    DALI_ENFORCE(This().TotalReducedElements() > 0, "Cannot calculate a mean from 0 elements");
-    return { static_cast<scale_t>(1.0 / This().TotalReducedElements()) };
+  Postprocessor GetPostprocessorImpl(int sample_index, bool reduce_batch) const {
+    int64_t reduced_elems = reduce_batch ? This().TotalReducedElements()
+                                         : This().ReducedElements(sample_index);
+    DALI_ENFORCE(reduced_elems > 0, "Cannot calculate a mean from 0 elements");
```
klecki, line 70: I still have a feeling, that those Postprocessors should just get the number of reduced elements, but maybe there will be some sample-specific behaviour in the future.

dali/kernels/reduce/reduce_gpu_impl.cuh
```diff
@@ -277,6 +277,34 @@ GetPreprocessorBanksHelper(bool_const<do_preprocess>, ...) {
...
+ *
+ * template <int non_reduced_dims>
+ * MyPreprocessorBank<non_reduced_dims> *GetPreprocessorBanksImpl(WorkArea &wa) const { ... };
+ * ```
+ *
+ * You may need to add extra arguments to Run/Setup, in which case just shadow the original
+ * function(s) - you should call them at some point in your customized Run/Setup.
+ *
+ * See mean_stddev_gpu_impl.cuh for usage examples.
+ */
```
klecki: :+1:

dali/kernels/reduce/reduce_gpu_impl.cuh
```diff
@@ -308,6 +308,50 @@ GetPreprocessorBanksHelper(bool_const<do_preprocess>, ...) {
...
+
+  /// returns a const reference to the actual derived class instance
+  const Actual &This() const noexcept { return static_cast<Actual &>(*this); }
+
+  /// Get stage - for testing
+  const ReductionStage &GetStage(int idx) const { return stages_[idx]; }
+  /// Get number of stages - for testing
+  int GetNumStages() const { return stages_.size(); }
+
+  /// Trye, if reduction across input tensors was requested
```
klecki, line 322:
```suggestion
  /// True, if reduction across input tensors was requested
```

dali/kernels/reduce/reduce_gpu.h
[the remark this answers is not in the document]
klecki: I agree, that the formula in doc wouldn't hurt.

dali/kernels/reduce/mean_stddev_gpu_impl.cuh
```diff
@@ -0,0 +1,340 @@
...
+    return { param_t(1.0 / reduced_elems), regularization_ };
+  }
+};
+
+/**
+ * @brief Implements regularize inverse standard  deviation reduction with externally provided mean
```
klecki, line 313:
```suggestion
 * @brief Implements regularized inverse standard  deviation reduction with externally provided mean
```

klecki:
Do you think it would be possible to gather setup and verification from the tests into one place or is the baseline reduction implementation not flexible enough?

mzient:
!build

dali/kernels/reduce/reduce_test.h
```diff
@@ -135,6 +142,34 @@ void RefReduce(const TensorView<StorageCPU, Out> &out, const TensorView<StorageC
...
+      OnlineReducer<Out, Reduction> red;
+      red.reset();
+      for (int i = 0; i < N; i++) {
+        red.add(reduced_samples.data[i][j], R);
+      }
+      out.data[0][j] = red.result();
+    }
+  } else {
+    assert(out.num_samples() == in.num_samples());
+    for (int i =0; i < in.num_samples(); i++) {
```
klecki, line 167: Linter nitpick:
```suggestion
    for (int i = 0; i < in.num_samples(); i++) {
```

PR #1923, opened 2020-04-30
Collect exit code from test_body

klecki:
The test_body now will still be executed with `set -e`
but the exit code would be captured and the line should
pass regardless

Previously only the last instruction in test_body
would return error.

#### Why we need this PR?
Fix CI error detection

 - What solution was applied:
     WAR for capturing exit code, keep `set -e`
     QA scripts
     Verify that it works
 - Validation and testing:
     CI run
     No


klecki:
!build

klecki:
!build

PR #1924, opened 2020-04-30
DALI as a git submodule

AlexBula:
#### Why we need this PR?
*Pick one, remove the rest*
- It fixes a bug: *It allows to add DALI to an another project as a git submodule and link against it*

 - What solution was applied:
     *Change CMakeLists paths to ${PROJECT_SOURCE_DIR}, add export of DALI*
     *Few path in the CMakeLists were changed because they assumed DALI was the main project that was currently built*
 - Validation and testing:
     *Checked if DALI is building after changes*

**JIRA TASK**: *NA*

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -198,4 +198,7 @@ include(${CMAKE_CURRENT_SOURCE_DIR}/cmake/lint.cmake)
 
 include(${CMAKE_CURRENT_SOURCE_DIR}/cmake/Install.cmake)
 
-export(PACKAGE dali)
+add_library(DALI::dali ALIAS dali)
+add_library(DALI::dali_core ALIAS dali_core)
+add_library(DALI::dali_kernels ALIAS dali_kernels)
+add_library(DALI::dali_operators ALIAS dali_operators)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -197,3 +197,5 @@ file(WRITE ${CMAKE_BINARY_DIR}/dali/__init__.py "")
 include(${CMAKE_CURRENT_SOURCE_DIR}/cmake/lint.cmake)
 
 include(${CMAKE_CURRENT_SOURCE_DIR}/cmake/Install.cmake)
+
+export(PACKAGE dali)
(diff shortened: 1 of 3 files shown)


CMakeLists.txt
```diff
@@ -197,3 +197,5 @@ file(WRITE ${CMAKE_BINARY_DIR}/dali/__init__.py "")
 include(${CMAKE_CURRENT_SOURCE_DIR}/cmake/lint.cmake)
 
 include(${CMAKE_CURRENT_SOURCE_DIR}/cmake/Install.cmake)
+
+export(PACKAGE dali)
```
szalpal, line 201: I'd approach it a little bit differently: I think we should export individual targets within some namespace, e.g. using this:
```
export(TARGETS [target1 [target2 [...]]] [NAMESPACE <namespace>]
       [APPEND] FILE <filename> [EXPORT_LINK_INTERFACE_LIBRARIES])
```
There you could export `dali, dali_operators, dali_core etc...`.

Usage could be following:
```
add_subdirectory(extern/DALI)
target_link_libraries(mylib DALI::dali_core DALI::dali_kernels)
```

On top of that, we can expose convenient target that gathers every DALI target, so that user can go:
```
add_subdirectory(extern/DALI)
target_link_libraries(mylib DALI::DALI)
```

But I'd like to hear what others think about it

AlexBula: you are probably right :)

AlexBula: however I have no idea how to do it differently

AlexBula: Changed so only the listed targets are exported

AlexBula:
!build

CMakeLists.txt
```diff
@@ -197,3 +197,8 @@ file(WRITE ${CMAKE_BINARY_DIR}/dali/__init__.py "")
 include(${CMAKE_CURRENT_SOURCE_DIR}/cmake/lint.cmake)
 
 include(${CMAKE_CURRENT_SOURCE_DIR}/cmake/Install.cmake)
+
+add_library(DALI::dali ALIAS dali)
+add_library(DALI::dali_core ALIAS dali_core)
+add_library(DALI::dali_kernels ALIAS dali_kernels)
+add_library(DALI::dali_operators ALIAS dali_operators)
```
szalpal, line 204: Is it possible to additionally create one target, that puts together all these four?

One to rule them all ;)

AlexBula: I do not think so, at least I did not find one

PR #1932, opened 2020-05-04
C API function: inherit parameters from serialized pipeline

szalpal:

#### Why we need this PR?
*Pick one, remove the rest*
- It adds a new function to the C API - allow pipeline to inherit default parameters, which were saved in serialized pipeline, without the need of explicitly providing them.

 - What solution was applied:
     Added new function
     C API
     NA
 - Validation and testing:
     GTest
     doxygen

### DIFF
--- a/dali/c_api/c_api.cc
+++ b/dali/c_api/c_api.cc
@@ -18,8 +18,8 @@
 #include <string>
 #include <utility>
 #include <vector>
-#include <include/dali/core/cuda_stream.h>
 
+#include "dali/core/cuda_stream.h"
 #include "dali/core/format.h"
 #include "dali/core/tensor_shape.h"
 #include "dali/pipeline/pipeline.h"
--- a/dali/c_api/c_api.cc
+++ b/dali/c_api/c_api.cc
@@ -18,6 +18,7 @@
 #include <string>
 #include <utility>
 #include <vector>
+#include <include/dali/core/cuda_stream.h>
 
 #include "dali/core/format.h"
 #include "dali/core/tensor_shape.h"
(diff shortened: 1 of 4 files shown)


dali/c_api/c_api.cc
```diff
@@ -105,20 +105,32 @@ void daliCreatePipeline(daliPipelineHandle* pipe_handle,
     pipe->SetQueueSizes(cpu_prefetch_queue_depth, gpu_prefetch_queue_depth);
   }
   pipe->Build();
-  pipe_handle->pipe = reinterpret_cast<void*>(pipe);
+  pipe_handle->pipe = reinterpret_cast<void *>(pipe);
```
mzient, line 108:
```suggestion
  pipe_handle->pipe = pipe;
```
Implicit cast to void* is allowed.

dali/c_api/c_api.cc
```diff
@@ -105,20 +105,32 @@ void daliCreatePipeline(daliPipelineHandle* pipe_handle,
...
   pipe_handle->ws = new dali::DeviceWorkspace();
   CUDA_CALL(cudaStreamCreateWithFlags(&pipe_handle->copy_stream, cudaStreamNonBlocking));
 }
 
-void daliPrefetchUniform(daliPipelineHandle* pipe_handle, int queue_depth) {
-  dali::Pipeline* pipeline = reinterpret_cast<dali::Pipeline*>(pipe_handle->pipe);
+
+void daliDeserializeDefault(daliPipelineHandle *pipe_handle, const char *serialized_pipeline,
+                            int length) {
+  dali::Pipeline *pipe = new dali::Pipeline(std::string(serialized_pipeline, length));
```
mzient, line 116: do we allow length < 0 to denote a null-terminated string? If so, you'd need to include that in string construction:
```suggestion
  std::string str = length >= 0 ? std::string(serialized_pipeline, length) : std::string(serialized_pipeline)
  dali::Pipeline *pipe = new dali::Pipeline(std::move(str));
```

dali/c_api/c_api.cc
```diff
@@ -105,20 +105,32 @@ void daliCreatePipeline(daliPipelineHandle* pipe_handle,
...
+  pipe->Build();
```
mzient, line 117: If this throws, resources will leak.
The best option is to use a unique_ptr and release it in the next line.

dali/c_api/c_api.cc
```diff
@@ -105,20 +105,32 @@ void daliCreatePipeline(daliPipelineHandle* pipe_handle,
...
+  pipe_handle->pipe = reinterpret_cast<void *>(pipe);
```
mzient, line 118: ...like this
```suggestion
  pipe_handle->pipe = reinterpret_cast<void *>(pipe.release());
```

dali/c_api/c_api.cc
```diff
@@ -105,20 +105,32 @@ void daliCreatePipeline(daliPipelineHandle* pipe_handle,
...
+  pipe_handle->ws = new dali::DeviceWorkspace();
+  CUDA_CALL(cudaStreamCreateWithFlags(&pipe_handle->copy_stream, cudaStreamNonBlocking));
```
mzient, line 120: This should go before releasing ownership of pipe.

szalpal: Done

dali/c_api/c_api.cc
szalpal: Done

dali/c_api/c_api.cc
szalpal: Done

dali/c_api/c_api.cc
szalpal: We decided not to do this, since the pipeline can contain `\0`'s inside

dali/c_api/c_api.cc
```diff
@@ -105,20 +105,32 @@ void daliCreatePipeline(daliPipelineHandle* pipe_handle,
...
   pipe_handle->ws = new dali::DeviceWorkspace();
   CUDA_CALL(cudaStreamCreateWithFlags(&pipe_handle->copy_stream, cudaStreamNonBlocking));
 }
 
-void daliPrefetchUniform(daliPipelineHandle* pipe_handle, int queue_depth) {
-  dali::Pipeline* pipeline = reinterpret_cast<dali::Pipeline*>(pipe_handle->pipe);
+
+void daliDeserializeDefault(daliPipelineHandle *pipe_handle, const char *serialized_pipeline,
+                            int length) {
+  auto pipeline = std::make_unique<dali::Pipeline>(std::string(serialized_pipeline, length));
```
JanuszL, line 116: Maybe this should be done inside the `daliCreatePipeline` as well.

szalpal: Generally there's no, exception handling in C API, so it should be done in every function. I can add it, but I don't want this PR to evolve into fixing C API ;)

dali/c_api/c_api.cc
```diff
@@ -105,20 +105,32 @@ void daliCreatePipeline(daliPipelineHandle* pipe_handle,
...
+  pipeline->Build();
+  CUDA_CALL(cudaStreamCreateWithFlags(&pipe_handle->copy_stream, cudaStreamNonBlocking));
+  pipe_handle->ws = new dali::DeviceWorkspace();
```
mzient, line 119: If this throws (e.g. a bad_alloc, copy_stream will leak).
Suggested solution - create all resources first, in managed handles - and then only release them:
```C++
CUDAStream stream = CUDAStream::Create();
auto ws = make_unique<dali::DeviceWorkspace>()`
pipe_handle->ws = ws.release();
pipe_handle->copy_stream = stream.release();
pipe_handle->pipe = pipe.release();
```

szalpal:
!build

szalpal:
!build

PR #1938, opened 2020-05-05
Move from http://xiph.org to GitHub for libflac, libvorbis and libogg

JanuszL:
- GitHub seems to be more reliable than http://xiph.org so move to it

#### Why we need this PR?
*Pick one, remove the rest*
- It fixes a problem with the http://xiph.org availability

 - What solution was applied:
     GitHub seems to be more reliable than http://xiph.org so move to it
     docker deps
     NA
 - Validation and testing:
     CI
     NA

**JIRA TASK**: *[NA]*


klecki:
- GitHub seems to be more reliable than http://xiph.org so move to it

Issue #1958, opened 2020-05-13
NumpyReader is unable to load uint8 files


RDruon:
Hello DALI Team,

It seems the new NumpyReader operator can't load numpy files saved as uint8 array.

Please find below the minimal reproducible example:
```
import os
from nvidia.dali.pipeline import Pipeline
from nvidia.dali.plugin.pytorch import DALIGenericIterator
import nvidia.dali.ops as ops
import numpy as np

class NumpyFilesPipeline(Pipeline):
    def __init__(self, files_path, batch_size=1, num_threads=1, device_id=0):
        super(NumpyFilesPipeline, self).__init__(batch_size,
                                         num_threads,
                                         device_id)
        self.input = ops.NumpyReader(file_root  = files_path)
    def define_graph(self):
        numpy_files = self.input(name="Reader")
        return numpy_files
    def iter_setup(self):
        pass

def create_img_numpy(width, height, output_path, idx, base_filename='img{}'):
    a = np.random.rand(width,height,3) * 255
    final_output_path = os.path.join(output_path, base_filename.format(idx) + ".npy")
    np.save(final_output_path, a.astype(np.uint8))

if __name__ == "__main__":
    # Create numpy file
    for j in range(10):
        os.makedirs('/tmp/numpy_test/class{}'.format(j))
        for i in range(10):
            create_img_numpy(256, 256, '/tmp/numpy_test/class{}'.format(j), i)
    # Create numpy pipeline
    pipeline = NumpyFilesPipeline('/tmp/numpy_test/')
    pipeline.build()
    pipeline.run()
```
This code produces the following error:

> RuntimeError: Critical error in pipeline: Error in thread 0: [/opt/dali/dali/pipeline/data/buffer.h:143] Assert on "IsValidType(type_)" failed: Buffer has no type, 'mutable_data<T>()' or 'set_type' must be called on non-const buffer to set valid type

Running the same code and changing `np.uint8` to `np.int32` (in create_img_numpy function) fix the issue.
I think it's related to [this file](https://github.com/NVIDIA/DALI/blob/e50d1e0daa847904d5e14cc473cbee97903e90e5/dali/operators/reader/loader/numpy_loader.cc#L26), numpy format "u1" (uint8) is missing.

**DALI version**: nvidia_dali-0.23.0.dev0-12345-cp36-cp36m-manylinux1_x86_64.whl (built using docker)
**Python version**: Python 3.6.9

szalpal:
[...]

PR #1961, opened 2020-05-13
Adding more Numpy data types

szalpal:

#### Why we need this PR?
*Pick one, remove the rest*
- It fixes a bug - missing numpy data types

Bug discovered in #1958 

### DIFF
--- a/dali/operators/reader/loader/numpy_loader.cc
+++ b/dali/operators/reader/loader/numpy_loader.cc
@@ -24,16 +24,18 @@
 
 namespace dali {
 TypeInfo TypeFromNumpyStr(const std::string &format) {
-  if (format == "i8")
-    return TypeInfo::Create<int64_t>();
-  else if (format == "i4")
-    return TypeInfo::Create<int32_t>();
-  else if (format == "f4")
-    return TypeInfo::Create<float>();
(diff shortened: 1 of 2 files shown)


klecki:
Is there a format string for `bool`? Is `bool` possible/valid here?

szalpal:
There is, but C++ doesn't really map to fixed-length type for `bool`, so I didn't include it. Should I?

szalpal:
!build

klecki:
> There is, but C++ doesn't really map to fixed-length type for `bool`, so I didn't include it. Should I?

Yes, apparently it's implementation defined.

JanuszL:
Can you add a regression test for that?

szalpal:
!build

dali/operators/reader/loader/numpy_loader.cc
```diff
@@ -24,16 +24,18 @@
 
 namespace dali {
 TypeInfo TypeFromNumpyStr(const std::string &format) {
```
JanuszL, line 26: Why not to reuse `TypeFromFormatStr` from https://github.com/NVIDIA/DALI/blob/master/dali/util/pybind.h#L103?

PR #1963, opened 2020-05-14
Auto apply dataset options

awolant:

#### Why we need this PR?
- It adds new features to prohibit end user from unintentionally moving data from DALI GPU buffers to TensorFlow CPU allocated memory.

 - What solution was applied:
     *Apply needed dataset options during its creation*
     *When dataset is created, check if TF device and DALI device match*
     *TF plugin and `tf.data.Dataset` tests*
     *Pipeline and C API*
 - Validation and testing:
     *Modified existing tests*

**JIRA TASK**: *NA*


awolant:
!build

JanuszL:
Please fix LGTM alert. Other than this, LGTM.

dali/test/python/test_utils_tensorflow.py
```diff
@@ -36,15 +36,4 @@ def available_gpus():
...
-    options = tf.data.Options()
-    try:
-        options.experimental_optimization.apply_default_optimizations = False
-        options.experimental_optimization.autotune = False
-    except:
-        print('Could not set TF Dataset Options')
-
-    return options
\ No newline at end of file
+    return devices
```
szalpal, line 39:
```suggestion
    return devices
    
```
Adding empty line at the end

dali/python/nvidia/dali/plugin/tf.py
```diff
@@ -128,6 +128,16 @@ def dataset_compatible_tensorflow():
   from tensorflow.python.data.util import structure
   import functools
 
+  def dataset_options():
+    options = tf.data.Options()
+    try:
+        options.experimental_optimization.apply_default_optimizations = False
+        options.experimental_optimization.autotune = False
+    except:
+        print('Could not set TF Dataset Options')
```
mzient, line 137: 1. Why are we ignoring this error?
2. If we print anything, how about printing the message from the underlying exception?

awolant:
!build

dali/c_api/c_api.cc
```diff
@@ -368,8 +368,16 @@ size_t daliMaxDimTensors(daliPipelineHandle* pipe_handle, int n) {
 }
 
 unsigned daliGetNumOutput(daliPipelineHandle* pipe_handle) {
-  dali::DeviceWorkspace* ws = reinterpret_cast<dali::DeviceWorkspace*>(pipe_handle->ws);
-  return ws->NumOutput();
+  dali::Pipeline *pipeline = reinterpret_cast<dali::Pipeline *>(pipe_handle->pipe);
+  return pipeline->num_outputs();
+}
```
awolant, line 373: I've changed the implementation, because before it was possible to call this function before build and it returned 0 without errors.

dali/test/python/test_dali_tf_dataset_mnist.py
```diff
@@ -290,14 +292,16 @@ def _test_estimators_wrapping_keras_single_device(device='cpu', device_id=0):
...
+#     _test_estimators_wrapping_keras_single_device('gpu', 0)
 
 
-@with_setup(clear_checkpoints, clear_checkpoints)
-def test_estimators_wrapping_keras_single_other_gpu():
-    _test_estimators_wrapping_keras_single_device('gpu', 1)
+# Temporarily disabled, waiting for TF updates
+# @with_setup(clear_checkpoints, clear_checkpoints)
+# def test_estimators_wrapping_keras_single_other_gpu():
+#     _test_estimators_wrapping_keras_single_device('gpu', 1)
```
awolant, line 304: Removed tests would worked without the assertion added in this PR, but they would copy DALI GPU output to TF CPU buffer. 9.9/10 this is not what we want at all.

dali/python/nvidia/dali/plugin/tf.py
awolant: This was leftover from when I was developing this. Basically, older TF versions did not have the second option to set, but it was fine, because the behavior of the whole thing was as we wanted it to be. I removed `try`, because all TF versions that we support now do have this.

dali/pipeline/pipeline.h
```diff
@@ -385,6 +385,16 @@ class DLL_PUBLIC Pipeline {
    */
   DLL_PUBLIC inline int device_id() const { return device_id_; }
 
+    /**
+   * @brief Returns number of outputs.
+   */
+  DLL_PUBLIC int num_outputs() const;
+
+  /**
+   * @brief Returns string describing device backing output specified by given id.
```
mzient, line 394:
```suggestion
   * @brief Returns a string describing the type device type backing the output specified by given id.
```

dali/pipeline/pipeline.h
```diff
@@ -385,6 +385,16 @@ class DLL_PUBLIC Pipeline {
...
+   */
+  DLL_PUBLIC std::string output_device(int id) const;
```
mzient, line 396:
```suggestion
  DLL_PUBLIC const std::string &output_device(int id) const;
```
This will avoid unnecessary copy when using it like `pipe.output_device(id) == "gpu"`.

dali_tf_plugin/dali_dataset_op.cc
```diff
@@ -484,4 +506,4 @@ Creates a DALI dataset compatible with tf.data.Dataset from a DALI pipeline.
 
 }  // namespace dali_tf_impl
 
-#endif  // TF_MAJOR_VERSION == 1 && TF_MINOR_VERSION >= 15
+#endif  // TF_MAJOR_VERSION == 1 && TF_MINOR_VERSION >= 15
```
mzient, line 509: Missing new line at the end of file.

dali/pipeline/pipeline.h
awolant: Done

dali/pipeline/pipeline.h
awolant: Done

dali_tf_plugin/dali_dataset_op.cc
awolant: Done

awolant:
!build

awolant:
!build

awolant:
!build

awolant:
!build

awolant:
!build

awolant:
!build

Issue #1957, opened 2020-05-13
Hard to build the project from source


weberxie:
Hi,

I'm trying to build the project from source, I met a lot of problems such as `no package devtoolset-4 available`, and still can't build it success now, because of the specific network environment in China.

So, could you push the related base images to the docker repository, then we can pull them directly instead of build them from source.

Thanks!

szalpal:
Hi,

Have you followed [Docker build tutorial](https://docs.nvidia.com/deeplearning/dali/master-user-guide/docs/compilation.html#compiling-dali-from-source-using-docker-builder-recommended), or are you building bare-metal version?

RDruon:
Hello,

I was able to successfully build DALI using the Docker build instruction.
I changed the devtoolset-4 to devtoolset-7 in `docker/0001-An-approximate-manylinux3_x86_64.patch` file to be able to do it. 
Please find below the patch file I used
[DALI-devtoolset7-mod.patch.txt](https://github.com/NVIDIA/DALI/files/4626677/DALI-devtoolset7-mod.patch.txt)

weberxie:
yes, it did work after changed the devtoolset-4 to devtoolset-7. thanks for you kind reply.

JanuszL:
Hi,

We are going to deprecate CUDA 9 as CUDA 11 will be available pretty soon.
As a part of this transition, we can move to manylinux2014. If you are not building for CUDA 9 (and this was the only reason why we sicked to devtolset-4) you can check https://github.com/NVIDIA/DALI/pull/1965.

PR #1965, opened 2020-05-14
Change from a custom manylinux3 to prebuild and upstream manylinux2014

JanuszL:
- when cuda 9 is deprecated we can move to manylinux2014 as we can use the newer compiler than gcc 5.x
- CUDA 10 still required gcc 7 at most so we need to stick to it for all CUDA versions

#### Why we need this PR?
*Pick one, remove the rest*
- It changes from a custom manylinux3 to prebuild and upstream  manylinux2014

 - What solution was applied:
     when cuda 9 is deprecated we can move to manylinux2014 as we can use the newer compiler than gcc 5.x
     CUDA 10 still required gcc 7 at most so we need to stick to it for all CUDA versions
     build system
     NA
 - Validation and testing:
     local test, CI
     updated build instruction

In response to https://github.com/NVIDIA/DALI/issues/1957

**JIRA TASK**: *[NA]*


docker/Dockerfile.deps
```diff
@@ -160,7 +160,18 @@ RUN LIBSND_VERSION=1.0.28 && cd /tmp
...
-    cd /tmp && rm -rf libsnd-$LIBSND_VERSION
+    cd /tmp && rm -rf libsndfile-$LIBSND_VERSION
 
 # CUDA
 COPY --from=cuda /usr/local/cuda /usr/local/cuda
+
+# add devtoolset-7 to paths, it won't break anyting even if devtoolset-7 are not installed
+ENV ADDITIONAL_DEVTOOLSET_ROOTPATH /opt/rh/devtoolset-7/root
+ENV PATH $ADDITIONAL_DEVTOOLSET_ROOTPATH/usr/bin:$PATH
+ENV LD_LIBRARY_PATH $ADDITIONAL_DEVTOOLSET_ROOTPATH/usr/lib64:$ADDITIONAL_DEVTOOLSET_ROOTPATH/usr/lib:$ADDITIONAL_DEVTOOLSET_ROOTPATH/usr/lib64/dyninst:$ADDITIONAL_DEVTOOLSET_ROOTPATH/usr/lib/dyninst:$LD_LIBRARY_PATH
```
szalpal, line 171:
```suggestion
# add devtoolset-7 to paths, it won't break anyting even if devtoolset-7 are not installed
ENV ADDITIONAL_DEVTOOLSET_ROOTPATH=/opt/rh/devtoolset-7/root    \
    PATH=$ADDITIONAL_DEVTOOLSET_ROOTPATH/usr/bin:$PATH         \
    LD_LIBRARY_PATH=$ADDITIONAL_DEVTOOLSET_ROOTPATH/usr/lib64:$ADDITIONAL_DEVTOOLSET_ROOTPATH/usr/lib:$ADDITIONAL_DEVTOOLSET_ROOTPATH/usr/lib64/dyninst:$ADDITIONAL_DEVTOOLSET_ROOTPATH/usr/lib/dyninst:$LD_LIBRARY_PATH
```
Thanks to this we'll have only one additional layer in docker

JanuszL: Done

szalpal: You missed the `=` char ;)
```
[...]ROOTPATH=/opt/rh[...]
             ^ this one
```
Linking ref to docs: https://docs.docker.com/engine/reference/builder/#env

JanuszL: Done

JanuszL:
> LGTM, only comment is that maybe the docs in `Dockerfile.deps:1` would use some update

Done

dali/kernels/erase/erase_cpu.h
```diff
@@ -111,7 +127,7 @@ class EraseCpu {
     const T *in_ptr = in.data;
     T *out_ptr = out.data;
     if (out_ptr != in_ptr) {
-      std::memcpy(out_ptr, in_ptr, volume(shape) * sizeof(T));
+      detail::CopyImpl(out_ptr, in_ptr, volume(shape) * sizeof(T));
```
klecki, line 130: Isn't this error for anything with `sizeof(T) > 1`?

dali/kernels/erase/erase_cpu.h
```diff
@@ -30,6 +30,22 @@ namespace kernels {
...
+template <typename O>
+inline std::enable_if_t<std::is_trivially_copyable<O>::value, void>
+CopyImpl(O *out, const O *in, size_t num) {
+  std::memcpy(out, in, num);
+}
+
+template <typename O>
+inline std::enable_if_t<!std::is_trivially_copyable<O>::value, void>
+CopyImpl(O *out, const O *in, size_t num) {
+  for (size_t i = 0; i < num; ++i) {
```
klecki, line 42: With the usage below, you will go out of bounds as `num` is `count * sizeof(O)`

dali/kernels/erase/erase_cpu.h
JanuszL: ... ,right

dali/kernels/erase/erase_cpu.h
JanuszL: Fixed

PR #1971, opened 2020-05-19
Erase GPU operator 

banasraf:
#### Why we need this PR?
*Pick one, remove the rest*
- It adds a GPU erase operator needed for audio support. 

 - What solution was applied:
     *The gpu erase kernel was modified to support different number of erase regions for each sample. The operator implementation is mostly just instantiating the kernel.*
     *GPU erase kernel and a new file with GPU operator.*
     *Instantiating the kernel.*
 - Validation and testing:
     *Existing python test was extended to GPU. Kernel test was extended to cover different number of erase regions per sample.*
     *N/A*

**JIRA TASK**: *[DALI-1245]*

### DIFF
--- a/dali/kernels/erase/erase_gpu.h
+++ b/dali/kernels/erase/erase_gpu.h
@@ -71,7 +71,7 @@ template <typename T, int ndim>
 struct erase_sample_desc {
   const T *in = nullptr;
   T* out = nullptr;
-  span<ibox<ndim>> erase_regions = {};
+  span<const ibox<ndim>> erase_regions = {};
   ivec<ndim> sample_shape;
   ivec<ndim> sample_stride;
 };
@@ -379,10 +379,15 @@ struct EraseGpu {
 
   KernelRequirements Setup(KernelContext &context,
                            const InListGPU<T, ndim> &in,
-                           const std::vector<std::vector<ibox<ndim>>> &erased_regions,
-                           const span<const T> fill_values = {}) {
+                           const InListGPU<int32_t, 3> &erased_regions,
+                           span<const T> fill_values = {}) {
     const int num_samples = in.num_samples();
-
+    for (int i = 0; i < num_samples; ++i) {
+      auto tshape = erased_regions.tensor_shape(i);
+      DALI_ENFORCE(tshape[1] == 2 && tshape[2] == ndim, "The `erased_regions` has to be a "
+                                                        "TensorList of shape "
+                                                        "(regions count, 2, ndims).");
+    }
     if (channel_dim == -1) {
       DALI_ENFORCE(fill_values.size() == 1 || fill_values.size() == 0,
                    make_string("Kernel is unaware of channel dimension, exactly 0 or 1 fill value "
@@ -416,10 +421,6 @@ struct EraseGpu {
       se.add<T>(AllocType::GPU, 1);
     }
 
-    for (auto &regions : erased_regions) {
-      se.add<ibox<ndim>>(AllocType::GPU, regions.size());
-    }
-
     req.output_shapes = {in.shape};
     req.scratch_sizes = se.sizes;
     return req;
@@ -437,7 +438,7 @@ struct EraseGpu {
   void Run(KernelContext &ctx,
            OutListGPU<T, ndim> &out,
            const InListGPU<T, ndim> &in,
-           const std::vector<std::vector<ibox<ndim>>> &erased_regions,
+           const InListGPU<int32_t, 3> &erased_regions,
            span<const T> fill_values = {}) {
     auto stream = ctx.gpu.stream;  // MAYBE some runtime check if it's not used?
     const int num_samples = in.num_samples();
@@ -505,8 +506,9 @@ struct EraseGpu {
       auto &sample = sample_desc_cpu[i];
       sample.in = in.data[i];
       sample.out = out.data[i];
-      auto *regions_ptr = ctx.scratchpad->ToGPU(stream, erased_regions[i]);
-      sample.erase_regions = make_span(regions_ptr, erased_regions[i].size());
+      const auto *box_ptr = reinterpret_cast<const ibox<ndim>*>(erased_regions.tensor_data(i));
+      auto n_boxes = erased_regions.tensor_shape(i)[0];
+      sample.erase_regions = make_cspan(box_ptr, n_boxes);
       for (int dim = 0; dim < ndim; dim++) {
         sample.sample_shape[dim] = in.tensor_shape_span(i)[dim];
       }
--- a/dali/kernels/erase/erase_gpu.h
+++ b/dali/kernels/erase/erase_gpu.h
@@ -379,15 +379,10 @@ struct EraseGpu {
 
   KernelRequirements Setup(KernelContext &context,
                            const InListGPU<T, ndim> &in,
-                           const InListGPU<int32_t, 3> &erased_regions,
+                           const InListGPU<ibox<ndim>, 1> &erased_regions,
                            span<const T> fill_values = {}) {
     const int num_samples = in.num_samples();
-    for (int i = 0; i < num_samples; ++i) {
-      auto tshape = erased_regions.tensor_shape(i);
-      DALI_ENFORCE(tshape[1] == 2 && tshape[2] == ndim, "The `erased_regions` has to be a "
-                                                        "TensorList of shape "
-                                                        "(regions count, 2, ndims).");
-    }
+
     if (channel_dim == -1) {
       DALI_ENFORCE(fill_values.size() == 1 || fill_values.size() == 0,
                    make_string("Kernel is unaware of channel dimension, exactly 0 or 1 fill value "
@@ -438,7 +433,7 @@ struct EraseGpu {
   void Run(KernelContext &ctx,
            OutListGPU<T, ndim> &out,
            const InListGPU<T, ndim> &in,
-           const InListGPU<int32_t, 3> &erased_regions,
+           const InListGPU<ibox<ndim>, 1> &erased_regions,
            span<const T> fill_values = {}) {
     auto stream = ctx.gpu.stream;  // MAYBE some runtime check if it's not used?
     const int num_samples = in.num_samples();
@@ -506,9 +501,8 @@ struct EraseGpu {
       auto &sample = sample_desc_cpu[i];
       sample.in = in.data[i];
       sample.out = out.data[i];
-      const auto *box_ptr = reinterpret_cast<const ibox<ndim>*>(erased_regions.tensor_data(i));
-      auto n_boxes = erased_regions.tensor_shape(i)[0];
-      sample.erase_regions = make_cspan(box_ptr, n_boxes);
+      sample.erase_regions = make_cspan(erased_regions.tensor_data(i),
+                                        erased_regions.tensor_shape(i).num_elements());
       for (int dim = 0; dim < ndim; dim++) {
         sample.sample_shape[dim] = in.tensor_shape_span(i)[dim];
       }
--- a/dali/kernels/erase/erase_gpu.h
+++ b/dali/kernels/erase/erase_gpu.h
@@ -69,8 +69,8 @@ using ibox = box<int32_t, ndim>;
 
 template <typename T, int ndim>
 struct erase_sample_desc {
-  const T *in = nullptr;
-  T* out = nullptr;
+  const T * __restrict__ in = nullptr;
+  T * __restrict__ out = nullptr;
   span<const ibox<ndim>> erase_regions = {};
   ivec<ndim> sample_shape;
   ivec<ndim> sample_stride;
@@ -309,8 +309,8 @@ struct do_copy_or_erase {
 };
 
 template <int channel_dim = -1, typename T, int ndim = 2>
-__global__ void erase_gpu_impl(erase_sample_desc<T, ndim> *samples, ivec<ndim> region_shape,
-                               span<T> fill_values) {
+__global__ void erase_gpu_impl(const erase_sample_desc<T, ndim> *samples,
+                               ivec<ndim> region_shape, span<T> fill_values) {
   const auto region_idx = blockIdx.x;
   const auto sample_idx = blockIdx.y;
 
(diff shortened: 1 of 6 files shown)


dali/kernels/erase/erase_gpu.h
```diff
@@ -376,10 +378,9 @@ struct EraseGpu {
   using sample_t = erase_sample_desc<T, ndim>;
 
   KernelRequirements Setup(KernelContext &context,
-                          //  const TensorListView<StorageGPU, const T, ndim> &in
                            const InListGPU<T, ndim> &in,
-                           const InListGPU<ibox<ndim>, 1> &erased_regions,
-                           span<const T> fill_values = {}) {
+                           const std::vector<std::vector<ibox<ndim>>> &erased_regions,
```
klecki, line 382: Not really a fan of vector<vector<>>, especially as this data can be easily packed as one, contiguous allocation.

dali/kernels/erase/erase_gpu.h
```diff
@@ -503,24 +505,22 @@ struct EraseGpu {
       auto &sample = sample_desc_cpu[i];
       sample.in = in.data[i];
       sample.out = out.data[i];
-      sample.erase_regions = erased_regions.data[i];
+      auto *regions_ptr = ctx.scratchpad->ToGPU(stream, erased_regions[i]);
+      sample.erase_regions = make_span(regions_ptr, erased_regions[i].size());
```
klecki, line 509: Aren't you doing `num_samples` memcopies to device here? You can do just one, and still point to specific offset corresponding to erased_region of given sample.

dali/operators/generic/erase/erase.cu
```diff
@@ -0,0 +1,162 @@
...
+template <typename T, int Dims, int ChDim>
+struct select_impl<T, Dims, ChDim> {
+  static std::unique_ptr<OpImplBase<GPUBackend>> make(int channel_dim, const OpSpec &spec) {
+    if (channel_dim == ChDim) {
+      return std::make_unique<EraseImplGpu<T, Dims, ChDim>>(spec);
+    } else {
+      DALI_FAIL("Unsopported layout. Only channel-first and channel-last layouts are supported.");
+    }
+  }
+};
```
klecki, line 134: Maybe add a comment that it's used to select one of the allowed ChannelDim positions for those of a fainter heart.

dali/operators/generic/erase/erase.cu
```diff
@@ -0,0 +1,162 @@
...
+
+template <typename T, int Dims, int channel_dim>
+class EraseImplGpu : public OpImplBase<GPUBackend> {
+ public:
+  using EraseKernel = kernels::EraseGpu<T, Dims, channel_dim>;
+
+  explicit EraseImplGpu(const OpSpec &spec)
+      : spec_(spec), batch_size_(spec.GetArgument<int>("batch_size")) {
+        kmgr_.Resize<EraseKernel>(1, 1);
+      }
```
klecki, line 56: The alignment seems off.

dali/operators/generic/erase/erase.cu
```diff
@@ -0,0 +1,162 @@
...
+namespace detail {
+
+template <int ndim>
+ivec<ndim> to_ivec(const TensorShape<ndim> &shape) {
+  ivec<ndim> result;
+  for (int d = 0; d < ndim; d++) {
+    result[d] = shape[d];
+  }
+  return result;
+}
```
klecki, line 38: There is already the same function in `erase_gpu.h`.

dali/kernels/erase/erase_gpu.h
mzient: There should be a possibility to specify the regions directly in GPU memory - to make it possible to use GPU Shapes operator, for example.

dali/kernels/erase/erase_gpu.h
```diff
@@ -376,10 +378,9 @@ struct EraseGpu {
   using sample_t = erase_sample_desc<T, ndim>;
 
   KernelRequirements Setup(KernelContext &context,
-                          //  const TensorListView<StorageGPU, const T, ndim> &in
                            const InListGPU<T, ndim> &in,
-                           const InListGPU<ibox<ndim>, 1> &erased_regions,
-                           span<const T> fill_values = {}) {
+                           const std::vector<std::vector<ibox<ndim>>> &erased_regions,
+                           const span<const T> fill_values = {}) {
```
mzient, line 383: This is unnecessary - constness of span contents is determined by the element type, whereas the span itself is passed by value.

dali/kernels/erase/erase_gpu.h
```diff
@@ -432,7 +437,7 @@ struct EraseGpu {
   void Run(KernelContext &ctx,
            OutListGPU<T, ndim> &out,
            const InListGPU<T, ndim> &in,
-           const InListGPU<ibox<ndim>, 1> &erased_regions,
+           const std::vector<std::vector<ibox<ndim>>> &erased_regions,
```
mzient, line 440: Likewise - please revert to GPU-based data.

dali/kernels/erase/erase_gpu.h
```diff
@@ -70,7 +71,7 @@ template <typename T, int ndim>
 struct erase_sample_desc {
   const T *in = nullptr;
   T* out = nullptr;
-  const ibox<ndim> *erase_regions = nullptr;
+  span<ibox<ndim>> erase_regions = {};
```
mzient, line 74:
```suggestion
  span<const ibox<ndim>> erase_regions;
```

dali/operators/generic/erase/erase.cu
```diff
@@ -0,0 +1,162 @@
...
+      regions_[i].reserve(args[i].rois.size());
+      const auto &rois = args[i].rois;
+      for (const auto &roi : args[i].rois) {
+        regions_[i].push_back(detail::make_box(roi.anchor, roi.shape));
+      }
+    }
+  }
+
+ private:
+  OpSpec spec_;
```
mzient, line 99: That smells of bad design - base Operator class has this field - maybe we should just rework OpImplBase into something that doesn't require this kind of ugly tricks.

dali/operators/generic/erase/erase.cu
```diff
@@ -0,0 +1,162 @@
...
+  std::vector<std::vector<kernels::ibox<Dims>>> regions_;
+  SmallVector<T, 3> fill_values_;
+  kernels::KernelManager kmgr_;
+  kernels::KernelContext ctx_;
+};
+
+namespace detail {
+
+template <typename T, int Dims, int... ChDims>
+struct select_impl {};
```
mzient, line 112: Why does it need to be variadic at all? Is it even correct?
I think the code below is much easier to follow:
```
template <typename T, int Dim, typename... Args>
std::unique_ptr<OpImplBase<GPUBackend>> select_impl(int channel_dim, std::integral_constant<int, -1> chdim, const Args&... args) {
  if (channel_dim == chdim)
    return std::make_unique<EraseImplGPU<T, Dims, ChDim>(args...);
}

template <typename T, int Dim, int ChDim = Dim-1, typename... Args>
std::unique_ptr<OpImplBase<GPUBackend>> select_impl(int channel_dim, std::integral_constant<int, ChDim> chdim, const Args&... args) {
  if (channel_dim == chdim)
    return std::make_unique<EraseImplGPU<T, Dims, ChDim>(args...);
  else
     return select_impl<T, Dim>(channel_dim, std::integral_constant<int, ChDim-1>(), args...);
}
```

dali/kernels/erase/erase_gpu.h
banasraf: Why? It's an input argument, it will be supplied on CPU and I think it's kernel's responsibility to allocate GPU memory and move it there.

dali/operators/generic/erase/erase.cu
banasraf: I limit possible layouts to channel-first, channel-last and no channel, because compilation gets crazy long with all the combinations.

dali/operators/generic/erase/erase.cu
banasraf: done

dali/operators/generic/erase/erase.cu
banasraf: done

dali/operators/generic/erase/erase.cu
banasraf: changed to `const OpSpec&`

dali/kernels/erase/erase_gpu.h
banasraf: done

banasraf:
!build

dali/kernels/erase/erase_gpu.h
```diff
@@ -309,7 +310,7 @@ struct do_copy_or_erase {
 
 template <int channel_dim = -1, typename T, int ndim = 2>
 __global__ void erase_gpu_impl(erase_sample_desc<T, ndim> *samples, ivec<ndim> region_shape,
```
mzient, line 312:
```suggestion
__global__ void erase_gpu_impl(const erase_sample_desc<T, ndim> *samples, ivec<ndim> region_shape,
```

dali/kernels/erase/erase_gpu.h
```diff
@@ -503,24 +501,22 @@ struct EraseGpu {
...
 
-    CUDA_CALL(
-      cudaMemcpyAsync(sample_desc_gpu, sample_desc_cpu, num_samples * sizeof(sample_t),
-                      cudaMemcpyHostToDevice, stream));
-    CUDA_CALL(
-      cudaMemcpyAsync(fill_values_gpu, fill_values_cpu, num_fill_values * sizeof(T),
-                      cudaMemcpyHostToDevice, stream));
+    auto *sample_desc_gpu = ctx.scratchpad->ToGPU(stream, make_span(sample_desc_cpu, num_samples));
+    auto* fill_values_gpu =
+      ctx.scratchpad->ToGPU(stream, make_span(fill_values_cpu, num_fill_values));
```
mzient, line 514: Better option would be to use `ToContiguousGPU` - it will issue just one cudaMemcpy.

It's not in the scope of this task, I guess, however, if we keep some sane limit on number of channels, then the fill value could be copied to a `__constant__` - it should improve the performance, since the fill_value won't have to be read from global memory and will not compete for cache with input.

dali/kernels/erase/erase_gpu.h
```diff
@@ -70,7 +71,7 @@ template <typename T, int ndim>
 struct erase_sample_desc {
   const T *in = nullptr;
   T* out = nullptr;
```
mzient, line 73:
```suggestion
  const T *__restrict__ in = nullptr;
  T *__restrict__ out = nullptr;
```

dali/kernels/erase/erase_gpu.h
banasraf: done

dali/kernels/erase/erase_gpu.h
banasraf: Used ToContiguousGPU

dali/kernels/erase/erase_gpu.h
banasraf: done

dali/operators/generic/erase/erase.cu
```diff
@@ -0,0 +1,163 @@
...
+TensorListView<Storage, kernels::ibox<ndim>, 1>
+as_boxes(TensorListView<Storage, int32_t, 3> tlv) {
+  TensorListShape<1> new_shape(tlv.shape.size());
+  for (int i = 0; i < tlv.shape.size(); ++i)
+    new_shape.set_tensor_shape(i, {tlv.shape[i][0]});
+  using ptrs_t = kernels::ibox<ndim>* const*;
+  return {reinterpret_cast<ptrs_t>(tlv.data.data()), new_shape};
+}
+
+template <int ndim, typename Storage>
```
mzient, line 48: I don't think this overload is necessary - a non-const TensorListView should convert implicitly to const one.

dali/kernels/scratch_copy_impl.h
```diff
@@ -126,8 +126,8 @@ std::tuple<std::remove_cv_t<element_t<Collections>>*...>
 ToContiguousGPUMem(Scratchpad &scratchpad, cudaStream_t stream, const Collections &... c) {
   const size_t N = sizeof...(Collections);
   static_assert(
-    all_of<std::is_pod<std::remove_cv_t<element_t<Collections>>>::value...>::value,
-    "ToContiguousGPUMem must be used with collections of POD types");
+    all_of<std::is_trivially_copyable<std::remove_cv_t<element_t<Collections>>>::value...>::value,
```
mzient, line 129: There's one more is_pod in this file - please change it, too.

dali/operators/generic/erase/erase.cu
```diff
@@ -0,0 +1,163 @@
...
+                   TensorLayout in_layout) {
+    fill_values_ = spec_.template GetRepeatedArgument<float>("fill_value");
+    auto args = detail::GetEraseArgs<T, Dims>(spec_, ws, in_shape, in_layout);
+    auto regions_shape = TensorListShape<3>(batch_size_);
+    for (int i = 0; i < batch_size_; ++i) {
+      auto n_regions = static_cast<int>(args[i].rois.size());
+      regions_shape.set_tensor_shape(i, {n_regions, 2, Dims});
+    }
+    TensorList<CPUBackend> regions_cpu;
+    regions_cpu.set_type(TypeTable::GetTypeInfo(TypeTable::GetTypeID<int32_t>()));
```
klecki, line 107: Why not:
```suggestion
    regions_cpu.set_type(TypeInfo::Create<ibox<Dims>());
```
?

dali/operators/generic/erase/erase.cu
```diff
@@ -0,0 +1,163 @@
...
+    regions_cpu.Resize(regions_shape);
+    auto regions_tlv = detail::as_boxes<Dims>(view<int32_t, 3>(regions_cpu));
```
klecki, line 109: And you don't need this `as_boxes` thing at all, you can take a `view<ibox<Dims>>(regions_cpu)` and place the data in it directly. AFAIK it should work.

banasraf:
!build

dali/operators/generic/erase/erase.cu
banasraf: done

dali/operators/generic/erase/erase.cu
banasraf: done

PR #1974, opened 2020-05-21
Normalize GPU kernel

mzient:
#### Why we need this PR?
*Pick one, remove the rest*
- It adds NormalizeImplGPU because we want a GPU-based normalization

 - What solution was applied:
     * DropDims as a way to recalculate data/param coordinates
     * trivial slicing by data chunks
     * broadcasting using DropDims
     * full broadcasting (from host-side scalars) is a special case
     * NOTE: pImpl front-end kernel is not a part of this PR, it's in #1981 
     * Reductions - ReduceDims now uses fast_div
     * The kernel?
 - Validation and testing:
     * Unit tests (GTest)
     * Doxygen

**JIRA TASK**: DALI-1267

### DIFF
--- a/dali/kernels/normalize/normalize_gpu_impl.cuh
+++ b/dali/kernels/normalize/normalize_gpu_impl.cuh
@@ -115,7 +115,7 @@ struct NormalizeScalar {
  *        be regularized and inversed.
  *
  * The output elements are calculated as:
- * mul = 1 / sqrt(sqr(stddev[param_offset]) + epsilon)
+ * mul = 1 / sqrt(square(stddev[param_offset]) + epsilon)
  * (in[offset] - mean[param_offset]) * mul * scale + shift
  */
 template <int max_dims, typename Out, typename In, typename Mean, typename StdDev>
@@ -148,7 +148,7 @@ struct NormalizeInvStdDevNonScalar {
  *        be regularized and inversed.
  *
  * The output elements are calculated as:
- * mul = 1 / sqrt(sqr(stddev[param_offset]) + epsilon)
+ * mul = 1 / sqrt(square(stddev[param_offset]) + epsilon)
  * (in[offset] - mean[param_offset]) * mul * scale + shift
  */
 template <int max_dims, typename Out, typename In, typename StdDev>
--- a/dali/kernels/normalize/normalize_gpu_impl.cuh
+++ b/dali/kernels/normalize/normalize_gpu_impl.cuh
@@ -1,0 +1,372 @@
+// Copyright (c) 2020, NVIDIA CORPORATION. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef DALI_KERNELS_NORMALIZE_NORMALIZE_GPU_IMPL_CUH_
+#define DALI_KERNELS_NORMALIZE_NORMALIZE_GPU_IMPL_CUH_
+
+#include "dali/core/convert.h"
+#include "dali/core/small_vector.h"
+#include "dali/kernels/reduce/reduce_drop_dims.h"
+#include "dali/kernels/kernel.h"
+
+namespace dali {
+namespace kernels {
+
+namespace normalize_impl {
+
+template <int max_dims, typename Out, typename In>
+struct NormalizeNonScalar {
+  Out *__restrict__ out;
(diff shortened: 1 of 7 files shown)


mzient:
!build

dali/kernels/reduce/mean_stddev_gpu_impl_test.cu
```diff
@@ -348,9 +347,9 @@ TEST(InvStdDevImplGPU, Outer_Batch_Regularized) {
   EXPECT_GE(test.kernel.GetNumStages(), 2);  // both reduced axes must be split
   test.FillData(-100, 100);
 
-  test.Run(fake_mean.gpu(), 120);
+  test.Run(fake_mean.gpu(), 12000);
```
JanuszL, line 350: Any particular reason behind this change?

dali/kernels/reduce/reduce_drop_dims.h
```diff
@@ -241,18 +279,23 @@ struct DropDims {
     // Warning: intentional fall-through!
     #define REINDEX_CASE(idx)\
       case 2*idx:\
-        out += index / div[idx] * mul[idx];\
+        if (idx < kMaxDims)\
+          out += static_cast<uint64_t>(index) / div(idx) * mul[cuda_min(idx, kMaxDims-1)];\
```
JanuszL, line 283: How about creating `mul` function as well to have a consistent notation?
Idea, not a must.

dali/kernels/reduce/mean_stddev_gpu_impl_test.cu
mzient: Yes, I stopped squaring epsilon. It's how BatchNorm is implemented in every single document or framework I've looked at, so I decided to add the epsilon to variance as they do, even though having it in stddev terms may have seemd more intuitive (for once, the unit would be the same and it would be directly proportional to the signal's dynamic range, not to its square).

dali/kernels/reduce/reduce_drop_dims.h
mzient: actually, if I used mul as a function here (and mod function three lines below) I could get away without this silly `min(idx, kMaxDims-1)` which unnecessary, but clang is just not smart enough to silence the warning.

dali/kernels/normalize/normalize_gpu_impl_test.cu
```diff
@@ -0,0 +1,366 @@
...
+/**
+ * @brief Reference normalization of a single tensor
+ *
+ * If base/scale has an extent of 1 in any given dimension, it's broadcast along this axis.
+ *
+ * @param calc_inv_stddev if true, `scale` is assumed to contain standard deviation, which
+ *                        is subsequently regularized using given epsilon value
+ */
+template <typename Out, typename In>
+void RefNormalize(
```
JanuszL, line 80: What value `template <bool calc_inv_stddev, typename Out, typename In>
void RefNormalize` adds comparedto `template <typename Out, typename In>
void RefNormalize(` ?

dali/kernels/normalize/normalize_gpu_impl.cuh
```diff
@@ -0,0 +1,575 @@
...
+    if (in_shape.sample_dim() != ndim_)
+      throw std::invalid_argument("The input tensor list has different dimensionality than the "
+        "shape passed to Setup");
+    if (in_shape.num_samples() != num_samples_)
+      throw std::invalid_argument("The input tensor list has different number of samples than the "
+        "shape passed to Setup");
+  }
+
+  std::pair<dim3, dim3> GetLaunchParams(const TensorListShape<> &data_shape) const {
+    int64_t block = 1024;
```
JanuszL, line 422: Maybe we should query the GPU for capabilities - I mean user should and provide the necessary values to the kernel?

mzient:
!build

dali/kernels/normalize/normalize_gpu_impl_test.cu
mzient: There's an `if` in the inner loop - and I like it being a compile-time one.

dali/kernels/normalize/normalize_gpu_impl.cuh
```diff
@@ -0,0 +1,579 @@
...
+      throw std::invalid_argument("The input tensor list has different dimensionality than the "
+        "shape passed to Setup");
+    if (in_shape.num_samples() != num_samples_)
+      throw std::invalid_argument("The input tensor list has different number of samples than the "
+        "shape passed to Setup");
+  }
+
+  template <typename Desc, typename KernelFunc>
+  std::pair<dim3, dim3>
+  GetLaunchParams(const TensorListShape<> &data_shape, KernelFunc func) const {
```
JanuszL, line 423: I wonder how much we can generalize and reuse across all kernel launches?

dali/kernels/reduce/reduce_drop_dims.h
```diff
@@ -43,14 +45,58 @@ namespace reduce_impl {
  *
  * The reindexing is done by either dividing and multiplying by old/new strides or by takind modulo.
```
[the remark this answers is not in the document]
mzient: I'll fix it in the follow up.

PR #1975, opened 2020-05-22
Fix out-of-source build (missing include file)

szalpal:

#### Why we need this PR?
`float16.h` file, placed in `include` has `#include "dali/util/half.hpp"`. When building out-of-source DALI, this triggers an error, because of missing include - file is located in-source. This PR moves `half.hpp` to out-of-source deps.

 - What solution was applied:
     NA
     NA
     NA
 - Validation and testing:
     CI
     NA


szalpal:
!build

PR #1992, opened 2020-06-02
Build DALI with OpenMP SIMD

mzient:
Add openmp compiler flag, so we can use `#pragma omp simd` as a quick solution for vectorization.
Modify CPU Normalize kernel to use omp simd, to test that the build actually works.

#### Why we need this PR?
*Pick one, remove the rest*
- It adds new feature needed because we want DALI to run faster!

 - What solution was applied:
       * Add openmp compiler flag, so we can use `#pragma omp simd` as a quick solution for vectorization.
      * Modify CPU Normalize kernel to use omp simd, to test that the build actually works.
     * Build system
     * N/A
 - Validation and testing:
     * CPU Normalize tests
     * N/A

**JIRA TASK**: DALI-1449

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -15,6 +15,7 @@
 cmake_minimum_required(VERSION 3.13)
 # allow usage of check_symbol_exists() macro
 include(CheckCXXSymbolExists)
+include(CheckCXXCompilerFlag)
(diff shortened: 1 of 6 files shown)


mzient:
!build

PR #1997, opened 2020-06-04
Enable python ExternalSource operator for the GPU data

JanuszL:
- adds Python side support for GPU data feed to ExterenlSource operator
- extends ExternalSource example

#### Why we need this PR?
*Pick one, remove the rest*
- It adds support a GPU input to  te ExternalSource operator via Python API

 - What solution was applied:
      adds Python side support for GPU data feed to ExterenlSource operator
     extends ExternalSource example
     ExternalSource
     Python API
     backend_impl
     NA
 - Validation and testing:
     new CI tests are added
     example is extended

**JIRA TASK**: *[DALI-182]*


JanuszL:
!build

JanuszL:
!build

JanuszL:
!build

JanuszL:
!build

JanuszL:
!build

JanuszL:
!builld

JanuszL:
!builld

JanuszL:
!build

JanuszL:
!build

docs/examples/general/data_loading/external_input.ipynb
```diff
@@ -95,7 +95,7 @@
    "outputs": [],
```
awolant, line 95: <p>I'm not sure what happened here, but I always thought it should be <code>&lt;class 'cupy.ndarray'&gt;</code> </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/NVIDIA/DALI/pull/1997/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='15'/>

JanuszL: Good catch, I'm calling next on wrong iterator.

JanuszL: Done

dali/test/python/test_operator_element_extract.py
```diff
@@ -0,0 +1,96 @@
...
+
+        expected_first = test_data[i][0]
+        assert out1.shape == out3.shape
+        np.testing.assert_array_equal( expected_first, out1 )
+        np.testing.assert_array_equal( expected_first, out3 )
+
+        expected_last = test_data[i][F-1]
+        assert out2.shape == out4.shape
+        np.testing.assert_array_equal( expected_last, out2 )
+        np.testing.assert_array_equal( expected_last, out4 )
```
jantonguirao, line 96: empty line missing

dali/test/python/test_operator_external_source.py
```diff
@@ -1,275 +1 @@
...
-    pipe = Pipeline(1, 3, 0)
-
-    def gen():
-        for i in range(5):
-            yield [np.array([i + 1.5], dtype=np.float32)]
-
-    fn.external_source(gen(), cycle = False)     # no cycle - OK
-    with assert_raises(TypeError):
-        fn.external_source(gen(), cycle = True)  # cycle over generator - error expected
+from test_internals_operator_external_source import *
```
jantonguirao, line 1: and here

dali/test/python/test_gpu_operator_external_source.py
```diff
@@ -0,0 +1,2 @@
+from test_internals_operator_external_source import *
+use_cupy()
```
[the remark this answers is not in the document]
jantonguirao: Can you add a comment saying how this works (the fact that test_internals_operator_external_source checks what has been imported and uses cupy/numpy accordingly. Also, that it'd be enough for nosetests to find the test cases by just importing this file

docs/examples/general/data_loading/external_input.ipynb
jantonguirao, line 95: <p>The external source <strong>operator</strong> can also accept <strong>GPU data from CuPy </strong>...</p><p>For <strong>the</strong> sake of this example, we will create <strong>an</strong> ExternalInputGpuIterator in such a way that it returns data on the GPU already.</p><p>... decode it outside of DALI on <strong>the</strong> CPU</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/NVIDIA/DALI/pull/1997/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='12'/>

jantonguirao, line 95: <p>to <strong>use the GPU version of ExternalSource operator, removing the </strong>decoder as we assume that ...</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/NVIDIA/DALI/pull/1997/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='14'/>

jantonguirao, line 95: <p>barely visible. can we contrast argument that would produce something more pleasing to the eye?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/NVIDIA/DALI/pull/1997/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='18'/>

dali/test/python/test_internals_operator_external_source.py
```diff
@@ -0,0 +1,489 @@
...
+
+cupy_used = False
+
+def use_cupy():
+    global cp
+    global cupy_used
+    global datapy
+    import cupy as cp
+    cupy_used = True
+    datapy = cp
```
klecki, line 21: I think this file requires explanation at the top (maybe same for the ones that actually execute this code), that we use data as either nupy or cupy to test CPU and GPU inputs.

dali/python/nvidia/dali/pipeline.py
```diff
@@ -470,11 +476,20 @@ def feed_input(self, data_node, data, layout=""):
         if isinstance(data, list):
             inputs = []
             for datum in data:
-                inputs.append(Tensors.TensorCPU(datum, layout))
-            self._pipe.SetExternalTensorInput(name, inputs)
+                if hasattr(datum, "__cuda_array_interface__"):
+                    inp = Tensors.TensorGPU(datum, layout)
```
klecki, line 480: I wondered why we don't specify the `device_id` argument here in any fashion, following up on that it appears to be set to current device in the constructor, as by default cupy is working on the current device (where all allocations and memory transfers happen). We discussed that it may be worth some comments.

dali/pipeline/operator/builtin/external_source.h
```diff
@@ -123,20 +122,17 @@ class ExternalSource : public Operator<Backend> {
...
-  template<typename SrcBackend>
-  inline void SetDataSource(const TensorList<SrcBackend> &tl, cudaStream_t stream = 0) {
+  template<typename SrcBackend, template<typename> class SourceDataType>
+  inline void SetDataSourceHelper(const SourceDataType<SrcBackend> &t, cudaStream_t stream = 0) {
     if (std::is_same<SrcBackend, GPUBackend>::value && std::is_same<Backend, CPUBackend>::value) {
-      std::string msg = "Incorrect Backends warning. Loading GPU-originated data into CPU "
-                        "ExternalSource operator is discouraged and might be inefficient.";
-      DALI_WARN(msg);
+      DALI_WARN("Incorrect Backends warning. Loading GPU-originated data into CPU " +
+                "ExternalSource operator is discouraged and might be inefficient.");
```
mzient, line 129: There's no point in adding strings this way - subsequent literals will be merged at compile time.

dali/pipeline/operator/builtin/external_source.h
```diff
@@ -123,20 +122,17 @@ class ExternalSource : public Operator<Backend> {
...
     }
-    DALI_ENFORCE(OperatorBase::batch_size_ == static_cast<int>(tl.ntensor()),
-                 "Data list provided to ExternalSource needs to have batch_size length.");
+    DALI_ENFORCE(OperatorBase::batch_size_ == static_cast<int>(t.ntensor()),
+                 "Data list provided to ExternalSource needs to have batch_size = " +
```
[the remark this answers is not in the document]
mzient: :+1:

dali/python/backend_impl.cc
```diff
@@ -1064,11 +1073,26 @@ PYBIND11_MODULE(backend_impl, m) {
...
+                                 ? UserStream::Get()->GetStream(tl)
+                                 : static_cast<cudaStream_t>(ctypes_void_ptr(cuda_stream));
+          p->SetExternalInput(name, tl, stream);
+        },
+        "name"_a,
+        "list"_a,
+        "cuda_stream"_a = py::none())
     .def("SetExternalTensorInput",
-        [](Pipeline *p, const string &name, py::list list) {
+        [](Pipeline *p, const string &name, py::list list, py::object cuda_stream) {
```
klecki, line 1095: I think this could be handled by two overloads accepting std::vector<Tensor<CPUBackend>> and std::vector<Tensor<GPUBackend>> instead of `py::list list`. Than you probably don't need to resort to the magic pybind caster (at least not explicitly).

dali/pipeline/operator/builtin/external_source.h
```diff
@@ -123,20 +122,17 @@ class ExternalSource : public Operator<Backend> {
...
-   */
-  template<typename SrcBackend>
-  inline void SetDataSource(const TensorList<SrcBackend> &tl, cudaStream_t stream = 0) {
+  template<typename SrcBackend, template<typename> class SourceDataType>
+  inline void SetDataSourceHelper(const SourceDataType<SrcBackend> &t, cudaStream_t stream = 0) {
     if (std::is_same<SrcBackend, GPUBackend>::value && std::is_same<Backend, CPUBackend>::value) {
-      std::string msg = "Incorrect Backends warning. Loading GPU-originated data into CPU "
-                        "ExternalSource operator is discouraged and might be inefficient.";
-      DALI_WARN(msg);
+      DALI_WARN("Incorrect Backends warning. Loading GPU-originated data into CPU " +
```
klecki, line 128: I have a feeling that we agreed that this will be an error instead of warning, yet it was merged this way.
Also, we will spam the user with the warning every iteration, won't we?

klecki: And if we're just with the warning it probably should warn against the scenario of 
GPU input -> CPU external source -> GPU operator which is the main problem here.

dali/pipeline/operator/builtin/external_source.h
```diff
@@ -148,7 +144,7 @@ class ExternalSource : public Operator<Backend> {
       copy_to_storage_event = copy_to_storage_events_.GetEmpty();
     }
 
-    data.front()->Copy(tl, stream);
+    data.front()->Copy(t, stream);
     if (std::is_same<SrcBackend, GPUBackend>::value) {
       cudaEventRecord(*copy_to_storage_event.front(), stream);
```
klecki, line 149: hmm, shouldn't we wait for any event other than CPU->CPU (sync copy)?

dali/pipeline/operator/builtin/external_source.h
```diff
@@ -157,51 +153,31 @@ class ExternalSource : public Operator<Backend> {
       std::lock_guard<std::mutex> busy_lock(busy_m_);
       tl_data_.AddBack(data);
```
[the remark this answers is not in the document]
JanuszL: https://github.com/NVIDIA/DALI/pull/2024 we will have TensorVector readded so I would keep this name as it is.

dali/pipeline/operator/builtin/external_source.h
```diff
@@ -123,20 +122,17 @@ class ExternalSource : public Operator<Backend> {
...
 
-
-  /**
-   * @brief Sets the data that should be passed out of the op
-   * on the next iteration.
-   */
-  template<typename SrcBackend>
-  inline void SetDataSource(const TensorList<SrcBackend> &tl, cudaStream_t stream = 0) {
+  template<typename SrcBackend, template<typename> class SourceDataType>
+  inline void SetDataSourceHelper(const SourceDataType<SrcBackend> &t, cudaStream_t stream = 0) {
```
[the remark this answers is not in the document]
mzient: I'd keep `tl` - as I understand it's either a TensorList or a TensorVector.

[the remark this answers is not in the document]
JanuszL: This template can accept either TensorList or TensorVector, so t as TensorSomething was my first choice.

dali/pipeline/operator/builtin/external_source.h
```diff
@@ -157,51 +153,31 @@ class ExternalSource : public Operator<Backend> {
...
-    }
-
-    data.front()->resize(t.size());
-    for (size_t i = 0; i < t.size(); ++i) {
-      (*(data.front()))[i].Copy(t[i], stream);
-    }
-    if (std::is_same<SrcBackend, GPUBackend>::value) {
-      cudaEventRecord(*copy_to_storage_event.front(), stream);
-    }
+  inline void SetDataSource(const TensorList<SrcBackend> &tl, cudaStream_t stream = 0) {
```
mzient, line 165: There's no overload that accepts TensorVector and that would be nice for completeness. Maybe this function is not necessary and `SetDataSourceHelper` should be renamed `SetDataSource`? Overload resolution will pick the right overload (generic or one accepting vector of tensors).

dali/python/backend_impl.cc
JanuszL: See the comment:
```
// Note: This is a hack to get around weird casting
          // issues w/ pybind and a non-copyable type (dali::Tensor).
          // We cannot use pybind::cast<Tensor<CPUBackend>>
          // because somewhere through the chain of templates
          // pybind returns the calling template type, which
          // tries to call the deleted copy constructor for Tensor.
          // instead, we cast to a reference type and manually
          // move into the vector.
```

dali/pipeline/operator/builtin/external_source.h
JanuszL: @szalpal ?
> CPU external source -> GPU operator 

It would require checking the graph, what is not very trivial now and beyond the scope of this PR.
I think this warning is good enough to let know the user that he should not do that, and if he get it every iteration it won't slow things down as it is already nit the most efficient approach anyway.

dali/pipeline/operator/builtin/external_source_test.cc
```diff
@@ -98,34 +104,83 @@ class ExternalSourceTest : public::testing::WithParamInterface<int>,
     return dynamic_cast<ExternalSource<GPUBackend> *>(this->graph_.Node(OpType::GPU, 0).op.get());
   }
 
+  TensorShape<> GetRandShape(int dims) {
+    vector<Index> shape(dims, 0);
```
mzient, line 108:
```suggestion
    TensorShape<> shape;
    shape.resize(dims);
```

dali/python/backend_impl.cc
klecki: I thought it's the magic things that are happening below, but if it's the reason why we accept py::list here, than fine.

dali/pipeline/operator/builtin/external_source_test.cc
```diff
@@ -98,34 +104,83 @@ class ExternalSourceTest : public::testing::WithParamInterface<int>,
...
+    for (auto &val : shape) {
+      val = this->RandInt(1, 15);
+    }
+    return shape;
+  }
+
+  template<typename T>
+  void FeedWithCpuVector(T *src_op) {
```
mzient, line 116: Isn't that template overly generic?
```suggestion
  template<typename Backend>
  void FeedWithCpuVector(ExternalSource<Backend> *src_op) {
```

dali/pipeline/operator/builtin/external_source_test.cc
```diff
@@ -98,34 +104,83 @@ class ExternalSourceTest : public::testing::WithParamInterface<int>,
...
+        data[i] = fill_counter_;
+        ++fill_counter_;
+      }
+    }
+    src_op->SetDataSource(vt_cpu_);
+  }
+
   template<typename T>
-  void FeedWithVector(T *src_op) {
+  void FeedWithGpuVector(T *src_op) {
```
mzient, line 133: likewise

dali/pipeline/operator/builtin/external_source_test.cc
```diff
@@ -98,34 +104,83 @@ class ExternalSourceTest : public::testing::WithParamInterface<int>,
...
+    cudaStreamSynchronize(0);
+    src_op->SetDataSource(vt_gpu_);
   }
 
   template<typename T>
-  void FeedWithList(T *src_op) {
-    tl_.set_type(TypeInfo::Create<int>());
-    TensorListShape<> shape = uniform_list_shape(this->batch_size_, {10, 10});
-    tl_.Resize(shape);
+  void FeedWithCpuList(T *src_op) {
```
mzient, line 152: and once more

dali/pipeline/operator/builtin/external_source_test.cc
```diff
@@ -98,34 +104,83 @@ class ExternalSourceTest : public::testing::WithParamInterface<int>,
...
+        ++fill_counter_;
       }
-      ++fill_counter_;
     }
-    src_op->SetDataSource(tl_);
+    src_op->SetDataSource(tl_cpu_);
+  }
+
+  template<typename T>
+  void FeedWithGpuList(T *src_op) {
```
mzient, line 168: ...and again.

dali/python/backend_impl.cc
```diff
@@ -864,6 +864,15 @@ py::dict MetaToDict(const ReaderMeta &meta) {
   return d;
 }
 
+template <typename Backend>
+void FeedPipeline(Pipeline *p, const string &name, py::list list, cudaStream_t stream) {
+  vector<Tensor<Backend>> tensors(list.size());
+  for (size_t i = 0; i < list.size(); ++i) {
+    tensors[i] = std::move(list[i].cast<Tensor<Backend>&>());
+  }
+  p->SetExternalInput(name, tensors, stream);
```
mzient, line 873: Since you're creating this vector by hand, you can create a TensorVector as well and avoid extra conversion step.

dali/python/nvidia/dali/external_source.py
```diff
@@ -172,13 +173,20 @@ class ExternalSource():
     If provided, sets the layout of the data. When ``num_outputs > 1``, layout can be a list
     containing a distinct layout for each output. If the list has fewer elements than ``num_outputs``,
     only the first outputs have the layout set, the reset have it cleared.
+
+`cuda_stream` : Any value that can be casted to cudaStream_t
+    CUDA stream to be used for the copy (for the GPU input, for CPU is disregarded)
+    (if not provided, an internal user stream will be selected)
```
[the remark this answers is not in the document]
mzient: What does "internal user stream" mean? Just "internal stream"?

dali/python/nvidia/dali/pipeline.py
```diff
@@ -455,6 +455,12 @@ def feed_input(self, data_node, data, layout=""):
             It should be a string of the length that matches the dimensionality of the data, batch
             dimension excluded. For a batch of channel-first images, this should be "CHW", for
             channel-last video it's "FHWC" and so on.
+
+        `cuda_stream` : Any value that can be casted to cudaStream_t
+                    CUDA stream to be used for the copy (for the GPU input, for CPU is disregarded)
+                    (if not provided, an internal user stream will be selected)
```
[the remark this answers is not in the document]
mzient: likewise - internal stream?

dali/python/backend_impl.cc
mzient, line 873: Can you create TensorVector here and avoid the extra conversion step?

dali/pipeline/operator/builtin/external_source.cc
```diff
@@ -18,53 +18,24 @@ namespace dali {
...
-    } else {
-        DALI_ENFORCE(!t_data_.IsEmpty(), "ExternalSource is empty. Need to feed data first.");
-        vector_tensor_elm = t_data_.PopFront();
-        DALI_ENFORCE(OperatorBase::batch_size_ ==
-                     static_cast<int>(vector_tensor_elm.front()->size()),
-          "Data list provided to ExternalSource needs to have batch_size length = " +
-          std::to_string(batch_size_) + " length, found " +
-          std::to_string(static_cast<int>(vector_tensor_elm.front()->size())) + " samples.");
-    }
+    cv_.wait(busy_lock, [&data = tl_data_]{return !data.IsEmpty();});
```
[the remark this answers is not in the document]
JanuszL: Done

dali/pipeline/operator/builtin/external_source.cu
```diff
@@ -47,26 +47,23 @@ struct ExternalSource<GPUBackend>::RecycleFunctor {
...
+  std::list<uptr_tl_type> tensor_list_elm;
   std::list<uptr_cuda_event_type> cuda_event, internal_copy_to_storage;
   {
     std::unique_lock<std::mutex> busy_lock(busy_m_);
-    cv_.wait(busy_lock, [&data = data_in_tl_]{return !data.empty();});
-    auto is_data_in_tl = data_in_tl_.front();
-    data_in_tl_.pop_front();
-    DALI_ENFORCE(is_data_in_tl, "Cannot feed non-contiguous data to GPU op.");
-    data = tl_data_.PopFront();
+    cv_.wait(busy_lock, [&data = tl_data_]{return !data.IsEmpty();});
```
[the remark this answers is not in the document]
JanuszL: Done

dali/pipeline/operator/builtin/external_source.h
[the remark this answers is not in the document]
JanuszL: Done

dali/python/backend_impl.cc
```diff
@@ -1077,14 +1101,25 @@ PYBIND11_MODULE(backend_impl, m) {
...
-          vector<Tensor<CPUBackend>> tensors(list.size());
           DALI_ENFORCE(p->batch_size() == static_cast<int>(list.size()),
              "Data list provided to feed_input needs to have batch_size length.");
-          for (size_t i = 0; i < list.size(); ++i) {
-            tensors[i] = std::move(list[i].cast<Tensor<CPUBackend>&>());
+
+          bool is_cpu_data = true;
+          // not the most beautiful  but at least it doesn't throw as plain cast<T>()
+          py::detail::make_caster<Tensor<CPUBackend>&> conv;
+          is_cpu_data = conv.load(static_cast<py::object>(list[0]), true);
```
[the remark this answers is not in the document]
JanuszL: Done

dali/python/nvidia/dali/external_source.py
```diff
@@ -172,13 +173,20 @@ class ExternalSource():
     If provided, sets the layout of the data. When ``num_outputs > 1``, layout can be a list
     containing a distinct layout for each output. If the list has fewer elements than ``num_outputs``,
     only the first outputs have the layout set, the reset have it cleared.
+
+`cuda_stream` : Any value that can be casted to cudaStream_t
+    CUDA stream to be used for the copy (for the GPU input, for CPU is disregarded)
```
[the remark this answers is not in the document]

dali/python/nvidia/dali/external_source.py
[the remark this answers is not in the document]
JanuszL: `internal stream`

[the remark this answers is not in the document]

dali/python/nvidia/dali/pipeline.py
```diff
@@ -455,6 +455,12 @@ def feed_input(self, data_node, data, layout=""):
             It should be a string of the length that matches the dimensionality of the data, batch
             dimension excluded. For a batch of channel-first images, this should be "CHW", for
             channel-last video it's "FHWC" and so on.
+
+        `cuda_stream` : Any value that can be casted to cudaStream_t
+                    CUDA stream to be used for the copy (for the GPU input, for CPU is disregarded)
```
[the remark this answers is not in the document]

dali/python/nvidia/dali/pipeline.py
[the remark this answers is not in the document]

dali/test/python/test_gpu_operator_external_source.py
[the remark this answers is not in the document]
JanuszL: Done

dali/test/python/test_operator_element_extract.py
JanuszL: Done

dali/test/python/test_operator_external_source.py
JanuszL: Done

docs/examples/general/data_loading/external_input.ipynb
JanuszL: Done

JanuszL: Done

JanuszL: Done

dali/test/python/test_internals_operator_external_source.py
JanuszL: Done

dali/python/nvidia/dali/pipeline.py
JanuszL: Done

dali/pipeline/operator/builtin/external_source.h
JanuszL: Done

dali/pipeline/operator/builtin/external_source.h
JanuszL: Done

dali/pipeline/operator/builtin/external_source_test.cc
JanuszL: Done

dali/pipeline/operator/builtin/external_source_test.cc
JanuszL: Done

dali/pipeline/operator/builtin/external_source_test.cc
JanuszL: Done

dali/pipeline/operator/builtin/external_source_test.cc
JanuszL: Done

dali/pipeline/operator/builtin/external_source_test.cc
JanuszL: Done

dali/python/backend_impl.cc
JanuszL: Done

dali/pipeline/operator/builtin/external_source.h
JanuszL: Added waiting for GPU to anything, CPU pinned to GPU.

JanuszL:
!build

dali/pipeline/operator/builtin/external_source.h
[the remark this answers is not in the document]
szalpal: How about `batch`? Might be even better

dali/pipeline/operator/builtin/external_source.h
szalpal: We in fact agreed to have an error here. However, I went with warning, so DaliBackend development reasons - needed to verify GPU input there and I thought this way be easier than figuring out how to combine them with not-ready-yet python-level GPU input.

On the second thought for our previous discussion, I'd go for warning. There's no point of forbidding something just because it's tad slow and on the other hand it might come in handy when developing (e.g. there's GPU input, but not all GPU ops are ready, so we need partially CPU pipeline)

dali/pipeline/operator/builtin/external_source.h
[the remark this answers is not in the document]
JanuszL: Could be.

szalpal:
> Same can probably happen for GPU data and CPU ExternalSource.
> The SetExternalInput is non-blocki, apparently there was idea that the user can reuse the memory by observing the stream he provided by this is not documented (apparently @szalpal can say more about this) - we probably need to highlight that in the doc.

Correct, when user calls `daliSetExternalInputAsync` and synchronizes on the stream provided there, the memory he provided should not be needed anymore by DALI. But that's rather something obvious, isn't it? I mean, there are in fact 2 functions - sync and async - provided, but that's only for convenience. Stream management here is added, because the copy might be expensive, but user might be able to use few different streams to perform it, to make it faster.

JanuszL:
> I think there are potential ownership issues if we call SetExternalInput and we deal with anything that does async copy as we synchronize on the event only in run (and we make the stream wait for it). So if we decide to do this for GPU->GPU copy and we change/deallocate the memory before DALI's async copy to ExternalSource queue is finished.
> 
> Same can probably happen for GPU data and CPU ExternalSource.
> The SetExternalInput is non-blocki, apparently there was idea that the user can reuse the memory by observing the stream he provided by this is not documented (apparently @szalpal can say more about this) - we probably need to highlight that in the doc.
> 
> Also, if the user doesn't provide the stream for setting the external input we use some UserStream that is DALI-internal thing. User now has no idea when he can touch his memory again in a safe manner.
> 
> One solution would be to make SetExternalInput blocking in such case (till the internal copy is finished) or providing some other mechanism.

I added a `sync` parameter that is used when user doesn't provide the stream.

dali/python/backend_impl.cc
```diff
@@ -302,7 +311,7 @@ void ExposeTensor(py::module &m) {
       layout : str
             Layout of the data
       device_id: int
-            Device of where this tensor resides
+            Device of where this tensor resides. If no is provided the current device is used.
```
jantonguirao, line 314:
```suggestion
            Device of where this tensor resides. If not provided, the current device is used.
```

dali/python/backend_impl.cc
```diff
@@ -618,16 +631,16 @@ void ExposeTensorList(py::module &m) {
...
       Parameters
       ----------
       object : object
             Python object that implement CUDA Array Interface
       layout : str
             Layout of the data
-      device_id: int
-            Device of where this lists of tensors resides
+      device_id : int
+            Device of where this tensor resides. If no is provided the current device is used.
```
jantonguirao, line 643: same here

include/dali/c_api.h
```diff
@@ -139,13 +139,14 @@ DLL_PUBLIC void daliDeserializeDefault(daliPipelineHandle *pipe_handle,
...
  *               provided stream.
+ * @param sync If block until data provided is copied to the internal DALI buffer
  */
 DLL_PUBLIC void
 daliSetExternalInputAsync(daliPipelineHandle *pipe_handle, const char *name,
                           device_type_t device, const void *data_ptr,
                           dali_data_type_t data_type, const int64_t *shapes,
                           int sample_dim, const char *layout_str,
-                          cudaStream_t stream);
+                          cudaStream_t stream, int sync);
```
klecki, line 149: There is a question if we still need both async and sync variants of this function if this can be handled by the parameter.

JanuszL: Sync has own stream, in this variant you still need to provide one.

dali/python/backend_impl.cc
JanuszL: Done

dali/python/backend_impl.cc
JanuszL: Done

include/dali/c_api.h
klecki: Ok, makes sense.

include/dali/c_api.h
```diff
@@ -139,13 +139,14 @@ DLL_PUBLIC void daliDeserializeDefault(daliPipelineHandle *pipe_handle,
  *                   Can be set to NULL.
  * @param stream CUDA stream to use when copying the data onto GPU. Remember to synchronize on the
  *               provided stream.
+ * @param sync If block until data provided is copied to the internal DALI buffer
```
klecki, line 142:
```suggestion
 * @param sync Whether to block until the provided data is copied to the internal DALI buffer
```

dali/pipeline/operator/builtin/external_source.h
klecki: Cool. Thanks for adding the comment.

dali/pipeline/operator/builtin/external_source.cc
```diff
@@ -19,11 +19,20 @@ namespace dali {
 template<>
 void ExternalSource<CPUBackend>::RunImpl(HostWorkspace &ws) {
   std::list<uptr_tl_type> tensor_list_elm;
+  std::list<uptr_cuda_event_type> internal_copy_to_storage;
   {
     std::unique_lock<std::mutex> busy_lock(busy_m_);
-    cv_.wait(busy_lock, [&data = tl_data_]{return !data.IsEmpty();});
+    cv_.wait(busy_lock, [&data = tl_data_, &blocking = blocking_] {
```
klecki, line 25: I think it's better to capture the `blocking_` as copy instead of reference as it wouldn't change.

dali/pipeline/operator/builtin/external_source.cc
```diff
@@ -19,11 +19,20 @@ namespace dali {
...
+        return !(data.IsEmpty() && blocking);
```
klecki, line 26: We can also alternatively:
```
if (blocking) {
	cv_.wait(busy_lock, [&data = tl_data_]{return !data.IsEmpty();});
} else {
    // we have the lock, fail if there is no data.
}
```

dali/python/backend_impl.cc
```diff
@@ -1075,7 +1089,7 @@ PYBIND11_MODULE(backend_impl, m) {
     .def("SetExternalTLInput",
         [](Pipeline *p, const string &name, const TensorList<CPUBackend> &tl,
            py::object /*cuda_stream*/) {
-          p->SetExternalInput(name, tl, 0);
+          p->SetExternalInput(name, tl, 0, true);
```
klecki, line 1092: I assume it is fine to pass a false here as well (or rather it just doesn't matter)?

klecki: But it's more aligned with the docs that claim to be blocking.

dali/pipeline/operator/builtin/external_source.cu
```diff
@@ -51,7 +51,12 @@ void ExternalSource<GPUBackend>::RunImpl(DeviceWorkspace &ws) {
   std::list<uptr_cuda_event_type> cuda_event, internal_copy_to_storage;
   {
     std::unique_lock<std::mutex> busy_lock(busy_m_);
-    cv_.wait(busy_lock, [&data = tl_data_]{return !data.IsEmpty();});
+    cv_.wait(busy_lock, [&data = tl_data_, &blocking = blocking_] {
+        return !(data.IsEmpty() && blocking);
+      });
+    if (!blocking_ && tl_data_.IsEmpty()) {
+      DALI_FAIL("No data was provided to the ExternalSource. Make sure to feed it properly.");
+    }
```
klecki, line 59: Same suggestion as for the CPU.

JanuszL:
> The `sync` for the SetExternalInput seems to solve the problems, I'm not sure how I feel about the name and behaviour of the `blocking` ExternalSource parameter, if it won't bring some confusion.

I'm open for suggestions regarding the naming.

dali/python/backend_impl.cc
JanuszL: It is CPU so it should sync anyway, so it is better to pass the true to be aligned with what happens under the hood.

dali/pipeline/operator/builtin/external_source.cu
JanuszL: Dona

dali/pipeline/operator/builtin/external_source.cc
JanuszL: Dona

dali/pipeline/operator/builtin/external_source.cc
JanuszL: Dona

include/dali/c_api.h
JanuszL: Dona

JanuszL:
!build

PR #2009, opened 2020-06-08
Separable convolution

klecki:
#### Why we need this PR?
Adds separable convolution using Convolution Kernel, passing over all axes.

 - What solution was applied:
     SeparableConvolution kernel build by using several passes of ConvolutionKernel
     Kernels, kernels tests
     Nothing fancy, mostly boilerplate
 - Validation and testing:
     Gtest for kernel, baseline implementation moved to separate file.

**JIRA TASK**: *[DALI-1425]*

### DIFF
--- a/dali/kernels/imgproc/convolution/separable_convolution_cpu.h
+++ b/dali/kernels/imgproc/convolution/separable_convolution_cpu.h
@@ -1,0 +1,202 @@
+// Copyright (c) 2020, NVIDIA CORPORATION. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef DALI_KERNELS_IMGPROC_CONVOLUTION_SEPARABLE_CONVOLUTION_CPU_H_
+#define DALI_KERNELS_IMGPROC_CONVOLUTION_SEPARABLE_CONVOLUTION_CPU_H_
+
+#include "dali/core/convert.h"
+#include "dali/core/format.h"
+#include "dali/core/tensor_view.h"
+#include "dali/kernels/common/utils.h"
+#include "dali/kernels/imgproc/convolution/convolution_cpu.h"
+#include "dali/kernels/kernel.h"
+#include "dali/pipeline/util/operator_impl_utils.h"
+
+namespace dali {
+namespace kernels {
+
+/**
+ * @brief Apply convolution in all axes, starting from the innermost to outermost.
+ *        If channel axis is pressent, the convolution is not applied there.
+ *
+ * If `Out` is same as `W` the intermediate stages are written to out,
+ * otherwise the temporary buffer of `W` is required and allocated in scratchpad
+ *
+ * `W` is currently assumed to be float
+ *
+ * `windows` and `scales` are specified per axis.
+ *
+ * Specialized for 1, 2 or 3 axes, to not go overboard with TMP for generic solutions
+ *
+ * TODO(klecki): For more dimension, fusing a permute step when writing the result
+ * could allow for processing all steps with innermost, contiguous dimension.
+ * For example DHWC->DWHC->HWDC->DHWC, while applying convolutions for W, H, D respectively.
+ */
+template <typename Out, typename In, typename W, int ndim, bool has_channels = true>
+struct SeparableConvolutionCpu;
+
+template <typename Out, typename W, int ndim>
+std::enable_if_t<!std::is_same<Out, W>::value, TensorView<StorageCPU, W, ndim>> GetInterStageView(
+    const TensorView<StorageCPU, Out, ndim>& out, W* scratch) {
+  return {scratch, out.shape};
+}
+
+template <typename W, int ndim>
+TensorView<StorageCPU, W, ndim> GetInterStageView(const TensorView<StorageCPU, W, ndim>& out,
+                                                  W* scratch) {
+  return out;
+}
+
+template <typename Out, typename In, typename W, int axes, bool has_channels>
+struct SeparableConvolutionCpuImpl;
+
+
+template <typename Out, typename In, typename W, bool has_channels>
+struct SeparableConvolutionCpuImpl<Out, In, W, 1, has_channels> {
+  static constexpr int axes = 1;
+  static constexpr int ndim = has_channels ? 2 : 1;
+
+  KernelRequirements Setup(KernelContext& ctx, const TensorShape<ndim>& in_shape,
+                           const std::array<int, axes>& window_sizes) {
+    KernelRequirements req;
+    req.output_shapes.push_back(uniform_list_shape<ndim>(1, in_shape));
+
+    auto req_conv = conv_.Setup(ctx, in_shape, window_sizes[0]);
+    req.AddInputSet(req_conv, false);
+
+    return req;
+  }
+
+  void Run(KernelContext& ctx, const TensorView<StorageCPU, Out, ndim> out,
+           const TensorView<StorageCPU, const In, ndim>& in,
+           const std::array<TensorView<StorageCPU, const W, 1>, axes>& windows,
+           const std::array<W, axes>& scales = uniform_array<axes, W>(1.f)) {
+    conv_.Run(ctx, out, in, windows[0], scales[0]);
+  }
+
+  ConvolutionCpu<Out, In, W, ndim, 0, has_channels> conv_;
+  static_assert(std::is_same<W, float>::value,
+                "Only floats as intermediate values are currently supported.");
+};
+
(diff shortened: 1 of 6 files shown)


klecki:
!build

klecki:
!build

dali/kernels/imgproc/convolution/separable_convolution_cpu.h
```diff
@@ -0,0 +1,202 @@
...
+    req.AddInputSet(req_middle, false);
+    req.AddInputSet(req_outer, false);
+
+    return req;
+  }
+
+  void Run(KernelContext& ctx, const TensorView<StorageCPU, Out, ndim> out,
+           const TensorView<StorageCPU, const In, ndim>& in,
+           const std::array<TensorView<StorageCPU, const W, 1>, axes>& windows,
+           const std::array<W, axes>& scales = uniform_array<axes, W>(1.f)) {
```
mzient, line 161: Any real life use case for per-axis scale? They will be multiplied anyway.

dali/kernels/imgproc/convolution/separable_convolution_cpu.h
```diff
@@ -0,0 +1,202 @@
...
+struct SeparableConvolutionCpuImpl<Out, In, W, 2, has_channels> {
+  static constexpr int axes = 2;
+  static constexpr int ndim = has_channels ? 3 : 2;
+
+  KernelRequirements Setup(KernelContext& ctx, const TensorShape<ndim>& in_shape,
+                           const std::array<int, axes>& window_sizes) {
+    KernelRequirements req;
+
+    ScratchpadEstimator se;
+    se.add<W>(AllocType::Host, volume(in_shape));
```
mzient, line 103: Why do you want the intermedate image to be of type W? What if you have `float` input/output and integer weights?

dali/kernels/imgproc/convolution/separable_convolution_cpu.h
```diff
@@ -0,0 +1,202 @@
...
+struct SeparableConvolutionCpu<Out, In, W, 3, true>
+    : public SeparableConvolutionCpuImpl<Out, In, W, 2, true> {};
+
+template <typename Out, typename In, typename W>
+struct SeparableConvolutionCpu<Out, In, W, 3, false>
+    : public SeparableConvolutionCpuImpl<Out, In, W, 3, false> {};
+
+template <typename Out, typename In, typename W>
+struct SeparableConvolutionCpu<Out, In, W, 4, true>
+    : public SeparableConvolutionCpuImpl<Out, In, W, 3, true> {};
```
mzient, line 197: What is it for? Why not just rename XxxImpl to Xxx?

dali/kernels/imgproc/convolution/separable_convolution_cpu.h
```diff
@@ -0,0 +1,202 @@
...
+ *
+ * `W` is currently assumed to be float
+ *
+ * `windows` and `scales` are specified per axis.
+ *
+ * Specialized for 1, 2 or 3 axes, to not go overboard with TMP for generic solutions
+ *
+ * TODO(klecki): For more dimension, fusing a permute step when writing the result
+ * could allow for processing all steps with innermost, contiguous dimension.
+ * For example DHWC->DWHC->HWDC->DHWC, while applying convolutions for W, H, D respectively.
```
mzient, line 44: I wouldn't mark it as TODO - it doesn't seem like a good idea to transpose on the fly; when we employ some automatic vectorization, it will most certainly be defeated by transposition.

dali/kernels/imgproc/convolution/separable_convolution_cpu.h
mzient: Also, you don't need this buffer when your intermediate element type is the same as Out.

dali/kernels/imgproc/convolution/separable_convolution_cpu.h
```diff
@@ -0,0 +1,202 @@
...
+struct SeparableConvolutionCpuImpl<Out, In, W, 3, has_channels> {
+  static constexpr int axes = 3;
+  static constexpr int ndim = has_channels ? 4 : 3;
+
+  KernelRequirements Setup(KernelContext& ctx, const TensorShape<ndim>& in_shape,
+                           const std::array<int, axes>& window_sizes) {
+    KernelRequirements req;
+
+    ScratchpadEstimator se;
+    se.add<W>(AllocType::Host, volume(in_shape));
```
mzient, line 143: Likewise - you may live without this buffer.

include/dali/core/util.h
```diff
@@ -344,6 +345,14 @@ static_assert(all_of<true, true, true>::value,
 static_assert(!all_of<true, false, true>::value,
               "Should return false_type when any of the values is false.");
 
+
+template <size_t N, typename T>
+std::array<T, N> uniform_array(const T& t) {
```
JanuszL, line 350: How it differs from `ConstantFill` in tensor_test_utils.h? If not please reuse.

dali/kernels/imgproc/convolution/separable_convolution_cpu.h
klecki: Not really, got carried away, will simplify.

dali/kernels/imgproc/convolution/separable_convolution_cpu.h
klecki: > Why do you want the intermediate image to be of type W? What if you have float input/output and integer weights?

For now I assume that `W` is float. It can be generalized as well, and we can parametrize every possible step here with configurable type. Do you want me to do it?

> Also, you don't need this buffer when your intermediate element type is the same as Out.

Yes, I don't need it, but the in-place for first step will be slower, and for everything other than W==Out it's still needed so I just opted for that.

dali/kernels/imgproc/convolution/separable_convolution_cpu.h
klecki: I know, but probably I would get a comment that it can be faster. I left it same for all variants for simplicity as well.

include/dali/core/util.h
klecki: ConstantFill is a testing utility. It takes a TensorView as a parameter and fills it based on it's shape.
This thing returns an std::array. I used it, because I wanted to have a way to express an uniform default value as a parameter to the filters.
I can't use span there, as it cannot be created from temporary. When I use const reference, the return value from this function is captured and it's lifetime is long enough to make it work.

dali/kernels/imgproc/convolution/separable_convolution_cpu.h
klecki: So the Kernel is parametrized with the number of actual dimensions and not only data dimensions + bool for the channels.

include/dali/core/util.h
JanuszL: https://github.com/NVIDIA/DALI/blob/master/dali/test/tensor_test_utils.h#L341 this one works for iterable, you can create a std::array outside and call `ConstantFill` on it, or I have missed something.

dali/kernels/imgproc/convolution/separable_convolution_cpu.h
klecki: As for the intermediate data, maybe it would indeed be better to just store the result of arithmetic operation. Will do that.

include/dali/core/util.h
klecki: The point here is to do
```
void foo(const std::array<float, 5> &value = uniform_array<5>(1.0f));
```
So you have a default value of array parameter to function. Not really possible with ConstantFill.

JanuszL: Understood. Still the question is if `include/dali/core/util.h`  is the place for the test utils.
@mzient ?

klecki: This is not test utility.

JanuszL: Right, after all we can use it to fill arguments array or something like this.

dali/kernels/imgproc/convolution/separable_convolution_cpu.h
klecki: Done

klecki:
!build

dali/kernels/imgproc/convolution/separable_convolution_cpu.h
mzient: It's still there - simply missed or intentional?

dali/kernels/imgproc/convolution/separable_convolution_cpu.h
mzient: It's a personal opinion, but I find it _harder_ to use this way - I mean, I find it more intuitive to parameterize 2D convolution with channels and without channels with <2>, not <2> or <3>. From the operator standpoint, you still need to handle it explicitly, either in value switch or an `if`, so there's no difference.

dali/kernels/imgproc/convolution/separable_convolution_cpu.h
klecki: Missed, now it's here

klecki:
!build

klecki:
!build

klecki:
!build

dali/kernels/imgproc/convolution/separable_convolution_cpu.h
klecki: Removed, now it's less code which has some benefits.

klecki:
!build

PR #2024, opened 2020-06-15
Add zero-copy to the ExternalSource operator

JanuszL:
- adds `no_copy` option so it is possible to skip copying data provided  by the user to the internal staging buffer but it is the user  responsibility to keep the provided memory alive and unmodified until it is  consumed by DALI

#### Why we need this PR?
*Pick one, remove the rest*
- It adds zero-copy to the ExternalSource operator

 - What solution was applied:
     adds zero-copy to the ExternalSource operator
     adds `no_copy` option so it is possible to skip copying data provided  by the user to the internal staging buffer but it is the user  responsibility to keep the provided memory alive and unmodified until it is  consumed by DALI
     external source, backend, pipeline
     changed signature of CanInferOutputs
     even if zero copy is provided and the input data is not continuous we will copy inside the RunImp - for GPU we must, for CPU it is an open question if we want to do so
 - Validation and testing:
     new tests added
     docs updated

**JIRA TASK**: *[DALI-1466]*


JanuszL:
Goes after #2023

JanuszL:
!builds

include/dali/c_api.h
```diff
@@ -205,13 +205,15 @@ daliSetExternalInputTensorsAsync(daliPipelineHandle *pipe_handle, const char *na
                                  device_type_t device, const void *const *data_ptr,
                                  dali_data_type_t data_type, const int64_t *shapes,
                                  int64_t sample_dim, const char *layout_str,
-                                 cudaStream_t stream, int sync, int is_pinned);
+                                 cudaStream_t stream, int sync, int is_pinned,
```
klecki, line 208: I wonder if in case of repetetive arguments of the same time like here: 
`int sync, int is_pinned, int no_copy` that can start to be confusing at the call site the more of them we add, maybe we should fill a config structure and pass it (similarly to what kernel does).

dali/python/nvidia/dali/pipeline.py
```diff
@@ -518,6 +518,11 @@ def feed_input(self, data_node, data, layout="", cuda_stream = None):
             If internal stream is used, the call to ``feed_input`` will block until the copy to
             internal buffer is complete, since there's no way to synchronize with this stream to
             prevent overwriting the array with new data in another stream.
+
+        `no_copy` : If DALI should copy the buffer then feed_input is called
+            If set to true it is the user responsibility to keep buffer alive and unmodified
+            until it is used in the pipeline. The rule of thumb is to keep alive 3
+            (number of DALI stages) * prefetch_queue_depth
```
klecki, line 525:
```suggestion
        `no_copy` : If DALI should copy the buffer when feed_input is called
            If ``no_copy`` is set to true instead of making a copy of the provided buffer, 
            DALI passes the user's memory directly in the Pipeline.
            It is user's responsibility to keep the buffer alive and unmodified
            until it is used in the pipeline. 
            
            The buffer can be modified again after the outputs of the iteration it was used in were consumed, which can happen `prefetch_queue_depth` iterations after the `feed_input` call.
```

The keeping alive 3 * prefetch_queue_depth is a bit enigmatic.

dali/pipeline/operator/builtin/external_source.h
```diff
@@ -265,14 +249,106 @@ class ExternalSource : public Operator<Backend> {
...
+    DALI_FAIL("No_copy is supported only for the same data source device type as operator.");
+  }
+
+  template<typename SrcBackend, template<typename> class SourceDataType>
+  inline void SetDataSourceHelper(const SourceDataType<SrcBackend> &batch, cudaStream_t stream = 0,
+                                  bool sync = false, bool no_copy = false) {
+    bool is_gpu_src = std::is_same<SrcBackend, GPUBackend>::value;
+    bool is_gpu_dst = std::is_same<Backend, GPUBackend>::value;
+    if (is_gpu_src && !is_gpu_dst) {
+      DALI_WARN("Incorrect Backends warning. Loading GPU-originated data into CPU " +
```
klecki, line 280: Can I complain about this again? :P

dali/pipeline/operator/builtin/external_source.cc
```diff
@@ -20,38 +20,71 @@ namespace dali {
...
-    internal_copy_to_storage = copy_to_storage_events_.PopFront();
+    copy_info = zero_copy_.front();
+    zero_copy_.pop_front();
+    if (copy_info.is_tensor_vector && copy_info.is_zero_copy) {
+      tensor_vector_elm = tv_data_.PopFront();
+    } else {
+      tensor_list_elm = tl_data_.PopFront();
+    }
+    if (!copy_info.is_zero_copy) {
+      internal_copy_to_storage = copy_to_storage_events_.PopFront();
```
klecki, line 36: Either it's somewhere hidden or you don't use this at all.

dali/pipeline/operator/builtin/external_source.cc
```diff
@@ -20,38 +20,71 @@ namespace dali {
...
 
-  // sort by the work size
-  sample_ids_.clear();
-  sample_ids_.reserve(batch_size_);
-  for (int sample_id = 0; sample_id < batch_size_; sample_id++) {
-    sample_ids_.emplace_back(volume(tensor_list_elm.front()->tensor_shape(sample_id)), sample_id);
+    if (copy_info.is_zero_copy && copy_info.is_tensor_vector) {
+      auto shape = tensor_vector_elm.front()->shape();
+      for (int sample_id = 0; sample_id < batch_size_; sample_id++) {
+        sample_ids_.emplace_back(volume(shape[sample_id]), sample_id);
```
klecki, line 54: Can't you share TensorVector with output TensorVector (so it will be non-continuos) -> in such case you don't need copy.

dali/test/python/test_external_source_impl.py
```diff
@@ -533,3 +533,108 @@ def iter_setup(self):
...
+            if by_name:
+                self.feed_input("src1", batch_1, no_copy=True)
+                self.feed_input("src2", batch_2, no_copy=True)
+            else:
+                self.feed_input(self.batch_1, batch_1, no_copy=True)
+                self.feed_input(self.batch_2, batch_2, no_copy=True)
+
+    iter_num = 5
+    # we don't have mixed stage so it is enough to keep only 2 * prefetch_queue_depth, but in the general
+    # case it should be 3 * prefetch_queue_depth
```
klecki, line 615: I think it's CPU_QUEUE_DEPTH * GPU_QUEUE_DEPTH

dali/python/nvidia/dali/pipeline.py
JanuszL: Mhhh, that is right

include/dali/c_api.h
JanuszL: Good idea. But I want to wait for the second opinion as this will shape our C APi.

dali/python/nvidia/dali/pipeline.py
```diff
@@ -477,7 +477,7 @@ def build(self, define_graph = None):
         self._pipe.Build(self._names_and_devices)
         self._built = True
 
-    def feed_input(self, data_node, data, layout="", cuda_stream = None):
+    def feed_input(self, data_node, data, layout="", cuda_stream = None, no_copy = False):
```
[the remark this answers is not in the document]
JanuszL: >  it should match between feed_input and the constructor of ExternalSource.

I don't think so. I would leave only one place where each argument can be set, either operator ot the `feed_input` call.

> Do you envision a scenario, where the data would be mixed between no_copy and copy?

No idea. It is just possible. But I'm okay with limiting that. Let us wait for the second opinion.

dali/python/nvidia/dali/external_source.py
```diff
@@ -303,15 +310,14 @@ def __call__(self, *, source = None, cycle = None, name = None, layout = None, c
 def _is_external_source_with_callback(op_instance):
     return isinstance(op_instance._op, ExternalSource) and op_instance._callback is not None
 
-def external_source(source = None, num_outputs = None, *, cycle = None, name = None, device = "cpu", layout = None, cuda_stream = None):
+def external_source(source = None, num_outputs = None, *, cycle = None, name = None, device = "cpu", layout = None, cuda_stream = None, no_copy = False):
```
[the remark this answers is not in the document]
JanuszL: > as we would need to keep the python reference to the data.

What if user will keep providing the same buffer but just modifies the content of it. I think there are some numpy in-place operations. Do we have any other option?

dali/pipeline/pipeline.h
```diff
@@ -212,15 +214,15 @@ class DLL_PUBLIC Pipeline {
...
+   * @param no_copy If no copy should be done inside the ExternalSource operator
    */
   template<typename Backend>
   DLL_PUBLIC inline void
   SetExternalInput(const string &name, const vector<Tensor<Backend>> &tl, cudaStream_t stream = 0,
-                   bool sync = false) {
-    SetExternalInputHelper(name, tl, stream, sync);
+                   bool sync = false, bool no_copy = false) {
+    SetExternalInputHelper(name, tl, stream, sync, no_copy);
   }
```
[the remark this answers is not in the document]
JanuszL: Maybe someone wants to use it from the C++ API. Or not. I don't have strong opinion either way.

dali/pipeline/operator/operator.h
```diff
@@ -126,9 +126,12 @@ class DLL_PUBLIC OperatorBase {
 
   /**
    * @brief If Operator can infer the output shapes it means that its output would use a single
-   * underlying allocation, especailly for CPU TensorVector will use contiguous mode.
+   * underlying allocation, especially for CPU TensorVector will use contiguous mode.
+   * `prerun` tell if the operator is queried during the runtime or during the pipeline build
+   * as the ability to infer outputs could depend on the runtime parameters
    */
-  DLL_PUBLIC virtual bool CanInferOutputs() const {
+  DLL_PUBLIC virtual bool CanInferOutputs(bool prerun = false) const {
```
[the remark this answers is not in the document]
JanuszL: If we prevent mixing `no_copy` with copy this will be no longer the case.

dali/pipeline/operator/builtin/external_source.cc
JanuszL: We always adding it inside `SetDataSourceHelper` to `copy_to_storage_events_` but maybe I can just limit that to the GPU variant.

dali/pipeline/operator/builtin/external_source.cc
JanuszL: I can but maybe it is better to have that continuous...
I don't have any strong opinion. I would welcome the second opinion.

JanuszL:
> Switching back and forth between copy and no-copy may result in additional allocations and deallocations - passing the data through will result in freeing whatever was allocated for output buffers, when we want to copy we need to allocate back.

It will happen when the user passes a continuous GPU buffer and non-continuous one after another and ask `no_copy=True`. We must copy in the second case and we can share in the first. Interleaving it will trash the output memory. Should we stick to the data format that was provided as the first one or just assumes that in the normal case user won't mix it?

dali/python/nvidia/dali/pipeline.py

dali/python/nvidia/dali/external_source.py
```diff
@@ -193,14 +194,20 @@ class ExternalSource():
     If internal stream is used, the call to ``feed_input`` will block until the copy to internal
     buffer is complete, since there's no way to synchronize with this stream to prevent
     overwriting the array with new data in another stream.
+
+`no_copy` : If DALI should copy the buffer then feed_input is called
+    If set to true it is the user responsibility to keep buffer alive and unmodified
+    until it is used in the pipeline.
+    The rule of thumb is to keep alive 3 (number of DALI stages) * prefetch_queue_depth
```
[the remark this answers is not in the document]

dali/python/backend_impl.cc
```diff
@@ -1328,17 +1329,18 @@ PYBIND11_MODULE(backend_impl, m) {
...
                                 : static_cast<cudaStream_t>(ctypes_void_ptr(cuda_stream));
-            FeedPipeline<GPUBackend>(p, name, list, stream, cuda_stream.is_none());
+            FeedPipeline<GPUBackend>(p, name, list, stream, cuda_stream.is_none(), no_copy);
           }
         },
         "name"_a,
         "list"_a,
-        "cuda_stream"_a = py::none())
+        "cuda_stream"_a = py::none(),
+        "bool"_a = false)
```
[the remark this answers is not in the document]
JanuszL: Done

dali/pipeline/operator/builtin/external_source.cc
JanuszL: Done

dali/test/python/test_external_source_impl.py
JanuszL: Done

dali/python/backend_impl.cc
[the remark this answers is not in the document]
mzient: It's still here.

dali/python/nvidia/dali/external_source.py
```diff
@@ -193,14 +194,25 @@ class ExternalSource():
     If internal stream is used, the call to ``feed_input`` will block until the copy to internal
     buffer is complete, since there's no way to synchronize with this stream to prevent
     overwriting the array with new data in another stream.
+
+``no_copy` : If DALI should copy the buffer when feed_input is called
+    If ``no_copy`` is set to true instead of making a copy of the provided buffer,
```
mzient, line 199:
```suggestion
    If true, then instead of making a copy of the provided buffer,
```

dali/python/nvidia/dali/external_source.py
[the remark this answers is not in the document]
mzient: I think that now we have no restrictions on what a callback / iterator / generator can do - it can be a partial function, with state attached to some object.

include/dali/c_api.h
mzient: IMO this calls for `unsigned flags`

```
enum {
  DALI_ext_default = 0,  // we can put whatever we like here
  DALI_ext_force_sync = (1<<0),
  DALI_ext_pinned = (1<<1),
  DALI_ext_no_copy =(1<<2)
};
```

dali/pipeline/data/tensor_vector.h
```diff
@@ -290,6 +290,18 @@ class TensorVector {
     }
   }
 
+  void ShareData(TensorList<Backend> *in_tl) {
+    state_ = State::contiguous;
+    pinned_ = in_tl->is_pinned();
+
+    tl_->ShareData(in_tl);
+
+    int N = in_tl->ntensor();
```
JanuszL, line 299: Remove

dali/python/nvidia/dali/external_source.py
JanuszL: Done

dali/pipeline/data/tensor_vector.h
JanuszL: Done

klecki:
> It will happen when the user passes a continuous GPU buffer and non-continuous one after another and ask no_copy=True. We must copy in the second case and we can share in the first. Interleaving it will trash the output memory. Should we stick to the data format that was provided as the first one or just assumes that in the normal case user won't mix it?

I _think_ that in most scenarios the user won't mix the data. We should certainly document the behaviour and provide some requirements if the user wants to use the `no_copy` option. Allowing users to use such data input formats that would still cause a copy is IMO not the best way to approach this.

JanuszL:
!build

include/dali/c_api.h
JanuszL: Done

dali/python/nvidia/dali/pipeline.py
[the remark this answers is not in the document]
JanuszL: Done. `no_copy` is an operator argument.

dali/python/backend_impl.cc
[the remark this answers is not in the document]
JanuszL: Done again, I hope for good this time.

dali/python/nvidia/dali/external_source.py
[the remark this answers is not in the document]
JanuszL: In this case let us leave it up to the user to do the right thing. If he cannot preserve the memory passed then it is his fault.

dali/pipeline/pipeline.h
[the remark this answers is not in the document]
JanuszL: Removed

dali/pipeline/operator/operator.h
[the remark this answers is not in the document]
JanuszL: This change is no longer needed after making `no_copy` an operator argument.
Adding `PreventOutputAllocation` should be a different task.

dali/pipeline/operator/builtin/external_source.h
```diff
@@ -220,20 +171,45 @@ class ExternalSource : public Operator<Backend> {
...
+      }
     }
     return true;
   }
 
-  bool CanInferOutputs()  const override {
-    return true;
+  bool CanInferOutputs(bool prerun) const override {
+    if (prerun) {
+      return false;
```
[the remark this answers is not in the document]
JanuszL: Not relevant anymore.

dali/pipeline/operator/builtin/external_source.h
JanuszL: You can.

dali/pipeline/operator/builtin/external_source.cc
JanuszL: Done that way.

JanuszL:
> > It will happen when the user passes a continuous GPU buffer and non-continuous one after another and ask no_copy=True. We must copy in the second case and we can share in the first. Interleaving it will trash the output memory. Should we stick to the data format that was provided as the first one or just assumes that in the normal case user won't mix it?
> 
> I _think_ that in most scenarios the user won't mix the data. We should certainly document the behaviour and provide some requirements if the user wants to use the `no_copy` option. Allowing users to use such data input formats that would still cause a copy is IMO not the best way to approach this.

Added an appropriate warning, but I don't think we should block the user from using it at all.

JanuszL:
!build

dali/test/python/test_external_source_cupy.py
```diff
@@ -32,3 +34,45 @@ def test_external_source_with_iter_cupy_stream():
...
+    pipe = Pipeline(batch_size, 3, 0)
+
+    pipe.set_outputs(fn.external_source(device="gpu", source=generator, no_copy=True))
+    pipe.build()
+
+    pattern = "ExternalSource operator should not mix continuous and noncontinuous inputs. " \
+              "In such case additional memory that gather provided data in continuous space " \
+              "will be trashed."
+    with check_output_pattern(pattern):
+        for _ in range(iterations):
```
jantonguirao, line 77: do you need the loop here? Shouldn't it fail at the first iteration?

dali/test/python/test_external_source_cupy.py
```diff
@@ -32,3 +34,45 @@ def test_external_source_with_iter_cupy_stream():
 
             for i in range(10):
                 check_output(pipe.run(), [np.array([attempt * 100 + i * 10 + 1.5], dtype=np.float32)])
+import sys
+def discard_stderr():
```
jantonguirao, line 38: not used?

dali/pipeline/operator/builtin/external_source.cc
```diff
@@ -64,7 +68,21 @@ DALI_SCHEMA(_ExternalSource)
...
-fail when it is not)code", false)
+fail when it is not)code", true)
+  .AddOptionalArg("no_copy",
+      R"code(If DALI should copy the buffer when feed_input is called
+If true, then instead of making a copy of the provided buffer,
+DALI passes the user's memory directly in the Pipeline.
+It is user's responsibility to keep the buffer alive and unmodified
+until it is used in the pipeline.
+
+The buffer can be modified again after the outputs of the iteration it was used in were
```
jantonguirao, line 79: How about:
```
The buffer can be modified or freed again after the relevant iteration output has been consumed. Effectively, it happens after ``prefetch_queue_depth * gpu_queue_depth`` iterations following the ``feed_input`` call.
```

dali/pipeline/operator/builtin/external_source.cc
```diff
@@ -81,6 +99,20 @@ where the last dimension represents the different channels).)code")
   .NumOutput(1)
   .AddOptionalArg("blocking",
       R"code(Whether external source should block until data is available or just
-fail when it is not)code", false);
+fail when it is not)code", false)
+  .AddOptionalArg("no_copy",
+      R"code(If DALI should copy the buffer when feed_input is called
+If true, then instead of making a copy of the provided buffer,
```
jantonguirao, line 105: same comments can be applied here and in the python documentation

dali/pipeline/operator/builtin/external_source.cu
```diff
@@ -52,18 +52,55 @@ void ExternalSource<GPUBackend>::RunImpl(DeviceWorkspace &ws) {
...
 
   auto &output = ws.Output<GPUBackend>(0);
-  cudaStream_t stream_used = ws.has_stream() ? ws.stream() : 0;
-  CUDA_CALL(cudaStreamWaitEvent(stream_used, *internal_copy_to_storage.front(), 0));
-  output.Copy(*(tensor_list_elm.front()), stream_used);
-  // record an event so Recycle can synchronize on it
-  cudaEventRecord(*cuda_event.front(), stream_used);
-  sync_worker_.DoWork(RecycleFunctor{this, std::move(cuda_event), std::move(tensor_list_elm),
-                                     std::move(internal_copy_to_storage)});
+  if (!no_copy_) {
```
jantonguirao, line 66: how about:
```
if (!no_copy_ || !tensor_list_elm.front()->shares_data()) {
  cudaStream_t stream_used = ws.has_stream() ? ws.stream() : 0;
  CUDA_CALL(cudaStreamWaitEvent(stream_used, *internal_copy_to_storage.front(), 0));
}

if (!no_copy_) { 
  output.Copy(*(tensor_list_elm.front()), stream_used);
  ...
} else if (!tensor_list_elm.front()->shares_data()) {
  void * ptr = tensor_list_elm.front()->raw_mutable_data();
  ...
} else {
  output.ShareData(tensor_list_elm.front().get());
  ...
}
```

dali/pipeline/operator/builtin/external_source.cu
```diff
@@ -52,18 +52,55 @@ void ExternalSource<GPUBackend>::RunImpl(DeviceWorkspace &ws) {
...
+  } else {
+    // if we copied from GPU TensorVector to TensorList we need to sync with that copy
+    if (!tensor_list_elm.front()->shares_data()) {
+      cudaStream_t stream_used = ws.has_stream() ? ws.stream() : 0;
+      CUDA_CALL(cudaStreamWaitEvent(stream_used, *internal_copy_to_storage.front(), 0));
+      // make a shared pointer which will recycle buffer upon destruction. So when pipeline
+      // no longer needs that buffer we can return it to the pool
+      void * ptr = tensor_list_elm.front()->raw_mutable_data();
+      int device_id = tensor_list_elm.front()->device_id();
+      output.ShareData(shared_ptr<void>(ptr, [
```
jantonguirao, line 83: maybe this would read better if we created the shared_ptr as a pre-step?

dali/pipeline/operator/builtin/external_source.h
```diff
@@ -215,25 +167,36 @@ class ExternalSource : public Operator<Backend> {
...
       } else {
-        if (tl_data_.IsEmpty()) {
+        if (state_.empty()) {
           DALI_FAIL("No data was provided to the ExternalSource. Make sure to feed it properly.");
         }
       }
-      output_desc[0].shape = tl_data_.PeekFront()->shape();
-      output_desc[0].type = tl_data_.PeekFront()->type();
+      state_.pop_front();
+      // for zero copy we don't want any shape inference to avoid any data allocation ahead
```
jantonguirao, line 185: I guess this line was meant to go to line 170 (?)

dali/pipeline/operator/builtin/external_source.h
```diff
@@ -265,21 +236,177 @@ class ExternalSource : public Operator<Backend> {
     }
   }
 
+  template<typename SrcBackend, template<typename> class SourceDataType>
+  inline std::enable_if_t<!std::is_same<SrcBackend, Backend>::value, bool>
+  ShareData(const SourceDataType<SrcBackend> &t, cudaStream_t /*stream = 0*/) {
+    DALI_FAIL("No_copy is supported only for the same data source device type as operator.");
```
jantonguirao, line 242:
```suggestion
    DALI_FAIL("no_copy is supported only for the same data source device type as operator.");
```

dali/pipeline/operator/builtin/external_source.h
```diff
@@ -265,21 +236,177 @@ class ExternalSource : public Operator<Backend> {
...
+    return false;
+  }
+
+  template<typename SrcBackend, template<typename> class SourceDataType>
+  inline std::enable_if_t<std::is_same<SrcBackend, Backend>::value &&
+                          std::is_same<SrcBackend, CPUBackend>::value, bool>
+  ShareData(const SourceDataType<SrcBackend> &t, cudaStream_t /*stream = 0*/) {
+    auto tv_elm = tv_data_.GetEmpty();
+    tv_elm.front()->ShareData(const_cast<SourceDataType<CPUBackend>*>(&t));
+    tv_data_.AddBack(tv_elm);
```
jantonguirao, line 252:
```suggestion
    tv_data_.PushBack(tv_elm);
```
?

dali/pipeline/operator/builtin/external_source.h
```diff
@@ -265,21 +236,177 @@ class ExternalSource : public Operator<Backend> {
...
+    auto tl_elm = tl_data_.GetEmpty();
+    if (t.IsContiguous()) {
+      t.ShareWith(const_cast<TensorList<Backend>*>(tl_elm.front().get()));
+      gpu_noncontinuous_zero_copy = true;
+    } else {
+      // it is not continuous so we need to copy
+      tl_elm.front()->Copy(t, stream);
+      if (gpu_noncontinuous_zero_copy) {
+        DALI_WARN("ExternalSource operator should not mix continuous and noncontinuous inputs. "
+                  "In such case additional memory that gather provided data in continuous space "
```
jantonguirao, line 269:
```suggestion
                  "In such a case the internal memory used to gather data in a contiguous chunk of memory would be trashed."
```

dali/pipeline/operator/builtin/external_source.h
```diff
@@ -265,21 +236,177 @@ class ExternalSource : public Operator<Backend> {
...
+  inline void CopyData(const SourceDataType<SrcBackend> &batch, cudaStream_t stream = 0,
+                       bool sync = false);
+
+  template<typename SrcBackend, template<typename> class SourceDataType>
+  inline void SetDataSourceHelper(const SourceDataType<SrcBackend> &batch, cudaStream_t stream = 0,
+                                  bool sync = false) {
+    bool is_gpu_src = std::is_same<SrcBackend, GPUBackend>::value;
+    bool is_gpu_dst = std::is_same<Backend, GPUBackend>::value;
+    if (is_gpu_src && !is_gpu_dst) {
+      DALI_WARN("Incorrect Backends warning. Loading GPU-originated data into CPU "
```
jantonguirao, line 298:
```suggestion
      DALI_WARN("Warning: Loading GPU-originated data into CPU "
```

dali/pipeline/operator/builtin/external_source.h
```diff
@@ -265,21 +236,177 @@ class ExternalSource : public Operator<Backend> {
...
+                "ExternalSource operator is discouraged and might be inefficient.");
+    }
+    DALI_ENFORCE(OperatorBase::batch_size_ == static_cast<int>(batch.ntensor()),
+                 "Data list provided to ExternalSource needs to have batch_size = " +
```
jantonguirao, line 302: make_string?

dali/pipeline/operator/builtin/external_source.h
```diff
@@ -265,21 +236,177 @@ class ExternalSource : public Operator<Backend> {
...
   detail::CachingList<uptr_cuda_event_type> cuda_events_, copy_to_storage_events_;
   struct RecycleFunctor;
 
   std::mutex busy_m_;
   std::condition_variable cv_;
-  bool blocking_ = false;
+  bool blocking_ = true;
+  bool no_copy_ = false;
+
+  // now it only indicates that there is a data in the ExternalSource, in the future
```
jantonguirao, line 338:
```suggestion
  // now it only indicates that there is data in the ExternalSource, in the future
```

dali/pipeline/operator/builtin/external_source.h
```diff
@@ -265,21 +236,177 @@ class ExternalSource : public Operator<Backend> {
...
+  // a per sample metadata could be stored there
```
jantonguirao, line 339:
```suggestion
  // a per sample metadata could be stored here
```

dali/c_api/c_api_test.cc
```diff
@@ -231,7 +231,7 @@ TYPED_TEST(CApiTest, ExternalSourceSingleAllocPipe) {
     pipe_ptr->SetExternalInput(input_name, input);
     daliSetExternalInputAsync(&handle, input_name.c_str(), backend_to_device_type<TypeParam>::value,
                               input.raw_data(), dali_data_type_t::DALI_UINT8, input_shape.data(),
-                              input_shape.sample_dim(), nullptr, cuda_stream, false, false);
+                              input_shape.sample_dim(), nullptr, cuda_stream, DALI_ext_pinned);
```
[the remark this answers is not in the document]
JanuszL: Wrong copy/paste. It was suppose to be `DALI_ext_default`.

dali/pipeline/operator/builtin/external_source.cu
```diff
@@ -52,18 +52,55 @@ void ExternalSource<GPUBackend>::RunImpl(DeviceWorkspace &ws) {
...
+    cudaStream_t stream_used = ws.has_stream() ? ws.stream() : 0;
+    CUDA_CALL(cudaStreamWaitEvent(stream_used, *internal_copy_to_storage.front(), 0));
+    output.Copy(*(tensor_list_elm.front()), stream_used);
+    // record an event so Recycle can synchronize on it
+    cudaEventRecord(*cuda_event.front(), stream_used);
+    sync_worker_.DoWork(RecycleFunctor{this, std::move(cuda_event), std::move(tensor_list_elm),
+                                       std::move(internal_copy_to_storage)});
+  } else {
+    // if we copied from GPU TensorVector to TensorList we need to sync with that copy
+    if (!tensor_list_elm.front()->shares_data()) {
```
klecki, line 76: That can probably be also indicated in the state?

dali/pipeline/data/tensor_vector.h
```diff
@@ -290,7 +290,66 @@ class TensorVector {
...
+  void Copy(const TensorVector<SrcBackend> &in_tv, cudaStream_t stream) {
+    SetContiguous(true);
+    tl_->Copy(in_tv, stream);
+
+    tensors_.clear();
+    views_count_ = 0;
+    UpdateViews();
+  }
+
+  void ShareData(TensorList<Backend> *in_tl) {
```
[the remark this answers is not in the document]
JanuszL: This is the signature of `ShareData` in Tensor and TensorList. I don't want do change this in this PR.

dali/pipeline/operator/builtin/external_source.h
```diff
@@ -265,21 +236,177 @@ class ExternalSource : public Operator<Backend> {
...
+  tl_elm.front()->Copy(batch, stream);
+  // record event for:
+  // - GPU -> GPU
+  // - pinned CPU -> GPU
+  if (std::is_same<SrcBackend, GPUBackend>::value || batch.is_pinned()) {
+    cudaEventRecord(*copy_to_storage_event.front(), stream);
+  }
+  // sync for pinned CPU -> GPU as well, because the user doesn't know when he can
+  // reuse provided memory anyway
+  if (sync || batch.is_pinned()) {
```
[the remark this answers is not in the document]
JanuszL: For the CPU I don't expect anyone to provide an stream, so it is better to sync.

dali/pipeline/operator/builtin/external_source.h
```diff
@@ -265,21 +236,177 @@ class ExternalSource : public Operator<Backend> {
...
+  }
+
+  template<typename SrcBackend>
+  inline std::enable_if_t<std::is_same<SrcBackend, Backend>::value &&
+                          std::is_same<SrcBackend, GPUBackend>::value, bool>
+  ShareData(const TensorVector<SrcBackend> &t, cudaStream_t stream = 0) {
+    auto tl_elm = tl_data_.GetEmpty();
+    if (t.IsContiguous()) {
+      t.ShareWith(const_cast<TensorList<Backend>*>(tl_elm.front().get()));
+      gpu_noncontinuous_zero_copy = true;
```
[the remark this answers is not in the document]
JanuszL: `ExtenralSourceState` is to pass state from `SetDataSource` to `RunImpl`. Here is it passing the knowledge from one run of `SetDataSource` to another one.

dali/pipeline/operator/builtin/external_source.h
```diff
@@ -265,21 +236,177 @@ class ExternalSource : public Operator<Backend> {
...
+    tv_data_.AddBack(tv_elm);
+    return false;
+  }
+
+  template<typename SrcBackend>
+  inline std::enable_if_t<std::is_same<SrcBackend, Backend>::value &&
+                          std::is_same<SrcBackend, GPUBackend>::value, bool>
+  ShareData(const TensorVector<SrcBackend> &t, cudaStream_t stream = 0) {
+    auto tl_elm = tl_data_.GetEmpty();
+    if (t.IsContiguous()) {
```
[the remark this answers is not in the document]
JanuszL: It is not only the Python-originated data we need to deal with.

dali/test/python/test_external_source_cupy.py
JanuszL: No. I(t fails when you provide contiguous first and then non-contiguous memory to the ExternalSource. So at lest two iterations are needed, when the first input is non-contiguous, then it is 3. Making it 4 just to be sure.

dali/test/python/test_external_source_cupy.py
JanuszL: Test code. Removed.

dali/pipeline/operator/builtin/external_source.h
JanuszL: Pre-rework leftover.

include/dali/c_api.h
```diff
@@ -126,6 +125,19 @@ DLL_PUBLIC void daliDeserializeDefault(daliPipelineHandle *pipe_handle,
...
+  DALI_ext_default = 0,
+  /**
+   * If memory transfer should be synchronous - applies to GPU memory
+   */
+  DALI_ext_force_sync = (1<<0),
+  /**
+   * If provided CPU memory is page-locked
+   */
+  DALI_ext_pinned = (1<<1)
+};
```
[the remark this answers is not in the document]
JanuszL: I don't think we have more of them for the ExternalSource. Or you mean something else?

dali/python/nvidia/dali/external_source.py
```diff
@@ -303,15 +320,14 @@ def __call__(self, *, source = None, cycle = None, name = None, layout = None, c
 def _is_external_source_with_callback(op_instance):
     return isinstance(op_instance._op, ExternalSource) and op_instance._callback is not None
 
-def external_source(source = None, num_outputs = None, *, cycle = None, name = None, device = "cpu", layout = None, cuda_stream = None):
+def external_source(source = None, num_outputs = None, *, cycle = None, name = None, device = "cpu", layout = None, cuda_stream = None, **kwargs):
     """Creates a data node which is populated with data from a Python source.
 The data can be provided by the ``source`` function or iterable, or it can be provided by
-``pipeline.feed_input(name, data, layout)`` inside ``pipeline.iter_setup``.
+``pipeline.feed_input(name, data, layout, cuda_stream, no_copy)`` inside ``pipeline.iter_setup``.
```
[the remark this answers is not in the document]
JanuszL: Leftover.

dali/pipeline/data/tensor_vector.h
```diff
@@ -290,7 +290,66 @@ class TensorVector {
...
+    SetContiguous(true);
+    pinned_ = in_tl->is_pinned();
+    tl_->ShareData(in_tl);
+
+    tensors_.clear();
+    views_count_ = 0;
+    UpdateViews();
+  }
+
+  void ShareWith(TensorList<Backend> *in_tl) const {
```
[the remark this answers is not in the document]
JanuszL: I want to keep that consistent with ShareData.

dali/pipeline/operator/builtin/external_source.h
```diff
@@ -265,21 +236,177 @@ class ExternalSource : public Operator<Backend> {
...
   std::mutex busy_m_;
   std::condition_variable cv_;
-  bool blocking_ = false;
+  bool blocking_ = true;
+  bool no_copy_ = false;
+
+  // now it only indicates that there is a data in the ExternalSource, in the future
+  // a per sample metadata could be stored there
+  struct ExtenralSourceState {
+    bool dummy;
```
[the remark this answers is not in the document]
JanuszL: We don't

dali/pipeline/operator/builtin/external_source.h
```diff
@@ -265,21 +236,177 @@ class ExternalSource : public Operator<Backend> {
...
+                  std::to_string(OperatorBase::batch_size_) + " length, found " +
+                  std::to_string(static_cast<int>(batch.ntensor())) +
+                  " samples.");
+    // Note: If we create a GPU source, we will need to figure
+    // out what stream we want to do this copy in. CPU we can
+    // pass anything as it is ignored.
+    std::list<uptr_tl_type> tl_elm;
+    std::list<uptr_tl_type> tv_elm;
+    if (no_copy_) {
+      std::lock_guard<std::mutex> busy_lock(busy_m_);
```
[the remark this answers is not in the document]
JanuszL: We do other things around share that needs to be lock protected. And as sharing is fast we can just taka a lock, do what we need and release. In the case of copy we don't want to hold the mutex during the copy itself so tighter control is required there.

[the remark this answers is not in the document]
JanuszL: But maybe moving this inside `ShareUserData` won't cause much of code duplication after all.

dali/pipeline/operator/builtin/external_source.h
```diff
@@ -265,21 +236,177 @@ class ExternalSource : public Operator<Backend> {
...
+      state_.push_back({});
+      auto copied = ShareData(batch, stream);
+      if (copied && std::is_same<SourceDataType<SrcBackend>, TensorVector<GPUBackend>>::value) {
+        std::list<uptr_cuda_event_type> copy_to_storage_event;
+        copy_to_storage_event = copy_to_storage_events_.GetEmpty();
+        cudaEventRecord(*copy_to_storage_event.front(), stream);
+        copy_to_storage_events_.AddBack(copy_to_storage_event);
```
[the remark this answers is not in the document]
JanuszL: I don't think you want to mix no_copy with sync. The fact that we do an auxiliary copy inside is just a way to make that working, in all other cases sync doesn't make any sense.

dali/pipeline/operator/builtin/external_source.cu
JanuszL: Done

dali/pipeline/operator/builtin/external_source.h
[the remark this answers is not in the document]
JanuszL: Done

dali/pipeline/operator/builtin/external_source.h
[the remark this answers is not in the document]
JanuszL: Done

dali/python/nvidia/dali/external_source.py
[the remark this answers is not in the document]

dali/pipeline/operator/builtin/external_source.cc
```diff
@@ -64,7 +68,21 @@ DALI_SCHEMA(_ExternalSource)
...
+It is user's responsibility to keep the buffer alive and unmodified
+until it is used in the pipeline.
+
+The buffer can be modified again after the outputs of the iteration it was used in were
+consumed, which can happen ``prefetch_queue_depth`` * ``gpu_queue_depth`` iterations
+after the ``feed_input`` call.
+
+It works for continuous and non-continuous buffers without an additional copy, with one
+exception for the non-continuous GPU buffer. In that case, an internal continuous copy
+is created asynchronously.)code", false)
```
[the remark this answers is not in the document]
JanuszL: Done

dali/pipeline/operator/builtin/external_source.cc
```diff
@@ -64,7 +68,21 @@ DALI_SCHEMA(_ExternalSource)
...
+fail when it is not)code", true)
+  .AddOptionalArg("no_copy",
+      R"code(If DALI should copy the buffer when feed_input is called
+If true, then instead of making a copy of the provided buffer,
+DALI passes the user's memory directly in the Pipeline.
+It is user's responsibility to keep the buffer alive and unmodified
+until it is used in the pipeline.
+
+The buffer can be modified again after the outputs of the iteration it was used in were
+consumed, which can happen ``prefetch_queue_depth`` * ``gpu_queue_depth`` iterations
```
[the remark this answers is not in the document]
JanuszL: Need to check tests again.

dali/pipeline/operator/builtin/external_source.h
JanuszL: Done

dali/pipeline/operator/builtin/external_source.h
JanuszL: Done

dali/pipeline/operator/builtin/external_source.h
JanuszL: Done

dali/pipeline/operator/builtin/external_source.h
JanuszL: Done

dali/pipeline/operator/builtin/external_source.h
JanuszL: Done

dali/pipeline/operator/builtin/external_source.h
JanuszL: Done

dali/pipeline/operator/builtin/external_source.h
JanuszL: Done

dali/pipeline/operator/builtin/external_source.h
JanuszL: Done

dali/pipeline/operator/builtin/external_source.cu
JanuszL: Done

dali/pipeline/operator/builtin/external_source.cu
JanuszL: Done

dali/pipeline/operator/builtin/external_source.cc
JanuszL: Done

dali/pipeline/operator/builtin/external_source.cc
JanuszL: Done

dali/test/python/test_external_source_cupy.py
JanuszL: Done

dali/test/python/test_utils.py
```diff
@@ -286,3 +287,30 @@ class py_holder(object):
         global cp
         import cupy as cp
         return cp.asanyarray(holder)
+
+class check_output_pattern():
```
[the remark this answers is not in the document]
JanuszL: Done

dali/test/python/test_external_source_impl.py
```diff
@@ -533,3 +539,108 @@ def iter_setup(self):
...
+    else:
+        run_and_check(pipe, source)
+
+def test_iter_setup_zero_copy():
+    for use_fn_api in [False, True]:
+        for by_name in [False, True]:
+            for as_tensor in [False, True]:
+                for device in ["cpu", "gpu"]:
+                    for additional_num_keep_samples in [-1, 0, 1]:
+                        yield _test_iter_setup_zero_copy, use_fn_api, by_name, as_tensor, device, additional_num_keep_samples
```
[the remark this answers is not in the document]
JanuszL: Done

dali/python/nvidia/dali/external_source.py
```diff
@@ -193,6 +193,23 @@ class ExternalSource():
     If internal stream is used, the call to ``feed_input`` will block until the copy to internal
     buffer is complete, since there's no way to synchronize with this stream to prevent
     overwriting the array with new data in another stream.
+
+``blocking`` : optional, Whether external source should block until data is available or just
+fail when it is not
+
+``no_copy` : If DALI should copy the buffer when feed_input is called
```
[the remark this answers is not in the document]

dali/pipeline/operator/builtin/external_source.h
```diff
@@ -265,21 +236,177 @@ class ExternalSource : public Operator<Backend> {
...
 
   std::mutex busy_m_;
   std::condition_variable cv_;
-  bool blocking_ = false;
+  bool blocking_ = true;
+  bool no_copy_ = false;
+
+  // now it only indicates that there is a data in the ExternalSource, in the future
+  // a per sample metadata could be stored there
+  struct ExtenralSourceState {
```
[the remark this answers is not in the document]
JanuszL: Done

dali/pipeline/operator/builtin/external_source.cc
```diff
@@ -64,7 +68,21 @@ DALI_SCHEMA(_ExternalSource)
...
   .AddOptionalArg("blocking",
       R"code(Whether external source should block until data is available or just
-fail when it is not)code", false)
+fail when it is not)code", true)
+  .AddOptionalArg("no_copy",
+      R"code(If DALI should copy the buffer when feed_input is called
+If true, then instead of making a copy of the provided buffer,
+DALI passes the user's memory directly in the Pipeline.
+It is user's responsibility to keep the buffer alive and unmodified
+until it is used in the pipeline.
```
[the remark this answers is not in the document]
JanuszL: Done

dali/pipeline/operator/builtin/external_source.cc
```diff
@@ -64,7 +68,21 @@ DALI_SCHEMA(_ExternalSource)
   .NumOutput(1)
   .AddOptionalArg("blocking",
       R"code(Whether external source should block until data is available or just
-fail when it is not)code", false)
+fail when it is not)code", true)
+  .AddOptionalArg("no_copy",
+      R"code(If DALI should copy the buffer when feed_input is called
+If true, then instead of making a copy of the provided buffer,
+DALI passes the user's memory directly in the Pipeline.
+It is user's responsibility to keep the buffer alive and unmodified
```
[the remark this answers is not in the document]
JanuszL: Done

dali/pipeline/operator/builtin/external_source.cc
```diff
@@ -64,7 +68,21 @@ DALI_SCHEMA(_ExternalSource)
   .NumOutput(1)
   .AddOptionalArg("blocking",
       R"code(Whether external source should block until data is available or just
-fail when it is not)code", false)
+fail when it is not)code", true)
+  .AddOptionalArg("no_copy",
+      R"code(If DALI should copy the buffer when feed_input is called
+If true, then instead of making a copy of the provided buffer,
```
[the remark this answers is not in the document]
JanuszL: Done

dali/pipeline/operator/builtin/external_source.cc
```diff
@@ -64,7 +68,21 @@ DALI_SCHEMA(_ExternalSource)
   .NumOutput(1)
   .AddOptionalArg("blocking",
       R"code(Whether external source should block until data is available or just
-fail when it is not)code", false)
+fail when it is not)code", true)
+  .AddOptionalArg("no_copy",
+      R"code(If DALI should copy the buffer when feed_input is called
```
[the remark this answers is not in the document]
JanuszL: Done

dali/c_api/c_api_test.cc
```diff
@@ -349,7 +350,7 @@ TYPED_TEST(CApiTest, ExternalSourceSingleAllocDifferentBackendsTest) {
     pipe_ptr->SetExternalInput(input_name, input);
     daliSetExternalInput(&handle, input_name.c_str(), backend_to_device_type<DataBackend>::value,
                          input.raw_data(), dali_data_type_t::DALI_UINT8, input_shape.data(),
-                         input_shape.sample_dim(), nullptr, false);
+                         input_shape.sample_dim(), nullptr, DALI_ext_pinned);
```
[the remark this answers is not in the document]
JanuszL: Done

dali/c_api/c_api_test.cc
```diff
@@ -310,7 +311,7 @@ TYPED_TEST(CApiTest, ExternalSourceMultipleAllocPipe) {
   daliSetExternalInputTensorsAsync(&handle, input_name.c_str(),
                                    backend_to_device_type<TypeParam>::value, data_ptrs.data(),
                                    dali_data_type_t::DALI_UINT8, input_shape.data(),
-                                   input_shape.sample_dim(), "HWC", cuda_stream, false, false);
+                                   input_shape.sample_dim(), "HWC", cuda_stream, DALI_ext_pinned);
```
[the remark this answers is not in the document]
JanuszL: Done

dali/c_api/c_api_test.cc
```diff
@@ -289,7 +289,8 @@ TYPED_TEST(CApiTest, ExternalSourceMultipleAllocPipe) {
     daliSetExternalInputTensorsAsync(&handle, input_name.c_str(),
                                      backend_to_device_type<TypeParam>::value, data_ptrs.data(),
                                      dali_data_type_t::DALI_UINT8, input_shape.data(),
-                                     input_shape.sample_dim(), nullptr, cuda_stream, false, false);
+                                     input_shape.sample_dim(), nullptr, cuda_stream,
+                                     DALI_ext_pinned);
```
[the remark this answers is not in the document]
JanuszL: Done

dali/c_api/c_api_test.cc
```diff
@@ -251,7 +251,7 @@ TYPED_TEST(CApiTest, ExternalSourceSingleAllocPipe) {
   pipe_ptr->SetExternalInput(input_name, input);
   daliSetExternalInputAsync(&handle, input_name.c_str(), backend_to_device_type<TypeParam>::value,
                             input.raw_data(), dali_data_type_t::DALI_UINT8, input_shape.data(),
-                            input_shape.sample_dim(), "HWC", cuda_stream, false, false);
+                            input_shape.sample_dim(), "HWC", cuda_stream, DALI_ext_pinned);
```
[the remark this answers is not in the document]
JanuszL: Done

dali/c_api/c_api_test.cc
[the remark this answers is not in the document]
JanuszL: Done

dali/test/python/test_external_source_impl.py
```diff
@@ -533,3 +539,108 @@ def iter_setup(self):
...
+    source = TestIteratorZeroCopy(iter_num, batch_size, [2, 3], as_tensor=as_tensor, num_keep_samples=num_keep_samples)
+    pipe = IterSetupPipeline(iter(source), 3, 0, device, prefetch_queue_depth)
+    pipe.build()
+
+    cupy_used = datapy != np
+    if (device == "cpu" and cupy_used) or (device == "gpu" and not cupy_used):
+        assert_raises(RuntimeError, pipe.run)
+    elif additional_num_keep_samples < 0:
+        # assert_raises doesn't work here
+        try:
```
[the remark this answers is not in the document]
JanuszL: For some reason using this approach all I get is:
```
FAIL: test_external_source_cupy.test_iter_setup_zero_copy(True, True, False, 'gpu', -1)
  File "/usr/lib/python3/dist-packages/nose/case.py", line 197, in runTest
    self.test(*self.arg)
  File "/home/jlisiecki/Dali/dali/dali/test/python/test_external_source_impl.py", line 633, in _test_iter_setup_zero_copy
    run_and_check(pipe, source)
AssertionError: AssertionError not raised
```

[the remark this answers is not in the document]
JanuszL: ...., ok, there may be a flaw in my code.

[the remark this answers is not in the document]
JanuszL: Even despite my bug this approach doesn't work for the assertions from the different module - test_utils.py in this case.

dali/pipeline/operator/builtin/external_source.cc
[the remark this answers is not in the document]
JanuszL: Done

JanuszL:
!build

include/dali/c_api.h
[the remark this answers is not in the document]
klecki: I meant DALI_EXT_DEFAULT etc (all caps for the flags names).

dali/pipeline/operator/builtin/external_source.h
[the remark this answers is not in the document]
klecki: I know, I'm just checking.

dali/pipeline/operator/builtin/external_source.h
```diff
@@ -237,56 +236,65 @@ class ExternalSource : public Operator<Backend> {
   }
 
   template<typename SrcBackend, template<typename> class SourceDataType>
-  inline std::enable_if_t<!std::is_same<SrcBackend, Backend>::value, bool>
-  ShareData(const SourceDataType<SrcBackend> &t, cudaStream_t /*stream = 0*/) {
-    DALI_FAIL("No_copy is supported only for the same data source device type as operator.");
-    return false;
+  inline std::enable_if_t<!std::is_same<SrcBackend, Backend>::value, void>
```
klecki, line 239: Small nitpick: you don't need that `void`, it's there by default.

klecki: Same in the two below.

dali/pipeline/operator/builtin/external_source.h
```diff
@@ -335,15 +334,15 @@ class ExternalSource : public Operator<Backend> {
...
+  // a per sample metadata could be stored here
+  struct ExternalSourceState  {
+    bool copied_shared_data = false;
   };
 
-  std::list<ExtenralSourceState> state_;
+  std::list<ExternalSourceState > state_;
 
-  bool gpu_noncontinuous_zero_copy = false;
+  bool gpu_noncontiguous_zero_copy = false;
```
klecki, line 345: Small nitpick:
```suggestion
  bool gpu_noncontiguous_zero_copy_ = false;
```
Also I find the name the opposite of what it actually did, maybe add a comment here?

JanuszL:
!build

dali/pipeline/operator/builtin/external_source.h
JanuszL: Done

dali/pipeline/operator/builtin/external_source.h
JanuszL: Done

dali/pipeline/data/tensor_list.h
```diff
@@ -388,7 +388,7 @@ class DLL_PUBLIC TensorList : public Buffer<Backend> {
 
   /**
    * @brief Checks whether the TensorList is
-   * continuous. It returns true if and only if
+   * contiguous. It returns true if and only if
    * all of the stored Tensors are densely packed in memory.
    */
   inline bool IsContinuousTensor() const {
```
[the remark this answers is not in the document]
JanuszL: Done

dali/pipeline/operator/builtin/external_source.cc
```diff
@@ -64,7 +69,21 @@ DALI_SCHEMA(_ExternalSource)
   .NumOutput(1)
   .AddOptionalArg("blocking",
       R"code(Whether external source should block until data is available or just
-fail when it is not)code", false)
+fail when it is not)code", true)
+  .AddOptionalArg("no_copy",
+      R"code(Whether DALI should copy the buffer when feed_input is called
+If True, DALI passes the user memory directly to the Pipeline, instead of copying.
+It is the user's responsibility to keep the buffer alive and unmodified
+until it isconsumed by the pipeline.
```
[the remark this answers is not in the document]
JanuszL: Done

JanuszL:
!build

JanuszL:
!build

JanuszL:
!build

JanuszL:
!build

PR #2028, opened 2020-06-16
Makes TF RN50 TL3 test to compile ahead of time

JanuszL:
- makes XLA compile ahead to the training to reduce the peak memory usage

#### Why we need this PR?
*Pick one, remove the rest*
- It fixes a OOM in TensorFlow RN50 test

 - What solution was applied:
     makes XLA compile ahead to the training to reduce the peak memory usage
     TL3 RN50 test
     NA
 - Validation and testing:
     CI TL3 RN50 test
     NA

**JIRA TASK**: *[NA]*

### DIFF
--- a/qa/TL3_RN50_convergence/test_tensorflow.sh
+++ b/qa/TL3_RN50_convergence/test_tensorflow.sh
@@ -22,6 +22,8 @@ LOG=dali.log
 OUT=${LOG%.log}.dir
 
 SECONDS=0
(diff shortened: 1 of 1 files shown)


cliffwoolley:
I suggest instead of dropping the batch size to try setting the env var `TF_XLA_FLAGS="--tf_xla_enable_lazy_compilation=false"` .

PR #2035, opened 2020-06-18
Add an ability to dump info about operator output buffer size

JanuszL:
- adds an option the executor that makes it gather size of the operator output buffer statistics so the user can select the right value for `bytes_per_sample_hint`

#### Why we need this PR?
*Pick one, remove the rest*
- It adds an ability to dump info about operator output buffer size

 - What solution was applied:
     adds an option the executor that makes it gather size of the operator output buffer statistics so the user can select the right value for `bytes_per_sample_hint`
     Executor, pipeline API across the stack
     How it is done in the executor
 - Validation and testing:
     CI, but no test targeting the prints
     API description updated

**JIRA TASK**: *[DALI-1024]*

### DIFF
--- a/dali/pipeline/executor/async_pipelined_executor.h
+++ b/dali/pipeline/executor/async_pipelined_executor.h
@@ -36,9 +36,11 @@ class DLL_PUBLIC AsyncPipelinedExecutor : public PipelinedExecutor {
                                            size_t bytes_per_sample_hint, bool set_affinity = false,
                                            int max_num_stream = -1,
                                            int default_cuda_stream_priority = 0,
-                                           QueueSizes prefetch_queue_depth = QueueSizes{2, 2})
+                                           QueueSizes prefetch_queue_depth = QueueSizes{2, 2},
+                                           bool get_memory_stats = false)
       : PipelinedExecutor(batch_size, num_thread, device_id, bytes_per_sample_hint, set_affinity,
-                          max_num_stream, default_cuda_stream_priority, prefetch_queue_depth),
+                          max_num_stream, default_cuda_stream_priority, prefetch_queue_depth,
+                          get_memory_stats),
         cpu_thread_(device_id, set_affinity),
         mixed_thread_(device_id, set_affinity),
         gpu_thread_(device_id, set_affinity),
--- a/dali/pipeline/executor/async_pipelined_executor.h
+++ b/dali/pipeline/executor/async_pipelined_executor.h
@@ -36,11 +36,9 @@ class DLL_PUBLIC AsyncPipelinedExecutor : public PipelinedExecutor {
                                            size_t bytes_per_sample_hint, bool set_affinity = false,
                                            int max_num_stream = -1,
                                            int default_cuda_stream_priority = 0,
-                                           QueueSizes prefetch_queue_depth = QueueSizes{2, 2},
-                                           bool get_memory_stats = false)
+                                           QueueSizes prefetch_queue_depth = QueueSizes{2, 2})
       : PipelinedExecutor(batch_size, num_thread, device_id, bytes_per_sample_hint, set_affinity,
-                          max_num_stream, default_cuda_stream_priority, prefetch_queue_depth,
-                          get_memory_stats),
+                          max_num_stream, default_cuda_stream_priority, prefetch_queue_depth),
         cpu_thread_(device_id, set_affinity),
         mixed_thread_(device_id, set_affinity),
         gpu_thread_(device_id, set_affinity),
--- a/dali/pipeline/executor/executor.h
+++ b/dali/pipeline/executor/executor.h
@@ -21,6 +21,7 @@
 #include <string>
 #include <utility>
 #include <vector>
+#include <unordered_map>
 
 #include "dali/core/common.h"
 #include "dali/core/error_handling.h"
@@ -38,6 +39,7 @@
 #include "dali/pipeline/workspace/host_workspace.h"
 #include "dali/pipeline/workspace/mixed_workspace.h"
 #include "dali/pipeline/workspace/workspace_data_factory.h"
+#include "dali/pipeline/data/backend.h"
 
 namespace dali {
 
@@ -83,7 +85,8 @@ class DLL_PUBLIC Executor : public ExecutorBase, public WorkspacePolicy, public 
   DLL_PUBLIC inline Executor(int batch_size, int num_thread, int device_id,
                              size_t bytes_per_sample_hint, bool set_affinity = false,
                              int max_num_stream = -1, int default_cuda_stream_priority = 0,
-                             QueueSizes prefetch_queue_depth = QueueSizes{2, 2})
+                             QueueSizes prefetch_queue_depth = QueueSizes{2, 2},
+                             bool get_memory_stats = false)
       : batch_size_(batch_size),
         device_id_(device_id),
         bytes_per_sample_hint_(bytes_per_sample_hint),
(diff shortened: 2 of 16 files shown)


JanuszL:
!build

JanuszL:
!build

JanuszL:
!build

dali/pipeline/executor/async_pipelined_executor.h
```diff
@@ -36,9 +36,11 @@ class DLL_PUBLIC AsyncPipelinedExecutor : public PipelinedExecutor {
                                            size_t bytes_per_sample_hint, bool set_affinity = false,
                                            int max_num_stream = -1,
                                            int default_cuda_stream_priority = 0,
-                                           QueueSizes prefetch_queue_depth = QueueSizes{2, 2})
+                                           QueueSizes prefetch_queue_depth = QueueSizes{2, 2},
+                                           bool get_memory_stats = false)
```
klecki, line 40: I am against converging to infinity with the number of arguments in Executor's or Pipeline's constructor.

I see that's already done in the pipeline, do you think that Executor can get the same treatment with a setter for this option?

dali/pipeline/executor/executor.h
```diff
@@ -438,6 +505,24 @@ void Executor<WorkspacePolicy, QueuePolicy>::RunGPU() {
...
+            out_size = ws.template Output<CPUBackend>(i).nbytes();
+            reserved_size = ws.template Output<CPUBackend>(i).capacity();
+          } else {
+            out_size = ws.template Output<GPUBackend>(i).nbytes();
+            reserved_size = ws.template Output<GPUBackend>(i).capacity();
+          }
+          stats[i].first = std::max(out_size, stats[i].first);
+          stats[i].second = std::max(reserved_size, stats[i].second);
+        }
+      }
```
klecki, line 525: Hi, I'm the same piece of code that you wrote for Mixed Ops. (shame, that it's different than CPU).

JanuszL:
!build

dali/pipeline/executor/executor.h
```diff
@@ -118,6 +131,19 @@ class DLL_PUBLIC Executor : public ExecutorBase, public WorkspacePolicy, public
   DISABLE_COPY_MOVE_ASSIGN(Executor);
 
  protected:
+  using mem_stats_map = std::unordered_map<std::string, std::vector<std::pair<size_t, size_t>>>;
```
klecki, line 134: Please use some meaningful struct instead of pair of numbers.

dali/python/nvidia/dali/pipeline.py
```diff
@@ -92,11 +92,15 @@ class Pipeline(object):
...
+`get_memory_stats`: bool, optional, default = False
+    If DALI should print operator output buffer statistics.
+    Usefull for `bytes_per_sample_hint` operator parameter.
 """
     def __init__(self, batch_size = -1, num_threads = -1, device_id = -1, seed = -1,
                  exec_pipelined=True, prefetch_queue_depth=2,
                  exec_async=True, bytes_per_sample=0,
-                 set_affinity=False, max_streams=-1, default_cuda_stream_priority = 0):
+                 set_affinity=False, max_streams=-1, default_cuda_stream_priority = 0,
+                 get_memory_stats=False):
```
klecki, line 103: As we're adding new arguments and we're using Python 3 only, you can maybe place a `*` before it, so it's only allowed to be a Keyword argument.
I don't think it's good to encourage usage like `Pipeline(2, 4, 1, 0, True, 3, True, 0, True, 4, 3, False)`

dali/pipeline/executor/executor.h
```diff
@@ -94,13 +97,23 @@ class DLL_PUBLIC Executor : public ExecutorBase, public WorkspacePolicy, public
...
   }
 
+  ~Executor() {
+    if (get_memory_stats_) {
+      std::cout << "Operator memory statistics" << std::endl;
+      PrintStats(cpu_memory_stats_);
+      PrintStats(mixed_memory_stats_);
+      PrintStats(gpu_memory_stats_);
+    }
+  }
```
klecki, line 115: How about querying the numbers? You can return a dictionary of those numbers to Python (albeit not to TF), similarly to the Reader stats or the testing thingy that we're adding for hw decoder.
Maybe it's worth to provide a generic mechanism for returning such dictionary with metadata about any type of statistics per operator and for whole pipeline?

dali/pipeline/pipeline.h
```diff
@@ -310,6 +309,15 @@ class DLL_PUBLIC Pipeline {
...
 
+  /**
+   * @brief Set if the DALI pipeline should operator output buffer statistics
+   *
+   * @param get_memory_stats If DALI should print operator output buffer statistics.
+   * Usefull for `bytes_per_sample_hint` operator parameter.
+   */
+  DLL_PUBLIC void DumpOperatorOutputMemoryStatistics(bool get_memory_stats = true) {
+    get_memory_stats_ = get_memory_stats;
+  }
```
klecki, line 320: That deserves a section in Advanced guide how to set the hint in valid way for the types of used operators based on output (I don't remember the exact code, but I think CPU divides the hint for all the Tensors in Batch).

dali/pipeline/executor/executor.h
klecki: Also with several Pipelines in Python you're not sure which one is which, with multi process approach like horovod you probably can mix the outputs of different Pipelines.

Alternatively some function that allows you to force print the statistics could probably help a bit.

JanuszL: > Maybe it's worth to provide a generic mechanism for returning such dictionary with metadata about any type of statistics per operator and for whole pipeline?

I would go for the pipeline as operators are not a visible entities from the python side.

JanuszL: The advantage of having this in the destructor is lack of synchronization. If you can query it on your own you need to introduce reader/writer lock.

klecki: > The advantage of having this in the destructor is lack of synchronization. If you can query it on your own you need to introduce reader/writer lock.

Yes, and a TF support without any hassle. This could be an option for the TF plugin, to dump the statistics at some point when destructing the pipeline.

> I would go for the pipeline as operators are not a visible entities from the python side.

Accessible through Pipeline yes, I just meant some dictionaries like:
{ "op_name" : <statistics> } where <statics> can be another dictionary so you have easy mapping from std:: to python types.
The contents should be dependent on some options, that way you will be able to have one entry point that would be extendable.

klecki:
Also, we need to keep in mind to not do stats gathering in perf-critical places.

dali/pipeline/executor/executor.h
klecki: And we probably would want some stats about the whole pipeline as well at some point.

dali/pipeline/executor/async_pipelined_executor.h
JanuszL: > Also, we need to keep in mind to not do stats gathering in perf-critical places.

I don't think it does now.

JanuszL: Done

dali/pipeline/executor/executor.h

dali/pipeline/executor/executor.h

dali/python/nvidia/dali/pipeline.py

dali/pipeline/executor/executor.h

dali/pipeline/pipeline.h
JanuszL: Done

JanuszL:
!build

JanuszL:
!build

JanuszL:
!build

PR #2038, opened 2020-06-19
Add GaussianBlur CPU Op

klecki:
~Split into #2053 so it lands well below 1k lines.~
#2053 merged, this is ready again.

#### Why we need this PR?
Add Gaussian Blur Op.

 - What solution was applied:

Add small utility for holding the separated Gaussian windows
* No recalculation without parameter change
* If all axes are the same, keep only one window

Generated window normalized to sum = 1, comparable with OCV.

Use SeparableConvolution Kernel to implement Gaussian Blur.
Support for up to 3 data axes and optional sequence-first,
optional channel-last.

Operator implementation, python testing utilities.

Window generation, argument gathering, sequence processing.

 - Validation and testing:

Gtest for window utlities

Tests in python, adjusted utilites for batch compare
to allow comparison with list of np arrays.

Schema, no examples

**JIRA TASK**: *[DALI-1425]*

### DIFF
--- a/dali/operators/image/convolution/gaussian_blur.cc
+++ b/dali/operators/image/convolution/gaussian_blur.cc
@@ -1,0 +1,272 @@
+// Copyright (c) 2020, NVIDIA CORPORATION. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include <memory>
+#include <string>
+#include <vector>
+
+#include "dali/core/static_switch.h"
+#include "dali/kernels/imgproc/convolution/separable_convolution_cpu.h"
+#include "dali/kernels/kernel_manager.h"
+#include "dali/operators/image/convolution/gaussian_blur.h"
+#include "dali/operators/image/convolution/gaussian_blur_params.h"
+#include "dali/pipeline/data/views.h"
+#include "dali/pipeline/operator/common.h"
+
+namespace dali {
+
+constexpr static const char* kSigmaArgName = "sigma";
+constexpr static const char* kWindowSizeArgName = "window_size";
+
+DALI_SCHEMA(GaussianBlur)
+    .DocStr(R"code(Apply Gaussian Blur to the input.
+Separable convolution with Gaussian Kernel is used to calculate the output.
+
+User can specify the sigma or kernel window size.
+If only the sigma is provided, the radius is of kernel is calculated ``ceil(3 * sigma)``,
+thus the kernel window size is ``2 * ceil(3 * sigma) + 1``.
+
+If only the kernel window size is provided, the sigma is calculated using the following formula::
+
+  radius = (window_size - 1) / 2
+  sigma = (radius - 1) * 0.3 + 0.8
+
+Both sigma and kernel window size can be specified as single value for all data axes
+or per data axis.
+
+When specifying the sigma or window size per axis, they are provided same as layouts: from outermost
+to innermost.
+The channel ``C`` and frame ``F`` dimensions are not considered data axes.
+
+For example, with ``HWC`` input, user can provide ``sigma=1.0`` or ``sigma=(1.0, 2.0)`` as there
+are two data axes H and W.
+
+The same input can be provided as per-sample tensors.
+)code")
+    .NumInput(1)
+    .NumOutput(1)
+    .AllowSequences()
+    .SupportVolumetric()
+    .AddOptionalArg(kWindowSizeArgName, "The diameter of kernel.", std::vector<int>{0}, true)
+    .AddOptionalArg<float>(kSigmaArgName, R"code(Sigma value for Gaussian Kernel.)code",
+                           std::vector<float>{0.f}, true);
+
+/**
+ * @brief Fill the result span with the argument which can be provided as:
+ * * ArgumentInput - {result.size()}-shaped Tensor
+ * * ArgumentInput - {1}-shaped Tensor, the value will be replicated `result.size()` times
+ * * Vector input - single "repeated argument" of length {result.size()} or {1}
+ * * scalar argument - it will be replicated `result.size()` times
+ *
+ * TODO(klecki): we may want to make this a generic utility and propagate the span-approach to
+ * the rest of the related argument gettters
+ */
+template <typename T>
+void GetGeneralizedArg(span<T> result, const std::string name, int sample_idx, const OpSpec& spec,
+                       const ArgumentWorkspace& ws) {
+  int argument_length = result.size();
+  if (spec.HasTensorArgument(name)) {
+    const auto& tv = ws.ArgumentInput(name);
+    const auto& tensor = tv[sample_idx];
+    DALI_ENFORCE(tensor.shape().sample_dim() == 1,
+                 make_string("Argument ", name, " for sample ", sample_idx,
+                             " is expected to be 1D, got: ", tensor.shape().sample_dim(), "."));
+    DALI_ENFORCE(tensor.shape()[0] == 1 || tensor.shape()[0] == argument_length,
+                 make_string("Argument ", name, " for sample ", sample_idx,
+                             " is expected to have shape equal {1} or {", argument_length,
+                             "}, got: ", tensor.shape(), "."));
+    if (tensor.shape()[0] == 1) {
+      for (int i = 0; i < argument_length; i++) {
+        result[i] = tensor.data<T>()[0];
+      }
+    } else {
+      memcpy(result.data(), tensor.data<T>(), sizeof(T) * argument_length);
+    }
+    return;
+  }
+  std::vector<T> tmp;
+  // we already handled the argument input, this handles spec-related arguments only
+  GetSingleOrRepeatedArg(spec, tmp, name, argument_length);
+  memcpy(result.data(), tmp.data(), sizeof(T) * argument_length);
+}
+
+template <int axes>
+GaussianSampleParams<axes> GetSampleParams(int sample, const OpSpec& spec,
+                                           const ArgumentWorkspace& ws) {
+  GaussianSampleParams<axes> params;
+  GetGeneralizedArg<float>(make_span(params.sigmas), kSigmaArgName, sample, spec, ws);
+  GetGeneralizedArg<int>(make_span(params.window_sizes), kWindowSizeArgName, sample, spec, ws);
+  for (int i = 0; i < axes; i++) {
+    DALI_ENFORCE(
+        !(params.sigmas[i] == 0 && params.window_sizes[i] == 0),
+        make_string("`sigma` and `window_size` shouldn't be 0 at the same time for sample: ",
+                    sample, ", axis: ", i, "."));
+    DALI_ENFORCE(params.sigmas[i] >= 0,
+                 make_string("`sigma` must have non-negative values, got ", params.sigmas[i],
+                             " for sample: ", sample, ", axis: ", i, "."));
+    DALI_ENFORCE(params.window_sizes[i] >= 0,
+                 make_string("`window_size` must have non-negative values, got ",
+                             params.window_sizes[i], " for sample: ", sample, ", axis : ", i, "."));
+    if (params.window_sizes[i] == 0 && params.sigmas[i] > 0.f) {
+      params.window_sizes[i] = GaussianSigmaToDiameter(params.sigmas[i]);
+    } else if (params.sigmas[i] == 0.f && params.window_sizes[i] > 0) {
+      params.sigmas[i] = GaussianDiameterToSigma(params.window_sizes[i]);
+    }
+  }
+  return params;
+}
+
+GaussianDimDesc ParseAndValidateDim(int ndim, TensorLayout layout) {
+  static constexpr int kMaxDim = 3;
+  if (layout.empty()) {
+    // assuming plain data with no channels
+    DALI_ENFORCE(ndim <= kMaxDim,
+                 make_string("Input data with empty layout cannot have more than ", kMaxDim,
+                             " dimensions, got input with ", ndim, " dimensions."));
+    return {0, ndim, false, false};
+  }
+  // not-empty layout
+  int axes_start = 0;
+  int axes_count = ndim;
+  bool has_channels = ImageLayoutInfo::HasChannel(layout);
+  if (has_channels) {
+    axes_count--;
+    DALI_ENFORCE(ImageLayoutInfo::IsChannelLast(layout),
+                 "Only input data with no channels or channel-last is supported.");
+  }
+  bool is_sequence = layout.find('F') >= 0;
+  if (is_sequence) {
+    axes_start++;
+    axes_count--;
+    DALI_ENFORCE(
+        layout.find('F') == 0,
+        make_string("For sequence inputs frames 'F' should be the first dimension, got layout: \"",
+                    layout.str(), "\"."));
+  }
+  DALI_ENFORCE(axes_count <= kMaxDim,
+               make_string("Too many dimensions, found: ", axes_count,
+                           " data axes, maximum supported is: ", kMaxDim, "."));
+  return {axes_start, axes_count, has_channels, is_sequence};
+}
+
+// axes here is dimension of element processed by kernel - in case of sequence it's 1 less than the
+// actual dim
+template <typename T, int axes, bool has_channels>
+class GaussianBlurOpCpu : public OpImplBase<CPUBackend> {
+ public:
+  using Kernel = kernels::SeparableConvolutionCpu<T, T, float, axes, has_channels>;
+  static constexpr int ndim = Kernel::ndim;
+
+  explicit GaussianBlurOpCpu(const OpSpec& spec, const GaussianDimDesc& dim_desc)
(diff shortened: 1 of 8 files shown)


klecki:
!build

dali/operators/image/convolution/gaussian_blur.cc
```diff
@@ -0,0 +1,272 @@
...
+#include "dali/pipeline/operator/common.h"
+
+namespace dali {
+
+constexpr static const char* kSigmaArgName = "sigma";
+constexpr static const char* kWindowSizeArgName = "window_size";
+
+DALI_SCHEMA(GaussianBlur)
+    .DocStr(R"code(Apply Gaussian Blur to the input.
+Separable convolution with Gaussian Kernel is used to calculate the output.
```
JanuszL, line 34: I don't know if that is relevant. Implementation detail.

dali/operators/image/convolution/gaussian_blur.cc
```diff
@@ -0,0 +1,272 @@
...
+  if (spec.HasTensorArgument(name)) {
+    const auto& tv = ws.ArgumentInput(name);
+    const auto& tensor = tv[sample_idx];
+    DALI_ENFORCE(tensor.shape().sample_dim() == 1,
+                 make_string("Argument ", name, " for sample ", sample_idx,
+                             " is expected to be 1D, got: ", tensor.shape().sample_dim(), "."));
+    DALI_ENFORCE(tensor.shape()[0] == 1 || tensor.shape()[0] == argument_length,
+                 make_string("Argument ", name, " for sample ", sample_idx,
+                             " is expected to have shape equal {1} or {", argument_length,
+                             "}, got: ", tensor.shape(), "."));
```
JanuszL, line 88:
```suggestion
                             "} as the number of data input dimensions, got: ", tensor.shape(), "."));
```

dali/operators/image/convolution/gaussian_blur.cc
```diff
@@ -0,0 +1,272 @@
...
+  // not-empty layout
+  int axes_start = 0;
+  int axes_count = ndim;
+  bool has_channels = ImageLayoutInfo::HasChannel(layout);
+  if (has_channels) {
+    axes_count--;
+    DALI_ENFORCE(ImageLayoutInfo::IsChannelLast(layout),
+                 "Only input data with no channels or channel-last is supported.");
+  }
+  bool is_sequence = layout.find('F') >= 0;
```
JanuszL, line 148: Maybe this can be made a function inside ImageLayoutInfo.

dali/operators/image/convolution/gaussian_blur.cc
```diff
@@ -0,0 +1,272 @@
...
+
+    for (int i = 0; i < input.shape().num_samples(); i++) {
+      int seq_elements = 1;
+      int64_t stride = 0;
+      if (dim_desc_.is_sequence) {
+        auto shape = input[i].shape();
+        seq_elements = shape[0];
+        stride = volume(shape.begin() + 1, shape.end());
+      }
+      for (int elem_idx = 0; elem_idx < seq_elements; elem_idx++) {
```
JanuszL, line 215: Can you sort work the the volume?

dali/test/python/test_operator_gaussian_blur.py
```diff
@@ -0,0 +1,221 @@
...
+        pipe.set_outputs(blurred, decoded)
+    pipe.build()
+
+    result, input = pipe.run()
+    if op_type == "gpu":
+        result = result.as_cpu()
+        input = input.as_cpu()
+    input = to_batch(input, batch_size)
+    baseline_cv = [gaussian_cv(img, sigma, window_size) for img in input]
+    check_batch(result, baseline_cv, batch_size, max_allowed_error=1)
```
JanuszL, line 99: Maybe it is better to run more than one iteration?

dali/test/python/test_operator_gaussian_blur.py
```diff
@@ -0,0 +1,221 @@
...
+    pipe.build()
+
+    result, input = pipe.run()
+    if op_type == "gpu":
+        result = result.as_cpu()
+        input = input.as_cpu()
+    input = to_batch(input, batch_size)
+    baseline = [gaussian_baseline(img, sigma, window_size, axes, "F" in layout, dtype=dtype) for img in input]
+    max_error = 1 if dtype != np.float32 else 1e-07
+    check_batch(result, baseline, batch_size, max_allowed_error=max_error)
```
JanuszL, line 132: More than one iteration?

dali/test/python/test_utils.py
```diff
@@ -76,41 +76,57 @@ def get_gpu_num():
     out_list = [elm for elm in out_list if len(elm) > 0]
     return len(out_list)
 
-def check_batch(batch1, batch2, batch_size, eps = 1e-07):
+
+def check_batch(batch1, batch2, batch_size, eps=1e-07, max_allowed_error=None):
+
+    def is_error(mean_err, max_err, eps, max_allowed_error):
```
JanuszL, line 82: Shouldn't be:
```
        if max_allowed_error is not None and max_err > max_allowed_error:
            return True
        if mean_err > eps:
            return True
        return False
```

dali/operators/image/convolution/gaussian_blur.cc
klecki: Actually, it's probably worth to mention in what order do we process the data axes (innermost to outermost), as although implementation detail in some cases, it still influences the result (order of floating point operations).

dali/operators/image/convolution/gaussian_blur.cc
klecki: It's not the dimension, the dimension is (rank, or whatever) is 1 (1D). 
We check the actual shape here, we expect either 1 number, or `argument_length` numbers (as we can get either uniform scalar argument or a "vector"/repeated argument for every sample).

dali/operators/image/convolution/gaussian_blur.cc
klecki: There is already IsSequence, but it just checks for existence of 'F'. I could probably make all this work for some interleaved layouts, but that would require generalizing kernels probably.

JanuszL: I think `layout.find('F') >= 0;` just check for the existence. Not the `F` first.

dali/operators/image/convolution/gaussian_blur.cc
JanuszL: So please reword.

dali/test/python/test_operator_gaussian_blur.py
klecki: Done

dali/test/python/test_operator_gaussian_blur.py
klecki: Done

dali/test/python/test_utils.py
klecki: Done

klecki: Errm, or not done. I want to check the max_allowed_error if it's not None, otherwise use the previous behaviour - so the mean_err is checked only if the max_allowed_error in not None.

klecki: Something like:
```
        if max_allowed_error is not None:
            if max_err > max_allowed_error:
                return True
        elif mean_err > eps:
            return True
        return False
```
may also work.

dali/operators/image/convolution/gaussian_blur.cc
klecki: Done, hope it will get replaced with DALI-1473

dali/operators/image/convolution/gaussian_blur.cc
klecki: Right, I have an enforce further down. Swapped to the check from VideoLayoutInfo

dali/test/python/test_utils.py
JanuszL: So it is either max_allowed_error  or mean_err? Not both?

klecki:
!build

dali/test/python/test_utils.py
klecki: Yes, I added a comment, as there is a lot of code that uses just the mean_error, but I want to use for max_error as well without rewriting this from scratch.

dali/operators/image/convolution/gaussian_blur.cc
klecki: I think it provides the shapes as it should, unless I don't see somethings.

dali/operators/image/convolution/gaussian_blur.cc
```diff
@@ -0,0 +1,287 @@
...
+    return {0, ndim, false, false};
+  }
+  // not-empty layout
+  int axes_start = 0;
+  int axes_count = ndim;
+  bool has_channels = ImageLayoutInfo::HasChannel(layout);
+  if (has_channels) {
+    axes_count--;
+    DALI_ENFORCE(ImageLayoutInfo::IsChannelLast(layout),
+                 "Only input data with no channels or channel-last is supported.");
```
mzient, line 149: There's nothing that stops us from supporting channel last - just treat it as a sequence.

dali/operators/image/convolution/gaussian_blur.cc
```diff
@@ -0,0 +1,287 @@
...
+
+For example, with ``HWC`` input, user can provide ``sigma=1.0`` or ``sigma=(1.0, 2.0)`` as there
+are two data axes H and W.
+
+The same input can be provided as per-sample tensors.
+)code")
+    .NumInput(1)
+    .NumOutput(1)
+    .AllowSequences()
+    .SupportVolumetric()
```
mzient, line 63: It would be nice to be able to convert type - at least to "float", meaningful conversion between integers would require a scaling factor.

klecki: I added support for dtype: unspecified (passes through the input), same as input and float.

dali/operators/image/convolution/gaussian_blur.cc
klecki: Done

dali/operators/image/convolution/gaussian_blur_params.h
```diff
@@ -0,0 +1,140 @@
...
+};
+
+template <int axes>
+struct GaussianSampleParams {
+  std::array<int, axes> window_sizes;
+  std::array<float, axes> sigmas;
+
+  bool IsUniform() const {
+    for (int i = 1; i < axes; i++) {
+      if (sigmas[i - 1] != sigmas[i] || window_sizes[i - 1] != window_sizes[i]) {
```
klecki, line 49:
```suggestion
      if (sigmas[0] != sigmas[i] || window_sizes[0] != window_sizes[i]) {
```

klecki:
!build

dali/operators/image/convolution/gaussian_blur_params.h
klecki: Done

dali/operators/image/convolution/gaussian_blur.cc
```diff
@@ -0,0 +1,309 @@
...
+namespace dali {
+
+using namespace gaussian_blur;  // NOLINT
+
+constexpr static const char* kSigmaArgName = "sigma";
+constexpr static const char* kWindowSizeArgName = "window_size";
+
+DALI_SCHEMA(GaussianBlur)
+    .DocStr(R"code(Apply Gaussian Blur to the input.
+Separable convolution with Gaussian kernels is used, proceeding from innermost to outermost
```
mzient, line 38: I think that the order is not important from end-user perspective and both clutters the documentation and takes away certain implementation freedom from us, should we want to change the order.

dali/operators/image/convolution/gaussian_blur.cc
```diff
@@ -0,0 +1,309 @@
...
+dimension to calculate the output.
+
+User can specify the sigma or kernel window size.
```
mzient, line 41: Why or? They should be both configurable - actually the cutoff point is an important performance tuning factor. In some applications, more than 3 sigma would be nice, whereas in others, smaller radius gives sufficient approximation.

dali/operators/image/convolution/gaussian_blur.cc
```diff
@@ -0,0 +1,309 @@
...
+template <>
+bool GaussianBlur<CPUBackend>::SetupImpl(std::vector<OutputDesc>& output_desc,
+                                         const workspace_t<CPUBackend>& ws) {
+  const auto& input = ws.template InputRef<CPUBackend>(0);
+  auto layout = input.GetLayout();
+  auto dim_desc = ParseAndValidateDim(input.shape().sample_dim(), layout);
+  dtype_ = dtype_ != DALI_NO_TYPE ? dtype_ : input.type().id();
+  DALI_ENFORCE(dtype_ == input.type().id() || dtype_ == DALI_FLOAT,
+               "Unsupported output data type. Output type can be same as input (inferred "
+               "automatically) or set to float.");
```
mzient, line 282: A bit convoluted sentence. Alternative wording:
```suggestion
               "Output data type must be same as input, FLOAT or skipped (defaults to input type)");
```

dali/operators/image/convolution/gaussian_blur.cc
```diff
@@ -0,0 +1,309 @@
...
+                 make_string("`sigma` must have non-negative values, got ", params.sigmas[i],
+                             " for sample: ", sample, ", axis: ", i, "."));
+    DALI_ENFORCE(params.window_sizes[i] >= 0,
+                 make_string("`window_size` must have non-negative values, got ",
+                             params.window_sizes[i], " for sample: ", sample, ", axis : ", i, "."));
+    if (params.window_sizes[i] == 0 && params.sigmas[i] > 0.f) {
+      params.window_sizes[i] = SigmaToDiameter(params.sigmas[i]);
+    } else if (params.sigmas[i] == 0.f && params.window_sizes[i] > 0) {
+      params.sigmas[i] = DiameterToSigma(params.window_sizes[i]);
+    }
```
mzient, line 134: Missing else - both sigma and window size specified. It's an important feature.

klecki: It's not missing. If the user provided both sigma and window, they are already in params.sigmas and params.window_sizes. If they are positive, we just leave them to be.

This if/else if covers a case when one of them is `0` - which means calculate the value based on the other one.

I think I can remove the checks for one of the params being positive, as it's already enforced that they are non-negative and can't both be 0.

dali/operators/image/convolution/gaussian_blur.cc
klecki: Inclusive or.

dali/operators/image/convolution/gaussian_blur.cc
klecki: Done.

dali/operators/image/convolution/gaussian_blur.cc
```diff
@@ -0,0 +1,305 @@
...
+    output_desc[0].shape.resize(nsamples, input.shape().sample_dim());
+
+    params_.resize(nsamples);
+    windows_.resize(nsamples);
+
+    kmgr_.template Resize<Kernel>(nthreads, nsamples);
+
+    for (int i = 0; i < nsamples; i++) {
+      params_[i] = GetSampleParams<axes>(i, spec_, ws);
+      windows_[i].PrepareWindows(params_[i]);
```
mzient, line 204: You're recalculating (likely) the same windows over and over - this can be quite wasteful for large windows.
```suggestion
      auto p = GetSampleParams<axes>(i, spec_, ws);
      if (params_[i] != p) {
        params_[i] = p;
        windows_[i].PrepareWindows(params_[i]);
      }
```

dali/test/python/test_operator_gaussian_blur.py
```diff
@@ -0,0 +1,253 @@
...
+        return dtype(image + 0.5)
+
+
+def check_gaussian_blur(batch_size, sigma, window_size, op_type="cpu"):
+    decoder_device = "cpu" if op_type == "cpu" else "mixed"
+    pipe = Pipeline(batch_size=batch_size, num_threads=4, device_id=0)
+    with pipe:
+        input, _ = fn.file_reader(file_root=images_dir, shard_id=0, num_shards=1)
+        decoded = fn.image_decoder(input, device=decoder_device, output_type=types.RGB)
+        blurred = fn.gaussian_blur(decoded, sigma=sigma, window_size=window_size)
```
mzient, line 90: :heart:

dali/operators/image/convolution/gaussian_blur.cc
```diff
@@ -0,0 +1,305 @@
...
+ * * ArgumentInput - {result.size()}-shaped Tensor
+ * * ArgumentInput - {1}-shaped Tensor, the value will be replicated `result.size()` times
+ * * Vector input - single "repeated argument" of length {result.size()} or {1}
+ * * scalar argument - it will be replicated `result.size()` times
+ *
+ * TODO(klecki): we may want to make this a generic utility and propagate the span-approach to
+ * the rest of the related argument gettters
+ */
+template <typename T>
+void GetGeneralizedArg(span<T> result, const std::string name, int sample_idx, const OpSpec& spec,
```
mzient, line 83: Why not `Operator::GetPerSampleArgument`?

klecki: It does something else - I need to obtain vectors per sample not scalars per sample.

dali/operators/image/convolution/gaussian_blur.cc
klecki: It's already done in PrepareWindows.

klecki:
!build

dali/test/python/test_operator_gaussian_blur.py
klecki: Yes, it so much easier :D

klecki:
!build

PR #2050, opened 2020-06-23
Fix handling of CUDA streams in Python frontend

mzient:

#### Why we need this PR?
*Pick one, remove the rest*
- It fixes a bug: gpu external source test failure under GPU load

 - What solution was applied:
    * When cuda_stream is None in ExternalSource and the object is a CuPy array, issue the copy on CuPy's current stream.
     * pipeline.py
     * N/A
 - Validation and testing:
     * Existing tests apply
    * N/A

**JIRA TASK**: DALI-1474

### DIFF
--- a/dali/python/backend_impl.cc
+++ b/dali/python/backend_impl.cc
@@ -132,14 +132,13 @@ TensorShape<> FillTensorData(const py::object object, TensorType *t, int device_
                 cu_a_interface.contains("data") &&
                 PyTuple_Check(cu_a_interface["data"].ptr()) &&
                 cu_a_interface["data"].cast<py::tuple>().size() >= 2 &&
-                cu_a_interface.contains("version") &&
-                // see https://numba.pydata.org/numba-doc/dev/cuda/cuda_array_interface.html
-                // this set of atributes is tagged as version 2
-                cu_a_interface["version"].cast<int>() >= 2,
+                cu_a_interface.contains("version"),
                 "Provided object doesn't have required cuda array interface "
                 "protocol fields of necessary type.");
   DALI_ENFORCE(!cu_a_interface.contains("strides") || cu_a_interface["strides"].is_none(),
                 "Strided data is not supported");
+  DALI_ENFORCE(!cu_a_interface.contains("mask") || cu_a_interface["mask"].is_none(),
+                "Masked tensors are not supported");
 
   // Create the Tensor and wrap the data
   TensorShape<> shape = shape_from_py(cu_a_interface["shape"].cast<py::tuple>());
--- a/dali/python/backend_impl.cc
+++ b/dali/python/backend_impl.cc
@@ -135,8 +135,6 @@ TensorShape<> FillTensorData(const py::object object, TensorType *t, int device_
                 cu_a_interface.contains("version"),
                 "Provided object doesn't have required cuda array interface "
                 "protocol fields of necessary type.");
-  DALI_ENFORCE(!cu_a_interface.contains("strides") || cu_a_interface["strides"].is_none(),
-                "Strided data is not supported");
   DALI_ENFORCE(!cu_a_interface.contains("mask") || cu_a_interface["mask"].is_none(),
                 "Masked tensors are not supported");
 
@@ -145,6 +143,19 @@ TensorShape<> FillTensorData(const py::object object, TensorType *t, int device_
 
   TypeInfo type = TypeFromFormatStr(cu_a_interface["typestr"].cast<py::str>());
   size_t bytes = volume(shape) * type.size();
+
+  if (cu_a_interface.contains("strides") && !cu_a_interface["strides"].is_none()) {
+    TensorShape<> strides = shape_from_py(cu_a_interface["strides"].cast<py::tuple>());
+    DALI_ENFORCE(strides.size() == shape.size(),
+      "There should be exactly as many strides as there are extents in array shape.");
+    int64_t stride_from_shape = type.size();
+    for (int i = strides.size() - 1; i >= 0; i--) {
+      DALI_ENFORCE(strides[i] == stride_from_shape,
+          make_string("Strided data not supported. Dimension ", i, " has stride ", strides[i],
+          " whereas densely packed data of this shape would have a stride ", stride_from_shape));
+      stride_from_shape *= shape[i];
+    }
+  }
 
   t->ShareData(PyLong_AsVoidPtr(cu_a_interface["data"].cast<py::tuple>()[0].ptr()), bytes);
   t->set_type(type);
--- a/dali/python/nvidia/dali/external_source.py
+++ b/dali/python/nvidia/dali/external_source.py
@@ -295,7 +295,7 @@ Keyword Args
 def _is_external_source_with_callback(op_instance):
     return isinstance(op_instance._op, ExternalSource) and op_instance._callback is not None
 
-def external_source(source = None, num_outputs = None, *, cycle = None, name = None, device = "cpu", layout = None):
+def external_source(source = None, num_outputs = None, *, cycle = None, name = None, device = "cpu", layout = None, cuda_stream = None):
     """Creates a data node which is populated with data from a Python source.
 The data can be provided by the ``source`` function or iterable, or it can be provided by
 ``pipeline.feed_input(name, data, layout)`` inside ``pipeline.iter_setup``.
@@ -315,7 +315,7 @@ The data can be provided by the ``source`` function or iterable, or it can be pr
                 "`external_source` nodes.")
 
     op = ExternalSource(device = device, num_outputs = num_outputs,
-                        source = source, cycle = cycle, layout = layout)
+                        source = source, cycle = cycle, layout = layout, cuda_stream = cuda_stream)
     return op(name = name)
 
 external_source.__doc__ += ExternalSource._args_doc
--- a/dali/python/nvidia/dali/external_source.py
+++ b/dali/python/nvidia/dali/external_source.py
@@ -178,15 +178,20 @@ Keyword Args
     containing a distinct layout for each output. If the list has fewer elements than ``num_outputs``,
     only the first outputs have the layout set, the reset have it cleared.
 
-`cuda_stream` : a cuda stream, which is going to be used for copying data to/from GPU source.
-    If not set, best effort will be taken to maintain correctness - i.e. if the data is provided
-    as a tensor/array from a recognized library (CuPy, PyTorch), the library's current stream is
-    used. This should work in typical scenarios, but advanced use cases (and code using unsupported
-    libraries) may still need to supply the stream handle explicitly.
+`cuda_stream` : a cuda stream, which is going to be used for copying data to GPU or from a GPU
+    source. If not set, best effort will be taken to maintain correctness - i.e. if the data
+    is provided as a tensor/array from a recognized library (CuPy, PyTorch), the library's
+    current stream is used. This should work in typical scenarios, but advanced use cases
+    (and code using unsupported libraries) may still need to supply the stream handle
+    explicitly.
 
     Special values:
-     *  0 - use default CUDA stream
-     * -1 - use internal stream
+    *  0 - use default CUDA stream
+    * -1 - use DALI's internal stream
+
+    If internal stream is used, the call to ``feed_input`` will block until the copy to internal
+    buffer is complete, since there's no way to synchronize with this stream to prevent
+    overwriting the array with new data in another stream.
 """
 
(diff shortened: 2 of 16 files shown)


mzient:
!build

klecki:
My points:
* we say in the doc, that if you don't provide the stream, we will use an internal one for the copy. If the user knows that memory is ready, he may want to preserve this behaviour.
* this is fix for CuPy test by adding a bit of CuPy-specific workaround. There are other sources of cuda memory we can get here that will still cause the bug.
* I think we should add a mention in the doc, that if the user doesn't provide the stream the memory should be ready.

Maybe we can make this configurable and by default extract the stream from whatever library we can recognize and know how to do it. Otherwise allow user to turn it off?

mzient:
!build

dali/python/nvidia/dali/external_source.py
```diff
@@ -178,11 +178,13 @@ class ExternalSource():
...
-    CUDA stream to be used for the copy (only relevant for GPU inputs)
-    If not provided, an internal stream will be selected.
-    In most cases, using the default internal user stream or stream 0
-    is expected.
+`cuda_stream` : a cuda stream, which is going to be used for copying data to/from GPU source.
+    If not set, best effort will be taken to maintain correctness - i.e. default streams
+    will be used for recognized libraries.
+
+    Special values:
+     *  0 - use default CUDA stream
```
JanuszL, line 186: If possible

dali/python/nvidia/dali/plugin/mxnet.py
```diff
@@ -62,9 +62,15 @@ def feed_ndarray(dali_tensor, arr, cuda_stream = None):
     # Get CTypes void pointer to the underlying memory held by arr
     ptr = ctypes.c_void_p()
     mx.base._LIB.MXNDArrayGetData(arr.handle, ctypes.byref(ptr))
+
+    if hasattr(cuda_stream, "cuda_stream"):  # torch
```
JanuszL, line 66: Extract to common utiliyu

dali/python/nvidia/dali/plugin/pytorch.py
```diff
@@ -57,10 +57,14 @@ def feed_ndarray(dali_tensor, arr, cuda_stream = None):
     assert dali_tensor.shape() == list(arr.size()), \
             ("Shapes do not match: DALI tensor has size {0}"
             ", but PyTorch Tensor has size {1}".format(dali_tensor.shape(), list(arr.size())))
-    #turn raw int to a c void pointer
+    if cuda_stream is torch.cuda.Stream:
```
JanuszL, line 60:
```suggestion
    if cuda_stream is torch.cuda.streams.Stream:
```

JanuszL:
!build

mzient:
!build

dali/python/nvidia/dali/pipeline.py
```diff
@@ -477,26 +493,48 @@ def feed_input(self, data_node, data, layout="", cuda_stream = None):
         from nvidia.dali.external_source import _check_data_batch
         _check_data_batch(data, self._batch_size, layout)
 
+        infer_stream = False
+        if cuda_stream is None:
+            infer_stream = True
+        if cuda_stream == -1:
+            cuda_stream = None
+        else:
+            cuda_stream = types._raw_cuda_stream(cuda_stream)
```
klecki, line 502: Can we maybe use another name for cuda_stream that is passed further as the argument to this function and the one we use later have a bit different meanings with the None and -1.

klecki: Maybe something like `stream_ptr`, and make the `_raw_cuda_stream` already return a `c_void_p` values?
You're packing it by hand in every invocation place.

dali/test/python/test_external_source_pytorch_gpu.py
```diff
@@ -0,0 +1,45 @@
...
+
+def test_external_source_callback_torch_stream():
+    with torch.cuda.stream(torch.cuda.Stream()):
+        for attempt in range(10):
+            t0 = torch.tensor([attempt * 100 + 1.5], dtype=torch.float32).cuda()
+            increment = torch.tensor([10], dtype=torch.float32).cuda()
+            pipe = Pipeline(1, 3, 0)
+
+            def gen_batch():
+                nonlocal t0
```
klecki, line 37: Why do you need non-local for t0 and not increment?

dali/python/nvidia/dali/pipeline.py
klecki: Or maybe it would be better to have some boolean for `SetExternalTensorInput` indicating if the stream should be generated internally?

dali/python/nvidia/dali/external_source.py
```diff
@@ -153,8 +153,8 @@ class ExternalSource():
     again when the generator reaches end of iteration.
     In the case of the GPU input, it is the user responsibility to modify the
     provided GPU memory content only using provided stream (DALI schedules a copy on it
-    and all work is properly queued). If no stream is provided feeding input blocks until the
-    provided memory is copied to the internal buffer
+    and all work is properly queued). If no stream is provided, DALI will use a default, with
+    best-effort approach at correctness (see ``cuda_stream`` argument documentation for details).
```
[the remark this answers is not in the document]
mzient: It's sort of orthogonal - we'll still block if using internal stream, because the kind of bug it protects against is even harder to detect. Maybe we should make it explicitly configurable. Let's discuss it on dev.

dali/python/nvidia/dali/pipeline.py
```diff
@@ -447,24 +457,30 @@ def feed_input(self, data_node, data, layout="", cuda_stream = None):
...
-        data : numpy.ndarray or a list thereof
+            The name of the :class:`nvidia.dali.ops.ExternalSource` node or a :class:`DataNode`
+            object returned by a call to that ExternalSource.
+
+        data : an ndarray or a list thereof
+            The array(s) may be one of:
+              * NumPy ndarray (CPU)
+              * MXNet ndarray (CPU)
+              * PyTorch tensor (CPU or GPU)
+              * CuPy array (GPU)
```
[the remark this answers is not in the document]
mzient: Good point.

dali/python/nvidia/dali/pipeline.py
mzient: Unpacking it as a raw pointer may cause hard errors in case of the following or similar scenario:
```
stream = torch.cuda.Stream()
fn.external_source(src, cuda_stream = stream)
# stream reference is forgotten
```
If we unwrap immediately, we'll lose the reference to the stream and will be destroyed - our stream pointer will be invalid and can even be recycled by the driver upon next stream creation, where we'd coincidentally have a _different_ stream, but still invalid. I don't want our users to debug THAT kind of erros.

dali/test/python/test_external_source_pytorch_gpu.py
mzient: Because I need the same tensor to change - simply returning a new one in PyTorch resulted in synchronization and the error could not be reproduced even when streams were wrong.

klecki:
!build

dali/python/nvidia/dali/external_source.py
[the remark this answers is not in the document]
JanuszL: I don't think we need that in python level. In C++ it is there. I would still mention here that if no stream is provided it will block.

dali/python/nvidia/dali/pipeline.py
```diff
@@ -438,33 +448,45 @@ def build(self, define_graph = None):
         self._built = True
 
     def feed_input(self, data_node, data, layout="", cuda_stream = None):
-        """Bind a NumPy array (or a list thereof) to an output of ExternalSource.
-        In the case of the GPU input, it is the user responsibility to modify the
-        provided GPU memory content only using provided stream (DALI schedules
-        a copy on it and all work is properly queued). If no stream is provided
-        feed_input blocks until the provided memory is copied to the internal buffer
+        """Pass a mutlidimensional array (or a list thereof) to an output of ExternalSource.
```
JanuszL, line 451: I would copy/paste more info from ExternalSource docs.

dali/python/nvidia/dali/pipeline.py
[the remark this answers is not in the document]
JanuszL: And Python Buffer Protocol (Numpy is just an example of it). I would add this info to the ExternalSource  docs as well.

dali/python/nvidia/dali/pipeline.py
JanuszL: > Or maybe it would be better to have some boolean for SetExternalTensorInput indicating if the stream should be generated internally?

I think None is fine.

dali/python/nvidia/dali/plugin/pytorch.py
```diff
@@ -57,10 +57,12 @@ def feed_ndarray(dali_tensor, arr, cuda_stream = None):
...
             ("Shapes do not match: DALI tensor has size {0}"
             ", but PyTorch Tensor has size {1}".format(dali_tensor.shape(), list(arr.size())))
-    #turn raw int to a c void pointer
+    cuda_stream = types._raw_cuda_stream(cuda_stream)
+
+    # turn raw int to a c void pointer
     c_type_pointer = ctypes.c_void_p(arr.data_ptr())
     if isinstance(dali_tensor, (TensorGPU, TensorListGPU)):
-        dali_tensor.copy_to_external(c_type_pointer, cuda_stream)
+        dali_tensor.copy_to_external(c_type_pointer, ctypes.c_void_p(cuda_stream))
```
JanuszL, line 65: Now pass nullptr instead of None. It break the logic in:
```
    .def("copy_to_external",
        [](Tensor<GPUBackend> &t, py::object p, py::object cuda_stream, bool non_blocking) {
          void *ptr = ctypes_void_ptr(p);
          cudaStream_t stream = cuda_stream.is_none()
                ? UserStream::Get()->GetStream(t)
                : static_cast<cudaStream_t>(ctypes_void_ptr(cuda_stream));
```
I would still pass None when no stream is provided.

dali/python/nvidia/dali/plugin/paddle.py
```diff
@@ -70,13 +70,16 @@ def feed_ndarray(dali_tensor, ptr, cuda_stream = None):
...
                     CUDA stream to be used for the copy
                     (if not provided, an internal user stream will be selected)
     """
+
+    cuda_stream = types._raw_cuda_stream(cuda_stream)
+
     c_type_pointer = ctypes.c_void_p(ptr)
     if isinstance(dali_tensor, (TensorGPU, TensorListGPU)):
-        dali_tensor.copy_to_external(c_type_pointer, cuda_stream)
+        dali_tensor.copy_to_external(c_type_pointer, ctypes.c_void_p(cuda_stream))
```
JanuszL, line 82: Now you will get nullptr instead of None. This breaks the logic in:
```
    .def("copy_to_external",
        [](Tensor<GPUBackend> &t, py::object p, py::object cuda_stream, bool non_blocking) {
          void *ptr = ctypes_void_ptr(p);
          cudaStream_t stream = cuda_stream.is_none()
                ? UserStream::Get()->GetStream(t)
                : static_cast<cudaStream_t>(ctypes_void_ptr(cuda_stream));
```

dali/python/nvidia/dali/plugin/pytorch.py
JanuszL, line 65: As above

dali/python/nvidia/dali/external_source.py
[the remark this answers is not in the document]
mzient: It's mentioned in the documentation of cuda_stream parameter.

mzient:
!build

dali/python/backend_impl.cc
```diff
@@ -132,21 +132,31 @@ TensorShape<> FillTensorData(const py::object object, TensorType *t, int device_
...
+  if (cu_a_interface.contains("strides") && !cu_a_interface["strides"].is_none()) {
+    TensorShape<> strides = shape_from_py(cu_a_interface["strides"].cast<py::tuple>());
+    DALI_ENFORCE(strides.size() == shape.size(),
+      "There should be exactly as many strides as there are extents in array shape.");
+    int64_t stride_from_shape = type.size();
+    for (int i = strides.size() - 1; i >= 0; i--) {
+      DALI_ENFORCE(strides[i] == stride_from_shape,
+          make_string("Strided data not supported. Dimension ", i, " has stride ", strides[i],
+          " whereas densely packed data of this shape would have a stride ", stride_from_shape));
+      stride_from_shape *= shape[i];
```
JanuszL, line 156: We use this check in a couple of places now. Can you extract it to the function?

PR #2051, opened 2020-06-23
Deprecate `output_dtype`, use `dtype`

klecki:
Unifies the argument used for Operators

#### Why we need this PR?
Unifies the codebase

 - What solution was applied:
     Removed `output_dtype` from C++ apis/schema
	 Added auto conversion from `output_dtype` to `dtype` in python with a warning (and error if both are provided).
     Ops that used `output_dtype`, processing Operator arguments in python.
     Check if there is no leftover
 - Validation and testing:
     Test for deprecation and error condition added
     Docs adjusted to new usages, removed mentions of the old ones.

**JIRA TASK**: *[NA]*


klecki:
!build

dali/test/python/test_pipeline.py
```diff
@@ -1584,3 +1587,57 @@ def calc_avg_max(val):
...
+    with pipe:
+        input = fn.external_source(data, layout="HWC")
+        cmn = fn.crop_mirror_normalize(input, device="cpu",
+                                                output_dtype=types.FLOAT,
+                                            dtype=types.FLOAT,
+                                                output_layout="HWC",
+                                                crop=(32, 32),
+                                                image_type=types.RGB,
+                                                mean=[128., 128., 128.],
+                                                std=[1., 1., 1.])
```
mzient, line 1639: Nitpick: formatting

klecki: done

klecki:
!build

dali/test/python/test_pipeline.py
```diff
@@ -1584,3 +1587,57 @@ def calc_avg_max(val):
...
+        pipe.set_outputs(cmn)
+    pipe.build()
+
```
JanuszL: Maybe you can add a test that would go over all operators inside`l _cpu_ops`, `_gpu_ops`, `_mixed_opsand` and check the output from `GetArgumentNames`?

klecki: I think it would be enough to just search for `output_dtype` in the SupportedOperators page. Unless you really want me to make sure it won't ever appear again?

JanuszL: Fine

PR #2053, opened 2020-06-24
Add Gaussian window calculation for Gaussian Op

klecki:
Needed for #2038.

#### Why we need this PR?
Needed for Gausian Blur CPU op to be used with Convolution Kernels.

 - What solution was applied:
     Calculate half of Gaussian Kernel window, duplicate for second half (symmetric) and normalize so it sums up to 1.
     Keep the state of the several windows based on provided params
     Operator utilites
    
 - Validation and testing:
     Gtest test added, compared with OpenCV.

**JIRA TASK**: *[DALI-1425]*

### DIFF
--- a/dali/operators/image/convolution/gaussian_blur_params.h
+++ b/dali/operators/image/convolution/gaussian_blur_params.h
@@ -1,0 +1,140 @@
+// Copyright (c) 2020, NVIDIA CORPORATION. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef DALI_OPERATORS_IMAGE_CONVOLUTION_GAUSSIAN_BLUR_PARAMS_H_
+#define DALI_OPERATORS_IMAGE_CONVOLUTION_GAUSSIAN_BLUR_PARAMS_H_
+
+#include <vector>
+
+#include "dali/pipeline/data/views.h"
+#include "dali/pipeline/operator/common.h"
+
+namespace dali {
+
+int GaussianSigmaToDiameter(float sigma) {
+  return 2 * ceilf(sigma * 3) + 1;
+}
+
+float GaussianDiameterToSigma(int diameter) {
+  // Based on OpenCV
+  int radius = (diameter - 1) / 2;
+  return (radius - 1) * 0.3 + 0.8;
+}
+
+struct GaussianDimDesc {
+  int usable_axes_start;
+  int usable_axes_count;
+  bool has_channels;
+  bool is_sequence;
+};
+
+template <int axes>
+struct GaussianSampleParams {
+  std::array<int, axes> window_sizes;
+  std::array<float, axes> sigmas;
+
+  bool IsUniform() const {
+    for (int i = 1; i < axes; i++) {
+      if (sigmas[i - 1] != sigmas[i] || window_sizes[i - 1] != window_sizes[i]) {
+        return false;
+      }
+    }
+    return true;
+  }
+
+  bool operator==(const GaussianSampleParams<axes> &other) const {
+    return window_sizes == other.window_sizes && sigmas == other.sigmas;
+  }
+
+  bool operator!=(const GaussianSampleParams<axes> &other) const {
+    return !(*this == other);
+  }
+};
+
+void FillGaussian(const TensorView<StorageCPU, float, 1> &window, float sigma) {
+  int r = (window.num_elements() - 1) / 2;
+  // 1 / sqrt(2 * pi * sigma^2) * exp(-(x^2) / (2 * sigma^2))
+  // the 1 / sqrt(2 * pi * sigma^2) coefficient disappears as we normalize the sum to 1.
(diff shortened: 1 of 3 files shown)


klecki:
!build

dali/operators/image/convolution/gaussian_blur_params.h
```diff
@@ -0,0 +1,140 @@
...
+#ifndef DALI_OPERATORS_IMAGE_CONVOLUTION_GAUSSIAN_BLUR_PARAMS_H_
+#define DALI_OPERATORS_IMAGE_CONVOLUTION_GAUSSIAN_BLUR_PARAMS_H_
+
+#include <vector>
+
+#include "dali/pipeline/data/views.h"
+#include "dali/pipeline/operator/common.h"
+
+namespace dali {
+
```
mzient, line 24: looks like you could use a namespace `gaussian_blur`, given how many names are prefixed with `Gaussian` here.

dali/operators/image/convolution/gaussian_blur_params.h
```diff
@@ -0,0 +1,140 @@
...
+  int usable_axes_start;
+  int usable_axes_count;
+  bool has_channels;
+  bool is_sequence;
+};
+
+template <int axes>
+struct GaussianSampleParams {
+  std::array<int, axes> window_sizes;
+  std::array<float, axes> sigmas;
```
mzient, line 45: This could be useful for GPU if not for std::array.

dali/operators/image/convolution/gaussian_blur_params.h
klecki: And remove the `Gaussian` prefix from everything I presume?

dali/operators/image/convolution/gaussian_blur_params.h
```diff
@@ -0,0 +1,140 @@
...
+  int r = (window.num_elements() - 1) / 2;
+  // 1 / sqrt(2 * pi * sigma^2) * exp(-(x^2) / (2 * sigma^2))
+  // the 1 / sqrt(2 * pi * sigma^2) coefficient disappears as we normalize the sum to 1.
+  double exp_scale = 0.5 / (sigma * sigma);
+  double sum = 0.;
+  // Calculate first half
+  for (int x = -r; x < 0; x++) {
+    *window(x + r) = exp(-(x * x * exp_scale));
+    sum += *window(x + r);
+  }
```
mzient, line 75: Does it need to be in double precision?

dali/operators/image/convolution/gaussian_blur_params.h
```diff
@@ -0,0 +1,140 @@
...
+  sum *= 2.;
+  sum += 1.0;
+  double scale = 1. / sum;
+  // place center, scaled element
+  *window(r) = scale;
+  // scale all elements so they sum up to 1, duplicate the second half
+  for (int x = 0; x < r; x++) {
+    *window(x) *= scale;
+    *window(2 * r - x) = *window(x);
+  }
```
mzient, line 86: SIMD killer award goes to....

dali/operators/image/convolution/gaussian_blur_params.h
```diff
@@ -0,0 +1,140 @@
...
+  return 2 * ceilf(sigma * 3) + 1;
+}
+
+float GaussianDiameterToSigma(int diameter) {
+  // Based on OpenCV
+  int radius = (diameter - 1) / 2;
+  return (radius - 1) * 0.3 + 0.8;
+}
+
+struct GaussianDimDesc {
```
mzient, line 35: This name is strange - if you insist on making it specific, I'd suggest GaussianBlurDimDesc - because this desc is not Gaussian.

dali/operators/image/convolution/gaussian_blur_params.h
```diff
@@ -0,0 +1,140 @@
...
+  int usable_axes_start;
+  int usable_axes_count;
+  bool has_channels;
+  bool is_sequence;
+};
+
+template <int axes>
+struct GaussianSampleParams {
```
mzient, line 43:
```suggestion
struct GaussianBlurSampleParams {
```
or a namespace

dali/operators/image/convolution/gaussian_blur_params.h
klecki: Oh, that one sneaked from the Op

dali/operators/image/convolution/gaussian_blur_params.h
klecki: Hmm, I added a namespace, but I will keep the name on the not-obvious ones like SampleParams. IMO it's clearer when they indicate that they are related to Gaussian.

dali/operators/image/convolution/gaussian_blur_params.h
klecki: Changed to DeviceArray.

dali/operators/image/convolution/gaussian_blur_params.h
klecki: Probably not, adjusted.

klecki:
!build

dali/operators/image/convolution/gaussian_blur_params.h
```diff
@@ -0,0 +1,143 @@
...
+};
+
+template <int axes>
+struct GaussianBlurParams {
+  DeviceArray<int, axes> window_sizes;
+  DeviceArray<float, axes> sigmas;
+
+  bool IsUniform() const {
+    for (int i = 1; i < axes; i++) {
+      if (sigmas[i - 1] != sigmas[i] || window_sizes[i - 1] != window_sizes[i]) {
```
JanuszL, line 51: I don't know how much it changes the perf but how about:
```
if (sigmas[0] != sigmas[i] || window_sizes[0] != window_sizes[i]) {
```

klecki: I can change it in the next one, as the build will soon finish.

PR #2099, opened 2020-07-09
Fix aarch64 builds that are sill gcc 5.x based

JanuszL:
- fixes compiler error when defining a template method of an explicitly specialized class template outside the body of the class
- fixes compiler error when moving unique_ptr in the lambda capture list -  now it used custom functor that is full movable

#### Why we need this PR?
*Pick one, remove the rest*
- It fixes an aarch64 builds that are sill gcc 5.x based

 - What solution was applied:
     fixes compiler error when defining a template method of an explicitly specialized class template outside the body of the class
     fixes compiler error when moving unique_ptr in the lambda capture list -  now it used custom functor that is full movable
     ExternalSource
     C++ magic?
 - Validation and testing:
     CI with aarch64 builds
     NA

**JIRA TASK**: *[NA]*

### DIFF
--- a/dali/pipeline/operator/builtin/external_source.cu
+++ b/dali/pipeline/operator/builtin/external_source.cu
@@ -41,7 +41,12 @@ struct ExternalSource<GPUBackend>::RecycleFunctor {
   std::list<uptr_cuda_event_type> event, copy_to_gpu;
   std::list<uptr_tl_type> ptr;
   void operator()() {
-    owner->RecycleBuffer(ptr, &event, &copy_to_gpu);
+    std::list<uptr_cuda_event_type> *event_ptr = nullptr;
+    std::list<uptr_cuda_event_type> *copy_to_gpu_ptr = nullptr;
+    if (event.size()) event_ptr = &event;
+    if (copy_to_gpu.size()) copy_to_gpu_ptr = &copy_to_gpu;
+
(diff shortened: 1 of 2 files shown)


dali/pipeline/operator/builtin/external_source.cu
```diff
@@ -80,21 +85,18 @@ void ExternalSource<GPUBackend>::RunImpl(DeviceWorkspace &ws) {
...
-    auto tmp_shr_ptr = shared_ptr<void>(ptr, [
-                          this,
-                          tensor_list = std::move(tensor_list_elm),
-                          copy_to_storage = std::move(internal_copy_to_storage)](void*) {
-                          RecycleBuffer(*(const_cast<std::list<uptr_tl_type>*>(&tensor_list)),
-                                        nullptr,
-                                        const_cast<std::list<uptr_cuda_event_type>*>
-                                                  (&copy_to_storage));
-                       });
+    RecycleFunctor funct{this, std::move(cuda_event), std::move(tensor_list_elm),
```
klecki, line 94: Oh, this thing has no-op copy constructor with an assert. Can you try it in debug with asserts to see if it indeed doesn't happen?
Checking local build with tests should be enough.

JanuszL: Double tested. Works with debug build as well.

PR #2103, opened 2020-07-10
Detect version of CUDA based on libcudart.so.* name

JanuszL:
- version.txt is deprecated by CUDA 11 so it cannot be used any longer  to detect installed CUDA version
- switches to parsing libcudart.so.* name to detect major CUDA version

#### Why we need this PR?
*Pick one, remove the rest*
- It fixes a version detection

 - What solution was applied:
     switches to parsing libcudart.so.* name to detect major CUDA version
     tests
     customop builder
     NA
 - Validation and testing:
     CI
     NA

**JIRA TASK**: *[NA]*


JanuszL:
Merged https://github.com/NVIDIA/DALI/pull/2105 with one fix on top.

PR #2112, opened 2020-07-15
Fix the lack of setting the right device in the ExternalSource

JanuszL:
- when SetDataSource is called it uses the default - 0 device to allocate a staging buffer. In multiple GPU scenario, it needs to set the right device the is the same as the DALI pipeline device

#### Why we need this PR?
*Pick one, remove the rest*
- It fixes the lack of setting the right device in the ExternalSource

 - What solution was applied:
     when SetDataSource is called it uses the default - 0 device to allocate a staging buffer. In multiple GPU scenario, it needs to set the right device the is the same as the DALI pipeline device
     ExternalSource
     NA
 - Validation and testing:
     NA
     NA

**JIRA TASK**: *[NA]*


JanuszL:
!build

PR #2115, opened 2020-07-15
Reduce Paddle RN50 test gpu mem fraction to 80%

klecki:

#### Why we need this PR?
Reduce size of test that gets OOM for some CI configurations

**JIRA TASK**: *[Use DALI-1508 or NA]*


klecki:
!build

PR #2120, opened 2020-07-15
Reduce RN50 paddle test size

klecki:

#### Why we need this PR?
Reduce size of test that gets OOM for some CI configurations

 - What solution was applied:
      Test size reduced
     Paddle RN50 test
     bash
 - Validation and testing:
     CI
     N/A

**JIRA TASK**: *[Use DALI-1508 or NA]*


klecki:
Alternative to #2115 - if it stops working, this one can be used, closing for now as not needed.

PR #2128, opened 2020-07-17
Optimize arithmetic op parameters

klecki:
#### Why we need this PR?
Adjust arithmetic op launch parameters and tile sizes for better performance.

 - What solution was applied:
     Test added, tile size and grid/block spec adjusted.
     Arithmetic Ops
     Better values?
 - Validation and testing:
     Various measurements
     NA

**JIRA TASK**: *[Use DALI-1514 or NA]*


klecki:
Cleaned up in #2137

PR #2130, opened 2020-07-20
Fix PaddlePaddle test broken by rarfile update not compatible with Python 3.5

JanuszL:
- with the recent release of rarfile version 3.2, PaddlePaddle depends on Python 3.5 support is broken so DALI need to pin version to <=3.1 to keep the compatibility as long as we support Python 3.5

#### Why we need this PR?
*Pick one, remove the rest*
- It fixes PaddlePaddle test broken by rarfile update not compatible with Python 3.5

 - What solution was applied:
     pins rarfile version to <=3.1
     PaddlePaddle tests
     NA
 - Validation and testing:
     CI
     NA

**JIRA TASK**: *[NA]*

### DIFF
--- a/qa/TL0_FW_iterators/test_paddle.sh
+++ b/qa/TL0_FW_iterators/test_paddle.sh
@@ -1,8 +1,7 @@
 #!/bin/bash -e
 # used pip packages
-
-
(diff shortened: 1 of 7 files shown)


JanuszL:
!build

PR #2135, opened 2020-07-21
Fix TF dataset notebook

awolant:

#### Why we need this PR?
*Pick one, remove the rest*
- It fixes a bug in TF dataset notebook related to latest new checks.

**JIRA TASK**: *[DALI-1505]*

### DIFF
--- a/docs/examples/frameworks/tensorflow/tensorflow-dataset.ipynb
+++ b/docs/examples/frameworks/tensorflow/tensorflow-dataset.ipynb
@@ -328,6 +328,7 @@
     "    with tf.device('/gpu:0'):\n",
     "        mnist_pipeline = MnistPipeline(BATCH_SIZE, device='gpu', device_id=0)\n",
     "        mnist_set = dali_tf.DALIDataset(\n",
+    "            fail_on_device_mismatch=False,\n",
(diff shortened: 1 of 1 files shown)


awolant:
!build

awolant:
!build

docs/examples/frameworks/tensorflow/tensorflow-dataset.ipynb
```diff
@@ -328,6 +328,7 @@
     "    with tf.device('/gpu:0'):\n",
```
JanuszL, line 328: <p>Why we need this now?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/NVIDIA/DALI/pull/2135/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='19'/>

awolant: Without it we would get CPU allocator in the Interator. So the situation here is that we get proper GPU allocator but the check for device compatibility fails saying that the op was placed on the CPU. This is why I decided to add is as an option to ignore this check.

Issue #2140, opened 2020-07-22
Why is _iterator_deprecation_warning() impossible to suppress?


dreavjr:
Does it mean that reader_name will very soon become mandatory? In that case, what is the design pattern to deal with ops.ExternalSource (especially in the case where num_outputs>1)?

If that is not the case (reader_name is not becoming mandatory, at least not very soon), I'd suggest the warning to have standard/expected Python behavior, making it possible to suppress.

JanuszL:
Hi,
It is recommended to not set the Iterator size manually. If you do then you see the warning.
If you want to use ExternalSource don't set `size` of just use -1 `Providing -1 means that the iterator will work until StopIteration is raised from the inside of iter_setup()` and let the ExternalSource drive the framework iterator. In that case, providing the `size` doesn't make any sense as the user can use any custom logic determining epoch length.

dreavjr:
Hi, thank you for the prompt answer!

I'm still a bit confused by what you mean. In the [template provided for using ExternalSource](https://docs.nvidia.com/deeplearning/dali/user-guide/docs/examples/frameworks/pytorch/pytorch-external_input.html) the logic is driven by using `StopIteration`, but `size`, `last_batch_padded`, and `fill_last_batch` are still set. Do you mean that in that example I could have omitted those parameters and the effect would've been the same?

dreavjr:
I did some tests, I found out that the template provided would _almost_ work as intended (without need to set `size`) by changing the functions `__iter__` and `__next__` in `ExternalInputIterator` to:

```
def __iter__(self):
    self.i = 0
    self.count = 0
    shuffle(self.files)
    return self

def __next__(self):
    batch = []
    labels = []

    if self.count >= self.n:
        raise StopIteration

    for _ in range(self.batch_size):
        jpeg_filename, label = self.files[self.i].split(' ')
        f = open(self.images_dir + jpeg_filename, 'rb')
        batch.append(np.frombuffer(f.read(), dtype = np.uint8))
        labels.append(np.array([label], dtype = np.uint8))
        self.i = (self.i + 1) % self.n
        self.count += 1
    return (batch, labels)
```

With those changes the following line in the main routine...
```
pii = PyTorchIterator(pipe, size=eii.size, last_batch_padded=True, fill_last_batch=False)
```
...can become simply:
```
pii = PyTorchIterator(pipe)
```

[...]

PR #2154, opened 2020-07-27
Add an ability to suppress _iterator_deprecation_warning

JanuszL:
- adds an ability to suppress _iterator_deprecation_warning by removing  catch_warnings call
- there are still valid use cases when the user wants to use ExternalSource and use fill_last_batch and last_batch_padded parameters by providing expected data set size - like in case of validation when the exact number of relevant samples matters

#### Why we need this PR?
*Pick one, remove the rest*
- It adds an ability to suppress _iterator_deprecation_warning

 - What solution was applied:
     adds an ability to suppress _iterator_deprecation_warning by removing  catch_warnings call
     FW base iterator
     NA
 - Validation and testing:
     CI
     NA

Related https://github.com/NVIDIA/DALI/issues/2140

**JIRA TASK**: *[NA]*

### DIFF
--- a/dali/python/nvidia/dali/plugin/base_iterator.py
+++ b/dali/python/nvidia/dali/plugin/base_iterator.py
@@ -19,14 +19,11 @@ import numpy as np
 import warnings
 
 def _iterator_deprecation_warning():
-    # show only this warning
-    with warnings.catch_warnings():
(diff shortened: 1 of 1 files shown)


JanuszL:
!build

PR #2155, opened 2020-07-27
Change DALI build command to use minor CUDA version as well

JanuszL:
- adds the need to provide cuda minor version digit as a part of CUDA version when issuing DALI build

#### Why we need this PR?
*Pick one, remove the rest*
- It adds the usage of CUDA minor version in the DALI build command

 - What solution was applied:
     adds the need to provide cuda minor version digit as a part of CUDA version when issuing DALI build
     build system
     NA
 - Validation and testing:
     NA
     compilation guide have been updated

**JIRA TASK**: *[NA]*


docker/build.sh
```diff
@@ -39,13 +39,13 @@ shift $((OPTIND - 1))
 export ARCH=${ARCH:-x86_64}
 export PYVER=${PYVER:-3.6}
 export PYV=${PYVER/./}
-export CUDA_VERSION=${CUDA_VERSION:-11}
+export CUDA_VERSION=${CUDA_VERSION:-110}
```
klecki, line 42: Would there be any benefit of handling it similarly to PYVER and allowing user to provide `11.1` instead of `111`?

JanuszL: Why not.

JanuszL: Done

docs/compilation.rst
```diff
@@ -29,7 +29,7 @@ Building Python wheel and (optionally) Docker image
 Change directory (``cd``) into ``docker`` directory and run ``./build.sh``. If needed, set the following environment variables:
 
 * PYVER - Python version used to create a docker image with DALI installed inside. The default is ``3.6``.
-* CUDA_VERSION - CUDA toolkit version (10 for 10.0 or 11 for 11.0). The default is ``11``. If the version is prefixed with `.` then any value ``XX`` can be passed and the user needs to make sure that Dockerfile.cudaXX.deps is present in `docker/` directory.
+* CUDA_VERSION - CUDA toolkit version (10.0 or11.0). The default is ``11.0``. If the version is prefixed with `.` then any value ``XX.Y`` can be passed and the user needs to make sure that Dockerfile.cudaXXY.deps is present in `docker/` directory.
```
klecki, line 32:
```suggestion
* CUDA_VERSION - CUDA toolkit version (10.0 or 11.0). The default is ``11.0``. If the version is prefixed with `.` then any value ``XX.Y`` can be passed and the user needs to make sure that Dockerfile.cudaXXY.deps is present in `docker/` directory.
```

klecki: I don't understand the "If the version is prefixed with `.` " part.

docs/compilation.rst
```diff
@@ -56,9 +56,9 @@ For example:
 
 .. code-block:: bash
 
-  PYVER=3.6 CUDA_VERSION=10 ./build.sh
+  PYVER=3.6 CUDA_VERSION=110 ./build.sh
```
klecki, line 59:
```suggestion
  PYVER=3.6 CUDA_VERSION=11.0 ./build.sh
```
(this will still work)

docs/compilation.rst
JanuszL: How about now?

docs/compilation.rst
JanuszL: Done

PR #2156, opened 2020-07-27
Span-of-arrays flattening + minor layout utils

mzient:
#### Why we need this PR?
*Pick one, remove the rest*
- It adds new feature: span flattening; expose IsSpatialDim and implement NumSpatialDims in terms of it.

 - What solution was applied:
     * utils: flatten - recursive flattening with no-op tail call; statically calculate new span extent, if possible
     * TensorLayout - factor out IsSpatialDim
     * core/util, TensorLayout
     * N/A
 - Validation and testing:
     * GTest
     * N/A

**JIRA TASK**: N/A

### DIFF
--- a/include/dali/core/span.h
+++ b/include/dali/core/span.h
@@ -192,6 +192,7 @@ DALI_HOST_DEV constexpr auto make_span(Collection &&c) {
   return make_span(c.data(), c.size());
 }
 
+DALI_NO_EXEC_CHECK
 template <typename T, size_t N>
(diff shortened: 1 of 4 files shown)


include/dali/core/span.h
```diff
@@ -192,6 +192,7 @@ DALI_HOST_DEV constexpr auto make_span(Collection &&c) {
   return make_span(c.data(), c.size());
 }
 
+DALI_NO_EXEC_CHECK
```
mzient, line 195: suppress NVCC warning when attempting to use this function in a .cu file

mzient:
!build

PR #2160, opened 2020-07-28
Make TF dataset build for 2.3.0

awolant:

#### Why we need this PR?
 - It adds support for TF 2.3.0
 - Updates other TF versions

**JIRA TASK**: *[DALI-1513]*

### DIFF
--- a/qa/setup_packages.py
+++ b/qa/setup_packages.py
@@ -322,8 +322,7 @@ all_packages = [PlainPackage("opencv-python", ["4.2.0.32"]),
                         "mxnet-cu{cuda_v}"),
                 CudaPackage("tensorflow-gpu",
                         { "100" : [
-                              PckgVer("1.15.2",  python_max_ver="3.7"),
-                              PckgVer("2.1.1",  python_max_ver="3.7"),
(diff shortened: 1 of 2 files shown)


awolant:
!build

qa/setup_packages.py
```diff
@@ -324,7 +324,8 @@ def get_pyvers_name(self, url, cuda_version):
                         { "100" : [
                               PckgVer("1.15.2",  python_max_ver="3.7"),
                               PckgVer("2.1.1",  python_max_ver="3.7"),
-                              "2.2.0"]
+                              "2.2.0",
+                              "2.3.0"]
```
awolant, line 328: With regards to this, what do you think about dropping tests for 2.1.1? And possibly moving from 1.15.2 to 1.15.3? Almost no difference there, but it is the latest version.

JanuszL: I'm for both ideas.

awolant:
!build

qa/setup_packages.py
awolant: Ok, done.

PR #2162, opened 2020-07-28
Workaround a bug in compiler, magically converting instance call to static call.

mzient:
Workaround a bug in compiler, magically converting instance call to static call.

#### Why we need this PR?
*Pick one, remove the rest*
- It works around a compiler bug: "cannot call member Release without object"

 - What solution was applied:
     * Added a proxy static function with object passed as argument
     * PerStreamPool
     * N/A
 - Validation and testing:
     * Now it compiles on machine with affected compiler
     * N/A

**JIRA TASK**: N/A


mzient:
!build

PR #2208, opened 2020-08-12
Fix failure when DALI runs on pre CUDA 11 driver using a compatibility layer

JanuszL:
- fixes crash caused by missing nvml symbol nvmlDeviceGetCpuAffinityWithinScope when DALI CUDA 11 build runs on pre 450.x driver with the compatibility layer enabled

#### Why we need this PR?
*Pick one, remove the rest*
- It fixes a failure when DALI runs on pre CUDA 11 driver using a compatibility layer

 - What solution was applied:
     fixes crash caused by missing nvml symbol nvmlDeviceGetCpuAffinityWithinScope when DALI CUDA 11 build runs on pre 450.x driver with the compatibility layer enabled
     nvml wraper
     if `is_driver_sufficient` is safe
 - Validation and testing:
     CI
     NA

**JIRA TASK**: *[NA]*


JanuszL:
!build

Issue #2213, opened 2020-08-15
Compiling the TF plugin (using docker) fails 


kindoblue:
I'm trying to compile Dali with the following command, in the docker directory

`BUILD_TF_PLUGIN=YES PYVER=3.7 CUDA_VERSION=10.0 ./build.sh`

Dali got compiled and the wheel generated but then the script starts to build the TF plugin and I get the following error. 

```bash
+ nvidia-docker run --name extract_dali_tf_prebuilt_manylinux1 nvidia/dali:cu100.build_tf_manylinux1 /bin/bash -c 'source /opt/dali/dali_tf_plugin/build_in_custom_op_docker.sh'
./build.sh: line 261: nvidia-docker: command not found
```

I don't recall reading in the documentation about installing nvidia-docker. Is it really needed for building a plugin within a docker image? 

On Ubuntu 20.04. Using docker script.

kindoblue:
I've installed nvidia-docker and the build process goes further but yet ends with an error

```bash
+ nvidia-docker run --name extract_dali_tf_prebuilt_manylinux1 nvidia/dali:cu100.build_tf_manylinux1 /bin/bash -c 'source /opt/dali/dali_tf_plugin/build_in_custom_op_docker.sh'
++ set -e
++ PYTHON_DIST_PACKAGES=($(python -c "import site; print(site.getsitepackages()[0])"))
+++ python -c 'import site; print(site.getsitepackages()[0])'
++ DALI_TOPDIR=/usr/local/lib/python2.7/dist-packages/nvidia/dali
+++ cat /usr/local/cuda/version.txt
+++ head -1
+++ sed 's/.*Version \([0-9]\+\)\.\([0-9]\+\).*/\1\2/'
++ CUDA_VERSION=100
+++ python ../qa/setup_packages.py -n -u tensorflow-gpu --cuda 100
Traceback (most recent call last):
  File "../qa/setup_packages.py", line 6, in <module>
    import urllib.parse
ImportError: No module named parse
++ LAST_CONFIG_INDEX=
```

klecki:
Hi @kindoblue,
sorry for the confusion with the nvidia-docker, I will update the documentation and script so it can support the new syntax from `NVIDIA Container Toolkit` (docker run --gpus all).

As for the second error, it looks like the script does not propagate the Python version properly and uses Python 2.7 for the TF Plugin containers. I will try to post a fix soon, will get back to you when U have a PR.

Thanks for reporting that.

[...]

PR #2214, opened 2020-08-17
Fix docker/build.sh to use Python 3 for TF plugin

klecki:
Adjust docs to mention that Python version is needed only for runner.
Add requirement for NVIDIA Container Toolkit when building TF plugin
with build.sh
Update build.sh to all 3 possible nvidia-docker commands.

#### Why we need this PR?
Fixes a bug reported in #2213. 

 - What solution was applied:
     TF plugin builder now uses Python 3 instead Python 2.
     Update build.sh to all 3 possible nvidia-docker commands.
     docker/build.sh and compilation docs.
	 Additional fixes for runner image
      
 - Validation and testing:
     Local workstation
     Compilation docs mention that the wheel is minor version agnostic.

**JIRA TASK**: *[NA]*

### DIFF
--- a/docker/Docker_run_cuda
+++ b/docker/Docker_run_cuda
@@ -11,13 +11,16 @@ ENV PYV=${PYV}
 RUN apt-get update && apt-get install -y --no-install-recommends \
         curl \
         git \
+        python3-distutils \
         python$PYVER \
         python$PYVER-dev && \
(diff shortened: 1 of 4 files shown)


klecki:
!build

awolant (approved):

PR #2215, opened 2020-08-18
Fix NVTX annotations

AntoineFroger:
Zero-initialize variables of type `nvtxEventAttributes_t`.

The `nvtxEventAttributes_t` type is trivial. When defined on the stack,
all non-initialized fields of a variable of this type will be set to
unexpected values. This might cause various bugs in tools as Nsight
Systems trying to interpret those random values.

Signed-off-by: Antoine Froger <afroger@nvidia.com>

#### Why we need this PR?

To generate correct NVTX annotations.

 - What solution was applied:
     *[Zero-initialize variables of type `nvtxEventAttributes_t`]*
     *[NVTX annotations generated by the `TimeRange` objects when NVTX is enabled in the build]*
 - Validation and testing:
     *[Nsight Systems can now trace the NVTX annotations generated by DALI without throwing a runtime exception]*

**JIRA TASK**: *[DALI-1573]*

### DIFF
--- a/include/dali/core/common.h
+++ b/include/dali/core/common.h
@@ -164,7 +164,7 @@ struct TimeRange {
 
   TimeRange(std::string name, const uint32_t rgb = kBlue) {  // NOLINT
 #if NVTX_ENABLED
-    nvtxEventAttributes_t att;
+    nvtxEventAttributes_t att = {};
     att.version = NVTX_VERSION;
     att.size = NVTX_EVENT_ATTRIB_STRUCT_SIZE;
     att.colorType = NVTX_COLOR_ARGB;


include/dali/core/common.h
```diff
@@ -165,6 +165,8 @@ struct TimeRange {
   TimeRange(std::string name, const uint32_t rgb = kBlue) {  // NOLINT
 #if NVTX_ENABLED
     nvtxEventAttributes_t att;
+    memset(&att, 0, sizeof(att));
```
mzient, line 168: Nitpick:
```suggestion
    nvtxEventAttributes_t att = {};
```
should do the trick and be "more C++"

AntoineFroger: I learned from experience that some older compilers have trouble with this notation.  For instance, compiling this simple line on a GCC 4.8.3 cross-compiler for PPC will lead to the following errors:

```
NvtxSimpleV3.cpp: In function 'void simple_nvtxMarker(nvtxDomainHandle_t)':
NvtxSimpleV3.cpp:125:42: error: missing initializer for member 'nvtxEventAttributes_v2::version' [-Werror=missing-field-initializers]
     nvtxEventAttributes_t eventAttrib = {};

NvtxSimpleV3.cpp:125:42: error: missing initializer for member 'nvtxEventAttributes_v2::size' [-Werror=missing-field-initializers]
NvtxSimpleV3.cpp:125:42: error: missing initializer for member 'nvtxEventAttributes_v2::category' [-Werror=missing-field-initializers]
NvtxSimpleV3.cpp:125:42: error: missing initializer for member 'nvtxEventAttributes_v2::colorType' [-Werror=missing-field-initializers]
NvtxSimpleV3.cpp:125:42: error: missing initializer for member 'nvtxEventAttributes_v2::color' [-Werror=missing-field-initializers]
NvtxSimpleV3.cpp:125:42: error: missing initializer for member 'nvtxEventAttributes_v2::payloadType' [-Werror=missing-field-initializers]
NvtxSimpleV3.cpp:125:42: error: missing initializer for member 'nvtxEventAttributes_v2::reserved0' [-Werror=missing-field-initializers]
NvtxSimpleV3.cpp:125:42: error: missing initializer for member 'nvtxEventAttributes_v2::payload' [-Werror=missing-field-initializers]
NvtxSimpleV3.cpp:125:42: error: missing initializer for member 'nvtxEventAttributes_v2::messageType' [-Werror=missing-field-initializers]
NvtxSimpleV3.cpp:125:42: error: missing initializer for member 'nvtxEventAttributes_v2::message' [-Werror=missing-field-initializers]
```

This is why I always take the same route with NVTX and initialize variables with `nvtxEventAttributes_v2` type using a simple `memset` on code bases that support multiple platforms and compiler versions.

Here I am afraid that since NVTX annotations are disabled by default in builds and DALI is supported in multiple of platforms, the issue might get unnoticed for quite some time.

Do you still want me to make this change?  I did verify that the code builds correctly on Linux x86_64 using `= {}`.

klecki: We're on C++14 and I don't think we use anything older than gcc 5 (and in most cases gcc 7+), I think we'll be fine with the more modern approach.

klecki: And by default NVTX markers are enabled starting from CUDA 10.0 and are enabled in all versions that we publish now.

AntoineFroger: Fixed, thanks for the precision.

klecki:
!build

klecki (approved):

awolant (approved):

Issue #2221, opened 2020-08-19
VideoReaderResize silently crashes on corrupted videos


vBLFTePebWNi6c:
I use next command to check if video corrupted:
```
ffmpeg -v error -i video.mp4 -f null - 2> errors.log
```
Log example for corrupted video:
```
[mov,mp4,m4a,3gp,3g2,mj2 @ 0x5627e352ee00] moov atom not found
video.mp4: Invalid data found when processing input
```

The problem is that VideoReaderResize crashes on initialization with this video without any meaningful message.

klecki:
Hi,
would you be able to share such invalid video so we can try to reproduce the error?
Is there any stack-trace/error message produced?

klecki:
Also, what DALI version, CUDA version, driver and GPU are you using?

vBLFTePebWNi6c:
@klecki it seems that I already deleted corrupted videos, but I wrote script that can do these:
```
#!/usr/bin/bash

INPUT_VIDEO="video.mp4"
OUTPUT_VIDEO="corrupted.mp4"

ffmpeg -y -i $INPUT_VIDEO -map v:0 -map -0:a -c:v libx264 -r 25 -crf 18 -pix_fmt yuv420p -g 5 -profile:v high $OUTPUT_VIDEO 2>/dev/null &
last_pid=$!
sleep 1s
kill -KILL $last_pid

ffmpeg -v error -i $OUTPUT_VIDEO -f null -
```
I checked that `VideoReaderResize` fails on `corrupted.mp4` so you can use this script to reproduce the error.
The pipeline was initialized in jupyter notebook and no stack-trace/error message produced -- kernel just silently died.

DALI version: 0.25.0.dev20200723
CUDA version: 10.2
Driver: 440.33.01
GPU: Geforce GTX 1080Ti

klecki:
Thank you for the script, I managed to reproduce the problem.
Unfortunately it is a crash inside ffmpeg, when calling `avformat_find_stream_info` on a corrupted file - it segufaults instead of returning an error.
We will be investigating whether we can somehow prevent this.

As a side note an internal ffmpeg error is printed in terminal where the jupyter is running.

klecki:
I found out that the previous call to `avformat_open_input` reports proper error, but we had comparison swapped (`<=` instead of `>=`). I will post a fix for that, thanks for reporting that and the repro.

JanuszL:
DALI 0.26 is available and should include the needed functionality.

PR #2223, opened 2020-08-19
Fix VideoReader error checking when opening files

klecki:
Error checking comparison was reversed on opening
video files resulting in not reporting an error
and further processing of corrupted file that
ended in segfault.

Now the error is caught and properly reported.
Test was added.

#### Why we need this PR?
It fixes a bug reported in #2221.

 - What solution was applied:
     Enforce check was reversed (errors have negative values).
     VideoReader
     Check if there are no more such mistakes. I went through the .cc file and it looks ok.
 - Validation and testing:
     A corrupted file was added to DALI_extra. Test with that file was added.
     N/A

**JIRA TASK**: *[NA]*


klecki:
!build

awolant (approved):

mzient (approved):

PR #2228, opened 2020-08-21
Unittest Bug Fix - `_test_estimators_single_device`

DEKHTIARJonathan:
Fixes the unittest bug. I left the old code commented out just in case

### DIFF
--- a/dali/test/python/test_dali_tf_dataset_mnist.py
+++ b/dali/test/python/test_dali_tf_dataset_mnist.py
@@ -73,24 +73,25 @@ class MnistPipeline(Pipeline):
 
 
 def _get_mnist_dataset(device='cpu', device_id=0, shard_id=0, num_shards=1, fail_on_device_mismatch=True):
-    mnist_pipeline = MnistPipeline(
-        4, data_path, device, device_id, shard_id, num_shards)
-    shapes = (
-        (BATCH_SIZE, IMAGE_SIZE, IMAGE_SIZE),
-        (BATCH_SIZE,))
-    dtypes = (
-        tf.float32,
-        tf.int32)
+    with tf.device('/{0}:{1}'.format(device, device_id)):
+        mnist_pipeline = MnistPipeline(
+            4, data_path, device, device_id, shard_id, num_shards)
+        shapes = (
+            (BATCH_SIZE, IMAGE_SIZE, IMAGE_SIZE),
+            (BATCH_SIZE,))
+        dtypes = (
+            tf.float32,
+            tf.int32)
 
-    daliset = dali_tf.DALIDataset(
-        pipeline=mnist_pipeline,
-        batch_size=BATCH_SIZE,
-        output_shapes=shapes,
-        output_dtypes=dtypes,
-        num_threads=4,
-        device_id=device_id,
-        fail_on_device_mismatch=fail_on_device_mismatch)
-    return daliset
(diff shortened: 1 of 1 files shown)


dali/test/python/test_dali_tf_dataset_mnist.py
```diff
@@ -224,15 +225,24 @@ def clear_checkpoints():
...
 def _test_estimators_single_device(model, device='cpu', device_id=0, fail_on_device_mismatch=True):
     def train_fn():
-        with tf.device('/{0}:{1}'.format(device, device_id)):
-            return _get_train_dataset(device, device_id, fail_on_device_mismatch=fail_on_device_mismatch).map(
-                lambda features, labels: ({'images': features}, labels))
+        # with tf.device('/{0}:{1}'.format(device, device_id)):
+        #     return _get_train_dataset(device, device_id, fail_on_device_mismatch=fail_on_device_mismatch).map(
+        #         lambda features, labels: ({'images': features}, labels))
+        ds = _get_train_dataset(device, device_id, fail_on_device_mismatch=fail_on_device_mismatch)
+        return ds.apply(tf.data.experimental.copy_to_device(
```
JanuszL, line 232: I don't think it is the way. It kills the idea to process data on the GPU if you will copy it to the CPU after all.

DEKHTIARJonathan: it's not that "it is" or "it is not" the way. What you guys want to do is just not possible.
the `map()` operation does not exist on GPU. I gave to @awolant the one that is compatible for testing. But it's an undocumented API. I never tried it myself.
This unittest is wrong 100%. Your code is not "legal". This PR fixes that. Now I agree it's kinda non-sense. But saying that doesn't help fixing the unittest ...

JanuszL: > Now I agree it's kinda non-sense. 

Then it is better to remove the wrong test that providing a wrong fix to a wrong test.

DEKHTIARJonathan: The fix is not "wrong". It fix the issue. However it might not test what you guys want to test ;) 
However, except if @awolant finds how to use the MapOnGPU OP, this is the "right" to execute `map()` after a DALI dataset ;)

DEKHTIARJonathan: To keep it short there was a number of issues here:
- only train_fn was using a device scope, not test_fn
- the device scope was automatically applied to MAP operation which doesn't exists on GPU, you can't do this
- Since DALI is on GPU device. Between the DALIDataset and the map() OP you need to re-copy the dataset on CPU device

And here is the MAP on GPU OP: https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/data/experimental/ops/prefetching_ops.py#L228

klecki: Alternatively, can we somehow live without using the `map`? Just make the model accept DALI outputs?

DEKHTIARJonathan: You could. But there's still an issue with TF iterator.
`iterator.get_next()` still triggers a memcpy from GPU to CPU and then from CPU to GPU...

JanuszL:
Closing due to inactivity. If this PR is still relevant and you have time to address the remaining updates, please feel free to reopen it.

PR #2240, opened 2020-08-31
Use SSE2 for single-channel audio resampling - 3x speedup.

mzient:

#### Why we need this PR?
*Pick one, remove the rest*
- It speeds up 1-channel audio resampling by a factor of 3 on x86 targets.

 - What solution was applied:
     * Linear samping of window function and its application using SSE2 intrinsics.
     * Audio resampling (decoder)
     * N/A
 - Validation and testing:
     * Existing tests apply
     * N/A

**JIRA TASK**: N/A

### DIFF
--- a/dali/kernels/signal/resampling.h
+++ b/dali/kernels/signal/resampling.h
@@ -1,4 +1,4 @@
-// Copyright (c) 2019, NVIDIA CORPORATION. All rights reserved.
+// Copyright (c) 2019-2020, NVIDIA CORPORATION. All rights reserved.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -15,6 +15,9 @@
 #ifndef DALI_KERNELS_SIGNAL_RESAMPLING_H_
 #define DALI_KERNELS_SIGNAL_RESAMPLING_H_
 
+#ifdef __SSE2__
+#include <emmintrin.h>
+#endif
 #include <cmath>
 #include <functional>
 #include <utility>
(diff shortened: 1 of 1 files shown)


dali/kernels/signal/resampling.h
```diff
@@ -116,9 +135,25 @@ struct Resampler {
...
           i1 = n_in - 1 - in_block_i;
         float f = 0;
-        float x = i0 - in_pos;
-        for (int i = i0; i <= i1; i++, x++) {
-          assert(in_block_ptr + i >= in && in_block_ptr + i < in + n_in);
+        int i = i0;
+
+#ifdef __SSE2__
+        __m128 f4 = _mm_setzero_ps();
+        __m128 x4 = _mm_setr_ps(i - in_pos, i+1 - in_pos, i+2 - in_pos, i+3 - in_pos);
```
JanuszL, line 142: Can we assume that we have always at least 4 elements available?

dali/kernels/signal/resampling.h
```diff
@@ -49,6 +52,22 @@ struct ResamplingWindow {
     return lookup[i] + di * (lookup[i + 1] - lookup[i]);
   }
 
+#ifdef __SSE2__
+  inline __m128 operator()(__m128 x) const {
+    __m128 fi = _mm_add_ps(x * _mm_set1_ps(scale), _mm_set1_ps(center));
```
JanuszL, line 57: Don't wee need to assure that `idx[3]+1 < lookup.size()`?

mzient: It's externally assured, just as it was for scalar variant that `i+1 < lookup.size()`. Or do you mean to add assert?

dali/kernels/signal/resampling.h
mzient: If we don't, the vectorized loop won't get executed at all, so we're safe.

mzient:
!build

dali/kernels/signal/resampling.h
JanuszL: I'm rather asking if we need such assert. If not then I'm fine too.

mzient: The assert was there in the scalar variant for debugging. I don't remember if it ever fired, but I kept it there. I'll probably put one here too, if we decide to extend SIMD support to multi-channel audio (now only 1ch variant got the upgrade). I think that for most scenarios it's OK, but there certainly are some where multi-channel audio is required.

JanuszL (approved):

klecki (approved):

PR #2244, opened 2020-09-01
Add ability to pass DALI TensorList or a list of DALI Tensors to exernal source

mzient:
* Add ability to pass DALI TensorList or a list of DALI Tensors to exernal source
* Keep external scalars at 0D

#### Why we need this PR?
*Pick one, remove the rest*
- It adds new feature: accept DALI Tensor/TenosrList needed for seamless chaining of DALI pipelines
- It fixes a bug: scalars silently extened to 1D 1-element tensors
- Refactoring: make shape validation from external data more robust (accept shape both as a property and as a function)

 - What solution was applied:
     * Add special case for DALI TensorListXPU and TenosrXPU in feed_input
     * Add special case for DALI TensorListXPU in _check_data_batch
     * Add handling for `shape` as a function in _check_data_batch
     * Split _check_data_batch into getting shape and validation of shape and layout
     * Python: pipeline, external source
     * C++: python backend
     * N/A
 - Validation and testing:
     * Python tests
     * Updated documentation for `Pipeline.feed_input` and `source` argument of ExternalSource

**JIRA TASK**: N/A

### DIFF
--- a/dali/python/backend_impl.cc
+++ b/dali/python/backend_impl.cc
@@ -553,13 +553,16 @@ void ExposeTensorList(py::module &m) {
         layout : str
               Layout of the data
         )code")
-    .def(py::init([](TensorList<CPUBackend> *tl, std::string *layout) {
+    .def(py::init([](TensorList<CPUBackend> *tl, py::object layout) {
           if (!tl)
             throw py::value_error("The source object must not be null");
           auto t = std::make_unique<TensorList<CPUBackend>>();
           t->ShareData(tl);
-          if (layout)
-            t->SetLayout(*layout);
+          if (!layout.is_none()) {
+            if (!py::isinstance<py::str>(layout))
+              throw py::type_error("`layout` must be a string or None");
+            t->SetLayout(std::string(layout.cast<py::str>()));
+          }
           return t.release();
         }),
       "tl"_a,
@@ -768,13 +771,16 @@ void ExposeTensorList(py::module &m) {
         layout : str
               Layout of the data
         )code")
-    .def(py::init([](TensorList<GPUBackend> *tl, std::string *layout) {
+    .def(py::init([](TensorList<GPUBackend> *tl, py::object layout) {
           if (!tl)
             throw py::value_error("The source object must not be null");
           auto t = std::make_unique<TensorList<GPUBackend>>();
           t->ShareData(tl);
-          if (layout)
-            t->SetLayout(*layout);
+          if (!layout.is_none()) {
+            if (!py::isinstance<py::str>(layout))
+              throw py::type_error("`layout` must be a string or None");
+            t->SetLayout(std::string(layout.cast<py::str>()));
+          }
           return t.release();
         }),
       "tl"_a,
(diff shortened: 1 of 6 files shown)


dali/python/backend_impl.cc
```diff
@@ -125,8 +125,8 @@ const TensorShape<> &ConvertShape(const TensorShape<> &shape,
 }
 
 template<typename TStrides, typename TShape>
-void CheckContiguousTensor(const TStrides &strides, size_t num_strides,
-                           const TShape &shape, size_t num_extents, size_t element_size) {
+void CheckContiguousTensor(const TStrides &strides, int num_strides,
+                           const TShape &shape, int num_extents, size_t element_size) {
```
mzient, line 129: Avoid wrapping of num_strides  - 1 to huge numbers.

mzient:
!build

dali/python/nvidia/dali/external_source.py
```diff
@@ -2,23 +2,32 @@
 from nvidia.dali import backend as _b
 import inspect
 
+def _get_batch_shape(data):
+    if isinstance(data, (list, tuple, _b.TensorListCPU, _b.TensorListGPU)):
+        if len(data) == 0:
+            return []
```
JanuszL, line 8:
```suggestion
            return [], None
```
?

JanuszL (approved):

mzient:
!build

mzient:
The last commit changes only a docstring - I don't think we need to run full CI and burn who-knows-how-many kWh just for that.

JanuszL (approved):

dali/python/nvidia/dali/pipeline.py
```diff
@@ -555,21 +558,29 @@ def to_numpy(x):
         # __cuda_array_interface__ doesn't provide any way to pass the information about the device
         # where the memory is located. It is assumed that the current device is the one that the memory belongs to,
         # unless the user sets the device explicitly creating TensorGPU/TensorListGPU
-        if isinstance(data, list):
+        if isinstance(data, (Tensors.TensorListCPU, Tensors.TensorListGPU)):
+            if layout is not None:
```
klecki, line 562: All looks correct, but I can't shake the feel that I just observed at least 3 distinct ways for handling the `None` layout. And, doesn't the TL constructor at least also handle the `None` layout only overwriting it if it was actually provided?

I'm not sure if I can suggest any better way of handling the `None` layout vs the default empty one. Maybe it should still be handled by the TL/T constructor.

klecki (approved):
Shouldn't it work before this change as well? It would just go through numpy/cuda_array_interface paths and now we can propagate the layouts, right?

klecki:
> The last commit changes only a docstring - I don't think we need to run full CI and burn who-knows-how-many kWh just for that.

For the peace of mind I would just run it to see the docs rendered.

We may consider docs-only pipelines at some point, I was thinking about that few times, shouldn't be that hard.

mzient:
!build

klecki (approved):

PR #2245, opened 2020-09-01
[Test] Test fix

awolant:

#### Why we need this PR?
*Pick one, remove the rest*

 - What solution was applied:
 - Validation and testing:


awolant:
!build

qa/TL1_ssd_training/test.sh
```diff
@@ -15,7 +15,7 @@ test_body() {
     # newer apex doexn't support Torch 1.1.0 we use for CUDA 9 tests
     git checkout 2ec84ebdca59278eaf15e8ddf32476d9d6d8b904
```
mzient, line 16: Do we need to fix the apex version since we no longer support CUDA 9?

PR #2251, opened 2020-09-04
Skip VP9 tests instead of failing if codec is not supported.

mzient:

#### Why we need this PR?
*Pick one, remove the rest*
- It fixes a bug: VP9 test failing on machines where nvcuvid does not support VP9

 - What solution was applied:
     * catch exception and skip test if the exception message matches one of the patterns for unsupported codec
     * Video reader test
     * N/A
 - Validation and testing:
     * Checked out locally, CI will cover it
     * N/A

**JIRA TASK**: N/A

### DIFF
--- a/dali/operators/reader/video_reader_op_test.cc
+++ b/dali/operators/reader/video_reader_op_test.cc
@@ -14,6 +14,7 @@
 
 #include <gtest/gtest.h>
 #include <cmath>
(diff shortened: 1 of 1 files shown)


mzient:
!build

JanuszL (approved):

jantonguirao (approved):

Issue #2263, opened 2020-09-09
How to specify the start frame without specifying the end in VideoReader?


MrNeither:
Hello. I have a video where the credits start (~10 seconds). I would not want to process them, and start immediately with 10 seconds. You have
https://docs.nvidia.com/deeplearning/dali/user-guide/docs/supported_ops.html?highlight=videoreader#nvidia.dali.ops.VideoReader
there are excellent explanations for how this can be done. But I don't know the length of the video. I can find it out, of course, but suddenly you can do it without calling additional methods.

Is there a way to do this? I tried
```python
file_list= video_directory + f'/{name}' + f' 0 100  -1 \n' 
```
and
```python
file_list= video_directory + f'/{name}' + f' 0 100  \n' 
```
but I get an error.

I would appreciate any information

JanuszL:
Hi,
It is not possible now, but this [PR](https://github.com/NVIDIA/DALI/pull/2264) should add this.

JanuszL:
Please test DALI nightly build after it is merged and see if that helps your use case.

MrNeither:
Thank you very much!! And how can I find out that PR was added to the nightly version?

JanuszL:
I would download the nightly build with the data that follows the date when the commit has been merged. It is still under review but I hope to get it merged soon.

MrNeither:
Hello. I saw that Your changes were added to the night version. I installed it, but for some reason nothing has changed

```python
print(nvidia.dali.__version__)
file_list = list(filter(lambda x: x.endswith(('.mp4', '.avi', '.mkv', '.m4v')), os.listdir(video_folder)))
file_list_txt = ""
for i, name in enumerate(file_list):
    file_list_txt += video_folder + f'/{name}' + f' {i} 200 \n'
tf = tempfile.NamedTemporaryFile()
tf.write(str.encode(file_list_txt))
tf.flush()

vr = VideoPipe(batch_size=32, num_threads=2, device_id=0, data=tf.name)
vr.build()
```

output
```
0.27.0dev.20200909
Traceback (most recent call last):
  File "/home/MrNeither/PycharmProjects/DALItest/Video2Features/main_DALI.py", line 194, in <module>
    vr.save_videos(video_folder=video_path, save_path=save_path)
  File "/home/MrNeither/PycharmProjects/DALItest/Video2Features/main_DALI.py", line 124, in save_videos
    pipe.build()
[...]

PR #2264, opened 2020-09-09
Allow passing a <= 0 values in the file list to allow more flexible frame indexing

JanuszL:
- allows passing <= 0 values in the file list to allow more flexible frame indexing. Positive value means the exact frame, negative is the number of frames from the end; equal start and end frame would yield an empty sequence and a warning

#### Why we need this PR?
*Pick one, remove the rest*
- It allow passing a default end frame in the video file list

 - What solution was applied:
     allows passing <= 0 values in the file list to allow more flexible frame indexing. Positive value means the exact frame, negative is the number of frames from the end; equal start and end frame would yield an empty sequence and a warning
     video reader
     NA
 - Validation and testing:
     video test is extended
     NA

In response https://github.com/NVIDIA/DALI/issues/2263

**JIRA TASK**: *[NA]*

### DIFF
--- a/dali/operators/reader/loader/video_loader.h
+++ b/dali/operators/reader/loader/video_loader.h
@@ -231,8 +231,6 @@ class VideoLoader : public Loader<GPUBackend, SequenceWrapper> {
           }
           if (end > 0) {
             end_frame = end;
-          } else if (end == 0) {
-            end_frame = file.frame_count_;
           } else {
             end_frame = file.frame_count_ + end;
           }
--- a/dali/operators/reader/loader/video_loader.h
+++ b/dali/operators/reader/loader/video_loader.h
@@ -168,6 +168,7 @@ class VideoLoader : public Loader<GPUBackend, SequenceWrapper> {
       stats_({0, 0, 0, 0, 0}),
       current_frame_idx_(-1),
       stop_(false) {
+    DALI_ENFORCE(stride_ > 0, "Stride should be > 0");
     if (step_ < 0)
       step_ = count_ * stride_;
 
@@ -252,8 +253,6 @@ class VideoLoader : public Loader<GPUBackend, SequenceWrapper> {
           }
           if (end > 0) {
             end_frame = static_cast<int>(std::floor(end * av_q2d(frame_rate)));
-          } else if (end == 0) {
-            end_frame = file.frame_count_;
           } else {
             end_frame = file.frame_count_ + static_cast<int>(std::floor(end * av_q2d(frame_rate)));
           }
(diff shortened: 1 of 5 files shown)


JanuszL:
!build

awolant:
Maybe we could reserve negative values for the future, if we ever wanted or needed to implement "skip n last frames"? (like indexing in numpy for example)

JanuszL:
> Maybe we could reserve negative values for the future, if we ever wanted or needed to implement "skip n last frames"? (like indexing in numpy for example)
> If not, than this is fine by me.

Good point. So 0 (I don't know how intuitive is that).

awolant:
For me, no value means the end, positive value means this many frames from beginning and negative value means this many frames from the end, but maybe some other opinion on this would be useful.

JanuszL:
!build

JanuszL:
> For me, no value means the end, positive value means this many frames from beginning and negative value means this many frames from the end, but maybe some other opinion on this would be useful.

How about now?

awolant:
If I'm not mistaken, currently passing 0 0 will give full video, while in Python indexing it gives empty list. I'm not sure, if we want to be compliant with this convention since empty video does not make much sense. 
Other than that looks good

awolant (approved):

mzient:
I'd prefer if 0,0 meant either empty sequence (skip the entry, perhaps with a warning) or just the first frame (if the end time is inclusive).

JanuszL:
> I'd prefer if 0,0 meant either empty sequence (skip the entry, perhaps with a warning) or just the first frame (if the end time is inclusive).

I would even go for N,N means empty sequence with the warning.

JanuszL:
!build

dali/operators/reader/loader/video_loader.h
```diff
@@ -221,17 +221,47 @@ class VideoLoader : public Loader<GPUBackend, SequenceWrapper> {
...
+            start_frame = start;
+          } else {
+            start_frame = file.frame_count_ + start;
+          }
+          if (end > 0) {
+            end_frame = end;
+          } else if (end == 0) {
+            end_frame = file.frame_count_;
+          } else {
+            end_frame = file.frame_count_ + end;
```
mzient, line 236: Nitpick
```suggestion
          } else {
            end_frame = file.frame_count_ + end;
```

mzient (approved):

dali/operators/reader/loader/video_loader.h
JanuszL: Fixed

JanuszL:
!build

awolant (approved):

JanuszL:
Build

PR #2304, opened 2020-09-28
Fix detection pipeline test on Ampere

awolant:

#### Why we need this PR?
*Pick one, remove the rest*
- It fixes a bug where detection pipeline test fails on Ampere.

 - What solution was applied:
     *`hw_decoder_load` set to 0 to avoid offloading part of the batch to hardware decoder, removed randomness from the test, fix deprecation warnings*

**JIRA TASK**: *[Use DALI-1620]*

### DIFF
--- a/dali/test/python/test_detection_pipeline.py
+++ b/dali/test/python/test_detection_pipeline.py
@@ -15,7 +15,6 @@
 import argparse
 import itertools
 import os
-import random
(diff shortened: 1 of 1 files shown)


awolant:
!build

JanuszL (approved):

banasraf (approved):

PR #2318, opened 2020-10-01
Use true scalars + bug fixes

mzient:
Fix usage of scalars - convert from {1} tensors to {}.
Fix a bug in executor - attempting to execute subsequent ops in the same stage after a failure.
Fix a bug in CocoReader - remove erroneuous assertion.

Classification readers are left out - changing the labels from 1-element tensors to scalars has merit, but I think we might have gone a bit too far there to change it now.

#### Why we need this PR?
*Pick one, remove the rest*
- It fixes bugs:
    * in executor - attempting to execute subsequent ops in the same stage after a failure.
    * in CocoReader - remove erroneuous assertion.
- Refactoring to improve handling of scalar values
    * allow 0D data as scalar argument inputs
    * random generators produce true scalars
    * arithmetic operators accept and produce true scalars
    * audio decoder produces sampling rate as a true scalar

 - What solution was applied:
     * Changed some tensors to 0D
     * Adjust tests
     * Changed the condition in OpSpec to accept 0D
     * OpSpec
     * ArithmOps
     * AudioDecoder
     * Random number generating ops (all of them)
     * N/A
 - Validation and testing:
     * Existing tests apply
     * Updated, where applicable

**JIRA TASK**: N/A


mzient:
!build

JanuszL (approved):

mzient:
!build

JanuszL:
Please fix LGTM allert.

dali/python/nvidia/dali/types.py
```diff
@@ -349,6 +351,7 @@ def ConstantNode(device, value, dtype, shape, layout, **kwargs):
         def _type_from_value_or_list(v):
             if not isinstance(v, (list, tuple)):
                 v = [v]
+                shape = []
```
JanuszL, line 354: Nor used?

JanuszL (approved):

mzient:
!build

mzient:
!build

dali/python/nvidia/dali/types.py
mzient: Used in line 403. This was actually a necessary fix.

JanuszL: Correct me if I'm wrong, but this is the scope of `_type_from_value_or_list` function and you are not changing the value of the `shape` variable from the outer scope.

mzient:
!build

jantonguirao (approved):

mzient:
!build

mzient:
!build

mzient:
!build

mzient:
!build

mzient:
!build

PR #2319, opened 2020-10-02
Fix build for LMDB disabled

JanuszL:
- adds a proper #if guard in loader_test for LMDB disabled
- adds additional check for write result in nemo_asr_loader_test

#### Why we need this PR?
*Pick one, remove the rest*
- It fixes build for LMDB disabled

 - What solution was applied:
     adds a proper #if guard in loader_test for LMDB disabled
     adds additional check for write result in nemo_asr_loader_test
     loader_test
     nemo_asr_loader_test
     NA
 - Validation and testing:
     CI
     NA

**JIRA TASK**: *[NA]*

### DIFF
--- a/dali/operators/reader/loader/nemo_asr_loader_test.cc
+++ b/dali/operators/reader/loader/nemo_asr_loader_test.cc
@@ -106,8 +106,9 @@ TEST(NemoAsrLoaderTest, WrongManifestPath) {
 void tempfile(std::string& filename, std::string content = "") {
   int fd = mkstemp(&filename[0]);
   ASSERT_NE(-1, fd);
-  if (!content.empty())
-    write(fd, content.c_str(), content.size());
+  if (!content.empty()) {
+    ASSERT_EQ(write(fd, content.c_str(), content.size()), content.size());
+  }
   close(fd);
 }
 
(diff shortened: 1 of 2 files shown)


jantonguirao (approved):

dali/operators/reader/loader/nemo_asr_loader_test.cc
```diff
@@ -106,8 +106,9 @@ TEST(NemoAsrLoaderTest, WrongManifestPath) {
 void tempfile(std::string& filename, std::string content = "") {
   int fd = mkstemp(&filename[0]);
   ASSERT_NE(-1, fd);
-  if (!content.empty())
-    write(fd, content.c_str(), content.size());
+  if (!content.empty()) {
+    ASSERT_EQ(write(fd, content.c_str(), content.size()), content.size());
```
mzient, line 110: Won't it raise a warning for comparing signed with unsigned (content.size())?

JanuszL: I don't see any.

mzient (approved):

dali/operators/reader/loader/nemo_asr_loader_test.cc
JanuszL: What I see it uses the function that might cast booth sizes to just signed int.

PR #2336, opened 2020-10-08
Fix error when VideoReader is prematurely terminated

JanuszL:
- simplifies the SequenceWrapper by removing unnecessary locking
- removes the false assumption that SequenceWrapper is thread-safe
- make sure that no prefetching is happening when VideoReader ends, and no memory access happens to the underlying VideoReader memory

#### Why we need this PR?
*Pick one, remove the rest*
- It fixes problem in hand in VideoReader destructor when there is an error

 - What solution was applied:
     simplifies the SequenceWrapper by removing unnecessary locking
     removes the false assumption that SequenceWrapper is thread-safe
     make sure that no prefetching is happening when VideoReader ends, and no memory access happens to the underlying VideoReader memory
     VideoReader
     SequenceWrapper
     SequenceWrapper changes
 - Validation and testing:
     CI
     NA

**JIRA TASK**: *[NA]*

### DIFF
--- a/dali/operators/reader/nvdecoder/sequencewrapper.h
+++ b/dali/operators/reader/nvdecoder/sequencewrapper.h
@@ -48,9 +48,7 @@ struct SequenceWrapper {
     timestamps.reserve(count);
 
     if (!event_) {
-      int dev;
-      CUDA_CALL(cudaGetDevice(&dev));
-      event_ = CUDAEvent::CreateWithFlags(cudaEventBlockingSync | cudaEventDisableTiming, dev);
+      event_ = CUDAEvent::CreateWithFlags(cudaEventBlockingSync | cudaEventDisableTiming);
     }
   }
 
@@ -83,11 +81,11 @@ struct SequenceWrapper {
   int label = -1;
   vector<double> timestamps;
   int first_frame_idx = -1;
-  DALIDataType dtype = {};
+  DALIDataType dtype = DALI_NO_TYPE;
   std::function<void(void)> read_sample_f;
 
  private:
-  CUDAEvent event_ = {};
+  CUDAEvent event_;
 };
 
 }  // namespace dali
--- a/dali/operators/reader/nvdecoder/sequencewrapper.h
+++ b/dali/operators/reader/nvdecoder/sequencewrapper.h
@@ -21,6 +21,7 @@
 #include <vector>
 
 #include "dali/core/common.h"
+#include "dali/core/cuda_event.h"
 #include "dali/core/static_switch.h"
 #include "dali/core/error_handling.h"
 #include "dali/pipeline/operator/argument.h"
@@ -34,8 +35,7 @@ namespace dali {
 // Struct that Loader::ReadOne will read
 struct SequenceWrapper {
  public:
-  SequenceWrapper()
-  : started_(false) {}
+  SequenceWrapper() = default;
 
   void initialize(int count, int height, int width, int channels, DALIDataType dtype) {
     this->count = count;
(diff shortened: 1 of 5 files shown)


JanuszL:
!build

dali/operators/reader/nvdecoder/sequencewrapper.h
```diff
@@ -47,44 +47,28 @@ struct SequenceWrapper {
     timestamps.clear();
     timestamps.reserve(count);
 
-    int dev;
-    CUDA_CALL(cudaGetDevice(&dev));
-    std::unique_lock<std::mutex> lock{started_lock_};
-    if (started_) {
-      CUDA_CALL(cudaEventDestroy(event_));
+    if (!initialized_) {
+      CUDA_CALL(cudaEventCreateWithFlags(&event_, cudaEventBlockingSync | cudaEventDisableTiming));
```
mzient, line 51: Why not use CUDAEvent instead? It has a proper destructor - no need to write your own.

dali/operators/reader/nvdecoder/sequencewrapper.h
```diff
@@ -47,44 +47,28 @@ struct SequenceWrapper {
...
-    started_cv_.notify_one();
   }
 
   void wait() const {
-    LOG_LINE << event_ << " wait to start" << std::endl;
-    wait_until_started_();
-    LOG_LINE << event_ << " waiting for sequence event" << std::endl;
-    CUDA_CALL(cudaEventSynchronize(event_));
-    LOG_LINE << event_ << " synchronized!" << std::endl;
+    if (initialized_) {
```
mzient, line 67:
```suggestion
    if (event_) {
```
However, is it no longer possible for `wait()` to be called before `set_started`? If not, this function will return immediately as if the event was signalled.

JanuszL: I checked the flow and we don't need to call `wait` before `set_started`. Maybe that was needed for nvvl, where different thread scheduled the work and different waited for it.

jantonguirao (approved):
LGTM, but I agree with Michal that it'd be better to use CUDAEvent which has a destructor

dali/operators/reader/nvdecoder/sequencewrapper.h
JanuszL: Legacy..., I have also forgot about our wrappers, fixed

JanuszL:
!build

dali/operators/reader/nvdecoder/sequencewrapper.h
```diff
@@ -97,26 +76,18 @@ struct SequenceWrapper {
...
-  void wait_until_started_() const {
-    std::unique_lock<std::mutex> lock{started_lock_};
-    started_cv_.wait(lock, [&](){return started_;});
-  }
-
-  mutable std::mutex started_lock_;
-  mutable std::condition_variable started_cv_;
-  cudaEvent_t event_;
-  bool started_;
+  CUDAEvent event_ = {};
```
mzient, line 90:
```suggestion
  CUDAEvent event_;
```
It has a default constructor.

dali/operators/reader/nvdecoder/sequencewrapper.h
```diff
@@ -97,26 +76,18 @@ struct SequenceWrapper {
...
+  int count = -1;
+  int height = -1;
+  int width = -1;
+  int channels = -1;
+  int label = -1;
   vector<double> timestamps;
-  int first_frame_idx;
-  DALIDataType dtype;
+  int first_frame_idx = -1;
+  DALIDataType dtype = {};
```
mzient, line 86:
```suggestion
  DALIDataType dtype = DALI_NO_TYPE;
```
The default is coincidentally DALI_UINT8 - do we want that?

dali/operators/reader/nvdecoder/sequencewrapper.h
```diff
@@ -47,44 +47,23 @@ struct SequenceWrapper {
...
-      try {
-        CUDA_CALL(cudaEventDestroy(event_));
-      } catch (const std::exception &) {
-        // Something went wrong with releasing resources. We'd better die now.
-        std::terminate();
-      }
+    if (!event_) {
+      int dev;
+      CUDA_CALL(cudaGetDevice(&dev));
+      event_ = CUDAEvent::CreateWithFlags(cudaEventBlockingSync | cudaEventDisableTiming, dev);
```
mzient, line 53:
```suggestion
      event_ = CUDAEvent::CreateWithFlags(cudaEventBlockingSync | cudaEventDisableTiming);
```
This will automatically create an event for current device.

dali/operators/reader/nvdecoder/sequencewrapper.h
JanuszL: I don't think we care that much, but changed.

dali/operators/reader/nvdecoder/sequencewrapper.h
JanuszL: Done

dali/operators/reader/nvdecoder/sequencewrapper.h
JanuszL: Done

JanuszL:
!build

mzient (approved):

PR #2345, opened 2020-10-09
Add support for N-dim tensors to OneHot

klecki:
Handle true scalar inputs
Extend tests a bit
TODO: Layout support

#### Why we need this PR?
Improve One Hot CPU implementation

 - What solution was applied:
     Added axis, loop over outer and inner flattened coordinate and fill the axis.
     OneHot CPU
  N/A
 - Validation and testing:
  nosetest   
     Yes

**JIRA TASK**: *[DALI-1658]*

### DIFF
--- a/dali/operators/generic/one_hot.cc
+++ b/dali/operators/generic/one_hot.cc
@@ -29,8 +29,8 @@ will fail.)code")
   .NumInput(1)
   .NumOutput(1)
   .AddOptionalArg("num_classes", R"code(Number of all classes in the data.)code", 0)
-  .AddOptionalArg<int>("axis", R"code(On which axis place the on-hot encoding axis of `num_classes`
-size. By default it's appended as the last dimension for non-scalar inputs, for scalar inputs,
+  .AddOptionalArg<int>("axis", R"code(Dimension to place the one-hot encoding axis of `num_classes`
+size. By default it's appended as the last dimension for non-scalar inputs. For scalar inputs,
 it becomes the only dimension.)code", -1)
   .AddOptionalArg(arg_names::kDtype, R"code(Output data type.)code", DALI_FLOAT)
   .AddOptionalArg("on_value",
@@ -59,8 +59,7 @@ TensorShape<> determine_shape(TensorShape<> in_shape, int num_classes, int axis,
   int inner_axes = in_shape.sample_dim() - axis;
   auto outer =  in_shape.first(outer_axes);
   auto inner = in_shape.last(inner_axes);
-  auto tmp = shape_cat(outer, num_classes);
-  return shape_cat(tmp, inner);
+  return shape_cat(shape_cat(outer, num_classes), inner);
 }
 
 }  // namespace
--- a/dali/operators/generic/one_hot.cc
+++ b/dali/operators/generic/one_hot.cc
@@ -26,24 +26,21 @@ DALI_SCHEMA(OneHot)
 
 Adds a new axis or converts scalar input into an axis of ``num_classes`` elements.
 
-For given input coordinate ``(x0, x1, ..., xn)``, and ``axis=k``, the output sample is specified as::
+For given input coordinate ``(x0, x1, ..., xn)``, and ``axis = k``, the output sample is specified as::
 
   cls = input[x0, x1, ..., xn]
-  output[x0, x1, ..., xk-1, :, xk+1, ..., xn] = off_value
-  output[x0, x1, ..., xk-1, cls, xk+1, ..., xn] = on_value
+  output[x0, x1, ..., xk-1, xk, xk+1, ..., xn] = on_value if xk == cls else off_value
 
-for all ``cls`` values in range ``[0, num_classes)``. The ``k-th`` axis will have ``num_classes`` elements.
+for all ``xk`` in range ``[0, num_classes)``.
 
-For scalars, for given input sample::
+For scalars, the output is set to ``on_value`` at the index taken from ``input`` and
+``off_value`` elsewhere::
 
-  output[:] = off_value
-  output[input] = on_value
+  output[i] = on_value if i == input else off_value
 
-The output will have 1 dimension of ``num_classes`` elements.
-
-For legacy support any input in which all tensors have volume equal to 1 is considered
-a scalar input (each of them have only one element). Legacy interpretation of tensors as scalars
-is not supported if ``axis`` argument is specified.)code")
+For backward compatibility, any input in which all tensors have only one element
+(regardless of the number of dimensions) is considered scalar. Legacy interpretation of tensors
+as scalars is not supported if ``axis`` argument is specified.)code")
   .NumInput(1)
   .NumOutput(1)
   .AddOptionalArg("num_classes", R"code(Number of all classes in the data.)code", 0)
--- a/dali/operators/generic/one_hot.cc
+++ b/dali/operators/generic/one_hot.cc
@@ -29,9 +29,9 @@ Adds a new axis or converts scalar input into an axis of ``num_classes`` element
 For given input coordinate ``(x0, x1, ..., xn)``, and ``axis = k``, the output sample is specified as::
 
   cls = input[x0, x1, ..., xn]
-  output[x0, x1, ..., xk-1, xk, xk+1, ..., xn] = on_value if xk == cls else off_value
+  output[x0, x1, ..., xk-1, i, xk, ..., xn] = on_value if i == cls else off_value
 
-for all ``xk`` in range ``[0, num_classes)``.
+for all ``i`` in range ``[0, num_classes)``.
 
 For scalars, the output is set to ``on_value`` at the index taken from ``input`` and
 ``off_value`` elsewhere::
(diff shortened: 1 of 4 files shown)


klecki:
!build

dali/operators/generic/one_hot.cc
```diff
@@ -29,6 +29,9 @@ will fail.)code")
   .NumInput(1)
   .NumOutput(1)
   .AddOptionalArg("num_classes", R"code(Number of all classes in the data.)code", 0)
+  .AddOptionalArg<int>("axis", R"code(On which axis place the on-hot encoding axis of `num_classes`
+size. By default it's appended as the last dimension for non-scalar inputs, for scalar inputs,
+it becomes the only dimension.)code", -1)
```
banasraf, line 34:
```suggestion
  .AddOptionalArg<int>("axis", R"code(Dimension to place the one-hot encoding axis of `num_classes`
size. By default it's appended as the last dimension for non-scalar inputs. For scalar inputs,
it becomes the only dimension.)code", -1)
```
?

dali/operators/generic/one_hot.cc
```diff
@@ -46,22 +49,45 @@ bool is_scalar(TensorShape<ndims> shape) {
...
+    return {num_classes};
   }
-  return shape_cat(in_shape, nclasses);
+  axis = axis < 0 ? in_shape.sample_dim() : axis;
+  int outer_axes = axis;
+  int inner_axes = in_shape.sample_dim() - axis;
+  auto outer =  in_shape.first(outer_axes);
+  auto inner = in_shape.last(inner_axes);
+  auto tmp = shape_cat(outer, num_classes);
+  return shape_cat(tmp, inner);
```
banasraf, line 63:
```suggestion
  return shape_cat(shape_cat(outer, num_classes), inner);
```
The sequence `outer, num_classes, inner` is visible and it gets rid of `tmp` which doesn't say anything.

banasraf (approved):

dali/operators/generic/one_hot.cc
klecki: Done

dali/operators/generic/one_hot.cc
klecki: Good idea, done.

dali/operators/generic/one_hot.cc
```diff
@@ -46,22 +49,44 @@ bool is_scalar(TensorShape<ndims> shape) {
...
+  DALI_ENFORCE(-1 <= axis_ && axis_ <= input_sample_dim,
+               make_string("Provided axis is outside of allowed range, got: ", axis_,
+                           ", expected to be in range: [-1, ", input_sample_dim, "]."));
+
+  // Legacy scalar-like support
+  // TODO(klecki): It was already buggy, maybe drop it
+  bool all_scalars = true;
+  for (int i = 0; i < num_samples; i++) {
+    all_scalars = all_scalars && is_scalar(input.shape()[i]);
+  }
```
mzient, line 80:
```suggestion
  bool all_scalars = !spec_.ArgumentDefined("axis");
  for (int i = 0; all_scalars && i < num_samples; i++) {
    all_scalars = all_scalars && is_scalar(input.shape()[i]);
  }
```

mzient: This will keep the legacy behavior when there's no axis, but will be much stricter when the axis is provided.
Also, no need to look at subsequent samples once we've found one non-scalar.

dali/operators/generic/one_hot.h
```diff
@@ -31,17 +31,24 @@ namespace detail {
...
-    int cls = in[i];
-    if (cls < 0 || cls >= num_classes)
-      continue;
-    int64_t out_idx = (i % inner) + cls * inner + (i / inner) * (inner * num_classes);
-    out[out_idx] = on_value;
+  axis = axis < 0 ? output.shape.sample_dim() - 1 : axis;
+  auto outer = output.shape.first(axis);
+  auto inner = output.shape.last(output.shape.sample_dim() - axis - 1);
+  auto volume_outer = volume(outer);
+  auto volume_inner = volume(inner);
```
mzient, line 41: We don't need to create the intermediate shapes if their only purpose is to calculate the volume.
```suggestion
  auto volume_outer = volume(outer.begin(), outer.begin() + axis);
  auto volume_inner = volume(outer.begin() + axis + 1, outer.ebd());
```
But won't push it - if you find current solution to be more readable, then keep it.

dali/operators/generic/one_hot.h
```diff
@@ -31,17 +31,24 @@ namespace detail {
...
   auto out = output.data;
-  for (int i = 0; i < volume(output.shape); ++i) out[i] = off_value;
-  int64_t inner = volume(input.shape.last(ndims - axis));
-  for (int64_t i = 0, n = input.num_elements(); i < n; i++) {
-    int cls = in[i];
-    if (cls < 0 || cls >= num_classes)
-      continue;
-    int64_t out_idx = (i % inner) + cls * inner + (i / inner) * (inner * num_classes);
-    out[out_idx] = on_value;
+  axis = axis < 0 ? output.shape.sample_dim() - 1 : axis;
```
mzient, line 37: Let the called do it (outside the sample loop)?

dali/operators/generic/one_hot.cc
klecki: Done. Also adjusted the comment.

dali/operators/generic/one_hot.h
klecki: Done, forgot about that overload of volume tbh.

dali/operators/generic/one_hot.h
klecki: done

mzient (approved):

klecki:
!build

klecki:
!build

klecki:
!build

dali/operators/generic/one_hot.cc
```diff
@@ -24,18 +24,38 @@ DALI_REGISTER_OPERATOR(OneHot, OneHot, CPU);
...
   .DocStr(R"code(Produces a one-hot encoding of the input.
 
-If the input is not a scalar (tensor consisting from one value per sample), the operator
-will fail.)code")
+Adds a new axis or converts scalar input into an axis of ``num_classes`` elements.
+
+For given input coordinate ``(x0, x1, ..., xn)``, and ``axis = k``, the output sample is specified as::
+
+  cls = input[x0, x1, ..., xn]
+  output[x0, x1, ..., xk-1, i, xk, ..., xn] = on_value if i == cls else off_value
```
JanuszL, line 32:
```suggestion
For given input coordinate ``input[0] = (x0, x1, ..., xn)``, and ``axis = k``, the output sample is specified as::

  output[0, x0, x1, ..., xk-1, i, xk, ..., xn] = on_value and off_value for all remaining output[0, ...]
```

dali/operators/generic/one_hot.h
```diff
@@ -31,17 +31,21 @@ namespace detail {
...
-    if (cls < 0 || cls >= num_classes)
-      continue;
-    int64_t out_idx = (i % inner) + cls * inner + (i / inner) * (inner * num_classes);
-    out[out_idx] = on_value;
+  auto volume_outer = volume(output.shape.begin(), output.shape.begin() + axis);
+  auto volume_inner = volume(output.shape.begin() + axis + 1, output.shape.end());
+  for (int i = 0; i < volume(output.shape); ++i) {
+    out[i] = off_value;
+  }
+  for (int64_t outer_coord = 0; outer_coord < volume_outer; outer_coord++) {
```
JanuszL, line 42: As I understand it is not one hot anymore, as per each sample you may have more than one element in the whole tensor set. Does it still make sense?

dali/test/python/test_operator_one_hot.py
```diff
@@ -15,39 +15,88 @@
...
+    results = []
+    for i in range(batch_size):
+        shape = insert_as_axis(input[i].shape, num_classes, axis, total_axes)
+        result = np.zeros(shape, dtype=np.int32)
+        shapes.append(shape)
+        for i0 in range(input[i].shape[0]):
+            for i1 in range(input[i].shape[1]):
+                for i2 in range(input[i].shape[2]):
+                    in_coord = (i0, i1, i2)
+                    out_coord = insert_as_axis(in_coord, input[i][in_coord], axis, total_axes)
```
JanuszL, line 56: I don't think this is how TF oneHot works. Now it is multiple hot. Can you provide rationale/reference for such behavior?

klecki:
```
If indices is a matrix (batch) with shape [batch, features], the output shape will be:

  batch x features x depth if axis == -1
  batch x depth x features if axis == 1
  depth x batch x features if axis == 0
```
from [tf.one_hot doc](https://www.tensorflow.org/api_docs/python/tf/one_hot)

dali/operators/generic/one_hot.h
klecki: It's as the doc says. We produce output with additional axis. For any (input) coordinate that has there a value `cls` we put the on_value in the extended coordinate.
If the coordinate is `x0, x1, ..., xn`, for `axis=-1` we would have on_value in `x0, x1, ..., xn, cls`; for axis=1, it's `x0, cls, x1, ... xn` etc output[x0, j, x1, ... xn]` for j != cls will have off_value.

dali/operators/generic/one_hot.cc
klecki: I explicitly mention sample here, the 0 doesn't make sense and the input is not equal to coordinate.

JanuszL (approved):

PR #2350, opened 2020-10-12
Tensor concatenation and stacking

mzient:
#### Why we need this PR?
*Pick one, remove the rest*
- It adds new feature: tensor concatenation and stacking operators

 - What solution was applied:
     * Use TensorJoinCPU/TenosrJoinGPU kernels in the operator
     * Use a simple copy/reshaed copy when there's just 1 input
     * Use `any` to store type-specific data in the operator and avoid writing full-blown pImpl
     * Use function overloads to provide backend-specific RunImpl and have just one class
     * Register the operator with different template arguments (backend and new_axis) as Cat/Stack CPU/GPU
     * TensorJoin operator
     * Constant operator (minor fix)
     * Axis handling
     * Use of `any` for inputs
 - Validation and testing:
     * Python tests
     * Docstrings
     * Jupyter

**JIRA TASK**: DALI-1620

### DIFF
--- a/dali/operators/generic/join.cc
+++ b/dali/operators/generic/join.cc
@@ -45,12 +45,12 @@ The shapes of respective tensors in the inputs must match.)")
   .AddOptionalArg<int>("axis", R"(The axis in the output tensor along which the inputs are stacked.
 
 The axis is inserted before a corresponding axis in the inputs. A value of 0 indicates that whole
-tensors are stacked. Speicfying ``axis`` equal to the number of dimensions in the inputs causes
+tensors are stacked. Specifying ``axis`` equal to the number of dimensions in the inputs causes
 the values from the inputs to be interleaved)", 0, false)
   .AddOptionalArg<string>("axis_name", R"(Name of the new axis to be inserted.
 
-A one-character that will denot the new axis in the output layout. The output layout will be
-constructed by inserting that character into the input layout at position indicated by ``axis``.
+A one-character string that will denote the new axis in the output layout. The output layout will be
+constructed by inserting that character into the input layout at the position indicated by ``axis``.
 For example, specifying ``axis = 0`` and ``axis_name = "C"`` with input layout "HW" will yield
 the output layout "CHW")", nullptr, false)
   .NumInput(1, 999)
@@ -80,7 +80,7 @@ bool TensorJoin<Backend, new_axis>::SetupImpl(
   SetupAxis();
   SetOutputLayout(ws);
 
-  // Run ove inputs and store them in a vector
+  // Run over the inputs and store them in a vector
   TYPE_SWITCH(out_type, type2id, T, TENSOR_JOIN_TYPES, (
     SetupTyped<T>(output_shape, ws);
   ), (DALI_FAIL(make_string("The element type ", out_type, " is not supported."))));  // NOLINT
@@ -160,7 +160,7 @@ void TensorJoin<Backend, new_axis>::SetOutputLayout(const workspace_t<Backend> &
 template <typename Backend, bool new_axis>
 void TensorJoin<Backend, new_axis>::SetupAxis() {
   // axis_name indicates the join axis for concatenation only;
-  // for stacking, it's the name of the new axiis
+  // for stacking, it's the name of the new axis
   if (has_axis_name_ && !new_axis) {
     axis_ = input_layout_.find(axis_name_arg_);
     DALI_ENFORCE(axis_ >= 0, make_string("``axis_name`` specifies an undefined axis '",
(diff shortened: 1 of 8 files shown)


mzient:
!build

dali/operators/generic/join.cc
```diff
@@ -0,0 +1,254 @@
...
+    }
+  } else {
+    output_layout_ = input_layout_;
+  }
+}
+
+template <typename Backend, bool new_axis>
+void TensorJoin<Backend, new_axis>::SetupAxis() {
+  // axis_name indicates the join axis for concatenation only;
+  // for stacking, it's the name of the new axiis
```
jantonguirao, line 163:
```suggestion
  // for stacking, it's the name of the new axis
```

jantonguirao (approved):

dali/operators/generic/join.cc
```diff
@@ -0,0 +1,254 @@
...
+  .DocStr(R"(Joins the input tensors along an existing axis.
+
+The shapes of the inputs must match in all dimensions except the concatenation axis.)")
+  .AddOptionalArg<int>("axis", "Axis along which the input tensors are concatenated.", 0, false)
+  .AddOptionalArg<string>("axis_name", R"(Name of the axis along which the tensors are concatenated.
+
+This argument is mutually exclusive with ``axis``.
+This argument requires that at least one input has a non-empty layout and that all non-empty
+input layouts match.)", nullptr, false)
+  .NumInput(1, 999)
```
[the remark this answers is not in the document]
JanuszL: Maybe we should have a define for an infinite number of inputs (999 + 1)?

dali/operators/generic/join.cc
```diff
@@ -0,0 +1,254 @@
...
+    DALI_ENFORCE(type_id == out_type, make_string(
+        "All inputs must have the same type.\nType of input #0: ", out_type,
+        "\nType of input #", i, ": ", type_id));
+  }
+
+  GetInputLayout(ws);
+  SetupAxis();
+  SetOutputLayout(ws);
+
+  // Run ove inputs and store them in a vector
```
JanuszL, line 83:
```suggestion
  // Run over inputs and store them in a vector
```

dali/operators/generic/join.cc
```diff
@@ -0,0 +1,254 @@
...
+
+template <typename Backend, bool new_axis>
+template <typename T>
+void TensorJoin<Backend, new_axis>::SetupTyped(
+      TensorListShape<> &output_shape, const workspace_t<Backend> &ws) {
+  auto &inputs = this->template inputs<T>();
+  inputs.clear();
+  int ninp = this->spec_.NumRegularInput();
+
+  copy_idx_ = 0;
```
JanuszL, line 98:
```suggestion
  copy_only_ = false;
```

dali/operators/generic/join.cc
```diff
@@ -0,0 +1,254 @@
...
+    if (new_axis)
+      out.Resize(shape);
+    out.SetLayout(output_layout_);
+    return;
+  }
+
+  out.SetLayout(output_layout_);
+  TYPE_SWITCH(out.type().id(), type2id, T, TENSOR_JOIN_TYPES, (
+    RunTyped(view<T>(out), ws);
+  ), (DALI_FAIL("Internal error: unsupported type reached RunImpl function")));  // NOLINT
```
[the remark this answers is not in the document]
mzient: If it was ever reached, then printing the type would shift attention from the real problem: that we've encountered a type that should have been rejected in Setup.

docs/examples/general/tensor_join.ipynb
```diff
@@ -0,0 +1,404 @@
+{
```
JanuszL, line 1: <p>metbhods -&gt; methods</p><p>combine separate combine separate coordinates into vectors -&gt; combine separate coordinates into vectors</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/NVIDIA/DALI/pull/2350/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='0'/>

dali/operators/generic/join.cc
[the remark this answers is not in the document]
mzient: That's a good idea, but maybe not for this PR...?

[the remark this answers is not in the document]
JanuszL: Sure

dali/operators/generic/join.cc
mzient: ??? That's a _number_!

dali/operators/generic/join.cc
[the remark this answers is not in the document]
JanuszL: I don't think the user cares how fatal the error is, and the more debug info we get from the user the better.

dali/operators/generic/join.cc
JanuszL: And I don't think it should be. As I understand it is to tell if it should just copy or not. Making it an int blurs the purpose.

mzient: But it also tell _which_ one should be copied.

dali/operators/generic/join.cc
```diff
@@ -0,0 +1,254 @@
...
+  auto &inputs = this->template inputs<T>();
+  inputs.clear();
+  int ninp = this->spec_.NumRegularInput();
+
+  copy_idx_ = 0;
+  for (int i = 0; i < ninp; i++) {
+    auto tlv = view<T>(ws.template InputRef<Backend>(i));
+    if (new_axis || tlv.num_elements() > 0) {  // when concatenating, we can skip empty inputs
+      if (inputs.empty())
+        copy_idx_ = i;
```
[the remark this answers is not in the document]
JanuszL: Ok

JanuszL (approved):

jantonguirao (approved):

PR #2377, opened 2020-10-19
Fix a problem with reference to "incomplete" type (error in Clang/CUDA).

mzient:
#### Why we need this PR?
*Pick one, remove the rest*
- It fixes a problem with use of incomplete type when compiling with clang

 - What solution was applied:
     * Used external (wrt class) template alias instead of a member.
     * Reduction kernels
     * N//A
 - Validation and testing:
     * Existing tests apply
     * N/A

**JIRA TASK**: N/A


mzient:
!build

jantonguirao (approved):

klecki (approved):

PR #2382, opened 2020-10-21
Generate signatures

klecki:
TODO: 

- [ ] deduplicate between modules
- [ ] See if explicit kwargs break something
- [ ] Does it help with IDE integration? 

#### Why we need this PR?
Set signatures of DALI calls.

 - What solution was applied:
     Decorate the `__call__` or `fn.op` and set the signature explicitly. Bind the arguments to signature to group and reorder them as needed by the operator (between named positional args and kwargs).

Reference: https://www.python.org/dev/peps/pep-0362/

     ops and fn Python APIs.
 - Validation and testing:

### DIFF
--- a/dali/python/nvidia/dali/fn.py
+++ b/dali/python/nvidia/dali/fn.py
@@ -19,6 +19,7 @@ from functools import wraps
 
 from nvidia.dali.data_node import DataNode as _DataNode
 from nvidia.dali import internal as _internal
+from nvidia.dali import backend as _b
 
 def _call_signature(op_name, include_self, add_kwargs=False):
     schema = _b.GetSchema(op_name)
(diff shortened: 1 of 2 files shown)


dali/python/nvidia/dali/fn.py
```diff
@@ -14,8 +14,45 @@
...
 #pylint: disable=no-member
 import sys
+import inspect
+from functools import wraps
+
 from nvidia.dali.data_node import DataNode as _DataNode
 from nvidia.dali import internal as _internal
+from nvidia.dali import backend as _b
+
+def _call_signature(op_name, include_self, add_kwargs=False):
```
JanuszL, line 24: Is there any relevant difference between fn and ops flavors?

dali/python/nvidia/dali/fn.py
```diff
@@ -14,8 +14,45 @@
...
+        for i in range(schema.MinNumInput(), schema.MaxNumInput()):
+            input_list.append(inspect.Parameter(schema.GetInputName(i), inspect.Parameter.POSITIONAL_OR_KEYWORD, default=None))
+    if add_kwargs:
+        for arg in schema.GetArgumentNames():
+            # providing any defult changes DALI semantics
+            input_list.append(inspect.Parameter(arg, inspect.Parameter.KEYWORD_ONLY))
+    input_list.append(inspect.Parameter("kwargs", inspect.Parameter.VAR_KEYWORD))
+    return inspect.Signature(input_list)
+
+def decorate_signature(op_name, include_self=False):
```
JanuszL, line 41: Same as above.

klecki:
Done in #5154

PR #2388, opened 2020-10-22
Add layout support to OneHot operator

stiepan:

#### Why we need this PR?
- For OneHot operator to handle layout properly.

 - What solution was applied:
    Added new parameter layout_axis_name that is used as name in the layout for dimension introduced by operator
    Only one hot operator is affected
    Take a look on dimension naming restrictions for the layouts. By default dimension is labeled with 'O'. If you want the operator to return output without layout you need to explicitly pass empty string in layout_axis_name. Output layout is going to be empty also if the output is more than one dimensional and the input has no layout.
 - Validation and testing:
    Added relevant tests in test_operator_one_hot.py file.
    There's new parameter in OneHot operator documentation.

```
import nvidia.dali as dali
import numpy as np

pipe = dali.pipeline.Pipeline(batch_size = 2, num_threads = 3, device_id = 0,)
with pipe:
    src = dali.ops.ExternalSource(source=[np.array([[[1, 2], [2, 1]], [[3, 0], [0, 3]]]).reshape((2, 2, 2, 1, 1))], layout="ABCD")
    one_hot = dali.fn.one_hot(src(), num_classes=4)
    pipe.set_outputs(one_hot)
pipe.build()
(out,) = pipe.run()
out.layout()
```

**DALI-1659**

### DIFF
--- a/dali/operators/generic/one_hot.cc
+++ b/dali/operators/generic/one_hot.cc
@@ -87,13 +87,13 @@ TensorShape<> determine_shape(TensorShape<> in_shape, int num_classes, int axis,
 
 TensorLayout OneHot::GetOutputLayout(const HostWorkspace &ws, int placement_axis,
                                      int output_sample_dim) const {
+  if (!new_axis_name_) {
+    return {};
+  }
   const auto &input = ws.template InputRef<CPUBackend>(0);
   auto in_layout = input.GetLayout();
   // .size method returns uint_8 which doesn't work well when 0 is printed in error message
   int in_layout_size = in_layout.size();
-  if (!new_axis_name_) {
-    return {};
-  }
   // Handles the legacy 'multi-dimensional' scalars-like
   if (output_sample_dim == 1) {
     return TensorLayout(&new_axis_name_, 1);
--- a/dali/operators/generic/one_hot.cc
+++ b/dali/operators/generic/one_hot.cc
@@ -59,7 +59,7 @@ class in the corresponding input coordinate.
 
 This value will be cast to the ``dtype`` type.)code", 0.f)
   .AddOptionalArg<std::string>("axis_name",
-                  R"code(Single character that will be used as a name for newly added
+                  R"code(Single character that will be used as a name for the newly added
 dimension in the output layout. If no character is provided, the output layout will be
 empty.)code", nullptr);
 
@@ -104,7 +104,7 @@ TensorLayout OneHot::GetOutputLayout(const HostWorkspace &ws, int placement_axis
   }
   DALI_FAIL(make_string("Input layout mismatch - expected input layout to be of size ",
                         output_sample_dim - 1, " but instead got \"", in_layout,
-                        "\" that is of size ", in_layout_size, "."));
+                        "\", which is of size ", in_layout_size, "."));
 }
 
 bool OneHot::SetupImpl(std::vector<OutputDesc> &output_desc, const HostWorkspace &ws) {
--- a/dali/operators/generic/one_hot.cc
+++ b/dali/operators/generic/one_hot.cc
@@ -89,7 +89,8 @@ TensorLayout OneHot::GetOutputLayout(const HostWorkspace &ws, int placement_axis
                                      int output_sample_dim) const {
   const auto &input = ws.template InputRef<CPUBackend>(0);
   auto in_layout = input.GetLayout();
-  auto in_layout_size = in_layout.size();
+  // .size method returns uint_8 which doesn't work well when 0 is printed in error message
+  int in_layout_size = in_layout.size();
   if (!new_axis_name_) {
     return {};
   }
--- a/dali/operators/generic/one_hot.cc
+++ b/dali/operators/generic/one_hot.cc
@@ -57,7 +57,11 @@ This value will be cast to the ``dtype`` type.)code", 1.f)
                   R"code(Value that will be used to fill the output to indicate the lack of given
 class in the corresponding input coordinate.
 
-This value will be cast to the ``dtype`` type.)code", 0.f);
+This value will be cast to the ``dtype`` type.)code", 0.f)
+  .AddOptionalArg("layout_axis_name",
+                  R"code(Single character (an upper case letter or a digit) that newly added
+dimension will be named with in the output layout. Provide empty string for empty output
+layout.)code", "O");
 
 namespace {
 
@@ -74,12 +78,32 @@ TensorShape<> determine_shape(TensorShape<> in_shape, int num_classes, int axis,
   axis = axis < 0 ? in_shape.sample_dim() : axis;
   int outer_axes = axis;
   int inner_axes = in_shape.sample_dim() - axis;
-  auto outer =  in_shape.first(outer_axes);
+  auto outer = in_shape.first(outer_axes);
   auto inner = in_shape.last(inner_axes);
   return shape_cat(shape_cat(outer, num_classes), inner);
 }
 
 }  // namespace
+
+TensorLayout OneHot::GetOutputLayout(const HostWorkspace &ws, int placement_axis) const {
+  const auto &input = ws.template InputRef<CPUBackend>(0);
+  const auto &output = ws.template OutputRef<CPUBackend>(0);
+  auto in_layout = input.GetLayout();
+  auto in_layout_size = in_layout.size();
+  auto out_sample_dim = output.shape().sample_dim();
+  if (layout_axis_name_.length() == 0) {
+    return {};
+  }
+  // Handles the legacy 'multi-dimensional' scalars-like
+  if (out_sample_dim == 1) {
+    return layout_axis_name_;
+  }
+  if (in_layout_size + 1 == out_sample_dim) {
+    return in_layout.first(placement_axis) + layout_axis_name_ +
+           in_layout.last(in_layout_size - placement_axis);
+  }
+  return {};
+}
 
 bool OneHot::SetupImpl(std::vector<OutputDesc> &output_desc, const HostWorkspace &ws) {
   const auto &input = ws.template InputRef<CPUBackend>(0);
@@ -113,13 +137,13 @@ bool OneHot::SetupImpl(std::vector<OutputDesc> &output_desc, const HostWorkspace
   return true;
 }
 
-// TODO(klecki): Handle layout, maybe introduce a parameter for how to name the new axis
 void OneHot::RunImpl(HostWorkspace &ws) {
   const auto &input = ws.template InputRef<CPUBackend>(0);
   auto &output = ws.template OutputRef<CPUBackend>(0);
   auto &tp = ws.GetThreadPool();
   auto in_shape = input.shape();
   int placement_axis = axis_ < 0 ? output.shape().sample_dim() - 1 : axis_;
+  output.SetLayout(GetOutputLayout(ws, placement_axis));
   TYPE_SWITCH(input.type().id(), type2id, InputType, ONE_HOT_TYPES, (
     TYPE_SWITCH(output_type_, type2id, OutputType, ONE_HOT_TYPES, (
 
--- a/dali/operators/generic/one_hot.cc
+++ b/dali/operators/generic/one_hot.cc
@@ -58,10 +58,10 @@ This value will be cast to the ``dtype`` type.)code", 1.f)
 class in the corresponding input coordinate.
 
 This value will be cast to the ``dtype`` type.)code", 0.f)
-  .AddOptionalArg("layout_axis_name",
-                  R"code(Single character (an upper case letter or a digit) that newly added
-dimension will be named with in the output layout. Provide empty string for empty output
-layout.)code", "O");
+  .AddOptionalArg<std::string>("axis_name",
+                  R"code(Single character that will be used as a name for newly added
+dimension in the output layout. If no character is provided, the output layout will be
+empty.)code", nullptr);
 
 namespace {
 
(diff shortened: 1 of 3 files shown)


stiepan:
!build

dali/operators/generic/one_hot.cc
```diff
@@ -57,7 +57,11 @@ This value will be cast to the ``dtype`` type.)code", 1.f)
                   R"code(Value that will be used to fill the output to indicate the lack of given
 class in the corresponding input coordinate.
 
-This value will be cast to the ``dtype`` type.)code", 0.f);
+This value will be cast to the ``dtype`` type.)code", 0.f)
+  .AddOptionalArg("layout_axis_name",
```
mzient, line 61: We use `axis_name` in other operators.

dali/operators/generic/one_hot.cc
```diff
@@ -57,7 +57,11 @@ This value will be cast to the ``dtype`` type.)code", 1.f)
...
+                  R"code(Single character (an upper case letter or a digit) that newly added
```
mzient, line 62:
```suggestion
                  R"code(Single character that newly added
```
It can be anything.

dali/operators/generic/one_hot.h
```diff
@@ -74,12 +87,14 @@ class OneHot : public Operator<CPUBackend> {
...
   void RunImpl(HostWorkspace &ws) override;
 
+  TensorLayout GetOutputLayout(const HostWorkspace &ws, int placement_axis) const;
 
   int num_classes_;
   int axis_;
   const DALIDataType output_type_;
   float on_value_;
   float off_value_;
+  const std::string layout_axis_name_;
```
mzient, line 97:
```suggestion
  char layout_axis_name_ = 0;
```
It's a single character.

dali/operators/generic/one_hot.cc
```diff
@@ -74,13 +78,33 @@ TensorShape<> determine_shape(TensorShape<> in_shape, int num_classes, int axis,
...
-  auto outer =  in_shape.first(outer_axes);
+  auto outer = in_shape.first(outer_axes);
   auto inner = in_shape.last(inner_axes);
   return shape_cat(shape_cat(outer, num_classes), inner);
 }
 
 }  // namespace
 
+TensorLayout OneHot::GetOutputLayout(const HostWorkspace &ws, int placement_axis) const {
+  const auto &input = ws.template InputRef<CPUBackend>(0);
```
mzient, line 89: You can short circuit here:
```Python
  if (!new_axis_name_)
    return {};
  }
```

dali/test/python/test_operator_one_hot.py
```diff
@@ -11,34 +11,74 @@
...
+
 num_classes = 20
 batch_size = 10
 
 
 def insert_as_axis(target, value, axis, total_axes):
     return target[0:axis] + (value,) + target[axis:total_axes]
 
+
+def get_initial_layout(sample_dim=0, insert_at=None, new_axis_name='O', base="ABCD"):
```
klecki, line 34: Nitpick:
```suggestion
def get_initial_layout(sample_dim=0, base="ABCD"):
```

dali/operators/generic/one_hot.h
```diff
@@ -57,7 +57,20 @@ class OneHot : public Operator<CPUBackend> {
         axis_(spec.GetArgument<int>("axis")),
         output_type_(spec.GetArgument<DALIDataType>(arg_names::kDtype)),
         on_value_(spec.GetArgument<float>("on_value")),
-        off_value_(spec.GetArgument<float>("off_value")) {}
+        off_value_(spec.GetArgument<float>("off_value")),
+        layout_axis_name_(spec.GetArgument<std::string>("layout_axis_name")) {
```
mzient, line 61: FYI: We're gradually departing from storing argument values in constructors. Layouts are likely to remain constant, though.

dali/operators/generic/one_hot.h
```diff
@@ -57,7 +57,20 @@ class OneHot : public Operator<CPUBackend> {
...
+    auto axis_name_len = layout_axis_name_.length();
+    unsigned char ch;
+    if (axis_name_len > 1 ||
+        (axis_name_len == 1 &&
+         (!std::isupper(ch = static_cast<unsigned char>(layout_axis_name_[0])) &&
+          !std::isdigit(ch)))) {
```
mzient, line 67: These constraints are excessive.

dali/test/python/test_operator_one_hot.py
```diff
@@ -66,37 +107,85 @@ def one_hot(input):
...
+    check_one_hot_operator(premade_batch, axis=-1,
+                           initial_layout="ABC", layout_axis_name="CD")
+
+
+@raises(RuntimeError)
+def test_lowercase_axis_name():
+    np.random.seed(42)
+    premade_batch = random_3d_tensors_batch()
+    check_one_hot_operator(premade_batch, axis=-1,
+                           initial_layout="ABC", layout_axis_name="q")
```
mzient, line 191:
```suggestion
                           initial_layout="ABC", layout_axis_name="")
```
Lowercase is OK, but I think that this should fail - we should distinguish "" from None.

dali/operators/generic/one_hot.cc
```diff
@@ -57,7 +57,11 @@ This value will be cast to the ``dtype`` type.)code", 1.f)
                   R"code(Value that will be used to fill the output to indicate the lack of given
 class in the corresponding input coordinate.
 
-This value will be cast to the ``dtype`` type.)code", 0.f);
+This value will be cast to the ``dtype`` type.)code", 0.f)
+  .AddOptionalArg("layout_axis_name",
+                  R"code(Single character (an upper case letter or a digit) that newly added
+dimension will be named with in the output layout. Provide empty string for empty output
+layout.)code", "O");
```
mzient, line 64:
```suggestion
layout.)code", nullptr);
```

dali/operators/generic/one_hot.cc
mzient, line 61:
```suggestion
  .AddOptionalArg<string>("layout_axis_name",
```

dali/operators/generic/one_hot.h
stiepan: What about std::isalnum so that '\n' or '\r' are rejected? Or it's not possible to pass them in the first place?

dali/test/python/test_operator_one_hot.py
```diff
@@ -11,34 +11,74 @@
...
 class OneHotPipeline(Pipeline):
-    def __init__(self, num_classes, input, axis=-1, num_threads=1):
+    def __init__(self, num_classes, input, axis=-1, num_threads=1, layout=None, layout_axis_name="O"):
         super(OneHotPipeline, self).__init__(batch_size,
                                              num_threads,
                                              0)
-        sample_dim = len(input[0].shape)
-        self.ext_src = ops.ExternalSource(source=[input], cycle=True, layout="ABCD"[0:sample_dim])
-        self.one_hot = ops.OneHot(num_classes=num_classes, axis=axis, dtype=types.INT32, device="cpu")
+        self.ext_src = ops.ExternalSource(source=[input], cycle=True, layout=layout)
```
klecki, line 73: I know this is a bit out of scope here, but can we maybe move the batch generation to here (pass the callbacks that you already created) - so we have several different inputs when iterating.

dali/test/python/test_operator_one_hot.py
```diff
@@ -66,37 +107,85 @@ def one_hot(input):
...
+    yield partial(check_one_hot_operator, axis=None, expected_output_dim=1, initial_layout=layout, layout_axis_name=""), premade_batch
+
+
+def test_one_hot_custom_layout_axis_name():
+    np.random.seed(42)
+    layout = get_initial_layout(3)
+    premade_batch = random_3d_tensors_batch()
+    yield partial(check_one_hot_operator, axis=-1, initial_layout=layout, layout_axis_name="X"), premade_batch
+    yield partial(check_one_hot_operator, axis=-1, initial_layout=layout, layout_axis_name=0), premade_batch
+    yield partial(check_one_hot_operator, axis=-1, initial_layout=layout, layout_axis_name=1), premade_batch
```
klecki, line 175: Use a loop

```suggestion
	for layout_axis_name in ["X", "x", "0"]:
        yield partial(check_one_hot_operator, axis=-1, initial_layout=layout, layout_axis_name="X"), premade_batch
        yield partial(check_one_hot_operator, axis=-1, initial_layout=layout, layout_axis_name=0), premade_batch
        yield partial(check_one_hot_operator, axis=-1, initial_layout=layout, layout_axis_name=1), premade_batch
```

dali/test/python/test_operator_one_hot.py
```diff
@@ -66,37 +107,85 @@ def one_hot(input):
...
+    premade_batch = random_scalar_like_tensors_batch(3)
+    yield partial(check_one_hot_operator, axis=None, expected_output_dim=1, initial_layout=layout, layout_axis_name=""), premade_batch
+
+
+def test_one_hot_custom_layout_axis_name():
+    np.random.seed(42)
+    layout = get_initial_layout(3)
+    premade_batch = random_3d_tensors_batch()
+    yield partial(check_one_hot_operator, axis=-1, initial_layout=layout, layout_axis_name="X"), premade_batch
+    yield partial(check_one_hot_operator, axis=-1, initial_layout=layout, layout_axis_name=0), premade_batch
```
klecki, line 174: I don't like this conversion from int to string here, I would suggest using explicit string.

dali/operators/generic/one_hot.h
mzient: I'd be more worried by non-ASCII characters than by non-alphanumeric. But so far we haven't seen any problems and it's not validated anywhere. The only thing that would wreak havoc is `\0`, but we don't guard against it, either - it would likely cause corrupted layout at some point, but nothing more severe (it wouldn't crash or violate memory access).

dali/test/python/test_operator_one_hot.py
stiepan: OK. So just to make sure we're on the same page - onehot by default should not add layout and only providing the param explicitly should enable the feature, is that right?

How I imagined it was that by default the 'O' is added the layout and only providing empty string disables the feature. I wasn't sure if it is possible to distinguish at the c++ operator level between user passing param=None and not providing param at all to the operator call, that's why I've chosen empty string instead for opting-out.

dali/operators/generic/one_hot.cc
```diff
@@ -57,7 +57,11 @@ This value will be cast to the ``dtype`` type.)code", 1.f)
                   R"code(Value that will be used to fill the output to indicate the lack of given
 class in the corresponding input coordinate.
 
-This value will be cast to the ``dtype`` type.)code", 0.f);
+This value will be cast to the ``dtype`` type.)code", 0.f)
+  .AddOptionalArg("layout_axis_name",
+                  R"code(Single character (an upper case letter or a digit) that newly added
+dimension will be named with in the output layout. Provide empty string for empty output
```
JanuszL, line 63: If you change the default to nullptr you would probably need to adjust the docs here as well.

dali/operators/generic/one_hot.cc
```diff
@@ -74,13 +78,33 @@ TensorShape<> determine_shape(TensorShape<> in_shape, int num_classes, int axis,
...
 }
 
 }  // namespace
 
+TensorLayout OneHot::GetOutputLayout(const HostWorkspace &ws, int placement_axis) const {
+  const auto &input = ws.template InputRef<CPUBackend>(0);
+  const auto &output = ws.template OutputRef<CPUBackend>(0);
+  auto in_layout = input.GetLayout();
+  auto in_layout_size = in_layout.size();
+  auto out_sample_dim = output.shape().sample_dim();
```
JanuszL, line 93: Maybe it should accept `out_sample_dim` as an argument, otherwise someone may call this function to the place where output.shape() is not defined yet.

dali/operators/generic/one_hot.h
JanuszL: We have a layout check on the executor level AFAIK. We can add something like this there as well. But just an idea for a separate PR.

dali/operators/generic/one_hot.cc
stiepan: done

dali/test/python/test_operator_one_hot.py
stiepan: done

dali/operators/generic/one_hot.cc
stiepan: done

dali/operators/generic/one_hot.cc
stiepan: done

dali/operators/generic/one_hot.cc
stiepan: done

dali/operators/generic/one_hot.cc
stiepan: OK, added it as a parameter.

stiepan:
!build

klecki (approved):

JanuszL (approved):

dali/operators/generic/one_hot.cc
```diff
@@ -57,7 +57,11 @@ This value will be cast to the ``dtype`` type.)code", 1.f)
                   R"code(Value that will be used to fill the output to indicate the lack of given
 class in the corresponding input coordinate.
 
-This value will be cast to the ``dtype`` type.)code", 0.f);
+This value will be cast to the ``dtype`` type.)code", 0.f)
+  .AddOptionalArg<std::string>("axis_name",
+                  R"code(Single character that will be used as a name for newly added
```
mzient, line 62:
```suggestion
                  R"code(Single character that will be used as a name for the newly added
```

dali/operators/generic/one_hot.cc
```diff
@@ -74,13 +78,35 @@ TensorShape<> determine_shape(TensorShape<> in_shape, int num_classes, int axis,
...
 }  // namespace
 
+TensorLayout OneHot::GetOutputLayout(const HostWorkspace &ws, int placement_axis,
+                                     int output_sample_dim) const {
+  const auto &input = ws.template InputRef<CPUBackend>(0);
+  auto in_layout = input.GetLayout();
+  // .size method returns uint_8 which doesn't work well when 0 is printed in error message
+  int in_layout_size = in_layout.size();
+  if (!new_axis_name_) {
+    return {};
```
mzient, line 95: You don't have to look at the inputs or extract layout to come to this conclusion. This can safely be the first thing this function does.

dali/operators/generic/one_hot.cc
```diff
@@ -74,13 +78,35 @@ TensorShape<> determine_shape(TensorShape<> in_shape, int num_classes, int axis,
...
+  if (output_sample_dim == 1) {
+    return TensorLayout(&new_axis_name_, 1);
+  }
+  if (in_layout_size + 1 == output_sample_dim) {
+    return in_layout.first(placement_axis) + TensorLayout(&new_axis_name_, 1) +
+           in_layout.last(in_layout_size - placement_axis);
+  }
+  DALI_FAIL(make_string("Input layout mismatch - expected input layout to be of size ",
+                        output_sample_dim - 1, " but instead got \"", in_layout,
+                        "\" that is of size ", in_layout_size, "."));
```
mzient, line 107:
```suggestion
                        "\", which is of size ", in_layout_size, "."));
```
Non-defining relative clause always uses "which" with a leading comma.

dali/operators/generic/one_hot.cc
stiepan: Good point, fixed

mzient (approved):

stiepan:
!build

PR #2395, opened 2020-10-24
More reductions 

awolant:
#### Why we need this PR?
- It adds new feature needed because we want to support more reductions

 - What solution was applied:
     *Added new ops: Mean, MeanSquare, RootMeanSquare, Std, Variance*
 - Validation and testing:
     *Added tests for new ops*
     *Docs in op schema*

**JIRA TASK**: *[DALI-1675, DALI-1676]*


JanuszL (approved):

awolant:
!build

dali/operators/generic/reduce/mean_square.h
```diff
@@ -0,0 +1,70 @@
...
+/// for MeanSquare operator. Is uesed as parameter to TYPE_MAP macro.
+#define MEAN_SQUARE_TYPES_MAP ( \
+    ((uint8_t), (uint8_t, float)), \
+    ((int8_t), (int8_t, float)), \
+    ((uint16_t), (uint16_t, float)), \
+    ((int16_t), (int16_t, float)), \
+    ((uint32_t), (uint32_t, float)), \
+    ((int32_t), (int32_t, float)), \
+    ((uint64_t), (uint64_t, float)), \
+    ((int64_t), (int64_t, float)), \
```
mzient, line 31: Calculating mean square and staying in a small output type is asking for trouble - by squaring we're doubling the number of bits required. We also _can_ support both signed and unsigned outputs for signed input just by reinterpreting the data - that would complicate the matters a bit, but would cut the number of instantiations without sacrificing functionality (compared to either supporting both signed and unsigned outputs or not supporting one of the type flavors).

dali/operators/generic/reduce/reduce_with_mean_input.h
```diff
@@ -0,0 +1,188 @@
...
+#define REDUCE_WITH_MEAN_TYPES_MAP ( \
+    ((uint8_t), (uint8_t, float)), \
+    ((int8_t), (int8_t, float)), \
+    ((uint16_t), (uint16_t, float)), \
+    ((int16_t), (int16_t, float)), \
+    ((uint32_t), (uint32_t, float)), \
+    ((int32_t), (int32_t, float)), \
+    ((uint64_t), (uint64_t, float)), \
+    ((int64_t), (int64_t, float)), \
+    ((float), (float)))
```
mzient, line 37: Likewise - the output types for Standard Deviation and for Variance would differ (variance needs a 2x larger output type).

dali/operators/generic/reduce/reduce.cc
```diff
@@ -30,14 +34,53 @@ Not providing any axis results in reduction of all elements.)code",
...
     DALI_NO_TYPE)
+  .AddParent("ReduceBase");
+
+  DALI_SCHEMA(ReduceWithMeanInput)
+  .AddOptionalArg("ddof",
+    R"code(Delat Degrees of Freedom. Adjusts the divisor used in calculations, which is `N - ddof`.)code",
+    0)
+  .AddParent("ReduceWithOutputType");
+
+DALI_SCHEMA(reductions__Std)
```
mzient, line 49: I don't like this capitalization - I think it should be StD, but it would require some special handling in _to_snake_case (it's too short to add to special case mapping with acceptable risk).

dali/operators/generic/reduce/reduce.cc
```diff
@@ -30,14 +34,53 @@ Not providing any axis results in reduction of all elements.)code",
...
-  .AddParent("ReduceBase");
+  .AddParent("ReduceWithOutputType");
+
+DALI_SCHEMA(reductions__MeanSquare)
+  .DocStr("Gets mean square of elements along provided axes.")
+  .NumInput(1)
+  .NumOutput(1)
+  .AddParent("ReduceWithOutputType");
+
+DALI_SCHEMA(reductions__RootMeanSquare)
```
mzient, line 73:
```suggestion
DALI_SCHEMA(reductions__RMS)
```
This is a well established term in the field. Also, if we keep it uppercase here it will convert to well to "rms" in fn.

awolant: Done

dali/operators/generic/reduce/reduce.cc
awolant: Done, changed to `STD`

mzient (approved):

JanuszL (approved):

awolant:
!build

PR #2408, opened 2020-10-27
Ops rework to prepare iter-to-iter batch size variability

szalpal:
### What happens in this PR:
This is a PR that prepares ground to allow iter-to-iter batch size variability in DALI. This is a situation, where a single `Pipeline` object can be ran with various batch sizes of the input data, but assuming that the batch size remains constant throughout a single iteration.
In order to allow such behaviour, I assumed that whatever user was passing up until now to the `Pipeline` ctor as the `batch_size` argument, now will denote maximum batch size which can be handled by `Pipeline`. This is not a hard assumption, meaning there are places it is not necessary to have such.

To explain what happens here, I'll firstly describe how it used to work:
`batch_size` typically was acquired in one 3 ways: from an input itself, from the field in `Operator` superclass and using `GetArgument("batch_size")` function. The first one is fine. I went through every such usage and changed the latter two:
1. I've decided whether this particular usage shall be `curr_batch_size` or `max_batch_size`. 
1. `max_batch_size` is set in the `Operator` superclass in a way the old `batch_size` was
1. For the `curr_batch_size`, I've used the new function in the workspace (`WorskpaceBase::get_batch_size(int)`, snake_case, because `GetBatchSize` is already taken; and I believe this name should be unique, in case someone in future want to implement fully variable batch size). This is a utility function, which acquires batch size from an input. 

Handling inputless operators is done in `get_batch_size` as well.

Not every operator is reworked so far. I've decided that the Decoders are too complex (please mind, that not only the rework matters, but also testing it) to be a part of this PR. The same goes for `NormalDistribution` and other random ops, which need the change in `BatchRng` to work. Those will be reworked in next PRs. 
Additionally, there are ops that aren't needed for Triton use case, which are Readers and BoxEncoder.
The ops that are not reworked still work, ofc. It's just assumed, that they always work on `max_batch_size`, therefore they won't work with anything smaller.

#### Testing
Approach to testing is described in `test_dali_variable_batch_size.py`. Please check #2481 

### Remarks
1. Operators, that are not reworked yet, assume that they always work on `max_batch_size`.
1. This PR doesn't suffice Triton use case. The operators, that are also required are `HostDecoder` and `NormalDistribution`. Reworking of these will happen in next PR(s).
1. The `SampleWorkspace` and the `get_batch_size`: generally, `SampleWorkspace` doesn't have a clue, that there exists such thing as a batch. I've added an explicit fail, if someone tries to call this function in the `SampleWorkspace`. I believe it would be incorrect to completely omit this one, since in this case the `OperatorBase`'s function would be called. Which actually might return meaningful value.

### Relevant for the reviewer:
1. The first commit retains previous behaviour, but just splits using `batch_size` into 2: `max_batch_size` and `curr_batch_size`. If you find the amount of files overwhelming, turn this one off
1. Please double-check, if I didn't miss any operator, that should be included in the variable_bs test in this PR.
Operators that are not and should not be included in this PR:
    - Readers
    - Decoders (except `AudioDecoder`)
    - Random ops
    - `transforms` module, since it doesn't really rely on batch size
    - `BoxEncoder`

 - What solution was applied:
     See above
     everything.
     See section above
 - Validation and testing:
     `test_dali_variable_batch_size.py`
     j/w

### DIFF
--- a/dali/benchmark/color_twist_bench.cc
+++ b/dali/benchmark/color_twist_bench.cc
@@ -36,7 +36,7 @@ BENCHMARK_DEFINE_F(OperatorBench, ColorTwistGPU)(benchmark::State& st) {
   this->RunGPU<uint8_t>(
     st,
     OpSpec("ColorTwist")
-      .AddArg("batch_size", batch_size)
+      .AddArg("max_batch_size", batch_size)
       .AddArg("num_threads", 1)
       .AddArg("device", "gpu")
       .AddArg("output_type", DALI_RGB)
@@ -56,7 +56,7 @@ BENCHMARK_DEFINE_F(OperatorBench, OldColorTwistGPU)(benchmark::State& st) {
   this->RunGPU<uint8_t>(
     st,
     OpSpec("OldColorTwist")
-      .AddArg("batch_size", batch_size)
+      .AddArg("max_batch_size", batch_size)
       .AddArg("num_threads", 1)
       .AddArg("device", "gpu")
       .AddArg("output_type", DALI_RGB)
--- a/dali/operators/audio/nonsilence_op.cc
+++ b/dali/operators/audio/nonsilence_op.cc
@@ -67,10 +67,11 @@ bool NonsilenceOperatorCpu::SetupImpl(std::vector<OutputDesc> &output_desc,
   TypeInfo output_type;
   output_type.SetType<int>(TypeTable::GetTypeID<int>());
   TensorShape<> scalar_shape = {};
+  auto curr_batch_size = ws.GetInputBatchSize(0);
 
   output_desc.resize(detail::kNumOutputs);
   for (int i = 0; i < detail::kNumOutputs; i++) {
-    output_desc[i].shape = uniform_list_shape(batch_size_, scalar_shape);
+    output_desc[i].shape = uniform_list_shape(curr_batch_size, scalar_shape);
     output_desc[i].type = output_type;
   }
   return true;
--- a/dali/operators/audio/nonsilence_op.h
+++ b/dali/operators/audio/nonsilence_op.h
@@ -150,9 +150,10 @@ class NonsilenceOperator : public Operator<Backend> {
 
 
   void AcquireArgs(const OpSpec &spec, const workspace_t<Backend> &ws) {
-    this->GetPerSampleArgument(cutoff_db_, "cutoff_db", ws);
+    auto curr_batch_size = ws.GetInputBatchSize(0);
+    this->GetPerSampleArgument(cutoff_db_, "cutoff_db", ws, curr_batch_size);
     if (spec.HasArgument("reference_power")) {
-      this->GetPerSampleArgument(reference_power_, "reference_power", ws);
+      this->GetPerSampleArgument(reference_power_, "reference_power", ws, curr_batch_size);
       for (const auto &val : reference_power_) {
         DALI_ENFORCE(val > 0, make_string("`reference_power` has to be positive. Got: ", val));
       }
@@ -209,9 +210,10 @@ class NonsilenceOperatorCpu : public NonsilenceOperator<CPUBackend> {
     const auto &input = ws.template InputRef<CPUBackend>(0);
     auto &output_begin = ws.OutputRef<CPUBackend>(0);
     auto &output_length = ws.OutputRef<CPUBackend>(1);
+    auto curr_batch_size = ws.GetInputBatchSize(0);
     auto &tp = ws.GetThreadPool();
     auto in_shape = input.shape();
-    for (int sample_id = 0; sample_id < batch_size_; sample_id++) {
+    for (int sample_id = 0; sample_id < curr_batch_size; sample_id++) {
       tp.AddWork(
               [&, sample_id](int thread_id) {
                   detail::Args<InputType> args;
--- a/dali/operators/audio/preemphasis_filter_op.cu
+++ b/dali/operators/audio/preemphasis_filter_op.cu
@@ -51,7 +51,7 @@ class PreemphasisFilterGPU : public PreemphasisFilter<GPUBackend> {
  public:
   explicit PreemphasisFilterGPU(const OpSpec &spec) : PreemphasisFilter<GPUBackend>(spec) {
     // void is OK here, pointer sizes are the same size
-    int64_t sz = batch_size_ * sizeof(detail::SampleDescriptor<void, void>);
+    int64_t sz = max_batch_size_ * sizeof(detail::SampleDescriptor<void, void>);
     scratch_mem_.set_type(TypeTable::GetTypeInfo(DALI_UINT8));
     scratch_mem_.Resize({sz});
   }
@@ -69,9 +69,10 @@ void PreemphasisFilterGPU::RunImplTyped(workspace_t<GPUBackend> &ws) {
   using SampleDesc = detail::SampleDescriptor<OutputType, InputType>;
   const auto &input = ws.InputRef<GPUBackend>(0);
   auto &output = ws.OutputRef<GPUBackend>(0);
+  auto curr_batch_size = ws.GetInputBatchSize(0);
 
-  std::vector<SampleDesc> samples_cpu(batch_size_);
-  for (int sample_idx = 0; sample_idx < batch_size_; sample_idx++) {
+  std::vector<SampleDesc> samples_cpu(curr_batch_size);
+  for (int sample_idx = 0; sample_idx < curr_batch_size; sample_idx++) {
     auto &sample = samples_cpu[sample_idx];
     sample.in = input.tensor<InputType>(sample_idx);
     sample.out = output.mutable_tensor<OutputType>(sample_idx);
@@ -79,7 +80,7 @@ void PreemphasisFilterGPU::RunImplTyped(workspace_t<GPUBackend> &ws) {
     sample.coeff = preemph_coeff_[sample_idx];
   }
 
-  int64_t sz = batch_size_ * sizeof(SampleDesc);
+  int64_t sz = curr_batch_size * sizeof(SampleDesc);
   scratch_mem_.set_type(TypeTable::GetTypeInfo(DALI_UINT8));
   scratch_mem_.Resize({sz});
   auto sample_descs_gpu = reinterpret_cast<SampleDesc*>(scratch_mem_.mutable_data<uint8_t>());
@@ -88,8 +89,8 @@ void PreemphasisFilterGPU::RunImplTyped(workspace_t<GPUBackend> &ws) {
     cudaMemcpyAsync(sample_descs_gpu, samples_cpu.data(), sz, cudaMemcpyHostToDevice, stream));
 
   int block = 256;
-  auto blocks_per_sample = std::max(32, 1024 / batch_size_);
-  dim3 grid(blocks_per_sample, batch_size_);
+  auto blocks_per_sample = std::max(32, 1024 / curr_batch_size);
+  dim3 grid(blocks_per_sample, curr_batch_size);
   detail::PreemphasisFilterKernel<<<grid, block, 0, stream>>>(sample_descs_gpu);
 }
 
--- a/dali/operators/bbox/bb_flip.h
+++ b/dali/operators/bbox/bb_flip.h
@@ -44,21 +44,6 @@ class BbFlip<CPUBackend> : public Operator<CPUBackend> {
 
  private:
   /**
-   * Checks, if argument provided by user is a scalar and,
-   * in such case, extends this scalar to entire tensor
-   * @tparam TensorDataType Underlying data type in tensor
-   */
-  template <typename TensorDataType>
-  void ExtendScalarToTensor(std::string argument_name, const OpSpec &spec,
-                            Tensor<CPUBackend> *tensor) {
-    tensor->Resize({batch_size_});
-    for (int i = 0; i < batch_size_; i++) {
-      tensor->mutable_data<TensorDataType>()[i] =
-          spec.GetArgument<TensorDataType>(argument_name);
-    }
-  }
-
-  /**
    * Bounding box can be represented in two ways:
    * 1. Upper-left corner, width, height (`wh_type`)
    *    (x1, y1,  w,  h)
--- a/dali/operators/decoder/cache/cached_decoder_impl.cc
+++ b/dali/operators/decoder/cache/cached_decoder_impl.cc
@@ -39,7 +39,7 @@ CachedDecoderImpl::CachedDecoderImpl(const OpSpec& spec)
         device_id_, cache_type, cache_size, cache_debug, cache_threshold);
 
       use_batch_copy_kernel_ = spec.GetArgument<bool>("cache_batch_copy");
-      auto batch_size = spec.GetArgument<int>("batch_size");
+      auto batch_size = spec.GetArgument<int>("max_batch_size");
       const size_t kMaxSizePerBlock = 1<<18;  // 256 kB per block
       scatter_gather_.reset(new kernels::ScatterGatherGPU(
         kMaxSizePerBlock, cache_size, batch_size));
--- a/dali/operators/decoder/host/fused/host_decoder_random_crop_test.cc
+++ b/dali/operators/decoder/host/fused/host_decoder_random_crop_test.cc
@@ -25,7 +25,7 @@ class ImageDecoderRandomCropTest_CPU : public DecodeTestBase<ImgType> {
   ImageDecoderRandomCropTest_CPU()
     : random_crop_attr(
       OpSpec("RandomCropAttr")
-        .AddArg("batch_size", this->batch_size_)
+        .AddArg("max_batch_size", this->batch_size_)
         .AddArg("seed", kSeed)) {}
 
  protected:
--- a/dali/operators/generic/constant.h
+++ b/dali/operators/generic/constant.h
@@ -81,10 +81,12 @@ class Constant : public Operator<Backend> {
   bool SetupImpl(std::vector<OutputDesc> &output_desc, const Workspace &ws) override {
     output_desc.resize(1);
     auto curr_batch_size = ws.GetRequestedBatchSize(0);
-    if (output_shape_.empty() || output_shape_ .num_elements() < curr_batch_size) {
+    DALI_ENFORCE(curr_batch_size == max_batch_size_,
+                 "This operator does not doesn't support batch smaller than maximum");
+    if (output_shape_.empty() || output_shape_.num_elements() != curr_batch_size) {
       output_shape_ = uniform_list_shape(curr_batch_size, shape_arg_);
     }
-    output_desc[0] = { output_shape_, TypeTable::GetTypeInfo(output_type_) };
+    output_desc[0] = {output_shape_, TypeTable::GetTypeInfo(output_type_)};
     return false;
   }
 
--- a/dali/operators/generic/constant.h
+++ b/dali/operators/generic/constant.h
@@ -80,9 +80,9 @@ class Constant : public Operator<Backend> {
 
   bool SetupImpl(std::vector<OutputDesc> &output_desc, const Workspace &ws) override {
     output_desc.resize(1);
-    if (output_shape_.empty()) {
-      int batch_size = this->spec_.template GetArgument<int>("batch_size");
-      output_shape_ = uniform_list_shape(batch_size, shape_arg_);
+    auto curr_batch_size = ws.GetRequestedBatchSize(0);
+    if (output_shape_.empty() || output_shape_ .num_elements() < curr_batch_size) {
+      output_shape_ = uniform_list_shape(curr_batch_size, shape_arg_);
     }
     output_desc[0] = { output_shape_, TypeTable::GetTypeInfo(output_type_) };
     return false;
@@ -91,6 +91,7 @@ class Constant : public Operator<Backend> {
   void RunImpl(Workspace &ws) override;
 
  private:
+  USE_OPERATOR_MEMBERS();
   std::vector<int> shape_arg_;
   std::vector<int> idata_;
   std::vector<float> fdata_;
--- a/dali/operators/generic/erase/erase.cu
+++ b/dali/operators/generic/erase/erase.cu
@@ -42,14 +42,11 @@ class EraseImplGpu : public OpImplBase<GPUBackend> {
  public:
   using EraseKernel = kernels::EraseGpu<T, Dims, channel_dim>;
 
-  explicit EraseImplGpu(const OpSpec &spec)
-  : spec_(spec)
-  , batch_size_(spec.GetArgument<int>("batch_size")) {
-      kmgr_.Resize<EraseKernel>(1, 1);
+  explicit EraseImplGpu(const OpSpec &spec) : spec_(spec) {
+    kmgr_.Resize<EraseKernel>(1, 1);
   }
 
-  bool SetupImpl(std::vector<OutputDesc> &output_desc,
-                 const workspace_t<GPUBackend> &ws) override {
+  bool SetupImpl(std::vector<OutputDesc> &output_desc, const workspace_t<GPUBackend> &ws) override {
     const auto &input = ws.template InputRef<GPUBackend>(0);
     auto layout = input.GetLayout();
     auto type = input.type();
@@ -79,10 +76,11 @@ class EraseImplGpu : public OpImplBase<GPUBackend> {
 
   void AcquireArgs(const DeviceWorkspace &ws, TensorListShape<> in_shape,
                    TensorLayout in_layout) {
+    auto curr_batch_size = ws.GetInputBatchSize(0);
     fill_values_ = spec_.template GetRepeatedArgument<float>("fill_value");
     auto args = detail::GetEraseArgs<T, Dims>(spec_, ws, in_shape, in_layout);
-    auto regions_shape = TensorListShape<1>(batch_size_);
-    for (int i = 0; i < batch_size_; ++i) {
+    auto regions_shape = TensorListShape<1>(curr_batch_size);
+    for (int i = 0; i < curr_batch_size; ++i) {
       auto n_regions = static_cast<int>(args[i].rois.size());
       regions_shape.set_tensor_shape(i, {n_regions});
     }
@@ -90,7 +88,7 @@ class EraseImplGpu : public OpImplBase<GPUBackend> {
     regions_cpu.set_type(TypeInfo::Create<kernels::ibox<Dims>>());
     regions_cpu.Resize(regions_shape);
     auto regions_tlv = view<kernels::ibox<Dims>, 1>(regions_cpu);
-    for (int i = 0; i < batch_size_; ++i) {
+    for (int i = 0; i < curr_batch_size; ++i) {
       auto regions_tv = regions_tlv[i];
       for (int j = 0; j < regions_tv.shape[0]; ++j) {
         auto box = detail::make_box(args[i].rois[j].anchor, args[i].rois[j].shape);
@@ -102,7 +100,6 @@ class EraseImplGpu : public OpImplBase<GPUBackend> {
 
  private:
   const OpSpec &spec_;
-  int batch_size_ = 0;
   std::vector<int> axes_;
   std::vector<kernels::EraseArgs<T, Dims>> args_;
   TensorList<GPUBackend> regions_gpu_;
--- a/dali/operators/generic/flip.h
+++ b/dali/operators/generic/flip.h
@@ -58,23 +58,26 @@ class Flip: public Operator<Backend> {
     return this->spec_.template GetArgument<int>("depthwise", &ws, idx);
   }
 
-  std::vector<int> GetHorizontal(const ArgumentWorkspace &ws) {
+  std::vector<int> GetHorizontal(const workspace_t<Backend> &ws, int curr_batch_size) {
     std::vector<int> result;
-    OperatorBase::GetPerSampleArgument(result, "horizontal", ws);
+    OperatorBase::GetPerSampleArgument(result, "horizontal", ws, curr_batch_size);
     return result;
   }
 
-  std::vector<int> GetVertical(const ArgumentWorkspace &ws) {
+  std::vector<int> GetVertical(const workspace_t<Backend> &ws, int curr_batch_size) {
     std::vector<int> result;
-    OperatorBase::GetPerSampleArgument(result, "vertical", ws);
+    OperatorBase::GetPerSampleArgument(result, "vertical", ws, curr_batch_size);
     return result;
   }
 
-  std::vector<int> GetDepthwise(const ArgumentWorkspace &ws) {
+  std::vector<int> GetDepthwise(const workspace_t<Backend> &ws, int curr_batch_size) {
     std::vector<int> result;
-    OperatorBase::GetPerSampleArgument(result, "depthwise", ws);
+    OperatorBase::GetPerSampleArgument(result, "depthwise", ws, curr_batch_size);
     return result;
   }
+
+ private:
+  USE_OPERATOR_MEMBERS();
 };
 
 }  // namespace dali
--- a/dali/operators/generic/slice/slice_attr.h
+++ b/dali/operators/generic/slice/slice_attr.h
@@ -34,10 +34,9 @@ namespace dali {
 class SliceAttr {
  public:
   explicit inline SliceAttr(const OpSpec &spec)
-      : batch_size__(spec.GetArgument<int>("batch_size"))
-      , normalized_anchor_(spec.GetArgument<bool>("normalized_anchor"))
-      , normalized_shape_(spec.GetArgument<bool>("normalized_shape"))
-      , crop_window_generators_(batch_size__) {
+      : normalized_anchor_(spec.GetArgument<bool>("normalized_anchor")),
+        normalized_shape_(spec.GetArgument<bool>("normalized_shape")),
+        crop_window_generators_(spec.GetArgument<int>("max_batch_size")) {
     const bool has_axes_arg = spec.HasArgument("axes");
     const bool has_axis_names_arg = spec.HasArgument("axis_names");
     // Process `axis_names` if provided, or if neither `axis_names` nor `axes` are
@@ -56,6 +55,7 @@ class SliceAttr {
     DALI_ENFORCE(ws.NumInput() == 3,
       "Expected 3 inputs. Received: " + std::to_string(ws.NumInput()));
     // slice args are CPU inputs
+    auto curr_batch_size = ws.GetInputBatchSize(0);
     const auto &crop_anchor = ws.template InputRef<CPUBackend>(1);
     const auto &crop_shape = ws.template InputRef<CPUBackend>(2);
     DALI_ENFORCE(crop_anchor.type().id() == crop_shape.type().id(),
@@ -65,7 +65,7 @@ class SliceAttr {
     TYPE_SWITCH(args_dtype, type2id, ArgsType, SLICE_ARGS_TYPES, (
       auto anchor_view = view<const ArgsType>(crop_anchor);
       auto shape_view = view<const ArgsType>(crop_shape);
-      for (size_t data_idx = 0; data_idx < batch_size__; data_idx++) {
+      for (int data_idx = 0; data_idx < curr_batch_size; data_idx++) {
         VerifyArgsShape(anchor_view.tensor_shape(data_idx), shape_view.tensor_shape(data_idx));
         ProcessArgumentsHelper(data_idx,
                                anchor_view.tensor_data(data_idx),
@@ -161,7 +161,6 @@ class SliceAttr {
   }
 
  private:
-  size_t batch_size__;
   bool normalized_anchor_, normalized_shape_;
   std::vector<CropWindowGenerator> crop_window_generators_;
   std::vector<int> axes_;
--- a/dali/operators/geometry/affine_transforms/transform_base_op.h
+++ b/dali/operators/geometry/affine_transforms/transform_base_op.h
@@ -121,6 +121,7 @@ class TransformBaseOp : public Operator<Backend> {
 
   bool SetupImpl(std::vector<OutputDesc> &output_descs, const workspace_t<Backend> &ws) override {
     has_input_ = ws.NumInput() > 0;
+    auto curr_batch_size = has_input_ ? ws.GetInputBatchSize(0) : ws.GetRequestedBatchSize(0);
     if (has_input_) {
       auto &input = ws.template InputRef<Backend>(0);
       const auto &shape = input.shape();
@@ -136,7 +137,7 @@ class TransformBaseOp : public Operator<Backend> {
           "(ndim, ndim+1) representing an affine transform. Got: ", shape));
       nsamples_ = shape.num_samples();
     } else {
-      nsamples_ = spec_.template GetArgument<int>("batch_size");
+      nsamples_ = curr_batch_size;
     }
 
     This().ProcessArgs(spec_, ws);
@@ -280,6 +281,7 @@ class TransformBaseOp : public Operator<Backend> {
   }
 
  protected:
+  USE_OPERATOR_MEMBERS();
   DALIDataType dtype_ = DALI_FLOAT;
   int ndim_ = -1;  // will be inferred from the arguments or the input
   int nsamples_ = -1;
@@ -287,8 +289,6 @@ class TransformBaseOp : public Operator<Backend> {
   bool reverse_order_ = false;
 
   Tensor<CPUBackend> matrix_data_;
-
-  using Operator<Backend>::spec_;
 };
 
 
--- a/dali/operators/geometry/coord_flip.cu
+++ b/dali/operators/geometry/coord_flip.cu
@@ -68,10 +68,11 @@ class CoordFlipGPU : public CoordFlip<GPUBackend> {
 void CoordFlipGPU::RunImpl(workspace_t<GPUBackend> &ws) {
   const auto &input = ws.InputRef<GPUBackend>(0);
   auto &output = ws.OutputRef<GPUBackend>(0);
+  auto curr_batch_size = ws.GetInputBatchSize(0);
 
   sample_descs_.clear();
-  sample_descs_.reserve(batch_size_);
-  for (int sample_id = 0; sample_id < batch_size_; sample_id++) {
+  sample_descs_.reserve(max_batch_size_);
+  for (int sample_id = 0; sample_id < curr_batch_size; sample_id++) {
     SampleDesc<float> sample_desc;
     sample_desc.in = input.tensor<float>(sample_id);
     sample_desc.out = output.mutable_tensor<float>(sample_id);
@@ -105,7 +106,7 @@ void CoordFlipGPU::RunImpl(workspace_t<GPUBackend> &ws) {
   }
 
   scratchpad_.set_type(TypeInfo::Create<uint8_t>());
-  int64_t sz = batch_size_ * sizeof(SampleDesc<float>);
+  int64_t sz = curr_batch_size * sizeof(SampleDesc<float>);
   scratchpad_.Resize({sz});
   auto sample_descs_gpu_ = reinterpret_cast<SampleDesc<float>*>(
       scratchpad_.mutable_data<uint8_t>());
@@ -114,8 +115,8 @@ void CoordFlipGPU::RunImpl(workspace_t<GPUBackend> &ws) {
     cudaMemcpyAsync(sample_descs_gpu_, sample_descs_.data(), sz, cudaMemcpyHostToDevice, stream));
 
   int block = 1024;
-  auto blocks_per_sample = std::max(32, 1024 / batch_size_);
-  dim3 grid(blocks_per_sample, batch_size_);
+  auto blocks_per_sample = std::max(32, 1024 / curr_batch_size);
+  dim3 grid(blocks_per_sample, curr_batch_size);
   CoordFlipKernel<<<grid, block, 0, stream>>>(sample_descs_gpu_, ndim_);
 }
 
--- a/dali/operators/image/color/brightness_contrast.cu
+++ b/dali/operators/image/color/brightness_contrast.cu
@@ -29,6 +29,7 @@ DALI_REGISTER_OPERATOR(BrightnessContrast, BrightnessContrastGpu, GPU)
 
 bool BrightnessContrastGpu::SetupImpl(std::vector<OutputDesc> &output_desc,
                                       const workspace_t<GPUBackend> &ws) {
+  KMgrResize(num_threads_, max_batch_size_);
   const auto &input = ws.template InputRef<GPUBackend>(0);
   const auto &output = ws.template OutputRef<GPUBackend>(0);
   output_desc.resize(1);
--- a/dali/operators/image/color/brightness_contrast.h
+++ b/dali/operators/image/color/brightness_contrast.h
@@ -57,17 +57,12 @@ class BrightnessContrastOp : public Operator<Backend> {
 
  protected:
   explicit BrightnessContrastOp(const OpSpec &spec)
-        : Operator<Backend>(spec)
-        , output_type_arg_(spec.GetArgument<DALIDataType>("dtype"))
-        , output_type_(DALI_NO_TYPE)
-        , input_type_(DALI_NO_TYPE) {
+      : Operator<Backend>(spec),
+        output_type_arg_(spec.GetArgument<DALIDataType>("dtype")),
+        output_type_(DALI_NO_TYPE),
+        input_type_(DALI_NO_TYPE) {
     if (spec.HasArgument("contrast_center"))
       contrast_center_ = spec.GetArgument<float>("contrast_center");
-    if (std::is_same<Backend, GPUBackend>::value) {
-      kernel_manager_.Resize(1, 1);
-    } else {
-      kernel_manager_.Resize(num_threads_, batch_size_);
-    }
   }
 
   bool CanInferOutputs() const override {
@@ -95,15 +90,21 @@ class BrightnessContrastOp : public Operator<Backend> {
   }
 
   void AcquireArguments(const workspace_t<Backend> &ws) {
-    this->GetPerSampleArgument(brightness_, "brightness", ws);
-    this->GetPerSampleArgument(brightness_shift_, "brightness_shift", ws);
-    this->GetPerSampleArgument(contrast_, "contrast", ws);
+    auto curr_batch_size = ws.GetInputBatchSize(0);
+    this->GetPerSampleArgument(brightness_, "brightness", ws, curr_batch_size);
+    this->GetPerSampleArgument(brightness_shift_, "brightness_shift", ws, curr_batch_size);
+    this->GetPerSampleArgument(contrast_, "contrast", ws, curr_batch_size);
 
     input_type_ = ws.template InputRef<Backend>(0).type().id();
-    output_type_ =
-        output_type_arg_ != DALI_NO_TYPE
-        ? output_type_arg_
-        : input_type_;
+    output_type_ = output_type_arg_ != DALI_NO_TYPE ? output_type_arg_ : input_type_;
+  }
+
+  void KMgrResize(int num_threads, int batch_size) {
+    if (std::is_same<Backend, GPUBackend>::value) {
+      kernel_manager_.Resize(1, 1);
+    } else {
+      kernel_manager_.Resize(num_threads, batch_size);
+    }
   }
 
   USE_OPERATOR_MEMBERS();
--- a/dali/operators/image/color/color_twist.cu
+++ b/dali/operators/image/color/color_twist.cu
@@ -32,6 +32,7 @@ DALI_REGISTER_OPERATOR(Saturation, ColorTwistGpu, GPU);
 DALI_REGISTER_OPERATOR(ColorTwist, ColorTwistGpu, GPU);
 
 bool ColorTwistGpu::SetupImpl(std::vector<OutputDesc> &output_desc, const DeviceWorkspace &ws) {
+  KMgrResize(num_threads_, max_batch_size_);
   const auto &input = ws.template InputRef<GPUBackend>(0);
   output_desc.resize(1);
   DetermineTransformation(ws);
--- a/dali/operators/image/color/color_twist.h
+++ b/dali/operators/image/color/color_twist.h
@@ -97,57 +97,48 @@ class ColorTwistBase : public Operator<Backend> {
 
  protected:
   explicit ColorTwistBase(const OpSpec &spec)
-          : Operator<Backend>(spec)
-          , output_type_arg_(spec.GetArgument<DALIDataType>(color::kOutputType))
-          , output_type_(DALI_NO_TYPE) {
-    if (std::is_same<Backend, GPUBackend>::value) {
-      kernel_manager_.Resize(1, 1);
-    } else {
-      kernel_manager_.Resize(num_threads_, batch_size_);
-    }
-  }
-
+      : Operator<Backend>(spec),
+        output_type_arg_(spec.GetArgument<DALIDataType>(color::kOutputType)),
+        output_type_(DALI_NO_TYPE) {}
 
   bool CanInferOutputs() const override {
     return true;
   }
 
   void AcquireArguments(const workspace_t<Backend> &ws) {
+    auto curr_batch_size = ws.GetInputBatchSize(0);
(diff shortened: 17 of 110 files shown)


szalpal:
!build

szalpal:
!build

szalpal:
!build

szalpal:
!build

szalpal:
!build

szalpal:
!build

szalpal:
!build

szalpal:
!build

dali/operators/generic/flip.h
```diff
@@ -58,23 +58,29 @@ class Flip: public Operator<Backend> {
     return this->spec_.template GetArgument<int>("depthwise", &ws, idx);
   }
 
-  std::vector<int> GetHorizontal(const ArgumentWorkspace &ws) {
+  std::vector<int> GetHorizontal(const workspace_t<Backend> &ws) {
+    auto curr_batch_size_ = ws.template get_batch_size<Backend>(0);
```
szalpal, line 62: `curr_batch_size_` can be an rval here

dali/operators/bbox/bb_flip.h
```diff
@@ -48,15 +48,15 @@ class BbFlip<CPUBackend> : public Operator<CPUBackend> {
...
-  }
+//  template <typename TensorDataType>
+//  void ExtendScalarToTensor(std::string argument_name, const OpSpec &spec,
+//                            Tensor<CPUBackend> *tensor) {
+//    tensor->Resize({curr_batch_size_});
+//    for (int i = 0; i < curr_batch_size_; i++) {
+//      tensor->mutable_data<TensorDataType>()[i] =
+//          spec.GetArgument<TensorDataType>(argument_name);
+//    }
+//  }
```
szalpal, line 59: Should I remove it?

szalpal:
!build

szalpal:
!build

dali/operators/audio/nonsilence_op.h
```diff
@@ -209,9 +210,10 @@ class NonsilenceOperatorCpu : public NonsilenceOperator<CPUBackend> {
     const auto &input = ws.template InputRef<CPUBackend>(0);
     auto &output_begin = ws.OutputRef<CPUBackend>(0);
     auto &output_length = ws.OutputRef<CPUBackend>(1);
+    auto curr_batch_size_ = ws.get_batch_size<CPUBackend>(0);
```
mzient, line 213:
```suggestion
    auto curr_batch_size = ws.get_batch_size<CPUBackend>(0);
```

dali/operators/audio/nonsilence_op.h
```diff
@@ -209,9 +210,10 @@ class NonsilenceOperatorCpu : public NonsilenceOperator<CPUBackend> {
...
     auto &tp = ws.GetThreadPool();
     auto in_shape = input.shape();
-    for (int sample_id = 0; sample_id < batch_size_; sample_id++) {
+    for (int sample_id = 0; sample_id < curr_batch_size_; sample_id++) {
```
mzient, line 216:
```suggestion
    for (int sample_id = 0; sample_id < batch_size; sample_id++) {
```

dali/operators/audio/preemphasis_filter_op.cu
```diff
@@ -69,17 +69,18 @@ void PreemphasisFilterGPU::RunImplTyped(workspace_t<GPUBackend> &ws) {
   using SampleDesc = detail::SampleDescriptor<OutputType, InputType>;
   const auto &input = ws.InputRef<GPUBackend>(0);
   auto &output = ws.OutputRef<GPUBackend>(0);
+  auto curr_batch_size_ = ws.get_batch_size<GPUBackend>(0);
```
mzient, line 72: This is not a member! Also, the "curr_" prefix serves very little purpose but to make the code _unnecessarily_ verbose.
```suggestion
  auto batch_size = ws.get_batch_size<GPUBackend>(0);
```

dali/operators/bbox/bb_flip.h
mzient: Yes, please.

dali/operators/decoder/cache/cached_decoder_impl.cc
```diff
@@ -39,7 +39,7 @@ CachedDecoderImpl::CachedDecoderImpl(const OpSpec& spec)
         device_id_, cache_type, cache_size, cache_debug, cache_threshold);
 
       use_batch_copy_kernel_ = spec.GetArgument<bool>("cache_batch_copy");
-      auto batch_size = spec.GetArgument<int>("batch_size");
+      auto batch_size = spec.GetArgument<int>("max_batch_size");
```
mzient, line 42: Now that's a bug, isn't it?

dali/operators/decoder/host/fused/host_decoder_random_crop_test.cc
```diff
@@ -25,7 +25,7 @@ class ImageDecoderRandomCropTest_CPU : public DecodeTestBase<ImgType> {
   ImageDecoderRandomCropTest_CPU()
     : random_crop_attr(
       OpSpec("RandomCropAttr")
-        .AddArg("batch_size", this->batch_size_)
+        .AddArg("max_batch_size", this->batch_size_)
```
mzient, line 28: Perhaps it would be better to rename this field, too?

dali/operators/generic/erase/erase.cu
```diff
@@ -42,14 +42,12 @@ class EraseImplGpu : public OpImplBase<GPUBackend> {
  public:
   using EraseKernel = kernels::EraseGpu<T, Dims, channel_dim>;
 
-  explicit EraseImplGpu(const OpSpec &spec)
-  : spec_(spec)
-  , batch_size_(spec.GetArgument<int>("batch_size")) {
-      kmgr_.Resize<EraseKernel>(1, 1);
+  explicit EraseImplGpu(const OpSpec &spec) : spec_(spec) {
+    curr_batch_size_ = spec.GetArgument<int>("max_batch_size");  // TODO(mszolucha) remove
```
mzient, line 46: Please remove, then.

dali/operators/generic/erase/erase.cu
```diff
@@ -79,18 +77,19 @@ class EraseImplGpu : public OpImplBase<GPUBackend> {
 
   void AcquireArgs(const DeviceWorkspace &ws, TensorListShape<> in_shape,
                    TensorLayout in_layout) {
+    auto curr_batch_size_ = ws.get_batch_size<GPUBackend>(0);
```
mzient, line 80:
```suggestion
    auto curr_batch_size = ws.get_batch_size<GPUBackend>(0);
```
This is not a member!
Remove the member with the same name.

dali/operators/generic/erase/erase.cu
```diff
@@ -101,8 +100,8 @@ class EraseImplGpu : public OpImplBase<GPUBackend> {
   }
 
  private:
+  int curr_batch_size_ = -1;  // TODO(mszolucha) remove
```
mzient, line 103: Remove!

dali/operators/generic/flip.h
mzient, line 62: Not a member!
```suggestion
    auto curr_batch_size = ws.template get_batch_size<Backend>(0);
```

dali/operators/generic/flip.h
```diff
@@ -58,23 +58,29 @@ class Flip: public Operator<Backend> {
...
     std::vector<int> result;
-    OperatorBase::GetPerSampleArgument(result, "horizontal", ws);
+    OperatorBase::GetPerSampleArgument(result, "horizontal", ws, curr_batch_size_);
     return result;
   }
 
-  std::vector<int> GetVertical(const ArgumentWorkspace &ws) {
+  std::vector<int> GetVertical(const workspace_t<Backend> &ws) {
+    auto curr_batch_size_ = ws.template get_batch_size<Backend>(0);
```
mzient, line 69:
```suggestion
    auto batch_size = ws.template get_batch_size<Backend>(0);
```

dali/operators/generic/flip.h
```diff
@@ -58,23 +58,29 @@ class Flip: public Operator<Backend> {
...
     std::vector<int> result;
-    OperatorBase::GetPerSampleArgument(result, "vertical", ws);
+    OperatorBase::GetPerSampleArgument(result, "vertical", ws, curr_batch_size_);
     return result;
   }
 
-  std::vector<int> GetDepthwise(const ArgumentWorkspace &ws) {
+  std::vector<int> GetDepthwise(const workspace_t<Backend> &ws) {
+    auto curr_batch_size_ = ws.template get_batch_size<Backend>(0);
```
mzient, line 76:
```suggestion
    auto batch_size = ws.template get_batch_size<Backend>(0);
```

dali/operators/generic/one_hot.cc
```diff
@@ -76,6 +76,7 @@ class OneHotCPU : public OneHot<CPUBackend> {
 void OneHotCPU::RunImpl(workspace_t<CPUBackend> &ws) {
   const auto &input = ws.template InputRef<CPUBackend>(0);
   auto &output = ws.template OutputRef<CPUBackend>(0);
+  auto curr_batch_size_ = ws.get_batch_size<CPUBackend>(0);
```
mzient, line 79:
```suggestion
  auto batch_size = ws.get_batch_size<CPUBackend>(0);
```

mzient (changes requested):
1.Please remove the obsolete members.
2. Remove the trailing underscore from non-member variables.
3. Use `batch_size` rather than the longer `curr_batch_size`. Unless you're dealing with current and previous/next, the curr_ prefix has no value.

Probably more to come, but these points need addressing.

szalpal:
> 1.Please remove the obsolete members.

Took care about (almost) all TODOs I've left in the PR. Sorry for leaving them there

> 2. Remove the trailing underscore from non-member variables.

Please read the description

> 3. Use `batch_size` rather than the longer `curr_batch_size`. Unless you're dealing with current and previous/next, the curr_ prefix has no value.

I have particular reasons for having it `curr_batch_size`. Let's discuss it offline

dali/operators/decoder/cache/cached_decoder_impl.cc
szalpal: As I wrote in the description, I'm not reworking decoders and random operators in this PR.

szalpal:
!build

szalpal:
!build

dali/pipeline/workspace/workspace.h
```diff
@@ -145,12 +145,38 @@ class WorkspaceBase : public ArgumentWorkspace {
...
+    return input.shape();
+  }
+
+  /**
+   * Returns batch size for given input
+   */
+  template <typename Backend>
+  int get_batch_size(int input_idx) const {
+    return NumInput() == 0 ? batch_size_ : GetInputShape<Backend>(input_idx).num_samples();
+  }
```
mzient, line 175: I find this function disturbing:
1. It returns something else than the argument says (when there are no inputs it returns something else) - perhaps there should be a parameterless overload.
2. It returns _input_ batch size, not batch size - and the two will differ quite soon (conditional ops?). Let's plan a few steps ahead and not make ourselves suffer lengthy rebuilds, API changes and conflicts. Now the function is not used yet by any upstream code and it's the right time to make the API future-proof.
3. Instead of manually providing backend, we can check which backend we have - this will save us a lot of template disambiguations: `ws.template get_batch_size<Backend>()` -> `ws.get_batch_size()`

Let's discuss it tomorrow.

szalpal:
!build

dali/pipeline/operator/common.h
```diff
@@ -137,19 +132,16 @@ void GetGeneralizedArg(span<T> result, const std::string name, int sample_idx, c
...
-                                         const OpSpec &spec,
+std::pair<int, int> GetShapeLikeArgument(std::vector<ExtentType> &out_shape, const OpSpec &spec,
                                          const std::string &argument_name,
-                                         const ArgumentWorkspace &ws,
-                                         int ndim = -1,
-                                         int batch_size = -1 /* -1 = get from "batch_size" arg */) {
-  if (batch_size < 0)
-    batch_size = spec.GetArgument<int>("batch_size");
+                                         const ArgumentWorkspace &ws, int ndim = -1,
+                                         int batch_size = -1) {
```
mzient, line 138: Weird grouping of arguments. Either keep them one per line or use logical grouping, like:
out_shape  // output
spec, arg_name, ws,  // source of the argument
ndim, batch_size  // shape

dali/pipeline/operator/common.h
```diff
@@ -202,14 +194,11 @@ std::pair<int, int> GetShapeLikeArgument(std::vector<ExtentType> &out_shape,
...
-                                     const OpSpec &spec,
-                                     const std::string &argument_name,
-                                     const ArgumentWorkspace &ws,
-                                     int ndim = out_ndim,
-                                     int batch_size = -1 /* -1 = get from "batch_size" arg */) {
-  auto ret = GetShapeLikeArgument<ArgumentType>(
-    out_tls.shapes, spec, argument_name, ws, ndim, batch_size);
+std::pair<int, int> GetShapeArgument(TensorListShape<out_ndim> &out_tls, const OpSpec &spec,
+                                     const std::string &argument_name, const ArgumentWorkspace &ws,
+                                     int ndim = out_ndim, int batch_size = -1) {
```
mzient, line 199: Same here.

dali/pipeline/operator/operator.cc
```diff
@@ -16,6 +16,16 @@
...
 
+template <typename Backend>
+void OperatorBase::EnforceConstantBatchSize(const workspace_t<Backend> &ws) const {
+  DALI_ENFORCE(ws.get_input_batch_size(0) == max_batch_size_,
+               "Batch size has to be constant throughout the processing");
+}
+
+template void OperatorBase::EnforceConstantBatchSize<CPUBackend>(const workspace_t<CPUBackend>&w) const;  // NOLINT
+template void OperatorBase::EnforceConstantBatchSize<GPUBackend>(const workspace_t<GPUBackend>&w) const;  // NOLINT
+template void OperatorBase::EnforceConstantBatchSize<MixedBackend>(const workspace_t<MixedBackend>&w) const;  // NOLINT
```
mzient, line 27: This will most likely not link with clang 6 (some DLL_PUBLIC would likely be required). Why not just keep it in the header?

dali/test/python/test_external_source_impl.py
```diff
@@ -416,26 +416,6 @@ def iter_setup(self):
...
-            return [self.batch]
-
-        def iter_setup(self):
-            batch = datapy.zeros([self.external_s_size_,4,5])
-            self.feed_input(self.batch, batch)
-
-    batch_size = 3
-    pipe = ExternalSourcePipeline(batch_size, batch_size - 1, 3, 0)
-    pipe.build()
-    assert_raises(RuntimeError, pipe.run)
```
mzient, line 438: Why was it removed? I think this still holds.

dali/pipeline/workspace/workspace.h
```diff
@@ -145,12 +145,54 @@ class WorkspaceBase : public ArgumentWorkspace {
...
+  template <typename Backend>
+  auto GetInputShape(int input_idx) const {
+    const auto& input = *InputHandle(input_idx, Backend{});
+    return input.shape();
+  }
+
+  /**
+   * Returns batch size for given input
+   */
+  int get_input_batch_size(int input_idx) const {
```
mzient, line 172: It's a nitpick, but why are these functions in snake_case while others in this class are in PascalCase? Not that I love PascalCase or sth, but still, these functions are not that different.

dali/pipeline/operator/builtin/external_source.h
```diff
@@ -377,13 +377,14 @@ class ExternalSource : public Operator<Backend> {
     bool is_gpu_src = std::is_same<SrcBackend, GPUBackend>::value;
     bool is_gpu_dst = std::is_same<Backend, GPUBackend>::value;
     if (is_gpu_src && !is_gpu_dst) {
-      DALI_WARN("Warning: Loading GPU-originated data into CPU "
-                "ExternalSource operator is discouraged and might be inefficient.");
+      DALI_WARN(
```
mzient, line 380: There were no meaningful changes made to this code. Why reformat?

dali/pipeline/operator/builtin/external_source.h
```diff
@@ -377,13 +377,14 @@ class ExternalSource : public Operator<Backend> {
...
+          "and might be inefficient.");
     }
-    DALI_ENFORCE(OperatorBase::batch_size_ == static_cast<int>(batch.ntensor()),
-                 make_string("Data list provided to ExternalSource needs to have batch_size = ",
-                             OperatorBase::batch_size_, " length, found ", batch.ntensor(),
-                             " samples."));
+    DALI_ENFORCE(
+        OperatorBase::max_batch_size_ >= static_cast<int>(batch.ntensor()),
+        make_string("Data list provided to ExternalSource needs to have batch_size <= ",
+                    OperatorBase::max_batch_size_, ", found ", batch.ntensor(), " samples."));
```
mzient, line 387: Likewise - you could have changed 3 letters, not 3 lines. Can you tone down your formatting tool a bit?

dali/pipeline/operator/common.h
```diff
@@ -45,24 +45,19 @@ inline void GetSingleOrRepeatedArg(const OpSpec &spec, vector<T> &result,
...
-                          const OpSpec &spec,
-                          const ArgumentWorkspace &ws,
-                          int batch_size = -1 /* -1 = get from "batch_size" arg */) {
-  if (batch_size < 0)
-    batch_size = spec.GetArgument<int>("batch_size");
-
+void GetPerSampleArgument(std::vector<T> &output, const std::string &argument_name,
+                          const OpSpec &spec, const ArgumentWorkspace &ws, int batch_size) {
+  DALI_ENFORCE(batch_size > 0,
+               make_string("Invalid batch size. Expected positive, actual: ", batch_size));
```
mzient, line 51: With random application, we must learn to work with empty batches.
I think this should be:
```suggestion
  DALI_ENFORCE(batch_size >= 0);
```
as I can see no way in which the user's code can produce a _negative_ batch size, much less cause an operator to request it.

dali/pipeline/operator/common.h
```diff
@@ -45,24 +45,19 @@ inline void GetSingleOrRepeatedArg(const OpSpec &spec, vector<T> &result,
...
     decltype(auto) shape = arg.shape();
     int N = shape.num_samples();
     if (N == 1) {
       bool is_valid_shape = volume(shape.tensor_shape(0)) == batch_size;
 
-      DALI_ENFORCE(is_valid_shape,
-        make_string("`", argument_name, "` must be a 1xN or Nx1 (N = ", batch_size,
-                    ") tensor list. Got: ", shape));
+      DALI_ENFORCE(is_valid_shape, make_string("`", argument_name, "` must be a 1xN or Nx1 (N = ",
+                                               batch_size, ") tensor list. Got: ", shape));
```
mzient, line 60: meh.

dali/pipeline/operator/common.h
```diff
@@ -137,19 +132,16 @@ void GetGeneralizedArg(span<T> result, const std::string name, int sample_idx, c
...
+std::pair<int, int> GetShapeLikeArgument(std::vector<ExtentType> &out_shape, const OpSpec &spec,
                                          const std::string &argument_name,
-                                         const ArgumentWorkspace &ws,
-                                         int ndim = -1,
-                                         int batch_size = -1 /* -1 = get from "batch_size" arg */) {
-  if (batch_size < 0)
-    batch_size = spec.GetArgument<int>("batch_size");
+                                         const ArgumentWorkspace &ws, int ndim = -1,
+                                         int batch_size = -1) {
+  DALI_ENFORCE(batch_size > 0, "Negative (default) batch_size is not supported anymore");
```
mzient, line 139: Likewise - it's not a run-time condition, it's a _coding error_:
```suggestion
  assert(batch_size >= 0);
```

dali/python/nvidia/dali/fn.py
```diff
@@ -76,6 +76,7 @@ def op_wrapper(*inputs, **arguments):
         return op_class(**init_args)(*inputs, **call_args)
 
     op_wrapper.__name__ = wrapper_name
+    op_wrapper.__qualname__ = wrapper_name
```
mzient, line 79: What is it?

dali_tf_plugin/daliop.cc
```diff
@@ -101,7 +101,7 @@ class DaliOp : public tf::OpKernel {
 
     int num_threads;
     int device_id;
-    int batch_size;
+    int max_batch_size;
```
mzient, line 104: @awolant Is it possible to have a variable batch size in Tensorflow?

dali/pipeline/workspace/workspace.h
szalpal: I've written in the description, `GetBatchSize` was taken and I needed unique name to grep through it just in case. I guess `GetInputBatchSize` is unique in the code base. I can change it if you want, but this I'll do after all the green ticks necessary. I really don't like to yet again rsync me whole code base and rebuild everything, it's just a waste of time ;)

dali/python/nvidia/dali/fn.py
szalpal: LMGIFY ;)
https://stackoverflow.com/questions/58108488/what-is-qualname-in-python

dali/test/python/test_external_source_impl.py
szalpal: This test check, if an exception is thrown when external source is fed with different batch sizes. We do not really want this scenario to throw an exception

dali/pipeline/operator/operator.cc
szalpal: This code passes the CI, so I guess we're covered. Basically, I need an "easy" way to turn this on/off, cause I'm testing whatever is here with a script that is outside this PR. I don't want to rebuilt whole DALI, which is going to happen if that's in header file...

mzient: The changes to TensorVector also passed CI, but didn't pass debug build in clang :(

mzient (commented):
Still some issues, but no show stoppers.

szalpal:
!build

szalpal:
!build

szalpal:
!build

dali/python/nvidia/dali/fn.py
mzient: OK, thanks for explanation.

dali/test/python/test_external_source_impl.py
mzient: OK, whatever, I'm not particularly attached to this test, however, the exception should still be thrown until we're reasonably sure that feeding batches of varying length actually works.

dali/operators/audio/nonsilence_op.cc
```diff
@@ -67,10 +67,11 @@ bool NonsilenceOperatorCpu::SetupImpl(std::vector<OutputDesc> &output_desc,
   TypeInfo output_type;
   output_type.SetType<int>(TypeTable::GetTypeID<int>());
   TensorShape<> scalar_shape = {};
+  auto curr_batch_size = ws.get_input_batch_size(0);
```
klecki, line 70: As I mentioned by this function, it would be nice to have a variant without arguments, that would return the common batch size and enforce that all inputs use the same one. With 1 input it would do the same as this call, but we could drop the 0 - most of our ops assume equal batches in all inputs.

klecki (commented):
Just started, I have suggestion to semantics of `get_input_batch`.

I will go over the operators that you changed next week.

dali/python/nvidia/dali/external_source.py
```diff
@@ -18,9 +18,10 @@ def _get_batch_shape(data):
 
 def _check_data_batch(data, batch_size, layout):
     shape, uniform = _get_batch_shape(data)
-    if len(shape) != batch_size:
+    if len(shape) > batch_size:
```
mzient, line 21: I think this change should work until we enable it for the users (and add tests).

dali/operators/audio/nonsilence_op.cc
szalpal: In an offline discussion we decided, that it's not necessary to add `GetInputBatchSize()` overload

dali/pipeline/workspace/sample_workspace.h
```diff
@@ -72,6 +72,12 @@ class DLL_PUBLIC SampleWorkspace : public WorkspaceBase<SampleInputType, SampleO
   template <typename Backend>
   DLL_PUBLIC Tensor<Backend>& Output(int idx);
 
+  int get_input_batch_size(int) const {
+    DALI_FAIL(
+        "Impossible function: Sample workspace is not aware, that there exists such thing as an "
+        "input batch");
+  }
```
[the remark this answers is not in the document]
szalpal: The L0 and L1 tests passed, as well as tests in #2481 pass. Additionally, I manually verified, that none of the SampleWorkspace-based ops have a notion of a batch size. Therefore I guess we're covered, but I encourage you to verify it by yourself in the review, to have a separate pair of eyes looking here ;)

dali/benchmark/color_twist_bench.cc
```diff
@@ -36,7 +36,7 @@ BENCHMARK_DEFINE_F(OperatorBench, ColorTwistGPU)(benchmark::State& st) {
   this->RunGPU<uint8_t>(
     st,
     OpSpec("ColorTwist")
-      .AddArg("batch_size", batch_size)
```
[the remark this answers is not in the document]
szalpal: I believe we do need: when I removed this line, `dali_benchmark` crashed on `max_batch_size>0` assert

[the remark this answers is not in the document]
klecki: Oh, I just realized, that it doesn't construct a pipe, it just runs the operator, so yeah, it needs to be there, sorry for the confusion.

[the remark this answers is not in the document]
JanuszL: Was it expected or there could be some error you in the logic you just hidden by this change in the test?

dali/operators/generic/constant.h
```diff
@@ -80,9 +80,9 @@ class Constant : public Operator<Backend> {
 
   bool SetupImpl(std::vector<OutputDesc> &output_desc, const Workspace &ws) override {
     output_desc.resize(1);
+    auto curr_batch_size = ws.GetRequestedBatchSize(0);
     if (output_shape_.empty()) {
```
JanuszL, line 84:
```suggestion
    if (output_shape_.empty() || output_shape_ .size() < curr_batch_size ) {
```
Because if you the rest iteration has batch size 1 and the next one 2, you will end up with `output_shape_ ` os size 1 once and for all.

dali/operators/generic/flip.h
```diff
@@ -58,23 +58,29 @@ class Flip: public Operator<Backend> {
     return this->spec_.template GetArgument<int>("depthwise", &ws, idx);
   }
 
-  std::vector<int> GetHorizontal(const ArgumentWorkspace &ws) {
+  std::vector<int> GetHorizontal(const workspace_t<Backend> &ws) {
+    auto curr_batch_size = ws.GetInputBatchSize(0);
```
JanuszL, line 62: Why not make `curr_batch_size` a function argument and ask the caller to provide it instead of querying it every time here?

dali/operators/geometry/affine_transforms/transform_base_op.h
```diff
@@ -280,15 +281,14 @@ class TransformBaseOp : public Operator<Backend> {
   }
 
  protected:
+  USE_OPERATOR_MEMBERS();
```
JanuszL, line 284: Do we need that here?

dali/operators/geometry/coord_flip.cu
```diff
@@ -68,10 +70,10 @@ class CoordFlipGPU : public CoordFlip<GPUBackend> {
 void CoordFlipGPU::RunImpl(workspace_t<GPUBackend> &ws) {
   const auto &input = ws.InputRef<GPUBackend>(0);
   auto &output = ws.OutputRef<GPUBackend>(0);
+  auto curr_batch_size = ws.GetInputBatchSize(0);
 
   sample_descs_.clear();
-  sample_descs_.reserve(batch_size_);
```
JanuszL, line 73: I would keep this:
```
sample_descs_.reserve(max_batch_size_);
```
here and remove L53.

dali/operators/image/color/brightness_contrast.cu
```diff
@@ -29,6 +29,8 @@ DALI_REGISTER_OPERATOR(BrightnessContrast, BrightnessContrastGpu, GPU)
 
 bool BrightnessContrastGpu::SetupImpl(std::vector<OutputDesc> &output_desc,
                                       const workspace_t<GPUBackend> &ws) {
+  auto curr_batch_size = ws.GetInputBatchSize(0);
```
JanuszL, line 32: It makes a false impression that it does anything meaningful for GPU, while it s always `kernel_manager_.Resize(1, 1);`.

dali/operators/image/color/color_twist.h
```diff
@@ -156,6 +147,13 @@ class ColorTwistBase : public Operator<Backend> {
     }
   }
 
+  void KMagResize(int num_threads, int batch_size) {
```
JanuszL, line 150: After all maybe we can make this some kind of utility as this pattern appears in more than one place.

dali/operators/image/crop/crop_attr.h
```diff
@@ -34,9 +34,8 @@ namespace dali {
 class CropAttr {
  public:
   static constexpr int kNoCrop = -1;
-  explicit inline CropAttr(const OpSpec &spec)
-    : spec__(spec)
-    , batch_size__(spec__.GetArgument<int>("batch_size")) {
+  explicit inline CropAttr(const OpSpec& spec) : spec__(spec) {
+    auto max_batch_size = spec__.GetArgument<int>("max_batch_size");
```
JanuszL, line 38: Maybe you can make it a member and obtain only once and reuse in `ProcessArguments`.

dali/operators/image/paste/paste.h
```diff
@@ -33,27 +33,25 @@ class Paste : public Operator<Backend> {
   // 6 values: in_H, in_W, out_H, out_W, paste_y, paste_x
   static const int NUM_INDICES = 6;
 
-  explicit inline Paste(const OpSpec &spec) :
-    Operator<Backend>(spec),
-    C_(spec.GetArgument<int>("n_channels")) {
+  explicit Paste(const OpSpec &spec)
+      : Operator<Backend>(spec), C_(spec.GetArgument<int>("n_channels")) {
```
JanuszL, line 37: Maybe you can make `curr_batch_size` a member?

dali/operators/reader/reader_op.h
```diff
@@ -308,15 +308,15 @@ class DataReader : public Operator<Backend> {
...
+    return curr_batch_producer_ == curr_batch_consumer_ && consumer_cycle_ == producer_cycle_;
   }
 
   bool IsPrefetchQueueFull() {
-    return curr_batch_producer_ == curr_batch_consumer_
-           && consumer_cycle_ != producer_cycle_;
+    return curr_batch_producer_ == curr_batch_consumer_ && consumer_cycle_ != producer_cycle_;
   }
 
+  USE_OPERATOR_MEMBERS();
```
JanuszL, line 318: Why we need that here?

dali/operators/ssd/box_encoder.cuh
```diff
@@ -77,6 +78,7 @@ class BoxEncoder<GPUBackend> : public Operator<GPUBackend> {
  private:
   static constexpr int kBoxesOutputDim = 2;
   static constexpr int kLabelsOutputDim = 1;
+  int curr_refactor_batch_size_;
```
JanuszL, line 81: Any better name?

dali/pipeline/data/tensor_vector.h
```diff
@@ -123,6 +123,12 @@ class DLL_PUBLIC TensorVector {
 
   DLL_PUBLIC void Resize(const TensorListShape<> &new_shape, const TypeInfo &new_type);
 
+  /**
+   * Change the number of tensors in TensorVector.
+   * @param new_size
+   */
+  void SetSize(int new_size);
```
JanuszL, line 130: Maybe it should just accept an unsigned value and prevent from the negative one?

dali/pipeline/operator/common.h
```diff
@@ -137,19 +132,16 @@ void GetGeneralizedArg(span<T> result, const std::string name, int sample_idx, c
...
+std::pair<int, int> GetShapeLikeArgument(std::vector<ExtentType> &out_shape, const OpSpec &spec,
                                          const std::string &argument_name,
-                                         const ArgumentWorkspace &ws,
-                                         int ndim = -1,
-                                         int batch_size = -1 /* -1 = get from "batch_size" arg */) {
-  if (batch_size < 0)
-    batch_size = spec.GetArgument<int>("batch_size");
+                                         const ArgumentWorkspace &ws, int ndim = -1,
+                                         int batch_size = -1) {
+  assert(batch_size >= 0);
```
JanuszL, line 139: Why not DALI_ENFORCE like in L50.

dali/pipeline/operator/operator.cc
```diff
@@ -16,6 +16,18 @@
 
 namespace dali {
 
+template <typename Backend>
+void OperatorBase::EnforceConstantBatchSize(const workspace_t<Backend> &ws) const {
+  if (ws.NumInput() > 0) {
+    DALI_ENFORCE(ws.GetInputBatchSize(0) == max_batch_size_,
```
JanuszL, line 22: Maybe it should iterate over all inputs and check all of it?

dali/pipeline/pipeline.h
```diff
@@ -394,10 +394,13 @@ class DLL_PUBLIC Pipeline {
   DLL_PUBLIC void SaveGraphToDotFile(const std::string &filename, bool show_tensors = false,
                                      bool show_ids = false, bool use_colors = false);
 
+  /// @{
   /**
-   * @brief Returns the batch size that will be produced by the pipeline.
+   * @brief Returns the maximum batch size that can be produced by the Pipeline
```
JanuszL, line 399:
```suggestion
   * @brief Returns the maximum batch size that can be processed by the Pipeline
```

dali/pipeline/workspace/sample_workspace.h
```diff
@@ -72,6 +72,12 @@ class DLL_PUBLIC SampleWorkspace : public WorkspaceBase<SampleInputType, SampleO
   template <typename Backend>
   DLL_PUBLIC Tensor<Backend>& Output(int idx);
 
+  int GetInputBatchSize(int) const {
+    DALI_FAIL(
+        "Impossible function: Sample workspace is not aware, that there exists such thing as an "
```
JanuszL, line 77:
```suggestion
        "Not implemented: Sample workspace is not aware, that there exists such thing as an "
``?

File: dali/python/nvidia/dali/pipeline.py [hunk 46]
```diff
@@ -51,10 +51,13 @@ class Pipeline(object):
...
 ----------
 `batch_size` : int, optional, default = -1
-    Batch size of the pipeline. Negative values for this parameter
+    Maximum batch size of the pipeline. Negative values for this parameter
     are invalid - the default value may only be used with
     serialized pipeline (the value stored in serialized pipeline
-    is used instead).
+    is used instead). In most cases, the actual batch size of the pipeline
+    will be equal to the maximum one. Running DALI pipeline with batch size
+    smaller than that is supported, but currently limited. Batch size
```
JanuszL, line 59:
```suggestion
    smaller than that is not supported. Batch size
```

dali/python/nvidia/dali/pipeline.py
```diff
@@ -51,10 +51,13 @@ class Pipeline(object):
...
+    might change from iteration to iteration of DALI pipeline.
```
JanuszL, line 60:
```suggestion
    smaller than that is not supported.
```

dali/test/python/test_external_source_impl.py
JanuszL: I would keep this  test in this PR, and remove when it actually doesn't apply anymore.

dali/test/python/test_external_source_impl.py
```diff
@@ -507,28 +487,6 @@ def iter_setup(self):
     pipe.build()
     assert_raises(RuntimeError, pipe.run)
 
-def test_external_source_fail_list():
```
JanuszL, line 510: As above.

dali/test/python/test_utils.py
```diff
@@ -112,11 +112,30 @@ def is_error(mean_err, max_err, eps, max_allowed_error):
...
+            shape = batch.as_array().shape if isinstance(batch,
+                                                         dali.backend.TensorListCPU) else batch.shape
+            assert shape[0] == batch_size, \
+                "Incorrect batch size. Expected: {}, actual: {}".format(batch_size, shape[0])
+        else:
+            tested_batch_size = 0
+            for sample in batch:  # not dense TensorListCPU has no other way of checking batch size
+                tested_batch_size += 1
+            assert tested_batch_size == batch_size, \
+                "Incorrect batch size. Expected: {}, actual: {}".format(batch_size, tested_batch_size)
```
JanuszL, line 128: why not:
```suggestion
        if  isinstance(batch, dali.backend.TensorListCPU) or isinstance(batch, dali.backend.TensorListGPU):
            tested_batch_size = len(batch)
        else:
            tested_batch_size = batch.shape[0]
            assert tested_batch_size == batch_size, \
                "Incorrect batch size. Expected: {}, actual: {}".format(batch_size, tested_batch_size)
```

dali/test/python/test_utils.py
```diff
@@ -177,22 +195,47 @@ def compare_pipelines(
...
                 current_expected_layout = expected_layout[i]
             else:
                 current_expected_layout = expected_layout
             check_batch(out1_data, out2_data, batch_size, eps, max_allowed_error,
                         expected_layout=current_expected_layout, compare_layouts=compare_layouts)
-    print("OK: ({} iterations)".format(N_iterations))
+    print("OK: ({} iterations)".format(n_iterations))
+
+
+def compare_pipelines(pipe1, pipe2, batch_size, N_iterations, eps=1e-07, max_allowed_error=None,
```
JanuszL, line 216: Can't you just add `pipe1.build()` to the original `compare_pipelines` and get rid of `compare_built_pipelines`?
If `build` has been called already pipeline will just treat it as a noop.

dali_tf_plugin/daliop.cc
awolant: Yes, there are ways to write a graph that can have variable batch size. You can do something like:
```
inputs = tf.placeholder(tf.float32, [None, 32, 32, 3])
batch_size = tf.shape(inputs[0])[0] 
```
Here `batch_size` is a tensor and the value will be known during the execution only.

For newer APIs (like tf.data, and TF 2.x in general) it's not so easy to do, but the `daliop` integrates with TF 1.x and graph approach, so I think it is reasonable to make it possible from DALI side.

dali/operators/ssd/box_encoder.cuh
szalpal: (changed)

dali/operators/image/color/color_twist.h
JanuszL: @mzient ?

dali/pipeline/workspace/sample_workspace.h
JanuszL: Can you make it a compile time assert if anyone uses that function?

dali/operators/image/color/brightness_contrast.cu
JanuszL: @mzient ?

JanuszL (approved):

dali/pipeline/workspace/sample_workspace.h
mzient: I still don't like the idea of having GetInputBatchSize at all, but I won't sweat it.
Regarding this function - why is it here? If it's wrong to call it - then it should not be called at all. In that case, why is it here at all?

dali/operators/image/color/brightness_contrast.h
```diff
@@ -95,15 +90,21 @@ class BrightnessContrastOp : public Operator<Backend> {
...
 
     input_type_ = ws.template InputRef<Backend>(0).type().id();
-    output_type_ =
-        output_type_arg_ != DALI_NO_TYPE
-        ? output_type_arg_
-        : input_type_;
+    output_type_ = output_type_arg_ != DALI_NO_TYPE ? output_type_arg_ : input_type_;
+  }
+
+  void KMagResize(int num_threads, int batch_size) {
```
mzient, line 102:
```suggestion
  void KMgrResize(int num_threads, int batch_size) {
```
`Mag` is not an abbreviation of "manager".

dali/operators/image/color/color_twist.h
mzient: It's not always 1, 1. Neither is it always num_threads, batch_size for CPU - and if it is, then it should be `(num_threads, max_batch_size_)`.

dali/operators/image/color/brightness_contrast.cu
mzient: 1. It should be KMgrResize - if we indeed need a function for that.
2. It should resize to max_batch_size_, not to current - otherwise you'll keep creating and destroying kernel instances - therefore, it can just stay in the constructor.

dali/pipeline/pipeline.h
```diff
@@ -394,10 +394,13 @@ class DLL_PUBLIC Pipeline {
...
                                      bool show_ids = false, bool use_colors = false);
 
+  /// @{
   /**
-   * @brief Returns the batch size that will be produced by the pipeline.
+   * @brief Returns the maximum batch size that can be processed by the Pipeline
    */
-  DLL_PUBLIC inline int batch_size() const { return batch_size_; }
+  DLL_PUBLIC inline int batch_size() const { return max_batch_size_; }
+  DLL_PUBLIC inline int max_batch_size() const { return max_batch_size_; }
```
klecki, line 402: Pipeline::SerializeToProtobuf() should probably use this call (`max_batch_size`) for `pipe.set_batch_size(...)` instead of `this->batch_size()`.

dali/pipeline/pipeline.h
```diff
@@ -394,10 +394,13 @@ class DLL_PUBLIC Pipeline {
   DLL_PUBLIC void SaveGraphToDotFile(const std::string &filename, bool show_tensors = false,
                                      bool show_ids = false, bool use_colors = false);
 
+  /// @{
   /**
-   * @brief Returns the batch size that will be produced by the pipeline.
+   * @brief Returns the maximum batch size that can be processed by the Pipeline
    */
-  DLL_PUBLIC inline int batch_size() const { return batch_size_; }
+  DLL_PUBLIC inline int batch_size() const { return max_batch_size_; }
```
klecki, line 401: I can see that this is mostly used in the c_api for obtaining shapes in the SetExternalInput calls. Do you intend to adjust it? It should probably start relying on the inputs to provide the batch size instead of what is stored in the pipeline (and return an error for now).

dali/operators/image/crop/bbox_crop.cc
```diff
@@ -43,7 +43,7 @@ void CollectShape(std::vector<TensorShape<>> &v,
                   const OpSpec& spec,
                   const workspace_t<CPUBackend>& ws,
                   int ndim) {
-  int batch_size = spec.GetArgument<int>("batch_size");
+  int batch_size = spec.GetArgument<int>("max_batch_size");
```
klecki, line 46: Shouldn't it introduce batch_size argument for the CollectShape function instead of using max_batch_size?

dali/operators/math/expressions/arithmetic.h
```diff
@@ -313,6 +313,7 @@ class ArithmeticGenericOp : public Operator<Backend> {
 
   bool SetupImpl(std::vector<OutputDesc> &output_desc, const workspace_t<Backend> &ws) override {
     output_desc.resize(1);
+    auto curr_batch_size = ws.GetRequestedBatchSize(0);
```
klecki, line 316: I guess we will probably plan even more broadcasting here, but this operator is perfectly fine with obtaining batch size from it's inputs right now instead of using the GetRequestedBatchSize.

In this case in currently can have [1, 3] inputs.

dali/pipeline/workspace/workspace.h
```diff
@@ -145,12 +145,54 @@ class WorkspaceBase : public ArgumentWorkspace {
...
+  template <typename Backend>
+  auto GetInputShape(int input_idx) const {
+    const auto& input = *InputHandle(input_idx, Backend{});
+    return input.shape();
+  }
+
+  /**
+   * Returns batch size for given input
+   */
+  int GetInputBatchSize(int input_idx) const {
```
klecki, line 172: Btw, I just found there is `NumInputAtIdx` and `NumOutputAtIdx` in the host workspace, which probably does something similar. Can you check if they are needed and if not remove them/replace with the API that we want to support?

dali/operators/python_function/dltensor_function.h
```diff
@@ -193,6 +193,7 @@ class DLTensorPythonFunctionImpl : public Operator<Backend> {
     std::lock_guard<std::mutex> operator_guard(operator_lock);
     py::gil_scoped_acquire interpreter_guard{};
     py::object output_o = py::none();
+    auto curr_batch_size = ws.GetInputBatchSize(0);
```
klecki, line 196: I don't think we can safely assume that there is Input 0 in this Operator.
It certainly requires more logic and probably in some cases would behave as supporting requested batch size which would result in some refactoring. Maybe for such problematic operators we can add an explicit check that it doesn't work with variable batch size and fail for cases where the BS != max_batch_size. That way we could quickly identify them and it would be probably fine to have several of such operators slowly getting updated.

dali/pipeline/operator/builtin/external_source.h
```diff
@@ -377,13 +377,14 @@ class ExternalSource : public Operator<Backend> {
...
+      DALI_WARN(
+          "Warning: Loading GPU-originated data into CPU ExternalSource operator is discouraged "
+          "and might be inefficient.");
     }
-    DALI_ENFORCE(OperatorBase::batch_size_ == static_cast<int>(batch.ntensor()),
-                 make_string("Data list provided to ExternalSource needs to have batch_size = ",
-                             OperatorBase::batch_size_, " length, found ", batch.ntensor(),
-                             " samples."));
+    DALI_ENFORCE(
+        OperatorBase::max_batch_size_ >= static_cast<int>(batch.ntensor()),
```
klecki, line 385: Just to complain that this comparison reads totally unnatural for me, the other way round would be easier to comprehend.

Untill we remove that constraint, I would keep any early checks for the user inputs with sane error messages. And maybe track them in one place, like list all in `EnforceConstantBatchSize` docstring.

dali/python/nvidia/dali/external_source.py
```diff
@@ -21,7 +21,8 @@ def _check_data_batch(data, batch_size, layout):
     shape, uniform = _get_batch_shape(data)
     if len(shape) != batch_size:
         raise RuntimeError("The external source callback returned an unexpected batch "
-        "size: {} instead of {}".format(len(shape), batch_size))
+                           "size. Expected batch_size <= {}, actual: {}".format(batch_size,
```
klecki, line 24: It should probably keep the old error message wit `==`

dali/python/nvidia/dali/pipeline.py
```diff
@@ -113,7 +116,7 @@ def __init__(self, batch_size = -1, num_threads = -1, device_id = -1, seed = -1,
                  *,
                  enable_memory_stats=False):
         self._sinks = []
-        self._batch_size = batch_size
+        self._max_batch_size = batch_size
```
klecki, line 116: There is `self._batch_size` still used in call_and_feed.

dali/operators/generic/constant.h
```diff
@@ -80,9 +80,9 @@ class Constant : public Operator<Backend> {
 
   bool SetupImpl(std::vector<OutputDesc> &output_desc, const Workspace &ws) override {
     output_desc.resize(1);
-    if (output_shape_.empty()) {
-      int batch_size = this->spec_.template GetArgument<int>("batch_size");
-      output_shape_ = uniform_list_shape(batch_size, shape_arg_);
+    auto curr_batch_size = ws.GetRequestedBatchSize(0);
+    if (output_shape_.empty() || output_shape_ .num_elements() < curr_batch_size) {
```
klecki, line 84:
```suggestion
    if (output_shape_.empty() || output_shape_ .num_elements() != curr_batch_size) {
```
Shouldn't this happen on change?

dali/operators/generic/constant.h
```diff
@@ -91,6 +91,7 @@ class Constant : public Operator<Backend> {
   void RunImpl(Workspace &ws) override;
```
klecki, line 91: Also, this operator probably requires some more work, I'm not convinced it will work with changing batch size now.

We only fill the output once (so we need to fill it either once in accordance to max_batch_size_, or refill every time it changes).

As AFAIK this operator works by replicating one sample, we probably should be able to share a part of prepared max batch.

dali/operators/generic/one_hot.cc
```diff
@@ -76,6 +76,7 @@ class OneHotCPU : public OneHot<CPUBackend> {
 void OneHotCPU::RunImpl(workspace_t<CPUBackend> &ws) {
   const auto &input = ws.template InputRef<CPUBackend>(0);
   auto &output = ws.template OutputRef<CPUBackend>(0);
+  auto curr_batch_size = ws.GetInputBatchSize(0);
```
klecki, line 79: We already have num_samples, this is unused.

dali/operators/generic/slice/slice_attr.h
```diff
@@ -34,10 +34,9 @@ namespace dali {
 class SliceAttr {
  public:
   explicit inline SliceAttr(const OpSpec &spec)
-      : batch_size__(spec.GetArgument<int>("batch_size"))
-      , normalized_anchor_(spec.GetArgument<bool>("normalized_anchor"))
-      , normalized_shape_(spec.GetArgument<bool>("normalized_shape"))
-      , crop_window_generators_(batch_size__) {
+      : normalized_anchor_(spec.GetArgument<bool>("normalized_anchor")),
+        normalized_shape_(spec.GetArgument<bool>("normalized_shape")),
+        crop_window_generators_(spec.GetArgument<int>("max_batch_size")) {
```
klecki, line 39: Any thoughts if maybe we should reserve & resize when needed?

dali/operators/image/crop/crop_attr.h
```diff
@@ -34,9 +34,8 @@ namespace dali {
 class CropAttr {
  public:
   static constexpr int kNoCrop = -1;
-  explicit inline CropAttr(const OpSpec &spec)
-    : spec__(spec)
-    , batch_size__(spec__.GetArgument<int>("batch_size")) {
+  explicit inline CropAttr(const OpSpec& spec) : spec__(spec) {
```
klecki, line 37: Marking where I have paused my review.

dali/pipeline/workspace/workspace.h
szalpal: Done

dali/operators/ssd/box_encoder.cuh
```diff
@@ -77,6 +78,7 @@ class BoxEncoder<GPUBackend> : public Operator<GPUBackend> {
  private:
   static constexpr int kBoxesOutputDim = 2;
   static constexpr int kLabelsOutputDim = 1;
+  int curr_batch_size_;
```
klecki, line 81: I think at some point you gave a rationale why not make the `curr_batch_size_` a member.

dali/pipeline/data/tensor_vector.h
klecki, line 130: I wouldn't change type to unsigned, but I would suggest some naming that differs more from Resize.
Maybe SetBatchSize?

klecki: Also, it needs more docs, and I already don't like the semantics. Where do you intend to use it? Why does resizing the batch size repeat the last shape? I'd say it's weird.

dali/pipeline/workspace/sample_workspace.h
```diff
@@ -72,6 +72,18 @@ class DLL_PUBLIC SampleWorkspace : public WorkspaceBase<SampleInputType, SampleO
   template <typename Backend>
   DLL_PUBLIC Tensor<Backend>& Output(int idx);
 
+  int GetInputBatchSize(int) const {
```
klecki, line 75: I got mixed feelings. If we don't allow the ops using SampleWorkspace to use them, maybe it's better to not add this? I know I was thinking about making those ops responsible for propagating the BS, but maybe it can be done in the default Setup/RunImpl for HostWorkspace.

dali/operators/image/color/color_twist.cu
```diff
@@ -32,6 +32,7 @@ DALI_REGISTER_OPERATOR(Saturation, ColorTwistGpu, GPU);
 DALI_REGISTER_OPERATOR(ColorTwist, ColorTwistGpu, GPU);
 
 bool ColorTwistGpu::SetupImpl(std::vector<OutputDesc> &output_desc, const DeviceWorkspace &ws) {
+  KMgrResize(num_threads_, max_batch_size_);
```
klecki, line 35: Why max and not current batch size? (I know it's probably more stable this way).

klecki (commented):
I don't like the mix of max and requested batch size in random ops. If we leave it for a followup, I would stick with using max_batch_size_ everywhere in that category of operators.

I also greped for uses of `ntensor()` and `num_samples()`, and found one place where the result is assigned to member variable:
```
dali/kernels/normalize/normalize_gpu_impl.cuh:238
```

I think the intention was to keep the current batch size as local variable so we don't carry over the previous bs by mistake (there is one class where curr_batch_size_ is a member additionally).

Can you check if the usage I mentioned is fine (or we want to rework it), and there are no other such problems?

szalpal:
!build

dali/operators/math/expressions/arithmetic.h
klecki: Still, it's not requested Batch Size. It's an input Batch Size.

klecki: The assert won't do, we need to calculate common batch size, and as we are passing it to PropagateShapes, make sure it's used there. Or move the calculation to PropagateShapes. The first is easier.

szalpal:
!build

szalpal:
!build

dali/operators/generic/constant.h
szalpal: Done

szalpal:
!build

dali/operators/python_function/dltensor_function.h
szalpal: I've put better batch size integrity checking, don't hesitate to take a look ;)

dali/operators/ssd/random_crop.cc
```diff
@@ -33,7 +33,8 @@ As an input, the operator accepts image, bounding boxes and labels. At the outpu
 cropped and valid bounding boxes and valid labels are returned.)code")
   .NumInput(3)   // [img, bbox, label]
   .NumOutput(3)  // [img, bbox, label]
-  .AddOptionalArg("num_attempts", R"code(Number of attempts.)code", 1);
+  .AddOptionalArg("num_attempts", R"code(Number of attempts.)code", 1)
+  .Deprecate("RandomBBoxCrop");
```
klecki, line 37: Add comment like 
```suggestion
  .Deprecate("RandomBBoxCrop");  // deprecated in DALI 0.30
```

dali/pipeline/data/tensor_vector.cc
```diff
@@ -233,8 +260,8 @@ void TensorVector<Backend>::reserve(size_t bytes_per_sample, int batch_size) {
   assert(batch_size > 0);
   state_ = State::noncontiguous;
   resize_tensors(batch_size);
-  for (auto t : tensors_) {
-    t->reserve(bytes_per_sample);
+  for (size_t i = 0; i < curr_tensors_size_; i++) {
```
klecki, line 263: reserves allows to reserve that many tensors as `batch_size` provided. If there is less tensors it should allocate more and reserve each of them to the desired state.

klecki (commented):
Unfortunately found small issue in TensorVector.

dali/pipeline/data/tensor_vector.cc
klecki: Ok resize_tensors adjust the curr_tenstors_size_.

klecki (approved):

PR #2415, opened 2020-10-29
Add a warning about the Python function incompatibility with TensorFlow

JanuszL:
- adds a note when num_outputs is set in the ExternalSource operator the output would be a list

#### Why we need this PR?
*Pick one, remove the rest*
- It adds a warning about the Python function incompatibility with TensorFlow

 - What solution was applied:
     adds a warning about the Python function incompatibility with TensorFlow
     adds a note when num_outputs is set in the ExternalSource operator the output would be a list
     Python function family docs
     ExternalSource docs
     NA
 - Validation and testing:
     CI
     docs have been updated

**JIRA TASK**: *[NA]*

### DIFF
--- a/dali/python/nvidia/dali/external_source.py
+++ b/dali/python/nvidia/dali/external_source.py
@@ -178,7 +178,8 @@ Args
 `num_outputs` : int, optional
     If specified, denotes the number of TensorLists that are produced by the source function.
 
-    If set the returned output would be a list.
+    If set, the operator returns a list of ``DataNode`` objects, otherwise a single ``DataNode``
+    object is returned.
 
 Keyword Args
 ------------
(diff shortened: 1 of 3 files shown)


JanuszL:
!build

dali/python/nvidia/dali/external_source.py
```diff
@@ -178,6 +178,8 @@ class ExternalSource():
 `num_outputs` : int, optional
     If specified, denotes the number of TensorLists that are produced by the source function.
 
+    If set the returned output would be a list.
```
mzient, line 181:
```suggestion
    If set, the return value is a list of ``DataNode`` objects, otherwise a single ``DataNode`` object is returned.
```

mzient (approved):

klecki (approved):

PR #2416, opened 2020-10-29
Compile DALI with Clang

klecki:
For developement purposes, allow DALI to be compiled only
with CLANG.

CMake adjustments: .cu files marked as CXX files to force
CMake to compile them with Clang. Clang compiles those files
as CUDA based on the .cu extension.

Some additional host/device attributes and overloads
for Clang.

Enable support for CMAKE_CUDA_ARCHITECTURES.

#### Why we need this PR?
Additional development tools

 - What solution was applied:

CMake adjustments: .cu files marked as CXX files to force
CMake to compile them with Clang. Clang compiles those files
as CUDA based on the .cu extension.

Some additional host/device attributes and overloads
for Clang.

 - Validation and testing:


dali/kernels/signal/fft/fft_postprocess.cuh
```diff
@@ -69,6 +69,27 @@ struct norm2 {
...
+}
+
+#else
+
+template <typename T>
+__host__ __device__ const T hostdev_max(const T& l, const T& r) {
+  return ::max(l, r);
+}
+
+#endif
```
mzient, line 91: We have cuda_min/cuda_max. Currently they return reference, but that could be changed.

include/dali/core/geom/vec.h
```diff
@@ -485,46 +485,48 @@ clamp(const vec<N, T> &in, const vec<N, T> &lo, const vec<N, T> &hi) {
   IMPL_VEC_ELEMENTWISE(clamp(in[i], lo[i], hi[i]));
 }
 
-#ifdef __CUDA_ARCH__
+#if (defined(__NVCC__) && defined(__CUDA_ARCH__)) || (defined(__clang__) && defined(__CUDA__))
 template <int N>
-__device__ vec<N> floor(const vec<N> &a) {
+constexpr DALI_DEVICE vec<N> floor(const vec<N> &a) {
```
mzient, line 490: I see no particular benefit of using DALI_DEVICE over __device__. It's compiled only in CUDA and therefore the keyword will always be available.

klecki:
!build

include/dali/core/geom/vec.h
klecki: Oh, I see the point now, we need DALI_HOST_DEV and DALI_HOST (or provide `__host__` and `__device__` without CUDA headers on our own). So the `DALI_DEVICE` is more or less for completeness.

mzient: > So the DALI_DEVICE is more or less for completeness.

That's right.

include/dali/core/float16.h
```diff
@@ -17,14 +17,14 @@
...
-#ifndef __CUDA_ARCH__
+#if !defined(__CUDA_ARCH__)
 #include "dali/util/half.hpp"
 #endif
 
 namespace dali {
 
 // For the GPU
-#ifdef __CUDA_ARCH__
+#if defined(__CUDA_ARCH__)
```
mzient, line 27: ?

include/dali/core/float16.h
```diff
@@ -17,14 +17,14 @@
 
 #include <cuda_fp16.h>  // for __half & related methods
 #include <type_traits>
-#ifndef __CUDA_ARCH__
+#if !defined(__CUDA_ARCH__)
```
mzient, line 20: ?

CMakeLists.txt
```diff
@@ -119,6 +122,21 @@ cmake_dependent_option(STATIC_LIBS "Build static libraries instead of shared-obj
 option(VERBOSE_LOGS "Adds verbose loging to DALI" OFF)
 option(WERROR "Treat all warnings as errors" OFF)
 
+cmake_dependent_option(DALI_CLANG_ONLY "Compile DALI using only Clang. Suitable only for developement."
+    OFF "CMAKE_CXX_COMPILER_ID STREQUAL Clang" OFF)
+
+if (DALI_CLANG_ONLY AND BUILD_NVDEC)
+  message(STATUS "NVDEC is not supportet when compiling only with Clang. Setting BUILD_NVDEC to OFF.")
```
JanuszL, line 129: Anything special in NVDEC that doesn't work?

CMakeLists.txt
```diff
@@ -119,6 +122,21 @@ cmake_dependent_option(STATIC_LIBS "Build static libraries instead of shared-obj
...
+  set(BUILD_NVDEC OFF)
```
JanuszL, line 130: Don't you have to use `CACHE` keyword here as well?

CMakeLists.txt
klecki: Yes, the texture sampling is not implemented in clang.

CMakeLists.txt
```diff
@@ -234,31 +265,58 @@ if ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "Clang")
...
+if (DALI_CLANG_ONLY)
+  set(USED_CUDA_COMPILER ${CMAKE_CXX_COMPILER})
+else()
+  set(USED_CUDA_COMPILER ${CMAKE_CUDA_COMPILER})
+endif()
+
+
+message(STATUS "CUDA .cu files compiler: ${USED_CUDA_COMPILER}")
+
+CUDA_find_supported_arch_values(CUDA_supported_archs ${USED_CUDA_COMPILER} ${CUDA_known_archs})
```
JanuszL, line 288: Isn't that duplication of L293?

dali/kernels/CMakeLists.txt
```diff
@@ -15,6 +15,7 @@
 project(dali_kernels CUDA CXX C)
 
 set(CMAKE_CUDA_SEPARABLE_COMPILATION ON)
+set(CMAKE_CUDA_RESOLVE_DEVICE_SYMBOLS ON)
```
JanuszL, line 18: What is that for?

include/dali/core/convert.h
```diff
@@ -165,14 +189,45 @@ DALI_HOST_DEV constexpr bool clamp(T value, ret_type<bool>) {
...
+DALI_HOST constexpr float16 clamp(T value, ret_type<float16>) {
   constexpr float f16_min = -65504.0f, f16_max = 65504.0f;
   float f = clamp(value, ret_type<float>());
   f = f < f16_min ? f16_min : f > f16_max ? f16_max : f;
   return static_cast<float16>(f);
 }
 
+template <typename T>
+DALI_DEVICE constexpr __half clamp(T value, ret_type<__half>) {
+  constexpr float f16_min = -65504.0f, f16_max = 65504.0f;
```
JanuszL, line 223: Is there any CUDA define for max/min value of half?

CMakeLists.txt
JanuszL: OK

include/dali/core/float16.h
klecki: at some point I've written some more complex incantations here but I can revert.

CMakeLists.txt
klecki: Works without it somehow. Should I add it?

CMakeLists.txt
klecki: No, It's the CUDA_supported_archs - what compiler supports from the predefined set,
vs the CUDA_targeted_archs - what the user requested. The logic was already there.

dali/kernels/CMakeLists.txt
klecki: https://cmake.org/cmake/help/latest/prop_tgt/CUDA_RESOLVE_DEVICE_SYMBOLS.html#prop_tgt:CUDA_RESOLVE_DEVICE_SYMBOLS
This forces the library to link using nvcc in newer CMakes which is able to resolve the symbols (as clang apparently is not able to perform device linking).  Still I use the dummy.cu in older CMakes, so probably this line can be removed, I will check it.

include/dali/core/convert.h
klecki: Not that I'm aware of.

CMakeLists.txt
JanuszL: Maybe not. Just wanted to be sure.

CMakeLists.txt
JanuszL: OK

JanuszL (approved):

include/dali/core/convert.h
```diff
@@ -70,6 +70,30 @@ struct needs_clamp {
...
+  static constexpr bool value = false;
+};
+
+template <typename From>
+struct needs_clamp<From, __half> {
+  static constexpr bool value = false;
+};
+
+template <typename T, typename U>
+using skip_template =
```
mzient, line 87: Perhaps use a more specific name for something that's in namesace `dali`:
```suggestion
using skip_clamp_template =
```

mzient (approved):

JanuszL (approved):

PR #2424, opened 2020-11-02
Update clang to 10.0

klecki:

#### Why we need this PR?
Update to newer clang in deps to use the newest clang that's works with the build image.
11.0 requires newer libc.

 - What solution was applied:
     Update deps
     Docker
      
 - Validation and testing:
     CI
     NA

### DIFF
--- a/dali/operators/image/color/brightness_contrast.h
+++ b/dali/operators/image/color/brightness_contrast.h
@@ -34,7 +34,7 @@ namespace detail {
 template <typename T>
 constexpr float FullRange() {
   return std::is_integral<T>::value
-    ? std::numeric_limits<T>::max()
+    ? static_cast<float>(std::numeric_limits<T>::max())
     : 1.0f;
 }
 
--- a/dali/operators/image/resize/resize_attr.cc
+++ b/dali/operators/image/resize/resize_attr.cc
@@ -216,7 +216,8 @@ void ResizeAttr::PrepareResizeParams(const OpSpec &spec, const ArgumentWorkspace
     GetShapeLikeArgument<float>(size_arg_, spec, "size", ws, spatial_ndim_, N);
   }
 
-  max_size_.resize(spatial_ndim_, static_cast<float>(std::numeric_limits<int>::max()));
+  max_size_.resize(spatial_ndim_,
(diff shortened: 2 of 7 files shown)


awolant:
Any reason for removing explicit version variable from the command?

klecki:
> Any reason for removing explicit version variable from the command?

It's based on specific url, I tried using clang 11 and it also changed the SLES version (it didn't work, so I went back to 10.0), so there is more than one version that could be "configured" for this url. That's why I removed it.

awolant (approved):

dali/operators/image/resize/resize_attr.cc
```diff
@@ -216,7 +216,7 @@ void ResizeAttr::PrepareResizeParams(const OpSpec &spec, const ArgumentWorkspace
     GetShapeLikeArgument<float>(size_arg_, spec, "size", ws, spatial_ndim_, N);
   }
 
-  max_size_.resize(spatial_ndim_, std::numeric_limits<int>::max());
+  max_size_.resize(spatial_ndim_, static_cast<float>(std::numeric_limits<int>::max()));
```
mzient, line 219:
```suggestion
  max_size_.resize(spatial_ndim_, std::numeric_limits<float>::max());
```

dali/operators/image/color/brightness_contrast.h
```diff
@@ -34,7 +34,7 @@ namespace detail {
 template <typename T>
 constexpr float FullRange() {
   return std::is_integral<T>::value
-    ? std::numeric_limits<T>::max()
+    ? static_cast<float>(std::numeric_limits<T>::max())
```
mzient, line 37: phhh.... rly?

JanuszL (approved):

dali/operators/image/resize/resize_attr.cc
klecki: done

dali/operators/image/color/brightness_contrast.h
klecki: To make the compiler happy.

JanuszL (approved):

PR #2435, opened 2020-11-05
TensorVector update for iter-to-iter variable batch size

szalpal:
#### Why we need this PR?
- To enable new feature - iter-to-iter (i2i) variable batch size

Small rework to `TensorVector`, that allows to use it with i2i variable batch size.

1. I retained the API of old TensorVector and I don't intend to change it. Generally, because it will be a greatly disputable change and this is a bigger topic, which will surely result in some discussions about `Tensor` and `TensorList` coexistence, and `Buffer` in general. I'm leaving the API as it is.
1. I retained the tests from the old `TensorVector`, with one change - removed the throwing-assertions on resize action of the `TensorVector` (obviously why).
1. I wrote some new tests for `TensorVector`, from the usage perspective. I guess they cover all the functions, that might be impacted by i2i variable batch size.
1. I've moved function definitions to `cc` file as a favour for me and future developers. Changing `tensor_vector.h` recompiles whole `operators` module, making it living hell to develop this way.
1. What generally happened here is that I've changed the approach to `tensors_` resizing, making it available on demand. And then I had to write couple of `asserts` and change some `ifs`.

### DIFF
--- a/dali/pipeline/data/tensor_vector.cc
+++ b/dali/pipeline/data/tensor_vector.cc
@@ -1,0 +1,415 @@
+// Copyright (c) 2020, NVIDIA CORPORATION. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "tensor_vector.h"
+
+namespace dali {
+
+template <typename Backend>
+TensorVector<Backend>::TensorVector()
+    : views_count_(0), tl_(std::make_shared<TensorList<Backend>>()) {}
+
+template <typename Backend>
+TensorVector<Backend>::TensorVector(int batch_size)
+    : views_count_(0),
+      tensors_(batch_size, nullptr),
+      tl_(std::make_shared<TensorList<Backend>>(batch_size)) {
+  for (auto &t : tensors_) {
+    t = std::make_shared<Tensor<Backend>>();
+  }
+}
+
+template <typename Backend>
+TensorVector<Backend>::TensorVector(std::shared_ptr<TensorList<Backend>> tl)
+    : views_count_(0), tl_(std::move(tl)) {
+  assert(tl_ && "Construction with null TensorList is illegal");
+  pinned_ = tl_->is_pinned();
+  type_ = tl_->type();
+  state_ = State::contiguous;
+  tensors_.resize(tl_->ntensor());
+  UpdateViews();
+}
+
+template <typename Backend>
+TensorVector<Backend>::TensorVector(TensorVector<Backend> &&other) noexcept {
+  state_ = other.state_;
+  pinned_ = other.pinned_;
+  tl_ = std::move(other.tl_);
+  type_ = other.type_;
+  views_count_ = other.views_count_.load();
+  tensors_ = std::move(other.tensors_);
+  for (auto &t : tensors_) {
+    if (t) {
+      if (auto *del = std::get_deleter<ViewRefDeleter>(t->data_)) del->ref = &views_count_;
+    }
+  }
+
+  other.views_count_ = 0;
+  other.tensors_.clear();
+}
+
+template <typename Backend>
+size_t TensorVector<Backend>::nbytes() const noexcept {
+  if (state_ == State::contiguous) {
+    return tl_->nbytes();
+  }
+  // else
+  size_t total_nbytes = 0;
+  for (const auto &t : tensors_) {
+    total_nbytes += t->nbytes();
+  }
+  return total_nbytes;
+}
+
+template <typename Backend>
+size_t TensorVector<Backend>::capacity() const noexcept {
+  if (state_ == State::contiguous) {
+    return tl_->capacity();
+  }
+  // else
+  size_t total_capacity = 0;
+  for (const auto &t : tensors_) {
+    total_capacity += t->capacity();
+  }
+  return total_capacity;
+}
+
+template <typename Backend>
+TensorListShape<> TensorVector<Backend>::shape() const {
+  if (state_ == State::contiguous) {
+    return tl_->shape();
+  }
+  if (tensors_.empty()) {
+    return {};
+  }
+  TensorListShape<> result(tensors_.size(), tensors_[0]->ndim());
+  for (size_t i = 0; i < tensors_.size(); i++) {
+    result.set_tensor_shape(i, tensors_[i]->shape());
+  }
+  return result;
+}
+
+template <typename Backend>
+void TensorVector<Backend>::Resize(const TensorListShape<> &new_shape) {
+  return Resize(new_shape, type());
+}
+
+template <typename Backend>
+void TensorVector<Backend>::Resize(const TensorListShape<> &new_shape, const TypeInfo &new_type) {
+  // N.B. There probably is nothing wrong with adjusting batchsize for give TensorVector
+  // (sparse tensor list), but the semantics of what type and pinned status should
+  // the new elements have or having some of them allocated and the new ones not
+  // complicates the logic even further so we disallow it
+  DALI_ENFORCE(tensors_.empty() || static_cast<int>(tensors_.size()) == new_shape.size(),
+               "Changing the batch size is prohibited. It should be set once.");
+  if (tensors_.empty()) {
+    allocate_tensors(new_shape.size());
+  }
+  if (state_ == State::contiguous) {
+    tl_->Resize(new_shape, new_type);
+    UpdateViews();
+    return;
+  }
+  for (int i = 0; i < new_shape.size(); i++) {
+    tensors_[i]->Resize(new_shape[i], new_type);
+  }
+}
+
+template <typename Backend>
+void TensorVector<Backend>::set_type(const TypeInfo &new_type) {
+  type_ = new_type;
+  tl_->set_type(new_type);
+  for (auto t : tensors_) {
+    t->set_type(new_type);
+  }
+  if (state_ == State::contiguous) {
+    UpdateViews();
+  }
+}
+
+template <typename Backend>
+const TypeInfo &TensorVector<Backend>::type() const {
+  if (state_ == State::contiguous) {
+    return tl_->type();
+  }
+  if (tensors_.size() > 0) {
+    return tensors_[0]->type();
+  }
+  return type_;
+}
+
+/** @brief Set uniform layout for all samples in the list */
+template <typename Backend>
+void TensorVector<Backend>::SetLayout(const TensorLayout &layout) {
+  if (state_ == State::noncontiguous) {
+    DALI_ENFORCE(!tensors_.empty(), "Layout cannot be set uniformly for empty batch");
+  }
+  tl_->SetLayout(layout);
+  for (auto t : tensors_) {
+    t->SetLayout(layout);
+  }
+}
+
+template <typename Backend>
+TensorLayout TensorVector<Backend>::GetLayout() const {
+  if (state_ == State::contiguous) {
+    auto layout = tl_->GetLayout();
+    if (!layout.empty()) return layout;
+  }
+  if (tensors_.size() > 0) {
+    auto layout = tensors_[0]->GetLayout();
+    for (size_t i = 1; i < tensors_.size(); i++) assert(layout == tensors_[i]->GetLayout());
+    return layout;
+  }
+  return {};
+}
+
+template <typename Backend>
+const DALIMeta &TensorVector<Backend>::GetMeta(int idx) const {
+  return tensors_[idx]->GetMeta();
+}
+
+template <typename Backend>
+void TensorVector<Backend>::SetMeta(int idx, const DALIMeta &meta) {
+  tensors_[idx]->SetMeta(meta);
+}
+
+template <typename Backend>
+void TensorVector<Backend>::set_pinned(bool pinned) {
+  // Store the value, in case we pin empty vector and later call Resize
+  pinned_ = pinned;
+  tl_->set_pinned(pinned);
+  for (auto &t : tensors_) {
+    t->set_pinned(pinned);
+  }
+}
+
+template <typename Backend>
+bool TensorVector<Backend>::is_pinned() const {
+  // TODO(klecki): do we enforce the same pin status between elements, or prohibit mixing APIs?
+  if (state_ == State::contiguous) {
+    return tl_->is_pinned();
+  }
+  if (tensors_.size() > 0) {
+    return tensors_[0]->is_pinned();
+  }
+  return pinned_;
+}
+
+/**
+ * @brief Reserve as contiguous tensor list internally
+ */
+template <typename Backend>
+void TensorVector<Backend>::reserve(size_t total_bytes) {
+  state_ = State::contiguous;
+  tl_->reserve(total_bytes);
+}
+
+/**
+ * @brief Reserve as vector of `batch_size` tensors internally
+ */
+template <typename Backend>
+void TensorVector<Backend>::reserve(size_t bytes_per_sample, int batch_size) {
+  DALI_ENFORCE(tensors_.empty() || static_cast<int>(tensors_.size()) == batch_size,
+               "Changing the batch size is prohibited. It should be set once.");
+  state_ = State::noncontiguous;
+  // If we didn't declare the batch size but tried to pin memory or set type
+  // we need to apply it to tensors
+  if (tensors_.empty()) {
+    allocate_tensors(batch_size);
+  }
+  for (auto t : tensors_) {
+    t->reserve(bytes_per_sample);
+  }
+}
+
+/**
+ * @brief If the TensorVector is backed by TensorList (contiguous memory)
+ */
+template <typename Backend>
+bool TensorVector<Backend>::IsContiguous() const noexcept {
+  // TODO(klecki): check the views_count as well?
+  return state_ == State::contiguous && static_cast<size_t>(views_count_) == size();
+}
+
+/**
+ * @brief Set the current state if further calls like Resize() or set_type
+ *        should use TensorList or std::vector<Tensor> as backing memory
+ */
+template <typename Backend>
+void TensorVector<Backend>::SetContiguous(bool contiguous) {
+  if (contiguous) {
+    state_ = State::contiguous;
+  } else {
+    state_ = State::noncontiguous;
+  }
+}
+
+template <typename Backend>
+void TensorVector<Backend>::Reset() {
+  if (IsContiguous()) {
+    type_ = {};
+    tensors_.resize(0);
+    views_count_ = 0;
+    tl_->Reset();
+  } else {
+    type_ = {};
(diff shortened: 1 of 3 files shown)


szalpal:
!build

dali/pipeline/data/tensor_vector.h
```diff
@@ -130,358 +99,85 @@ class TensorVector {
...
-  inline const TypeInfo &type() const {
-    if (state_ == State::contiguous) {
-      return tl_->type();
-    }
-    if (tensors_.size() > 0) {
-      return tensors_[0]->type();
-    }
-    return type_;
-  }
+  const TypeInfo &type() const;
```
mzient, line 114: Please leave such simple getters as inline functions.

dali/pipeline/data/tensor_vector.h
```diff
@@ -130,358 +99,85 @@ class TensorVector {
...
-      result.set_tensor_shape(i, tensors_[i]->shape());
-    }
-    return result;
-  }
+  TensorListShape<> shape() const;
 
-  DLL_PUBLIC inline void Resize(const TensorListShape<> &new_shape) {
-    return Resize(new_shape, type());
-  }
+  DLL_PUBLIC void Resize(const TensorListShape<> &new_shape);
```
mzient, line 108: Keep it inline.

dali/pipeline/data/tensor_vector.cc
```diff
@@ -346,57 +368,52 @@ TensorVector<Backend> &TensorVector<Backend>::operator=(TensorVector<Backend> &&
...
-  // we need to apply it to tensors
-  tensors_.resize(batch_size, nullptr);
-  for (auto &t : tensors_) {
-    t = create_tensor();
+void TensorVector<Backend>::resize_tensors(int new_size) {
+  auto old_size = tensors_.size();
+  tensors_.resize(new_size);
+  for (int i = old_size; i < new_size; i++) {
+    tensors_[i] = std::move(std::make_shared<Tensor<Backend>>());
+    tensors_[i]->set_pinned(pinned_);
```
JanuszL, line 402: Shouldn't `is_pinned()` be used here?

dali/pipeline/data/tensor_vector.cc
```diff
@@ -138,18 +136,22 @@ void TensorVector<Backend>::set_type(const TypeInfo &new_type) {
...
   }
-  if (tensors_.size() > 0) {
-    return tensors_[0]->type();
+  if (tensors_.size() == 0) {
+    return type_;
   }
-  return type_;
+  for (size_t i = 1; i < tensors_.size(); i++) {
+    assert(tensors_[0]->type() == tensors_[i]->type());
+  }
```
mzient, line 150: I think it's a bit excessive to test it in every access to the type.

dali/pipeline/data/tensor_vector.cc
```diff
@@ -195,18 +203,22 @@ void TensorVector<Backend>::set_pinned(bool pinned) {
...
   if (state_ == State::contiguous) {
     return tl_->is_pinned();
   }
-  if (tensors_.size() > 0) {
-    return tensors_[0]->is_pinned();
+  if (tensors_.size() == 0) {
+    return pinned_;
+  }
+  for (size_t i = 1; i < tensors_.size(); i++) {
+    assert(tensors_[i]->is_pinned() == tensors_[0]->is_pinned());
```
mzient, line 216: Perhaps instead of enforcing uniformity (which may or may not be wanted), you should check that all relevant tensors are pinned.

dali/pipeline/data/tensor_vector.cc
```diff
@@ -12,35 +12,34 @@
...
+
 template <typename Backend>
-TensorVector<Backend>::TensorVector(int batch_size)
-    : views_count_(0),
-      tensors_(batch_size, nullptr),
-      tl_(std::make_shared<TensorList<Backend>>(batch_size)) {
-  for (auto &t : tensors_) {
-    t = std::make_shared<Tensor<Backend>>();
-  }
+TensorVector<Backend>::TensorVector(int max_batch_size)
```
mzient, line 25: This is changing the semantics of TensorVector and it's not expected - either based on previous behavor or by analogy to std::vector.

dali/pipeline/data/tensor_vector.cc
```diff
@@ -346,57 +368,52 @@ TensorVector<Backend> &TensorVector<Backend>::operator=(TensorVector<Backend> &&
...
-void TensorVector<Backend>::allocate_tensors(int batch_size) {
-  DALI_ENFORCE(tensors_.empty(), "Changing the batch size is prohibited. It should be set once.");
-  // If we didn't declare the batch size but tried to pin memory or set type
-  // we need to apply it to tensors
-  tensors_.resize(batch_size, nullptr);
-  for (auto &t : tensors_) {
-    t = create_tensor();
+void TensorVector<Backend>::resize_tensors(int new_size) {
+  auto old_size = tensors_.size();
+  tensors_.resize(new_size);
```
mzient, line 399: This is not a vector-like resize. You should either store these tensors in some kind of free list or maintain a separate actual size and never shrink `tensors_`.

dali/pipeline/data/tensor_vector.cc
```diff
@@ -346,57 +368,52 @@ TensorVector<Backend> &TensorVector<Backend>::operator=(TensorVector<Backend> &&
...
+  for (int i = old_size; i < new_size; i++) {
+    tensors_[i] = std::move(std::make_shared<Tensor<Backend>>());
```
mzient, line 401: No need for `move` for an r-value.

dali/pipeline/data/tensor_vector.cc
```diff
@@ -346,57 +368,52 @@ TensorVector<Backend> &TensorVector<Backend>::operator=(TensorVector<Backend> &&
...
-  }
+  assert(static_cast<size_t>(idx) < tensors_.size());
+  assert(static_cast<size_t>(idx) < tl_->ntensor());
+
+  tensors_[idx]->set_type(type_);
+
   auto *ptr = tl_->raw_mutable_tensor(idx);
 
-  TensorShape<> shape = tl_->tensor_shape(idx);
+  TensorShape<> shape = tl_->shape()[idx];
```
mzient, line 416: No go. This is going to be too inefficient. Copying a TensorListShape _per sample_ when updating views is waaaay too much.

dali/pipeline/data/tensor_vector.cc
szalpal: It should, fixed

dali/pipeline/data/tensor_vector.h
szalpal: Done

dali/pipeline/data/tensor_vector.h
szalpal: I don't know, if this getter is yet a simple one

dali/pipeline/data/tensor_vector.cc
szalpal: But it's only in debug conf, right? Cause compiler with optimize out this for loop in release. In this case, I think it'd be nice to have this extra sanity check

dali/pipeline/data/tensor_vector.cc
szalpal: Added `curr_tensors_size_`

dali/pipeline/data/tensor_vector.cc
szalpal: Done

szalpal:
!build

szalpal:
!build

dali/pipeline/data/tensor_vector.cc
szalpal: Maybe enforced uniformity better fits the `curr_tensors_size_` approach? I don't know if we envision non-uniform pin status of tensors in batch at all...

dali/pipeline/data/tensor_vector.cc
szalpal: Added `batch_size` to TensorList ctor

dali/pipeline/data/tensor_vector.cc
```diff
@@ -256,33 +249,56 @@ void TensorVector<Backend>::SetContiguous(bool contiguous) {
...
 
+
+template <typename Backend>
+template <typename SrcBackend>
+void TensorVector<Backend>::Copy(const TensorList<SrcBackend> &in_tl, cudaStream_t stream) {
+  SetContiguous(true);
+  tl_->Copy(in_tl, stream);
+
+  resize_tensors(tl_->ntensor());
+  views_count_ = 0;
```
mzient, line 272: I don't think it's OK to reset it by hand.

dali/pipeline/data/tensor_vector.cc
```diff
@@ -395,18 +389,21 @@ std::shared_ptr<TensorList<Backend>> TensorVector<Backend>::AsTensorList(bool ch
...
-    tensors_[i]->set_pinned(pinned_);
+  if (new_size > tensors_.size()) {
+    auto old_size = curr_tensors_size_;
+    tensors_.resize(new_size);
+    for (int i = old_size; i < new_size; i++) {
+      tensors_[i] = std::move(std::make_shared<Tensor<Backend>>());
+      tensors_[i]->set_pinned(is_pinned());
+    }
   }
+  curr_tensors_size_ = new_size;
```
mzient, line 402:
```C++
} else if (new_size < curr_tensors_size_) {
  for (int i = new_size; i < curr_tensors_size_; i++) {
    if (tensors_[i]->shares_data())
      tensors_[i]->Reset();
  }
}
```
this is critical for:
- maintaining correct `views_count_`
- properly releasing reference to data shared with external object.

dali/pipeline/data/tensor_vector.cc
```diff
@@ -395,18 +389,21 @@ std::shared_ptr<TensorList<Backend>> TensorVector<Backend>::AsTensorList(bool ch
 
 template <typename Backend>
 void TensorVector<Backend>::resize_tensors(int new_size) {
-  auto old_size = tensors_.size();
-  tensors_.resize(new_size);
-  for (int i = old_size; i < new_size; i++) {
-    tensors_[i] = std::move(std::make_shared<Tensor<Backend>>());
-    tensors_[i]->set_pinned(pinned_);
+  if (new_size > tensors_.size()) {
```
mzient: signed/unsigned comparison - will cause an error (-Werror) in clang
```suggestion
  if (new_size > static_cast<int>(tensors_.size()) {
```

dali/pipeline/data/tensor_vector.cc
```diff
@@ -0,0 +1,439 @@
...
+  } else {
+    DALI_FAIL("Cannot share non contiguous TensorVector with TensorList");
+  }
+}
+
+
+template <typename Backend>
+void TensorVector<Backend>::ShareData(TensorVector<Backend> *tv) {
+  DALI_ENFORCE(tv != nullptr, "Input TensorVector is nullptr");
+  state_ = tv->state_;
```
mzient, line 320:
```suggestion
  type_ = tv->type_;
  state_ = tv->state_;
```
Your change in `type()` has uncovered that setting `type_` is missing here (and probably in some other places - check Copy and ShareData).

dali/pipeline/data/tensor_vector.cc
```diff
@@ -0,0 +1,439 @@
...
+  views_count_ = 0;
+  UpdateViews();
+}
+
+
+template <typename Backend>
+void TensorVector<Backend>::ShareData(TensorList<Backend> *in_tl) {
+  DALI_ENFORCE(in_tl != nullptr, "Input TensorList is nullptr");
+  SetContiguous(true);
+  pinned_ = in_tl->is_pinned();
```
mzient, line 293: set type

dali/pipeline/data/tensor_vector.cc
```diff
@@ -0,0 +1,439 @@
...
+  views_count_ = 0;
+  UpdateViews();
+}
+
+
+template <typename Backend>
+template <typename SrcBackend>
+void TensorVector<Backend>::Copy(const TensorVector<SrcBackend> &in_tv, cudaStream_t stream) {
+  SetContiguous(true);
+  tl_->Copy(in_tv, stream);
```
mzient, line 281: set type

dali/pipeline/data/tensor_vector.cc
```diff
@@ -0,0 +1,439 @@
...
+  }
+  for (size_t i = 1; i < curr_tensors_size_; i++) {
+    assert(tensors_[i]->is_pinned() == tensors_[0]->is_pinned());
+  }
+  return tensors_[0]->is_pinned();
+}
+
+
+template <typename Backend>
+void TensorVector<Backend>::reserve(size_t total_bytes) {
```
mzient, line 220: That's buggy.
1. If previous state was non-contiguous, then we should probably clear `tensors_` (I think that's the only place we should do that).
2. After `tl_->reserve(total_bytes)` there should be a call to UpdateViews.

szalpal:
!build

szalpal:
!build

dali/pipeline/data/tensor_vector.cc
szalpal: Done

dali/pipeline/data/tensor_vector.cc
szalpal: Done

dali/pipeline/data/tensor_vector.cc
szalpal: Done

dali/pipeline/data/tensor_vector.cc
szalpal: Done

dali/pipeline/data/tensor_vector.cc
szalpal: Done

dali/pipeline/data/tensor_vector.cc
szalpal: Done

dali/pipeline/data/tensor_vector.cc
szalpal: In fact, UpdateViews does that. I removed redundant `views_count_=0` calls. The remained ones are I guess necessary: in `Reset` and in `ShareData(TensorVector)`, which doesn't do UpdateViews (for obvious reasons).

jantonguirao (approved):

dali/pipeline/data/tensor_vector.cc
```diff
@@ -346,70 +370,81 @@ TensorVector<Backend> &TensorVector<Backend>::operator=(TensorVector<Backend> &&
...
 }
 
+
 template <typename Backend>
 void TensorVector<Backend>::update_view(int idx) {
-  if (!tensors_[idx]) {
-    tensors_[idx] = create_tensor();
-  }
+  assert(static_cast<size_t>(idx) < curr_tensors_size_);
+  assert(static_cast<size_t>(idx) < tl_->ntensor());
```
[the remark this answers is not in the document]
szalpal: I've put this for a documentation purposes, helped me figure out what and where should happen. Thus I'll retain it here, for future generations ;)

dali/pipeline/data/tensor_vector.cc
```diff
@@ -346,70 +370,81 @@ TensorVector<Backend> &TensorVector<Backend>::operator=(TensorVector<Backend> &&
...
+
   auto *ptr = tl_->raw_mutable_tensor(idx);
 
-  TensorShape<> shape = tl_->tensor_shape(idx);
+  TensorShape<> shape = tl_->shape()[idx];
```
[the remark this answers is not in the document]
szalpal: Nothing, I guess? Went back

dali/pipeline/data/tensor_vector_test.cc
```diff
@@ -116,13 +118,183 @@ TYPED_TEST(TensorVectorSuite, BatchResize) {
...
+  return {};  // To please the compiler
+}
+
+template <>
+TensorShape<> get_shape<TensorList<CPUBackend>>(const TensorList<CPUBackend> &tl, int idx) {
+  return tl.tensor_shape(idx);
+}
+template <>
+TensorShape<> get_shape<TensorVector<CPUBackend>>(const TensorVector<CPUBackend> &tv, int idx) {
+  return tv.shape()[idx];
```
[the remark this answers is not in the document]
szalpal: Good idea

dali/pipeline/data/tensor_vector.h
```diff
@@ -123,368 +92,110 @@ class TensorVector {
...
-    }
-    // else
-    size_t total_capacity = 0;
-    for (const auto &t : tensors_) {
-      total_capacity += t->capacity();
-    }
-    return total_capacity;
+  TensorShape<> tensor_shape(int idx) const {
+    assert(state_ == State::contiguous && curr_tensors_size_ == views_count_ ||
+           state_ == State::noncontiguous);
```
mzient, line 110: This check isn't really necessary in this function, is it?

dali/pipeline/data/tensor_vector.h
```diff
@@ -123,368 +92,110 @@ class TensorVector {
...
-    if (state_ == State::contiguous) {
-      return tl_->capacity();
-    }
-    // else
-    size_t total_capacity = 0;
-    for (const auto &t : tensors_) {
-      total_capacity += t->capacity();
-    }
-    return total_capacity;
+  TensorShape<> tensor_shape(int idx) const {
```
mzient, line 108:
```suggestion
  const TensorShape<> &tensor_shape(int idx) const {
```

dali/pipeline/data/tensor_vector.h
```diff
@@ -123,368 +92,110 @@ class TensorVector {
...
-    }
-    if (tensors_.empty()) {
-      return {};
-    }
-    TensorListShape<> result(tensors_.size(), tensors_[0]->ndim());
-    for (size_t i = 0; i < tensors_.size(); i++) {
-      result.set_tensor_shape(i, tensors_[i]->shape());
-    }
-    return result;
+  const void *raw_tensor(int idx) const {
```
mzient, line 114: Perhaps add `raw_mutable_tensor` for a complete API?
I know it would be a bit more involved, as it would potentially require updating the views (and most certainly we shouldn't do that on every call to `raw_mutable_tensor`).

dali/pipeline/data/tensor_vector.h
szalpal: As you noticed in a comment below, this only makes sense to use if `UpdateViews` was correctly invoked. I wanted to add a debug check here, but I guess this one is incorrect... You have some other idea?

dali/pipeline/data/tensor_vector.h
szalpal: TensorList returns by value, so I guess we should keep them consistent. If we would like to go to `const&`, that'd be out of scope of this PR.

dali/pipeline/data/tensor_vector.h
szalpal: IMHO there should be an `assert` for this, but I don't have good idea how to do this...

mzient: Ufff... I've checked that _raw_ _mutable_data does not silently allocate, so we're good whatever the state.

dali/pipeline/data/tensor_vector.h
mzient: Because it doesn't have them. Please change it.

mzient: Tensors can return const-ref and we shouldn't force an extra copy if we don't have to. We're not implementing a virtual interface here, more like a concept.

mzient (approved):

szalpal:
!build

PR #2449, opened 2020-11-10
Fix dummy nvidia-tensorflow being installed instead of the real one

TheTimemaster:
Docker image Dockerfile.customopbuilder.clean does not currently have NVIDIA Private Python Package Index installed in the build process.
That makes the dependency downloader command install a dummy package (https://pypi.org/project/nvidia-tensorflow/) instead of the real one.

#### Why we need this PR?
- It fixes a bug where docker would try to install a dummy package and fail the build process

 - What solution was applied:
     nvidia-pyindex is properly installed before installing dependencies

### DIFF
--- a/docker/Dockerfile.customopbuilder.clean
+++ b/docker/Dockerfile.customopbuilder.clean
@@ -73,6 +73,8 @@ ENV LD_LIBRARY_PATH=/usr/local/cuda/lib64:/usr/local/cuda/lib64/stubs/:${LD_LIBR
 WORKDIR /opt/dali
 COPY qa/setup_packages.py qa/setup_packages.py
 
+# Install NVIDIA Private Python Package Index in order to make sure next step will install proper packages
+RUN pip install --user nvidia-pyindex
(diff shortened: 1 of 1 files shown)


docker/Dockerfile.customopbuilder.clean
```diff
@@ -73,6 +73,8 @@ ENV LD_LIBRARY_PATH=/usr/local/cuda/lib64:/usr/local/cuda/lib64/stubs/:${LD_LIBR
 WORKDIR /opt/dali
 COPY qa/setup_packages.py qa/setup_packages.py
 
+# Install NVIDIA Private Python Package Index in order to make sure next step will install proper packages
```
JanuszL, line 76: I would move this after L66.

JanuszL: As we want to reduce the number of docker layers.

JanuszL:
https://github.com/NVIDIA/DALI/pull/2451 should deal with that, so your PR seems to be no longer needed.

PR #2469, opened 2020-11-16
Per sample ExternalSource

mzient:
* add `batch` argument that changes how `source` operates
* update docs
* add tests.

#### Why we need this PR?
*Pick one, remove the rest*
- It adds new feature needed as a prerequisite for parallel external source

 - What solution was applied:
     * Add `batch` argument
     * Add tests
     * External source, pipeline
     * N/A
 - Validation and testing:
     * Python tests
     * Updated docs

**JIRA TASK**: DALI-1734

### DIFF
--- a/dali/python/nvidia/dali/external_source.py
+++ b/dali/python/nvidia/dali/external_source.py
@@ -193,7 +193,7 @@ Args
 
     A per-batch source may accept one positional argument. If it does, it is the current iteration
     iteration number and consecutive calls will be ``source(0)``, ``source(1)``, and so on.
-    A per-sample source may accept a :class:`nvidia.dali.SampleInfo` structure.
+    A per-sample source may accept a :class:`nvidia.dali.types.SampleInfo` structure.
 
     If the source is a generator function, the function is invoked and treated as an iterable.
     However, unlike a generator, the function can be used with ``cycle``. In this case, the function
--- a/dali/python/nvidia/dali/external_source.py
+++ b/dali/python/nvidia/dali/external_source.py
@@ -102,8 +102,7 @@ class _ExternalSourceGroup(object):
             self.current_sample += batch_size
             self.current_iter += 1
         except StopIteration:
-            self.current_sample = 0
-            self.current_iter = 0
+            self.reset_indices()
             raise
 
         if self.is_multioutput:
@@ -191,20 +190,33 @@ Args
     value is 1), the data is expected to a be tuple, or list, where each element corresponds to
     respective return value of the external_source.
 
-    A per-batch source may accept one positional argument. If it does, it is the current iteration
-    iteration number and consecutive calls will be ``source(0)``, ``source(1)``, and so on.
-    A per-sample source may accept a :class:`nvidia.dali.types.SampleInfo` structure.
+    The data samples must be in one of the compatible array types:
+        * NumPy ndarray (CPU)
+        * MXNet ndarray (CPU)
+        * PyTorch tensor (CPU or GPU)
+        * CuPy array (GPU)
+        * objects implementing ``__cuda_array_interface__``
+        * DALI `Tensor` object
+    Batch sources must produce entire batches of data. This can be achieved either by adding a new
+    outermost dimension to an array or by returning a list of arrays (in which case they can be of
+    different size, but must have the same rank and element type). A batch source can also
+    produce a DALI `TensorList` object, which can be an output of another DALI pipeline.
+
+    A per-batch source may accept one positional argument. If it does, it is the index of current
+    iteration within epoch and consecutive calls will be ``source(0)``, ``source(1)``, and so on.
+
+    A per-sample source may accept one positional argument of type
+    :class:`nvidia.dali.types.SampleInfo`, which contains index of the sample in current epoch and
+    in the batch, as well as current iteration number.
 
     If the source is a generator function, the function is invoked and treated as an iterable.
     However, unlike a generator, the function can be used with ``cycle``. In this case, the function
     will be called again when the generator reaches the end of iteration.
 
-    For the GPU input, it is a user's responsibility to modify the provided GPU memory content
+    For GPU inputs, it is a user's responsibility to modify the provided GPU memory content
     only in the provided stream. DALI schedules a copy on this stream, and all work is properly
     queued. If no stream is provided, DALI will use a default, with a best-effort approach at
     correctness. See the ``cuda_stream`` argument documentation for more information.
-    The data item produced by ``source`` may be anything that's accepted by
-    :meth:`nvidia.dali.pipeline.Pipeline.feed_input`
 
 `num_outputs` : int, optional
     If specified, denotes the number of TensorLists that are produced by the source function.
@@ -215,7 +227,7 @@ Args
 Keyword Args
 ------------
 
-`cycle`: bool
+`cycle`: bool, optional
     If set to True, the source will be wrapped.
 
     If set to False, StopIteration is raised when the end of data is reached.
@@ -230,8 +242,7 @@ Keyword Args
     Used when feeding the data in ``iter_setup`` and can be omitted if
     the data is provided by ``source``.
 
-
-`layout` : :ref:`layout str<layout_str_doc>` or list/tuple thereof
+`layout` : :ref:`layout str<layout_str_doc>` or list/tuple thereof, optional
     If provided, sets the layout of the data.
 
     When ``num_outputs > 1``, the layout can be a list that contains a distinct layout for each
@@ -255,17 +266,17 @@ Keyword Args
     buffer is complete, since there's no way to synchronize with this stream to prevent
     overwriting the array with new data in another stream.
 
-`use_copy_kernel` : optional, `bool`
+`use_copy_kernel` : bool, optional
     If set to True, DALI will use a CUDA kernel to feed the data instead of cudaMemcpyAsync (default).
 
     .. note::
         This is applicable only when copying data to and from GPU memory.
 
-`blocking` : optional
+`blocking` : bool, optional
     Determines whether the external source should wait until data is available or just fail
     when the data is not available.
 
-`no_copy` : optional
+`no_copy` : boo, optional
     Determines whether DALI should copy the buffer when feed_input is called.
 
     If set to True, DALI passes the user memory directly to the pipeline, instead of copying it.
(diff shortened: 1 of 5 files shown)


mzient:
!build

dali/python/nvidia/dali/pipeline.py
```diff
@@ -963,7 +963,7 @@ def _run_input_callbacks(self):
             return
 
         for group in self._input_callbacks:
-            group.call_and_feed(self, self._iter)
+            group.call_and_feed(self, self._batch_size, self._iter)
```
stiepan, line 966: You've mentioned in the morning the idea of inferring batch size from other externalsources if there are a few of them and some work in batch mode (falling back to pipeline batch size as a last resort when there are no external sources working in batch mode) - do you handle that/do we want such a feature?

mzient: No, it's not a part of this PR.

dali/python/nvidia/dali/external_source.py
```diff
@@ -62,31 +62,51 @@ def __next__(self):
...
+            callback_out = [self.callback(*self.callback_args(i, current_iter)) for i in range(batch_size)]
 
         if self.is_multioutput:
             for op in self.instances:
-                data = callback_out[op._output_index]
+                if self.batch:
+                    data = callback_out[op._output_index]
+                else:
+                    # extract a single output
+                    data = [callback_out[i][op._output_index] for i in range(batch_size)]
```
stiepan, line 109: I am not sure if it matters later on in the pipeline but maybe user should be able to pass optional collate function to decide how exactly the list of multioutput samples should be collated here? E.g. to return [array(128x128x3)], matrxi(batch_size x 1) rather than [array(128x128x3)], [array(1)]? Or lists are just fine here?

dali/python/nvidia/dali/pipeline.py
stiepan: Ok!

dali/python/nvidia/dali/external_source.py
mzient: I don't quite follow.
The values from multi-output external source are returned as:
```(batch1, batch2, ...)``` for batched variant and as 
```(sample_from_batch1, sample_from_batch2, ...)``` for per-sample variant.
One of the reason why the batched variant is expected to produce a tuple of batches (as opposed to a batch of tuples) is that the batches can be contiguous tensors where the outermost dimension is batch:
```(batch_tensor, [sample_tensor, sample_tensor, sample_tensor])```

Comment (stiepan) [reply dali/python/nvidia/dali/external_source.py:109]: Okok, so in the per-sample variant in case of two outputs, we'll be now necessarily returning tuple of (two) lists (each with length of batch size), right? And what I've been wondering is if we should allow user to pass here some additional callback that would for example squeeze one of this lists into batch tensor so that the output would be (batch_tensor, [sample1_tensor, sample2_tensor, sample3_tensor...]).

Comment (stiepan) [reply dali/python/nvidia/dali/external_source.py:109]: Basically I am referring to something like collate_fn from pytorch https://pytorch.org/docs/stable/data.html that given callback_out that is [sample_from_batch1, sample_from_batch2, ...] leaves it to the user how exactly it should be transformed into tuple of batches.

Comment (mzient) [reply dali/python/nvidia/dali/external_source.py:109]: > Okok, so in the per-sample variant in case of two outputs, we'll be now necessarily returning tuple of (two) lists (each with length of batch size)
In a two-output, per-sample case the callback returns a tuple of two _samples_ (e.g. an image and a label).
In a two-output, per-batch case, the callback returns two _batches_ (e.g. a batch of images and a batch of labels).

> Basically I am referring to something like collate_fn 
I think it's not necessary at this point - when you want some specific way of assembling the batch, you can just feed the batches directly. It _might_ help in some cases when we have parallel processing, but I think that would be shifting the functionality of DALI pipeline into external source. We can discuss that later.

Comment (stiepan) [reply dali/python/nvidia/dali/external_source.py:109]: Okok!

**Review by stiepan — DISMISSED** [review]
Looks good to me:)

File: dali/python/nvidia/dali/external_source.py [hunk 3]
```diff
@@ -251,10 +277,14 @@ class ExternalSource():
     For the GPU, to avoid extra copy, the provided buffer must be contiguous. If you provide a list
     of separate Tensors, there will be an additional copy made internally, consuming both memory
     and bandwidth.
+
+`batch` : optional
+    If set to ``True`` or ``None``, the ``source`` is expected to produce an entire batch at once.
+    If set to ``False``, the ``source`` is called per-sample. Its first parameter is sample index.
```
szalpal, line 283: How would you know the batch size in this case?

dali/python/nvidia/dali/external_source.py
```diff
@@ -251,10 +277,14 @@ class ExternalSource():
...
 """
 
     def __init__(self, source = None, num_outputs = None, *, cycle = None, layout = None, name = None, device = "cpu",
-                 cuda_stream = None, use_copy_kernel = None, **kwargs):
+                 cuda_stream = None, use_copy_kernel = None, batch = None, **kwargs):
```
szalpal, line 287: How about calling it `batch_mode`? Seems like a better description of what it does

szalpal (commented):
Please add also a test cases with the generator

dali/python/nvidia/dali/external_source.py
mzient: I think we already boolean argument called `batch`.

mzient:
!build

mzient:
!build

dali/python/nvidia/dali/external_source.py
```diff
@@ -62,31 +63,56 @@ def __next__(self):
...
+    def call_and_feed(self, pipeline, batch_size):
+        try:
+            if self.batch:
+                callback_out = self.callback(*self.callback_args(None))
+            else:
+                callback_out = [self.callback(*self.callback_args(i)) for i in range(batch_size)]
+            self.current_sample += batch_size
+            self.current_iter += 1
+        except StopIteration:
+            self.current_sample = 0
```
stiepan, line 105: Could it be self.reset_indices() call?

stiepan (approved):

dali/python/nvidia/dali/external_source.py
mzient: It could. I guess I won't re-run CI just for that, so I'll wait for another review - if there are some more changes requested, I'll adjust this one.

dali/python/nvidia/dali/external_source.py
mzient: The whole point is to not have to know it - you just produce consecutive samples and that's it.

dali/python/nvidia/dali/external_source.py
```diff
@@ -159,11 +185,16 @@ class ExternalSource():
...
-    If the source is a callable that accepts a positional argument, it is assumed to be the current
+    the source can supply one or more data items. The data item can be a whole batch (default) or
+    a single batch entry (when ``batch==False``). If ``num_outputs`` is not set, the ``source``
+    is expected to return one item (a batch or a sample). If this value is specified (even if its
+    value is 1), the data is expected to a be tuple, or list, where each element corresponds to
+    respective return value of the external_source.
+
+    A per-batch source may accept one positional argument. If it does, it is the current iteration
     iteration number and consecutive calls will be ``source(0)``, ``source(1)``, and so on.
+    A per-sample source may accept a :class:`nvidia.dali.types.SampleInfo` structure.
```
JanuszL, line 196: May or should?

dali/python/nvidia/dali/external_source.py
JanuszL: It is for normalized, but I don't know if we can call it a corresponding example.

dali/python/nvidia/dali/pipeline.py
```diff
@@ -804,6 +804,9 @@ def reset(self):
             self._first_iter = True
             self._last_iter = False
             self._iter = 0
+            if self._input_callbacks:
+                for group in self._input_callbacks:
+                    group.reset_indices()
```
JanuszL, line 809: Isn't that a breaking change? So far we haven't reset the supplied indices.

dali/python/nvidia/dali/external_source.py
```diff
@@ -159,11 +185,16 @@ class ExternalSource():
 
     The source is polled for data (via a call ``source()`` or ``next(source)``)
     when the pipeline needs input for the next iteration. Depending on the value of ``num_outputs``,
-    the source can supply one or more data batches. If ``num_outputs`` is not set, the ``source``
-    is expected to return one batch. If this value is specified, the data is expected to a be tuple,
-    or list, where each element corresponds to respective return value of the external_source.
-    If the source is a callable that accepts a positional argument, it is assumed to be the current
+    the source can supply one or more data items. The data item can be a whole batch (default) or
```
JanuszL, line 188: As we previously talked I would put the list of supported types here as well, I bet no one check fee_inputs looking for it.

dali/python/nvidia/dali/external_source.py
mzient: May = if it takes a parameter, then it's that structure. But it may also take no parameters at all.

dali/python/nvidia/dali/pipeline.py
mzient: Maybe it is, but it seriously doubt anyone has ever used the callback with argument - I don't think it's shown in any example.

dali/python/nvidia/dali/external_source.py
mzient: In any case, the documentation is rather clear. I'll change `batch : optional` to  `batch : bool, optional`.

dali/python/nvidia/dali/external_source.py
mzient: Done.

mzient:
!build

dali/python/nvidia/dali/external_source.py
```diff
@@ -199,8 +242,7 @@ class ExternalSource():
     Used when feeding the data in ``iter_setup`` and can be omitted if
     the data is provided by ``source``.
 
-
```
mzient, line 202: Docs were broken with this extra new line...

JanuszL (approved):

PR #2470, opened 2020-11-16
Hide hidden ops and improve Enum docs quality

klecki:
#### Why we need this PR?
- Improve the docs quality (mostly for Enums)
- Truly hide hidden ops 

 - What solution was applied:
     Custom documenter for enums was registered in Sphinx generator to bridge the gap between PyBind Enums and true Python Enums.
     Used mechanism to skip part of auto generated docstring that listed py::enum_ members as plain text.
	Ops that have hidden docs are now imported under true .hidden module and appear as "reimported" into parent module. Sphinx skips them from listing.
     Sphinx doc generator
     ops.py, fn.py for loading hidden Ops.
    The API for hiding and how it is used     
 - Validation and testing:
     PLZ :eyes:
     Local experiments with loading DALI.
     Yes

**JIRA TASK**: *[DALI-1725]*

### DIFF
--- a/dali/python/backend_impl.cc
+++ b/dali/python/backend_impl.cc
@@ -1099,7 +1099,7 @@ PYBIND11_MODULE(backend_impl, m) {
   types_m.add_object("CPU_ONLY_DEVICE_ID", PyLong_FromLong(CPU_ONLY_DEVICE_ID));
 
   // DALIDataType
-  py::enum_<DALIDataType>(types_m, "DALIDataType", "Data type of image")
+  py::enum_<DALIDataType>(types_m, "DALIDataType", "Data type of image.\n<SPHINX_IGNORE>")
     .value("NO_TYPE",       DALI_NO_TYPE)
     .value("UINT8",         DALI_UINT8)
     .value("UINT16",        DALI_UINT16)
@@ -1131,7 +1131,7 @@ PYBIND11_MODULE(backend_impl, m) {
     .export_values();
 
   // DALIImageType
-  py::enum_<DALIImageType>(types_m, "DALIImageType", "Image type")
+  py::enum_<DALIImageType>(types_m, "DALIImageType", "Image type\n<SPHINX_IGNORE>")
     .value("RGB", DALI_RGB)
     .value("BGR", DALI_BGR)
     .value("GRAY", DALI_GRAY)
@@ -1140,7 +1140,7 @@ PYBIND11_MODULE(backend_impl, m) {
     .export_values();
 
   // DALIInterpType
-  py::enum_<DALIInterpType>(types_m, "DALIInterpType", "Interpolation mode")
+  py::enum_<DALIInterpType>(types_m, "DALIInterpType", "Interpolation mode\n<SPHINX_IGNORE>")
     .value("INTERP_NN", DALI_INTERP_NN)
     .value("INTERP_LINEAR", DALI_INTERP_LINEAR)
     .value("INTERP_CUBIC", DALI_INTERP_CUBIC)
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -256,12 +256,11 @@ class EnumDocumenter(ClassDocumenter):
     # Produce .. py:class:: fields in the RST doc
     directivetype = 'class'
 
-    def __init__(self, *args: Any) -> None:
+    def __init__(self, *args):
         super().__init__(*args)
 
     @classmethod
-    def can_document_member(cls, member: Any, membername: str, isattr: bool, parent: Any
-                            ) -> bool:
+    def can_document_member(cls, member, membername, isattr, parent):
         """Verify that we handle only the registered DALI enums. Pybind doesn't subclass Enum class,
         so we need an explicit list.
         """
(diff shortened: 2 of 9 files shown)


klecki:
!build

dali/python/backend_impl.cc
```diff
@@ -1099,7 +1099,7 @@ PYBIND11_MODULE(backend_impl, m) {
   types_m.add_object("CPU_ONLY_DEVICE_ID", PyLong_FromLong(CPU_ONLY_DEVICE_ID));
 
   // DALIDataType
-  py::enum_<DALIDataType>(types_m, "DALIDataType", "Data type of image")
+  py::enum_<DALIDataType>(types_m, "DALIDataType", "Data type of image.\n<SPHINX_IGNORE>")
```
JanuszL, line 1102: Meh

docs/conf.py
```diff
@@ -238,7 +241,91 @@
...
+    def __init__(self, *args: Any) -> None:
+        super().__init__(*args)
+
+    @classmethod
+    def can_document_member(cls, member: Any, membername: str, isattr: bool, parent: Any
+                            ) -> bool:
+        """Verify that we handle only the registered DALI enums. Pybind doesn't subclass Enum class,
+        so we need an explicit list.
+        """
+        return membername in _dali_enums and isinstance(parent, ClassDocumenter)
```
JanuszL, line 269: Do we need to check `isinstance(parent, ClassDocumenter)` ?
As `EnumDocumenter` inherits from `ClassDocumenter`, can parent be different?

docs/conf.py
```diff
@@ -238,7 +241,91 @@
...
+                return member_value.value
+            else:
+                return int(member_value)
+        filtered.sort(key = get_member_value)
+
+        return filtered
+
+    def sort_members(self, documenters, order):
+        """Ignore the order. Here we have access only to documenters that carry the name
+        and not the object. We need to sort based on the enum values and we du it in
```
JanuszL, line 291:
```suggestion
        and not the object. We need to sort based on the enum values and we do it in
```

docs/conf.py
```diff
@@ -238,7 +241,91 @@
...
+        self.filter_members()
+        """
+        return documenters
+
+class EnumAttributeDocumenter(AttributeDocumenter):
+    # Give us higher priority over Sphinx native AttributeDocumenter
+    priority = 12
```
JanuszL, line 298: Why not 14?

docs/conf.py
```diff
@@ -238,7 +241,91 @@
...
+        with just their value.
+        """
+        super(AttributeDocumenter, self).add_directive_header(sig)
+        # sourcename = self.get_sourcename()
+        # # print(self.object)
+        # if isinstance(self.object, Enum):
+        #     objrepr = str(self.object)
+        # else:
+        #     objrepr = str(self.object)
+        # self.add_line('   :value: ' + objrepr, sourcename)
```
JanuszL, line 319: ?

docs/conf.py
```diff
@@ -238,7 +241,91 @@
...
+    objtype = 'enum'
+    # Produce .. py:class:: fields in the RST doc
+    directivetype = 'class'
+    # priority = 10
+
+    def __init__(self, *args: Any) -> None:
+        super().__init__(*args)
+
+    @classmethod
+    def can_document_member(cls, member: Any, membername: str, isattr: bool, parent: Any
```
JanuszL, line 264: Do you need to use input/output annotations here? I don't see that in other methods/functions.

dali/python/backend_impl.cc
klecki: I can probably change the docstring in the documenter of enum by hand, knowing that I need to skip everything starting from the `Members:`, but if we update pybind it can change.

docs/conf.py
klecki: It can be anything bigger than 10 (AttributeDocumenter priority) or 11 (more specialized like InstanceAttributeDocumenter, though I'm not sure if we need to be higher than them).
https://github.com/sphinx-doc/sphinx/blob/7db3633778a24c4fc90c8a0ceb2318d6df7cf4e3/sphinx/ext/autodoc/__init__.py#L2027

So I've picked 12. I can make it 42 if you want.

docs/conf.py
klecki: Leftover when I was overriding sphinx code. Will remove them.

docs/conf.py
JanuszL: 12 sound good in that case. Just please add more comment how it was picked and how to adjust if anything changes in Sphinx.

dali/python/backend_impl.cc
JanuszL: Ok, let us leave it as it is.

docs/conf.py
klecki: I'm not sure, but at some point a lot of elements were checked against this class, I can check if it's really necessary, but maybe it's safer to just keep it.

JanuszL: Sure.

klecki:
!build

docs/conf.py
klecki: Removed

docs/conf.py
klecki: Done

dali/python/backend_impl.cc
mzient: But won't it display this garbage in Python's help or IDEs?

JanuszL (approved):

dali/python/backend_impl.cc
klecki: I wouldn't call it garbage, no more than what the pybind does itself. We already sprinkle the docs with loads of Sphinx tags a annotations, and the main use case is the HTML page.

mzient: Still meh, but so be it.

mzient (approved):

PR #2481, opened 2020-11-18
iter-to-iter variable batch size

szalpal:
TODO:
- [x] Add deprecation warning to `pipeline.py::batch_size(self)` and add `@property max_batch_size()` 
- [x] Update documentation to `batch_size` arg in `pipeline.py`
- [x] Remove external_source tests that fail on variable batch size
- [x] Adjust `_check_data_batch` in `external_source.py`: expected batch size condition and error message (~L24)
- [x] https://github.com/NVIDIA/DALI/pull/2408#discussion_r537532851
- [x] https://github.com/NVIDIA/DALI/pull/2408#discussion_r537484299
- [x] figure out inputless ops
- [x] `ImageDecoder` rework
- [x] C API
- [x] Document, how `max_batch_size` setting might explode the memory consumption

### What happened in this PR:
1. Update C API to support i2i variable batch size
2. Update nvJpegDecoder and its flavours to support variable BS
3. Update Constant op for the same reason
4. Partially split Executor into `h` and `cc` file, to make development convenient
5. Also add variable batch size routines to Executor (batch size queues, PreRun with BS inferring from the graph)
6. Introduce `BatchSizeProvider` interface and add it to `ExternalSource`
7. Add a feature to `CachingList` (inner `ExternalSource` memory), so that it can traverse over the data list asynchronously w.r.t current head and tail of this list
8. Modify constraint that stipulates constant batch size to stipulate uniform batch size within single iteration
9. Python API adjustment
10. Adding enormous test for i2i variable batch size (every op is tested)

#### Why we need this PR?
*Pick one, remove the rest*
- To enable the i2i variable batch size

### DIFF
--- a/dali/c_api/c_api.cc
+++ b/dali/c_api/c_api.cc
@@ -35,13 +35,37 @@ namespace {
 
 bool dali_initialized = false;
 
-template<typename Backend>
+/**
+ * Maps operator name to the batch size set prior to daliSetExternal... call.
+ * Typically, this operator will be BatchSizeProvider.
+ * Negative values denote max batch size (default state).
+ * Typical usage:
+ * auto batch_sizes_map = reinterpret_cast<batch_size_map_t*>(handle->batch_sizes_map);
+ */
+using batch_size_map_t = std::unordered_map<std::string /* op_name */, int /* batch_size */>;
+
+
+int PopCurrBatchSize(batch_size_map_t *batch_size_map, int max_batch_size,
+                     const std::string &op_name) {
+  auto it = batch_size_map->find(op_name);
+  auto exists = it != batch_size_map->end();
+  auto ret = !exists || it->second < 0 ? max_batch_size : it->second;
+  if (exists) {
+    it->second = -1;
+  }
+  return ret;
+}
+
+
+template <typename Backend>
 void SetExternalInput(daliPipelineHandle *pipe_handle, const char *name, const void *data_ptr,
                       dali_data_type_t data_type, const int64_t *shapes, int sample_dim,
                       const char *layout_str, cudaStream_t stream = 0, unsigned int flags = 0) {
   dali::Pipeline *pipeline = reinterpret_cast<dali::Pipeline *>(pipe_handle->pipe);
-  std::vector<int64_t> shapes_tmp(shapes, shapes + sample_dim * pipeline->batch_size());
-  dali::TensorListShape<> tl_shape(std::move(shapes_tmp), pipeline->batch_size(), sample_dim);
+  auto bs_map = reinterpret_cast<batch_size_map_t *>(pipe_handle->batch_sizes_map);
+  auto curr_batch_size = PopCurrBatchSize(bs_map, pipeline->max_batch_size(), name);
+  std::vector<int64_t> shapes_tmp(shapes, shapes + sample_dim * curr_batch_size);
+  dali::TensorListShape<> tl_shape(std::move(shapes_tmp), curr_batch_size, sample_dim);
   dali::TensorLayout layout{};
   if (layout_str != nullptr) {
     layout = dali::TensorLayout(layout_str);
@@ -68,16 +92,18 @@ void SetExternalInputTensors(daliPipelineHandle *pipe_handle, const char *name,
                              const int64_t *shapes, int64_t sample_dim, const char *layout_str,
                              cudaStream_t stream = 0, unsigned int flags = 0) {
   dali::Pipeline *pipeline = reinterpret_cast<dali::Pipeline *>(pipe_handle->pipe);
-  std::vector<int64_t> shapes_tmp(shapes, shapes + sample_dim * pipeline->batch_size());
-  dali::TensorListShape<> tl_shape(std::move(shapes_tmp), pipeline->batch_size(), sample_dim);
+  auto bs_map = reinterpret_cast<batch_size_map_t *>(pipe_handle->batch_sizes_map);
+  auto curr_batch_size = PopCurrBatchSize(bs_map, pipeline->max_batch_size(), name);
+  std::vector<int64_t> shapes_tmp(shapes, shapes + sample_dim * curr_batch_size);
+  dali::TensorListShape<> tl_shape(std::move(shapes_tmp), curr_batch_size, sample_dim);
   dali::TensorLayout layout{};
   if (layout_str != nullptr) {
     layout = dali::TensorLayout(layout_str);
   }
-  dali::TensorVector<Backend> data(pipeline->batch_size());
+  dali::TensorVector<Backend> data(curr_batch_size);
   const auto &type_info = dali::TypeTable::GetTypeInfo(static_cast<dali::DALIDataType>(data_type));
   auto elem_sizeof = type_info.size();
-  for (int i = 0; i < pipeline->batch_size(); i++) {
+  for (int i = 0; i < curr_batch_size; i++) {
     // We cast away the const from data_ptr, as there is no other way of passing it to the
     // Tensor as we must also set the shape and type metadata.
     // The vector that we pass to pipeline is const.
@@ -113,22 +139,15 @@ void daliInitialize() {
   std::call_once(init_flag, init);
 }
 
-
-void daliCreatePipeline(daliPipelineHandle *pipe_handle,
-                        const char *serialized_pipeline,
-                        int length,
-                        int batch_size,
-                        int num_threads,
-                        int device_id,
-                        int separated_execution,
-                        int prefetch_queue_depth,
-                        int cpu_prefetch_queue_depth,
-                        int gpu_prefetch_queue_depth,
+void daliCreatePipeline(daliPipelineHandle *pipe_handle, const char *serialized_pipeline,
+                        int length, int max_batch_size, int num_threads, int device_id,
+                        int separated_execution, int prefetch_queue_depth,
+                        int cpu_prefetch_queue_depth, int gpu_prefetch_queue_depth,
                         int enable_memory_stats) {
   bool se = separated_execution != 0;
-  auto pipeline = std::make_unique<dali::Pipeline>(std::string(serialized_pipeline, length),
-                                                   batch_size, num_threads, device_id, true,
-                                                   prefetch_queue_depth, true);
+  auto pipeline =
+      std::make_unique<dali::Pipeline>(std::string(serialized_pipeline, length), max_batch_size,
+                                       num_threads, device_id, true, prefetch_queue_depth, true);
   pipeline->SetExecutionTypes(true, se, true);
   if (se) {
     pipeline->SetQueueSizes(cpu_prefetch_queue_depth, gpu_prefetch_queue_depth);
@@ -143,6 +162,9 @@ void daliCreatePipeline(daliPipelineHandle *pipe_handle,
   pipe_handle->ws = ws.release();
   pipe_handle->copy_stream = stream.release();
   pipe_handle->pipe = pipeline.release();
+
+  auto bs_map = std::make_unique<batch_size_map_t>();
+  pipe_handle->batch_sizes_map = bs_map.release();
 }
 
 
@@ -183,13 +205,19 @@ void daliPrefetchSeparate(daliPipelineHandle *pipe_handle,
 }
 
 
+void daliSetExternalInputBatchSize(daliPipelineHandle *pipe_handle, const char *name,
+                                   int batch_size) {
+  auto *bs_map = reinterpret_cast<batch_size_map_t *>(pipe_handle->batch_sizes_map);
+  (*bs_map)[name] = batch_size;
+}
+
+
 void daliSetExternalInput(daliPipelineHandle *pipe_handle, const char *name, device_type_t device,
                           const void *data_ptr, dali_data_type_t data_type, const int64_t *shapes,
                           int sample_dim, const char *layout_str, unsigned int flags) {
   daliSetExternalInputAsync(pipe_handle, name, device, data_ptr, data_type, shapes, sample_dim,
                             layout_str, pipe_handle->copy_stream, flags | DALI_ext_force_sync);
 }
-
 
 void daliSetExternalInputAsync(daliPipelineHandle *pipe_handle, const char *name,
                                device_type_t device, const void *data_ptr,
@@ -479,6 +507,7 @@ void daliCopyTensorListNTo(daliPipelineHandle *pipe_handle, void *dst, int outpu
 void daliDeletePipeline(daliPipelineHandle* pipe_handle) {
   dali::Pipeline *pipeline = reinterpret_cast<dali::Pipeline *>(pipe_handle->pipe);
   dali::DeviceWorkspace *ws = reinterpret_cast<dali::DeviceWorkspace *>(pipe_handle->ws);
+  auto *bs_map = reinterpret_cast<batch_size_map_t *>(pipe_handle->batch_sizes_map);
   DALI_ENFORCE(pipeline != nullptr && ws != nullptr, "Pipeline already deleted");
   if (pipe_handle->copy_stream) {
     CUDA_CALL(cudaStreamDestroy(pipe_handle->copy_stream));
@@ -486,8 +515,10 @@ void daliDeletePipeline(daliPipelineHandle* pipe_handle) {
   pipe_handle->copy_stream = nullptr;
   delete ws;
   delete pipeline;
+  delete bs_map;
   pipe_handle->ws = nullptr;
   pipe_handle->pipe = nullptr;
+  pipe_handle->batch_sizes_map = nullptr;
 }
 
 void daliLoadLibrary(const char* lib_path) {
--- a/dali/c_api/c_api_test.cc
+++ b/dali/c_api/c_api_test.cc
@@ -108,7 +108,8 @@ std::unique_ptr<Pipeline> GetTestPipeline(bool is_file_reader, const std::string
 // Allows only for uint8_t CPU/GPU output data to be compared
 template <typename Backend>
 void ComparePipelinesOutputs(daliPipelineHandle &handle, Pipeline &baseline,
-                             unsigned int copy_output_flags = DALI_ext_default) {
+                             unsigned int copy_output_flags = DALI_ext_default,
+                             int batch_size = dali::batch_size) {
   dali::DeviceWorkspace ws;
   baseline.Outputs(&ws);
   daliOutput(&handle);
@@ -230,6 +231,7 @@ TYPED_TEST(CApiTest, ExternalSourceSingleAllocPipe) {
     // Unnecessary copy in case of CPUBackend, makes the code generic across Backends
     input.Copy(input_cpu, cuda_stream);
     pipe_ptr->SetExternalInput(input_name, input);
+    daliSetExternalInputBatchSize(&handle, input_name.c_str(), input_shape.num_samples());
     daliSetExternalInputAsync(&handle, input_name.c_str(), backend_to_device_type<TypeParam>::value,
                               input.raw_data(), dali_data_type_t::DALI_UINT8, input_shape.data(),
                               input_shape.sample_dim(), nullptr, cuda_stream, DALI_ext_default);
@@ -261,8 +263,62 @@ TYPED_TEST(CApiTest, ExternalSourceSingleAllocPipe) {
 }
 
 
+TYPED_TEST(CApiTest, ExternalSourceSingleAllocVariableBatchSizePipe) {
+  TensorListShape<> reference_input_shape = {{37, 23, 3}, {12, 22, 3}, {42, 42, 3}, {8, 8, 3},
+                                             {64, 32, 3}, {32, 64, 3}, {20, 20, 3}, {64, 64, 3},
+                                             {10, 10, 3}, {60, 50, 3}, {10, 15, 3}, {48, 48, 3}};
+  int max_batch_size = reference_input_shape.num_samples();
+  std::vector<TensorListShape<>> trimmed_input_shapes = {
+      subshape(reference_input_shape, 0, max_batch_size / 2),
+      subshape(reference_input_shape, 0, max_batch_size / 4),
+      subshape(reference_input_shape, 0, max_batch_size),
+  };
+
+  auto pipe_ptr = GetTestPipeline<TypeParam>(false, this->output_device_);
+  auto serialized = pipe_ptr->SerializeToProtobuf();
+
+  daliPipelineHandle handle;
+  daliCreatePipeline(&handle, serialized.c_str(), serialized.size(), batch_size, num_thread,
+                     device_id, false, prefetch_queue_depth, prefetch_queue_depth,
+                     prefetch_queue_depth, false);
+
+  for (auto &input_shape : trimmed_input_shapes) {
+    pipe_ptr = GetTestPipeline<TypeParam>(false, this->output_device_);
+    pipe_ptr->Build();
+
+    TensorList<CPUBackend> input_cpu;
+    TensorList<TypeParam> input;
+    input_cpu.Resize(input_shape, TypeInfo::Create<uint8_t>());
+
+    for (int i = 0; i < prefetch_queue_depth; i++) {
+      SequentialFill(view<uint8_t>(input_cpu), 42 * i);
+      // Unnecessary copy in case of CPUBackend, makes the code generic across Backends
+      input.Copy(input_cpu, cuda_stream);
+      pipe_ptr->SetExternalInput(input_name, input);
+      daliSetExternalInputBatchSize(&handle, input_name.c_str(), input_shape.num_samples());
+      daliSetExternalInputAsync(&handle, input_name.c_str(),
+                                backend_to_device_type<TypeParam>::value, input.raw_data(),
+                                dali_data_type_t::DALI_UINT8, input_shape.data(),
+                                input_shape.sample_dim(), nullptr, cuda_stream, DALI_ext_default);
+    }
+
+    for (int i = 0; i < prefetch_queue_depth; i++) {
+      pipe_ptr->RunCPU();
+      pipe_ptr->RunGPU();
+    }
+    daliPrefetchUniform(&handle, prefetch_queue_depth);
+
+    dali::DeviceWorkspace ws;
+    for (int i = 0; i < prefetch_queue_depth; i++) {
+      ComparePipelinesOutputs<TypeParam>(handle, *pipe_ptr, DALI_ext_default,
+                                         input_shape.num_samples());
+    }
+  }
+}
+
+
 TYPED_TEST(CApiTest, ExternalSourceMultipleAllocPipe) {
-  TensorListShape<> input_shape = {{37, 23, 3}, {12, 22, 3}, {42, 42, 3}, {8,  8,  3},
+  TensorListShape<> input_shape = {{37, 23, 3}, {12, 22, 3}, {42, 42, 3}, {8, 8, 3},
                                    {64, 32, 3}, {32, 64, 3}, {20, 20, 3}, {64, 64, 3},
                                    {10, 10, 3}, {60, 50, 3}, {10, 15, 3}, {48, 48, 3}};
   TensorList<CPUBackend> input_cpu;
--- a/dali/operators/ssd/random_crop.cc
+++ b/dali/operators/ssd/random_crop.cc
@@ -34,7 +34,7 @@ cropped and valid bounding boxes and valid labels are returned.)code")
   .NumInput(3)   // [img, bbox, label]
   .NumOutput(3)  // [img, bbox, label]
   .AddOptionalArg("num_attempts", R"code(Number of attempts.)code", 1)
-  .Deprecate("RandomBBoxCrop");
+  .Deprecate("RandomBBoxCrop");  // deprecated in DALI 0.30
 
 /*
  * # This function is from https://github.com/kuangliu/pytorch-ssd.
--- a/dali/pipeline/data/type_traits.h
+++ b/dali/pipeline/data/type_traits.h
@@ -1,0 +1,69 @@
+// Copyright (c) 2020, NVIDIA CORPORATION. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef DALI_PIPELINE_DATA_TYPE_TRAITS_H_
+#define DALI_PIPELINE_DATA_TYPE_TRAITS_H_
+
+namespace dali {
+
+template <typename Backend>
+class TensorVector;
+
+template <typename Backend>
+class TensorList;
+
+class CPUBackend;
+class MixedBackend;
+class GPUBackend;
+
+template <typename Backend>
+struct is_backend {
+  static constexpr bool value = std::is_same<Backend, CPUBackend>::value ||
+                                std::is_same<Backend, MixedBackend>::value ||
+                                std::is_same<Backend, GPUBackend>::value;
+};
+
+template <template <typename> class MaybeTensorVector, typename Backend>
+struct is_tensor_vector {
+  static constexpr bool value =
+      std::is_same<MaybeTensorVector<Backend>, TensorVector<Backend>>::value;
+};
+
+template <template <typename> class MaybeTensorList, typename Backend>
+struct is_tensor_list {
+  static constexpr bool value = std::is_same<MaybeTensorList<Backend>, TensorList<Backend>>::value;
+};
+
+/**
+ * Verifies, that T is proper batch container for DALI
+ *
+ * Batch container is proper when it has a defined backend and is TensorVector or a TensorList
+ */
+template <template <typename Backend_> class T, typename Backend>
+struct is_batch_container {
+  static constexpr bool value =
+      is_backend<Backend>::value &&
+      (is_tensor_vector<T, Backend>::value || is_tensor_list<T, Backend>::value);
+};
+
+namespace test {
+static_assert(is_batch_container<TensorVector, CPUBackend>::value, "Test failed");
+static_assert(is_batch_container<TensorVector, GPUBackend>::value, "Test failed");
+static_assert(is_batch_container<TensorList, CPUBackend>::value, "Test failed");
+static_assert(is_batch_container<TensorList, GPUBackend>::value, "Test failed");
+}  // namespace test
+
+}  // namespace dali
+
+#endif  // DALI_PIPELINE_DATA_TYPE_TRAITS_H_
--- a/dali/pipeline/executor/executor.cc
+++ b/dali/pipeline/executor/executor.cc
@@ -1,4 +1,4 @@
-// Copyright (c) 2017-2018, NVIDIA CORPORATION. All rights reserved.
+// Copyright (c) 2020, NVIDIA CORPORATION. All rights reserved.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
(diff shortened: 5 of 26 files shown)


szalpal:
!build

szalpal:
!build

dali/pipeline/operator/builtin/external_source_test.cc
```diff
@@ -467,11 +467,69 @@ TEST(ExternalSourceTestNoInput, Throw) {
   vector<string> outputs = {"data_out_cpu"};
 
   exe->Build(&graph, outputs);
-  exe->RunCPU();
-  exe->RunMixed();
-  exe->RunGPU();
-  DeviceWorkspace ws;
-  ASSERT_THROW(exe->Outputs(&ws), std::runtime_error);
+  ASSERT_THROW(exe->RunCPU(), std::exception);
```
jantonguirao, line 470: can we narrow down the exception that is thrown?

dali/pipeline/pipeline.h
```diff
@@ -438,6 +440,13 @@ class DLL_PUBLIC Pipeline {
 
   DLL_PUBLIC DISABLE_COPY_MOVE_ASSIGN(Pipeline);
 
+  /**
+   * Maps operator name (typically ExternalSource) to the
```
jantonguirao, line 444: can it be anything different than external source??

dali/pipeline/workspace/workspace.h
```diff
@@ -189,7 +190,7 @@ class WorkspaceBase : public ArgumentWorkspace {
   /**
    * Set requested batch size for all outputs
    */
-  void set_batch_size(int batch_size) {
+  void SetBatchSizes(int batch_size) {
```
jantonguirao, line 193:
```suggestion
  void SetBatchSize(int batch_size) {
```
You are still setting one batch size, to all outputs

dali/python/nvidia/dali/pipeline.py
```diff
@@ -51,10 +51,13 @@ class Pipeline(object):
 Parameters
 ----------
 `batch_size` : int, optional, default = -1
-    Batch size of the pipeline. Negative values for this parameter
+    Maximum batch size of the pipeline. Negative values for this parameter
     are invalid - the default value may only be used with
     serialized pipeline (the value stored in serialized pipeline
-    is used instead).
+    is used instead). In most cases, the actual batch size of the pipeline
+    will be equal to the maximum one. Running DALI pipeline with batch size
```
jantonguirao, line 58: I got confused reading this. How about (just a suggestion):
```
Running the DALI Pipeline with a smaller batch size is also supported. The batch size might change from iteration to iteration.
```

dali/test/python/test_dali_variable_batch_size.py
```diff
@@ -0,0 +1,622 @@
...
+
+    def peek_image_shape_pipe(max_batch_size, input_data, device):
+        pipe = Pipeline(batch_size=max_batch_size, num_threads=4, device_id=0)
+        encoded = fn.external_source(source=input_data, num_outputs=1, cycle=False, device='cpu')
+        shape = fn.peek_image_shape(encoded, device=device)
+        pipe.set_outputs(shape)
+        return pipe
+
+    # TODO add tiff
+    image_decoder_extensions = ['.jpg', '.bmp', '.png', '.pnm', '.jp2']
```
jantonguirao, line 589: why TIFF is not included?

dali/test/python/test_dali_variable_batch_size.py
```diff
@@ -0,0 +1,622 @@
...
+    if isinstance(sample_shape, tuple):
+        size = sample_shape
+    elif inspect.isgeneratorfunction(sample_shape):
+        size = sample_shape()
+    else:
+        raise RuntimeError(
+            "`sample_shape` shall be either a tuple or a callable. Provide `(val,)` tuple for 1D shape")
+
+    if np.issubdtype(dtype, np.integer):
+        return [np.random.randint(lo, hi, size=(1, bs) + size, dtype=dtype) for bs in
```
jantonguirao, line 81:
```suggestion
        return [np.random.randint(lo, hi, size=(bs) + size, dtype=dtype) for bs in
```
You don't need the size=1 trailing dimension.

dali/test/python/test_dali_variable_batch_size.py
```diff
@@ -0,0 +1,622 @@
...
+    :param input_epoch: List of numpy arrays, where every item is a single batch
+    :param pipeline_fn: Function, that returns created (but not built) pipeline.
+                        Its signature should be (at least):
+                        pipeline_fn(max_batch_size, input_data, device, /, ...)
+    :param devices: Devices to run the check on
+    :param pipeline_fn_args: Additional args to pipeline_fn
+    """
+    for device in devices:
+        n_iter = len(input_epoch)
+        max_bs = max(batch.shape[1] for batch in input_epoch)
```
jantonguirao, line 120:
```suggestion
        max_bs = max(batch.shape[0] for batch in input_epoch)
```
If you followed my advice to remove the leading dimension

dali/test/python/test_dali_variable_batch_size.py
```diff
@@ -0,0 +1,622 @@
...
+    for op in random_ops:
+        print("\nTesting ", op)
+        random_op_helper(**op)
+
+
+def test_batch_permute():
+    def pipe(max_batch_size, input_data, device):
+        pipe = Pipeline(batch_size=max_batch_size, num_threads=4, device_id=0)
+        perm = fn.batch_permutation(seed=420)
+        data = fn.external_source(source=input_data, num_outputs=1, cycle=False, device=device)
```
jantonguirao, line 272:
```suggestion
        data = fn.external_source(source=input_data, cycle=False, device=device)
```
Then  your output will not be a tuple. You can remove all the `num_outputs=1` occurrences.

dali/pipeline/operator/batch_size_provider.h
```diff
@@ -0,0 +1,46 @@
...
+   * multiple times for the same batch, before Advance() invocation.
+   */
+  virtual int NextBatchSize() = 0;
+
+  /**
+   * Advances to next batch.
+   *
+   * On the event, that there's no further data available, Advance() shall throw std::out_of_range
+   */
+  virtual void Advance() = 0;
```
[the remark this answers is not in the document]
szalpal: I thought about it, but from my perspective that's inaccurate. This function does not advance the "next batch size". It does not advance "batch size" either. It advances to the next batch, so it could be this way: `AdvanceToNextBatch`. But that "next batch" is not the "next batch" from the `NextBatchSize` function. Anyway, I believe simple `Advance` with proper documentation is accurate enough.

dali/pipeline/operator/batch_size_provider.h
```diff
@@ -0,0 +1,46 @@
...
+   * Typically, you'll need to inherit from this interface virtually:
+   *
+   * template<typename Backend>
+   * class MyOperatorBsp : public OperatorBase<Backend>, virtual public BatchSizeProvider {};
+   */
+ public:
+  /**
+   * Returns next batch size.
+   *
+   * Implementation shall assure, that it's possible to call NextBatchSize()
```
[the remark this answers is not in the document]
szalpal: According to [this](https://www.myenglishteacher.eu/blog/english-grammar-shall-vs-should-with-examples/), shall is the word here. Dunno how may I believe this source of truth, it's just first record in my DDG search ;)

dali/pipeline/operator/builtin/external_source_test.cc
szalpal: Unfortunately no. `RunCPU` throws both `runtime_error` and `logic_error`, so `exception` is first common parent

dali/pipeline/pipeline.h
szalpal: Yes, in fact it will be `BatchSizeProvider` op. I'll update the comment

dali/python/nvidia/dali/pipeline.py
szalpal: Done

dali/test/python/test_dali_variable_batch_size.py
szalpal: My bad, fixed

dali/pipeline/executor/executor.cc
```diff
@@ -28,6 +28,301 @@
...
+void Executor<WorkspacePolicy, QueuePolicy>::PreRun() {
+  auto batch_size = InferBatchSize(batch_size_providers_);
+  batch_sizes_cpu_.push(batch_size);
+  batch_sizes_mixed_.push(batch_size);
+  batch_sizes_gpu_.push(batch_size);
+}
+
+template <typename WorkspacePolicy, typename QueuePolicy>
+void Executor<WorkspacePolicy, QueuePolicy>::RunCPU() {
+  PreRun();
```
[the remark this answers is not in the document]
szalpal: I'd prefer to extract it. After all, there might be more to PreRun in the future...

dali/pipeline/executor/executor.cc
```diff
@@ -28,6 +28,301 @@
...
+    auto bsp = dynamic_cast<BatchSizeProvider *>(graph_->Node(i).op.get());
+    if (!bsp) continue;
+    batch_size_providers_.emplace_back(bsp);
+  }
+}
+
+
+template <typename WorkspacePolicy, typename QueuePolicy>
+int Executor<WorkspacePolicy, QueuePolicy>::InferBatchSize(
+    const std::vector<BatchSizeProvider *> &batch_size_providers) const {
```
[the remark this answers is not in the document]
szalpal: Done

dali/pipeline/workspace/workspace.h
szalpal: My way of thinking here is, that it is in fact single value of batch size, but it sets it for number of outputs. Possibly in future there might be more value**s** of batch sizes, e.g. one per every output.

After all, that's what this function does:
```c++
    batch_sizes_.clear();
    batch_sizes_.resize(NumOutput(), batch_size);
```
fills `batch_sizes_` container with (multiple) value**s**

dali/operators/ssd/random_crop.cc
```diff
@@ -34,7 +34,7 @@ cropped and valid bounding boxes and valid labels are returned.)code")
   .NumInput(3)   // [img, bbox, label]
   .NumOutput(3)  // [img, bbox, label]
   .AddOptionalArg("num_attempts", R"code(Number of attempts.)code", 1)
-  .Deprecate("RandomBBoxCrop");
+  .Deprecate("RandomBBoxCrop");  // deprecated in DALI 0.30
```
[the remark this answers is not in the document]
szalpal: Actually, this was deprecated on 7th December 2020 ;)
image

dali/c_api/c_api.cc
```diff
@@ -35,13 +35,26 @@ namespace {
 
 bool dali_initialized = false;
 
-template<typename Backend>
+
+int GetCurrBatchSize(dali::Pipeline *pipe, std::string op_name) {
```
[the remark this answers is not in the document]
szalpal: Done

dali/c_api/c_api.cc
```diff
@@ -35,13 +35,26 @@ namespace {
...
+  auto it = pipe->batch_sizes_c_api_.find(op_name);
+  auto exists = it != pipe->batch_sizes_c_api_.end();
+  auto ret = !exists || it->second < 0 ? pipe->max_batch_size() : it->second;
+  if (exists) {
```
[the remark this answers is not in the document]
szalpal: But these 2 functions reuse the same code :( How about changing the name to `PopCurrBatchSize()`?

dali/operators/decoder/nvjpeg/decoupled_api/nvjpeg_decoder_decoupled_api.h
```diff
@@ -206,6 +198,8 @@ class nvJPEGDecoder : public Operator<MixedBackend>, CachedDecoderImpl {
     }
 #endif  // NVJPEG2K_ENABLED
 
+    PrepareMetadataContainers();
```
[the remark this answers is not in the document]
szalpal: How about `ReserveSampleContainers`?

dali/test/python/test_dali_variable_batch_size.py
szalpal: Done

dali/test/python/test_dali_variable_batch_size.py
szalpal: Done

dali/test/python/test_dali_variable_batch_size.py
szalpal: Done

dali/pipeline/executor/executor.cc
```diff
@@ -28,6 +28,301 @@
...
+  if (bsps.empty()) {
+    return max_batch_size_;
+  }
+  for (size_t i = 1; i < bsps.size(); i++) {
+    DALI_ENFORCE(bsps[0]->NextBatchSize() == bsps[i]->NextBatchSize(),
+                 "Batch size must be uniform across an iteration");
+  }
+  auto batch_size = bsps[0]->NextBatchSize();
+  assert(batch_size > 0);
+  bsps[0]->Advance();
```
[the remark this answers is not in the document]
szalpal: Done

dali/operators/decoder/nvjpeg/decoupled_api/nvjpeg_decoder_decoupled_api.h
```diff
@@ -909,6 +893,35 @@ class nvJPEGDecoder : public Operator<MixedBackend>, CachedDecoderImpl {
     RegisterDiagnostic("using_hw_decoder", &using_hw_decoder_);
   }
 
+  int CalcHwDecoderBatchSize(float hw_decoder_load, int curr_batch_size) {
+    auto hw_batch_size = static_cast<int>(std::round(hw_decoder_load * curr_batch_size));
```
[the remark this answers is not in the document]
szalpal: Done

dali/operators/decoder/nvjpeg/decoupled_api/nvjpeg_decoder_decoupled_api.h
```diff
@@ -73,23 +73,15 @@ class nvJPEGDecoder : public Operator<MixedBackend>, CachedDecoderImpl {
                      spec.GetArgument<bool>("affine")) {
 #if NVJPEG_VER_MAJOR >= 11
     // if hw_decoder_load is not present in the schema (crop/sliceDecoder) then it is not supported
+    bool try_init_hw_decoder = false;
```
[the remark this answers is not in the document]
szalpal: You might be right. I don't have hw decoder at my desk, so I'd say this would deserve separate task to write a test and verify it.

szalpal:
!build

dali/test/python/test_dali_variable_batch_size.py
```diff
@@ -0,0 +1,620 @@
...
+    def resize(data):
+        data += 13
+        return data
+
+    def pipe(max_batch_size, input_data, device, /):
+        pipe = Pipeline(batch_size=max_batch_size, num_threads=4, device_id=0, exec_async=False,
+                        exec_pipelined=False)
+        with pipe:
+            data = fn.external_source(source=input_data, cycle=False, device=device)
+            processed = fn.python_function(data, function=resize, num_outputs=1)
```
jantonguirao, line 616:
```suggestion
            processed = fn.python_function(data, function=resize)
```

dali/pipeline/executor/executor.cc
```diff
@@ -28,6 +28,302 @@
...
+int Executor<WorkspacePolicy, QueuePolicy>::InferBatchSize(
+    const std::vector<BatchSizeProvider *> &bsps) const {
+  if (bsps.empty()) {
+    return max_batch_size_;
+  }
+  for (size_t i = 1; i < bsps.size(); i++) {
+    DALI_ENFORCE(bsps[0]->NextBatchSize() == bsps[i]->NextBatchSize(),
+                 "Batch size must be uniform across an iteration");
+  }
+  auto batch_size = bsps[0]->NextBatchSize();
```
jantonguirao, line 321: suggestion: move the variable to line 317 and use it in the loop

jantonguirao (approved):

szalpal:
!build

szalpal:
!build

include/dali/c_api.h
```diff
@@ -96,24 +96,18 @@ DLL_PUBLIC void daliInitialize();
 
 /// @{
 /**
- * @brief Create DALI pipeline. Setting batch_size,
+ * @brief Create DALI pipeline. Setting max_batch_size,
```
banasraf, line 99: I don't understand the naming of this parameter. It's called max_batch_size, but there is no constrain on the External Input batch size provided by the user. Should be this parameter (and the corresponding field inside the Pipeline class) called something like *default batch size*? I don't see this *max* having any meaning here.

dali/python/nvidia/dali/pipeline.py
```diff
@@ -51,10 +51,15 @@ class Pipeline(object):
 Parameters
 ----------
 `batch_size` : int, optional, default = -1
-    Batch size of the pipeline. Negative values for this parameter
+    Maximum batch size of the pipeline. Negative values for this parameter
```
banasraf, line 54: Again, does this need to be called maximum batch size? This will encourage  users to set it to some super high values and it's especially misleading when the readers still don't support variable batch size and will always read the maximal batch size. 'Default batch size' would be better here, imo.

include/dali/c_api.h
szalpal: My intention was that if user tries to feed batch size bigger than `max_batch_size`, he'll get an error. The reason is that DALI preallocates data according to this value, so setting it too high explodes memory consumption.

This strategy also exists in Triton, where you provide `max_batch_size` for given model. I guess it's for the same reason - to preallocate some of buffers needed

dali/python/nvidia/dali/pipeline.py
szalpal: As above, setting `max_batch_size` allows for preallocations, which wouldn't be possible with `default_batch_size`.

Additionally, with `max_batch_size` we'll be able to catch some part of OOM errors, which in case of `default` one would emerge only after particular amount of training epochs. I guess user would appreciate knowing this beforehand ;)

include/dali/c_api.h
banasraf: Is there any `ENFORCE` for `batch_size < max_batch_size`? I can't see it anywhere.

dali/pipeline/operator/batch_size_provider.h
banasraf, line 31:
```suggestion
   * Implementation shall assure that it's possible to call NextBatchSize()
```

dali/pipeline/operator/batch_size_provider.h
```diff
@@ -0,0 +1,46 @@
...
+   * multiple times for the same batch, before Advance() invocation.
+   */
+  virtual int NextBatchSize() = 0;
+
+  /**
+   * Advances to next batch.
+   *
+   * On the event, that there's no further data available, Advance() shall throw std::out_of_range
```
banasraf, line 39:
```suggestion
   * When there's no further data available, Advance() shall throw std::out_of_range
```

dali/pipeline/data/type_traits.h
```diff
@@ -0,0 +1,74 @@
...
+
+template <typename Backend>
+class TensorList;
+
+class CPUBackend;
+class MixedBackend;
+class GPUBackend;
+
+template <typename Backend>
+struct is_backend {
```
banasraf, line 31: Maybe?
```c++
template <typename Backend>
struct is_backend {
  static constexpr bool value = 
    std::is_same<Backend, CPUBackend>::value || std::is_same<Backend, MixedBackend>::value ||
    std::is_same<Backend, GPUBackend>::value;
};
```
It seems like a tool that could be reused so we should make it more like standard traits.

Also this way of writing it imposes some inconvenience: 
```c++
bool b = is_backend<int>();   // <-- won't compile, cannot convert
```

dali/test/python/test_utils.py
```diff
@@ -435,3 +434,30 @@ def np_type_to_dali(type):
         np.float64: types.FLOAT64,
     }
     return np_types_to_dali_dict[type]
+
+
+def read_file_bin(filename):
```
banasraf, line 439: `numpy.fromfile` does this
```python
data = np.fromfile(filename, dtype='uint8')
```

szalpal: Good point

dali/pipeline/data/type_traits.h
szalpal: Done

dali/pipeline/operator/batch_size_provider.h
szalpal: Done

dali/pipeline/operator/batch_size_provider.h
szalpal: Done

include/dali/c_api.h
szalpal: Good point. Added to `RunHelper`

banasraf (approved):

dali/pipeline/operator/op_spec.h
```diff
@@ -441,16 +441,7 @@ class DLL_PUBLIC OpSpec {
...
-        shape.num_elements() == batch_size &&
-        (shape.num_samples() == batch_size || shape.num_samples() == 1);
-    if (should_throw) {
-      DALI_ENFORCE(valid_shape, make_string(
-                   "Unexpected shape of argument \"", name, "\". Expected batch of ", batch_size,
-                   " scalars or a batch of tensors containing one element per sample or a single "
-                   "tensor with ", batch_size, " elements. Got:\n", shape));
-    }
-    return valid_shape;
+    return true;
```
klecki, line 444: This worries me. Why don't you do any verification? Where is this used?

dali/pipeline/operator/builtin/external_source.h
```diff
@@ -84,13 +98,62 @@ class CachingList {
...
+    return *prophet_;
+  }
+
+  void AdvanceProphet() {
+    if (prophet_ == full_data_.end())
+      throw std::out_of_range(
+          "Attempted to step over the last element in the list. This operation is forbidden. Add "
+          "more elements to CachingList before calling AdvanceProphet.");
+    apprentice_ = prophet_;
+    advance(prophet_, 1);
```
klecki, line 125: What's wrong with `prophet_++`?
Also, what if our prophet (or `apprentice_`) is popped out of the list and put in the empty list? Wouldn't we start advancing in the set of empty nodes?

dali/pipeline/operator/builtin/external_source.h
```diff
@@ -415,9 +494,10 @@ class ExternalSource : public Operator<Backend> {
    */
   struct ExternalSourceState {
     bool copied_shared_data = false;
+    size_t batch_size = 0;
```
klecki, line 497: I see only modifications putting the current bs in, but no access to it?

dali/test/python/test_dali_variable_batch_size.py
```diff
@@ -0,0 +1,620 @@
...
+and compare them with ad-hoc created Pipelines for given (constant) batch sizes.
+This can be easily done using `check_batch` function below.
+
+On top of that, there are some utility functions and routines to help with some
+common cases:
+1. If the operator is typically processing image-like data (i.e. 3-dim, uint8,
+   0-255, with shape like [640, 480, 3]) and you want to test default arguments
+   only, please add a record to the `ops_image_default_args` list
+2. If the operator is typically processing image-like data (i.e. 3-dim, uint8,
+   0-255, with shape like [640, 480, 3]) and you want to specify any number of
```
klecki, line 41: How about more demanding scenarios (where maybe the shape is a bit more fuzzy?).
Also, didn't look for it yet, but do you test stuff like FCHW sequences? There might be some weirdness in sequence processing operators.

klecki (commented):
Few files still missing, I would like to check a bit more about the side-casts on my own.
My major gripe is mixing the C-API with the C++-API, not enough tests for the former one.

Also the prophet part probably doesn't happen in any scenario, but looking at the isolated CachingList it looks like a bug.

dali/c_api/c_api.cc
```diff
@@ -35,13 +35,26 @@ namespace {
 
 bool dali_initialized = false;
 
-template<typename Backend>
+
+int PopCurrBatchSize(dali::Pipeline *pipe, const std::string& op_name) {
+  auto it = pipe->batch_sizes_c_api_.find(op_name);
+  auto exists = it != pipe->batch_sizes_c_api_.end();
+  auto ret = !exists || it->second < 0 ? pipe->max_batch_size() : it->second;
```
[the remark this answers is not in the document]
szalpal: `||` has precedence over ternary, so when `!exists` is true, the `ret` is `pipe->max_batch_size()` :)

dali/pipeline/executor/executor.cc
```diff
@@ -28,6 +28,304 @@
...
+  auto &op = *op_node.op;
+  output_desc.clear();
+  const auto &spec = op.GetSpec();
+  const auto &schema = spec.GetSchema();
+  SmallVector<int, 16> empty_layout_in_idxs;
+
+  for (int i = 0; i < spec.NumInput(); i++) {
+    DALI_ENFORCE(ws.GetInputBatchSize(i) <= max_batch_size_,
+                 make_string("Expected batch size lower or equal to max batch size. Actual: ",
+                             ws.GetInputBatchSize(i), " <= ", max_batch_size_));
```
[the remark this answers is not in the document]
szalpal: Done

dali/c_api/c_api.cc
```diff
@@ -183,14 +190,20 @@ void daliPrefetchSeparate(daliPipelineHandle *pipe_handle,
 }
 
 
+void daliSetExternalInputBatchSize(daliPipelineHandle *pipe_handle, const char *name,
+                                   int batch_size) {
+  auto pipeline = reinterpret_cast<dali::Pipeline *>(pipe_handle->pipe);
```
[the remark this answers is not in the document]
szalpal: Done

dali/c_api/c_api.cc
```diff
@@ -183,14 +190,20 @@ void daliPrefetchSeparate(daliPipelineHandle *pipe_handle,
...
+  pipeline->batch_sizes_c_api_[name] = batch_size;
```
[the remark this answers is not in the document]
szalpal: But then I can't use `std::map`, because it's in C header :(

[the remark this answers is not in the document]
szalpal: Added with uuuugly casting

include/dali/c_api.h
```diff
@@ -135,15 +129,26 @@ enum {
...
   /**
    * If provided, a CUDA copy kernel will be used to feed external source instead of cudaMemcpyAsync
    * Only relevant when the input is either pinned host memory or device memory
    */
-  DALI_use_copy_kernel = (1<<2),
+  DALI_use_copy_kernel = (1 << 2),
 };
 
+/**
+ * @brief Set the batch size for the upcoming call to `daliSetExternalInput*(...)`
```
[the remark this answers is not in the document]
szalpal: I believe it's clear from reading doc for this function and `daliSetExternalInput...`

dali/pipeline/executor/executor.cc
```diff
@@ -28,6 +28,304 @@
...
+void Executor<WorkspacePolicy, QueuePolicy>::PreRun() {
+  auto batch_size = InferBatchSize(batch_size_providers_);
+  batch_sizes_cpu_.push(batch_size);
+  batch_sizes_mixed_.push(batch_size);
+  batch_sizes_gpu_.push(batch_size);
+}
+
+template <typename WorkspacePolicy, typename QueuePolicy>
+void Executor<WorkspacePolicy, QueuePolicy>::RunCPU() {
+  PreRun();
```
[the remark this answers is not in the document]
szalpal: We decided, that PreRun belongs here after all. It is weird, but we don't have better idea ATM

dali/pipeline/operator/batch_size_provider.h
```diff
@@ -0,0 +1,46 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef DALI_PIPELINE_OPERATOR_BATCH_SIZE_PROVIDER_H_
+#define DALI_PIPELINE_OPERATOR_BATCH_SIZE_PROVIDER_H_
+
+namespace dali {
+
+class BatchSizeProvider {
```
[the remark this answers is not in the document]
szalpal: Done

dali/pipeline/operator/builtin/external_source.h
szalpal: Removed

dali/pipeline/operator/builtin/external_source.h
```diff
@@ -16,16 +16,19 @@
 #define DALI_PIPELINE_OPERATOR_BUILTIN_EXTERNAL_SOURCE_H_
 
 #include <atomic>
-#include <string>
-#include <vector>
-#include <memory>
+#include <chrono>
```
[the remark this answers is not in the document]
szalpal: Removed

dali/pipeline/operator/operator.cc
```diff
@@ -18,9 +18,9 @@ namespace dali {
 
 template <typename Backend>
 void OperatorBase::EnforceConstantBatchSize(const workspace_t<Backend> &ws) const {
-  for (int i = 0; i < ws.NumInput(); i++) {
-    DALI_ENFORCE(ws.GetInputBatchSize(i) == max_batch_size_,
-                 "Batch size has to be constant throughout the processing");
+  for (int i = 1; i < ws.NumInput(); i++) {
```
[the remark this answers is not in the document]
szalpal: I believe it's not possible. In case of the output, there are operators, which arbitrarily choose what batch size to produce, e.g. `CoinFlip` and `NormalDist`. The effect is that the output batch size does not need to match input - user provides no input, hence we assume that the operator should request for a batch size, but the operator actually produces batch size from the info in the ArgumentInput. The problem is, that there are so many unknowns...

Do you happen to know, what kind of cases we should guard against? We could write tests for them to know, if this shall work as it is or not.

dali/test/python/test_dali_variable_batch_size.py
```diff
@@ -0,0 +1,620 @@
...
+
+def test_arithm_ops():
+    def pipe(max_batch_size, input_data, device):
+        pipe = Pipeline(batch_size=max_batch_size, num_threads=4, device_id=0)
+        data = fn.external_source(source=input_data, cycle=False, device=device)
+        data = data * 2
+        data = data + 3
+        data = data - 4
+        data = data / 5
+        data = data // 6
```
klecki, line 428: Can you check unary `-`, ternary `math.clamp`, and throw in a binary op with two DALI tensor arguments, so something like `data + data2`?

klecki (commented):
Finished the rest, there appear to be some issues in ImageDecoder and I would like to see the tests for sequences and with non-uniform input batches.

dali/pipeline/operator/builtin/external_source.h
szalpal: Done

dali/operators/decoder/nvjpeg/decoupled_api/nvjpeg_decoder_decoupled_api.h
```diff
@@ -909,6 +894,36 @@ class nvJPEGDecoder : public Operator<MixedBackend>, CachedDecoderImpl {
     RegisterDiagnostic("using_hw_decoder", &using_hw_decoder_);
   }
 
+  int CalcHwDecoderBatchSize(float hw_decoder_load, int curr_batch_size) {
+    if (hw_decoder_load == .0f) return 0;
```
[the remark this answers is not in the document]
szalpal: Done

dali/test/python/test_dali_variable_batch_size.py
```diff
@@ -0,0 +1,620 @@
...
+
+    :param max_batch_size: Actual sizes of every batch in the epoch will be less or equal to max_batch_size
+    :param n_iter: Number of iterations in the epoch
+    :param sample_shape: If sample_shape is callable, shape of every sample will be determined by
+                         calling sample_shape. In this case, every call to sample_shape has to
+                         return a tuple of integers. If sample_shape is a tuple, this will be a
+                         shape of every sample.
+    :param lo:
+    :param hi:
+    :param dtype: 'int' for uint8 or 'float' for float32
```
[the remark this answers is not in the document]
szalpal: Done

dali/test/python/test_dali_variable_batch_size.py
szalpal: Done

dali/test/python/test_dali_variable_batch_size.py
```diff
@@ -0,0 +1,620 @@
...
+    if isinstance(sample_shape, tuple):
+        size = sample_shape
+    elif inspect.isgeneratorfunction(sample_shape):
+        size = sample_shape()
+    else:
+        raise RuntimeError(
+            "`sample_shape` shall be either a tuple or a callable. Provide `(val,)` tuple for 1D shape")
+
+    if np.issubdtype(dtype, np.integer):
+        return [np.random.randint(lo, hi, size=(bs,) + size, dtype=dtype) for bs in
```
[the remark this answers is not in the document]
szalpal: We do not want non-uniform shapes at this point ;)

dali/test/python/test_dali_variable_batch_size.py
```diff
@@ -0,0 +1,620 @@
...
+    :param hi:
+    :param dtype: 'int' for uint8 or 'float' for float32
+    :return: An epoch of data
+    """
+    batch_sizes = np.array([max_batch_size // 2, max_batch_size // 4, max_batch_size])
+
+    if isinstance(sample_shape, tuple):
+        size = sample_shape
+    elif inspect.isgeneratorfunction(sample_shape):
+        size = sample_shape()
```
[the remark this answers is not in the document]
szalpal: Done

dali/test/python/test_dali_variable_batch_size.py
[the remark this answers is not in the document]
szalpal: Done, differently, but there's a non-uniformity

dali/test/python/test_dali_variable_batch_size.py
szalpal: Added sequence tests

dali/c_api/c_api_test.cc
```diff
@@ -230,6 +230,7 @@ TYPED_TEST(CApiTest, ExternalSourceSingleAllocPipe) {
     // Unnecessary copy in case of CPUBackend, makes the code generic across Backends
     input.Copy(input_cpu, cuda_stream);
     pipe_ptr->SetExternalInput(input_name, input);
+    daliSetExternalInputBatchSize(&handle, input_name.c_str(), input_shape.num_samples());
```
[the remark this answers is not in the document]
szalpal: Done

szalpal:
!build

dali/operators/decoder/nvjpeg/decoupled_api/nvjpeg_decoder_decoupled_api.h
```diff
@@ -73,23 +73,15 @@ class nvJPEGDecoder : public Operator<MixedBackend>, CachedDecoderImpl {
...
-    int tail = hw_decoder_bs_ % kNumHwDecoders;
-    if (tail > 0) {
-      hw_decoder_bs_ = hw_decoder_bs_ + kNumHwDecoders - tail;
-    }
-    if (hw_decoder_bs_ > max_batch_size_) {
-      hw_decoder_bs_ = max_batch_size_;
-    }
-
-    if (hw_decoder_bs_ > 0 &&
+    if (try_init_hw_decoder &&
```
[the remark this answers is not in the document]
szalpal: Done

dali/pipeline/operator/op_spec.h
szalpal: Fixed. Added removed lines back and added one more check to `operator.cc::EnforceUniformBatchSize`

dali/pipeline/operator/operator.h
```diff
@@ -364,14 +368,15 @@ class Operator<GPUBackend> : public OperatorBase {
...
-    EnforceConstantBatchSize<GPUBackend>(ws);
+    EnforceUniformInputBatchSize<GPUBackend>(ws);
     return SetupImpl(output_desc, ws);
   }
 
   void Run(DeviceWorkspace &ws) override {
     CheckInputLayouts(ws, spec_);
     SetupSharedSampleParams(ws);
     RunImpl(ws);
+//    EnforceUniformOutputBatchSize<GPUBackend>(ws);
```
klecki, line 379: Same here?

dali/pipeline/operator/operator.cc
```diff
@@ -17,16 +17,40 @@
...
+    DALI_ENFORCE(
+        curr_batch_size == static_cast<decltype(curr_batch_size)>(arg.second.tvec->ntensor()),
+        "ArgumentInput has to have the same batch size as an input.");
+  }
+}
+
+
+template <typename Backend>
+void OperatorBase::EnforceUniformOutputBatchSize(const workspace_t<Backend> &ws) const {
+  static_assert(!std::is_same<Backend, MixedBackend>::value,
```
klecki, line 37: It should be possible for Mixed AFAIK.

dali/pipeline/pipeline.h
```diff
@@ -428,7 +429,8 @@ class DLL_PUBLIC Pipeline {
   DLL_PUBLIC int num_outputs() const;
 
   /**
-   * @brief Returns a string describing the type device type backing the output specified by given id.
+   * @brief Returns a string describing the type device type backing the output specified by given
```
klecki, line 432: I still don't understand `the type device type`

klecki (approved):

szalpal:
!build

szalpal:
!build

PR #2501, opened 2020-11-27
Reenable nvJPEG2000

JanuszL:

#### Why we need this PR?
*Pick one, remove the rest*
- It reenables nvJPEG2000

 - What solution was applied:
     updates nvJPEG2000 version in the dependencies
     switch nvJPEG2000 support to be on by default
     build system
     deps docker image
     NA
 - Validation and testing:
     CI - current tests applies
     NA

**JIRA TASK**: *[NA]*

### DIFF
--- a/docker/Dockerfile.cuda111.x86_64.deps
+++ b/docker/Dockerfile.cuda111.x86_64.deps
@@ -9,7 +9,8 @@ RUN curl -LO https://developer.download.nvidia.com/compute/cuda/11.1.1/local_ins
     ./cuda_*.run --silent --no-opengl-libs --toolkit && \
     rm -f cuda_*.run;
 
-RUN apt-get update && \
+RUN NVJPEG2K_VERSION=0.1.0 && \
+    apt-get update && \
     apt-get install wget software-properties-common -y && \
     wget -qO - https://developer.download.nvidia.com/compute/cuda/repos/ubuntu1804/x86_64/7fa2af80.pub | apt-key add - && \
     add-apt-repository "deb https://developer.download.nvidia.com/compute/cuda/repos/ubuntu1804/x86_64/ /" && \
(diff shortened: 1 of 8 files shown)


JanuszL:
!build

docker/Dockerfile.cuda111.x86_64.deps
```diff
@@ -9,7 +9,8 @@ RUN curl -LO https://developer.download.nvidia.com/compute/cuda/11.1.1/local_ins
     ./cuda_*.run --silent --no-opengl-libs --toolkit && \
     rm -f cuda_*.run;
 
-RUN apt-get update && \
+RUN CUDA_VERSION=0.1.0 && \
```
klecki, line 12: Why do you need this? Looks weird.

JanuszL: Because docker caches its content based on the Dockerfile commands. So if the installation command is the same even if the underlying package is updated in the server docker build command will use the cached layer. Now, whenever we want to update nvJPEG2000 installation we will change this value and ask the docker to reinstall the necessary package.

klecki: Can we name it NVJPEG2K_VERSION? I presume, it's that.

klecki (approved):

docker/Dockerfile.cuda111.x86_64.deps
JanuszL: My bad. Fixed.

banasraf:
The preprocessor def `BUILD_NVJPEG2K_ENABLED`  doesn't work. If I remember correctly it needs to be changed to `NVJPEG2K_ENABLED`

JanuszL:
> The preprocessor def `BUILD_NVJPEG2K_ENABLED` doesn't work. If I remember correctly it needs to be changed to `NVJPEG2K_ENABLED`

Done

banasraf (approved):

PR #2504, opened 2020-11-27
Fix NVJPEG2K_ENABLED test macros

klecki:
It should use #if instead of #ifdef, as the
define uses a value not the fact of it being defined

#### Why we need this PR?
Fix NVJPEG2K_ENABLED test macros

 - What solution was applied:
     Search and swap
     NVJPEG, CUDA 10 build
     NA
 - Validation and testing:
     NA
     NA


klecki:
!build

klecki:
Build phase now works, merging.

PR #2516, opened 2020-12-01
Add GTest Event Listener with CUDA validation after TEST

klecki:

#### Why we need this PR?
- It fixes a bug/It adds new feature: Automatic error checking for CUDA after each Test Case.

 - What solution was applied:
	Used event listener to allow per test case error checking without the need to modify code.
     GTest
     N/A
 - Validation and testing:
     CI, adjustment in the testing framework
     N/A

**JIRA TASK**: *[Use DALI-1724 or NA]*

### DIFF
--- a/dali/test/dali_cuda_finalize_test.h
+++ b/dali/test/dali_cuda_finalize_test.h
@@ -17,6 +17,8 @@
 
 #include <cuda_runtime.h>
 #include <gtest/gtest.h>
+#include <cstring>
+
 #include "dali/core/cuda_utils.h"
 
 namespace dali {
@@ -33,12 +35,14 @@ namespace dali {
  */
 class CudaFinalizeEventListener : public ::testing::EmptyTestEventListener {
   void OnTestEnd(const ::testing::TestInfo& test_info) override {
-    auto sync_result = cudaDeviceSynchronize();
-    EXPECT_EQ(sync_result, cudaSuccess) << "CUDA error: \"" << cudaGetErrorName(sync_result)
-                                        << "\" - " << cudaGetErrorString(sync_result);
-    auto err = cudaGetLastError();
-    EXPECT_EQ(err, cudaSuccess) << "CUDA error: \"" << cudaGetErrorName(err) << "\" - "
-                                << cudaGetErrorString(err);
+    if (std::strstr(test_info.test_suite_name(), "CpuOnlyTest") == nullptr) {
+      auto sync_result = cudaDeviceSynchronize();
+      EXPECT_EQ(sync_result, cudaSuccess) << "CUDA error: \"" << cudaGetErrorName(sync_result)
+                                          << "\" - " << cudaGetErrorString(sync_result);
+      auto err = cudaGetLastError();
+      EXPECT_EQ(err, cudaSuccess) << "CUDA error: \"" << cudaGetErrorName(err) << "\" - "
+                                  << cudaGetErrorString(err);
+    }
   }
 };
 
--- a/dali/test/dali_cuda_finalize_test.h
+++ b/dali/test/dali_cuda_finalize_test.h
@@ -1,0 +1,47 @@
+// Copyright (c) 2020, NVIDIA CORPORATION. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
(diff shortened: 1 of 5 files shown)


klecki:
!build

dali/test/dali_cuda_finalize_test.h
```diff
@@ -0,0 +1,47 @@
...
+ *
+ * We don't want to use TearDown methods that would need to be either injected into every
+ * Fixture or defined globally and would run once. Events allow us to define additional
+ * checks with any granularity we want.
+ *
+ * We can use GTest assertion macros in the events with exception of OnTestPartResult.
+ */
+class CudaFinalizeEventListener : public ::testing::EmptyTestEventListener {
+  void OnTestEnd(const ::testing::TestInfo& test_info) override {
+    auto sync_result = cudaDeviceSynchronize();
```
szalpal, line 36: `sync_result` is always `cudaSuccess` IIRC

Also I'm wondering, could this actually catch the error from another thread or this synchronization secures this scenario

dali/test/dali_cuda_finalize_test.h
```diff
@@ -0,0 +1,47 @@
...
+    EXPECT_EQ(sync_result, cudaSuccess) << "CUDA error: \"" << cudaGetErrorName(sync_result)
```
szalpal, line 37: I'd add also `OnTestStart` and check `cudaGetLastError` there also, if by any chance it's not `cudaSuccess`

dali/test/dali_cuda_finalize_test.h
JanuszL: The docs is pretty clear:
```
Blocks until the device has completed all preceding requested tasks. cudaDeviceSynchronize() returns an error if one of the preceding tasks has failed.
```

klecki: It theoretically can catch error from other thread (apparently use of threading is on by default), but this sync should limit the leaking of error to something close. And I observed a scenario when cudaDeviceSynchronize returned success when there was launch error and a case where both calls failed.

dali/test/dali_cuda_finalize_test.h
klecki: That would yield maybe an early exit in some cases, but still, the error will be propagated to the point where it is checked - so either the test will do it explicitly (some does) or it will be checked after.
I will leave the error checking in one point only.

dali/test/dali_cuda_finalize_test.h
JanuszL: I think the error is reported only once. If the launch error appeared then it won't be reported again. @mzient

klecki: I forced a kernel launch failure and cudaDeviceSynchronize didn't report it, but the cudaGetLastError did (and this clears it for further calls).

dali/test/dali_cuda_finalize_test.h
szalpal: What I had in mind is that with this some race conditions or something like that, that aren't easy to extract a single unit test that break, may be easier to debug. I'm fine either way

szalpal (approved):

klecki:
!build

dali/test/dali_cuda_finalize_test.h
```diff
@@ -0,0 +1,51 @@
...
+ *
+ * We don't want to use TearDown methods that would need to be either injected into every
+ * Fixture or defined globally and would run once. Events allow us to define additional
+ * checks with any granularity we want.
+ *
+ * We can use GTest assertion macros in the events with exception of OnTestPartResult.
+ */
+class CudaFinalizeEventListener : public ::testing::EmptyTestEventListener {
+  void OnTestEnd(const ::testing::TestInfo& test_info) override {
+    if (std::strstr(test_info.test_suite_name(), "CpuOnlyTest") == nullptr) {
```
JanuszL, line 38: Maybe we can do something like:
```
(cuDeviceGet(&dummy, 0) != CUDA_ERROR_SHARED_OBJECT_SYMBOL_NOT_FOUND)
```
to make sure that driver is accessible. I'm just afraid that `CpuOnlyTest` asking every test without GPU to hold this name patter maybe too much.

klecki: I followed the pattern from our tests that assumes that substring.

klecki:
!build

dali/test/dali_cuda_finalize_test.h
```diff
@@ -0,0 +1,50 @@
...
+ *
+ * We don't want to use TearDown methods that would need to be either injected into every
+ * Fixture or defined globally and would run once. Events allow us to define additional
+ * checks with any granularity we want.
+ *
+ * We can use GTest assertion macros in the events with exception of OnTestPartResult.
+ */
+class CudaFinalizeEventListener : public ::testing::EmptyTestEventListener {
+  void OnTestEnd(const ::testing::TestInfo& test_info) override {
+    if (cuInitChecked()) {
```
JanuszL, line 37: You can add a comment why this is tested here

JanuszL (approved):

dali/test/dali_cuda_finalize_test.h
klecki: done

klecki:
!build

PR #2540, opened 2020-12-10
Vectorize CPU resampling

mzient:

#### Why we need this PR?
*Pick one, remove the rest*
- It fixes performance issues when using CPU resize after introducing 3D resampling

 - What solution was applied:
     * Add SIMD utility header to kernels/common
     * Use multi-vector approach to utilize full width of input and output vectors when resampling vertically
     * Use multi-vector approach to utilize full with of the output when resampling horizontally
     * Tune tile size for vertical resampling
     * CPU resampling backend
     * N/A
 - Validation and testing:
     * Existing tests siffuce
     * N/A

**JIRA TASK**: DALI-1776

### DIFF
--- a/dali/kernels/common/simd.h
+++ b/dali/kernels/common/simd.h
@@ -1,0 +1,209 @@
+// Copyright (c) 2020, NVIDIA CORPORATION. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef DALI_KERNELS_COMMON_SIMD_H_
+#define DALI_KERNELS_COMMON_SIMD_H_
+
+#ifdef __SSE2__
+#include <emmintrin.h>
+#endif
+
+#include <cstddef>
+#include <type_traits>
+#include "dali/core/force_inline.h"
+
+namespace dali {
+namespace kernels {
+namespace simd {
+
+#ifdef __SSE2__
+
+template <int n>
+struct float4x {
+  __m128 v[n];  // NOLINT
+};
+
+template <int n>
+struct i128x {
+  __m128i v[n];  // NOLINT
+};
+
+using float4x1 = float4x<1>;
+using float4x2 = float4x<2>;
+using float4x4 = float4x<4>;
+using i128x1 = i128x<1>;
+using i128x2 = i128x<2>;
+using i128x4 = i128x<4>;
+
+inline __m128i clamp_round(__m128 f, float lo, float hi) {
+  f = _mm_min_ps(_mm_max_ps(f, _mm_set1_ps(lo)), _mm_set1_ps(hi));
+  return _mm_cvtps_epi32(f);  // round
+}
+
+inline void store(int8_t *i8, i128x4 iv) {
+  __m128i sv0 = _mm_packs_epi32(iv.v[0], iv.v[1]);
+  __m128i sv1 = _mm_packs_epi32(iv.v[2], iv.v[3]);
+  _mm_storeu_si128(reinterpret_cast<__m128i*>(i8), _mm_packs_epi16(sv0, sv1));
+}
+
+inline void store(uint8_t *u8, i128x4 iv) {
+  __m128i sv0 = _mm_packs_epi32(iv.v[0], iv.v[1]);
+  __m128i sv1 = _mm_packs_epi32(iv.v[2], iv.v[3]);
+  _mm_storeu_si128(reinterpret_cast<__m128i*>(u8), _mm_packus_epi16(sv0, sv1));
+}
+
+inline void store(int16_t *i16, i128x2 iv) {
+  _mm_storeu_si128(reinterpret_cast<__m128i*>(i16), _mm_packs_epi32(iv.v[0], iv.v[1]));
+}
+
+inline void store(uint16_t *u16, i128x2 iv) {
+  __m128i sh = _mm_set_epi32(0, 0, 0, 1);
+  __m128i one = _mm_set1_epi32(1);
+  // avoid signed saturation by shifting right and saving LSB, then reconsitituting it
+  __m128i hi0 = _mm_srl_epi32(iv.v[0], sh);
+  __m128i hi1 = _mm_srl_epi32(iv.v[1], sh);
+  __m128i lo0 = _mm_and_si128(iv.v[0], one);
+  __m128i lo1 = _mm_and_si128(iv.v[0], one);
+  __m128i lo = _mm_packs_epi32(lo0, lo1);
+  __m128i hi = _mm_packs_epi32(hi0, hi1);
+  __m128i out = _mm_or_si128(_mm_sll_epi16(hi, sh), lo);
+  _mm_storeu_si128(reinterpret_cast<__m128i*>(u16), out);
+}
+
+inline float4x4 load_f(const uint8_t *u8) {
+  __m128i in = _mm_loadu_si128(reinterpret_cast<const __m128i *>(u8));
+  __m128i zero = _mm_setzero_si128();
+  __m128i lo16 = _mm_unpacklo_epi8(in, zero);
+  __m128i hi16 = _mm_unpackhi_epi8(in, zero);
+  __m128i i32_0 = _mm_unpacklo_epi8(lo16, zero);
+  __m128i i32_1 = _mm_unpackhi_epi8(lo16, zero);
+  __m128i i32_2 = _mm_unpacklo_epi8(hi16, zero);
+  __m128i i32_3 = _mm_unpackhi_epi8(hi16, zero);
+  return {{ _mm_cvtepi32_ps(i32_0),
+            _mm_cvtepi32_ps(i32_1),
+            _mm_cvtepi32_ps(i32_2),
+            _mm_cvtepi32_ps(i32_3) }};
+}
+
+inline __m128i sext8_32(__m128i i) {
+  __m128i sh = _mm_set_epi32(0, 0, 0, 24);
+  return _mm_sra_epi32(_mm_sll_epi32(i, sh), sh);
+}
+
+inline float4x4 load_f(const int8_t *i8) {
+  __m128i in = _mm_loadu_si128(reinterpret_cast<const __m128i *>(i8));
+  __m128i zero = _mm_setzero_si128();
+  __m128i lo16 = _mm_unpacklo_epi8(in, zero);
+  __m128i hi16 = _mm_unpackhi_epi8(in, zero);
+  __m128i i32_0 = _mm_unpacklo_epi8(lo16, zero);
+  __m128i i32_1 = _mm_unpackhi_epi8(lo16, zero);
+  __m128i i32_2 = _mm_unpacklo_epi8(hi16, zero);
+  __m128i i32_3 = _mm_unpackhi_epi8(hi16, zero);
+  return {{ _mm_cvtepi32_ps(sext8_32(i32_0)),
+            _mm_cvtepi32_ps(sext8_32(i32_1)),
+            _mm_cvtepi32_ps(sext8_32(i32_2)),
+            _mm_cvtepi32_ps(sext8_32(i32_3)) }};
+}
+
+inline __m128i sext16_32(__m128i i) {
+  __m128i sh = _mm_set_epi32(0, 0, 0, 16);
+  return _mm_sra_epi32(_mm_sll_epi32(i, sh), sh);
+}
+
+inline float4x2 load_f(const uint16_t *u16) {
+  __m128i in = _mm_loadu_si128(reinterpret_cast<const __m128i *>(u16));
+  __m128i zero = _mm_setzero_si128();
+  __m128i i32_0 = _mm_unpacklo_epi16(in, zero);
+  __m128i i32_1 = _mm_unpackhi_epi16(in, zero);
+  return {{ _mm_cvtepi32_ps(i32_0), _mm_cvtepi32_ps(i32_1) }};
+}
+
+inline float4x2 load_f(const int16_t *i16) {
+  __m128i in = _mm_loadu_si128(reinterpret_cast<const __m128i *>(i16));
+  __m128i zero = _mm_setzero_si128();
+  __m128i i32_0 = _mm_unpacklo_epi16(in, zero);
+  __m128i i32_1 = _mm_unpackhi_epi16(in, zero);
+  return {{ _mm_cvtepi32_ps(sext16_32(i32_0)), _mm_cvtepi32_ps(sext16_32(i32_1)) }};
+}
+
+inline float4x1 load_f(const float *f) {
+  return {{ _mm_loadu_ps(f) }};
+}
+
+template <typename Out>
+inline std::enable_if_t<std::is_integral<Out>::value>
+store(Out *out, float4x<sizeof(float)/sizeof(Out)> f) {
+  constexpr int nvec = sizeof(float)/sizeof(Out);
+  constexpr float lo = min_value<Out>();
+  constexpr float hi = max_value<Out>();
+  i128x<nvec> iv;
+  for (int i = 0; i < nvec; i++)
+    iv.v[i] = clamp_round(f.v[i], lo, hi);
+  store(out, iv);
+}
+
+inline void store(float *out, float4x1 f) {
+  _mm_storeu_ps(out, f.v[0]);
+}
+
+template <int num_vecs>
+struct multivec : float4x<num_vecs> {
+  DALI_FORCEINLINE static multivec zero() noexcept  {
+    multivec m;
+    for (int i = 0; i < num_vecs; i++)
+      m.v[i] = _mm_setzero_ps();
+    return m;
+  }
+
+  DALI_FORCEINLINE static multivec load(const float *in) noexcept  {
+    multivec m;
+    for (int i = 0; i < num_vecs; i++)
+      m.v[i] = _mm_loadu_ps(in + 4*i);
+    return m;
+  }
+
+  template <typename In>
+  DALI_FORCEINLINE static multivec load(const In *in) noexcept  {
+    constexpr int load_lanes = 16 / sizeof(In);
+    constexpr int load_vecs = load_lanes / 4;
+    multivec m;
+    for (int i = 0; i < num_vecs; i += load_vecs) {
+      auto tmp = simd::load_f(in + i * load_lanes);
+      for (int j = 0; j < load_vecs; j++)
+        m.v[i + j] = tmp.v[j];
+    }
+    return m;
+  }
+};
+
+
+template <int num_vecs, typename Out>
+DALI_FORCEINLINE static void store(Out *out, multivec<num_vecs> m) noexcept {
+  constexpr int store_lanes = 16 / sizeof(Out);
+  constexpr int store_vecs = store_lanes / 4;
+  for (int i = 0; i < num_vecs; i += store_vecs) {
+    float4x<store_vecs> slice;
+    for (int j = 0; j < store_vecs; j++)
+      slice.v[j] = m.v[i + j];
+    simd::store(out + i * store_lanes, slice);
+  }
+}
+
+#endif  // __SSE2__
+
+}  // namespace simd
+}  // namespace kernels
+}  // namespace dali
+
+#endif  // DALI_KERNELS_COMMON_SIMD_H_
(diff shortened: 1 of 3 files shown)


dali/kernels/common/simd.h
```diff
@@ -0,0 +1,209 @@
...
+            _mm_cvtepi32_ps(i32_2),
+            _mm_cvtepi32_ps(i32_3) }};
+}
+
+inline __m128i sext8_32(__m128i i) {
+  __m128i sh = _mm_set_epi32(0, 0, 0, 24);
+  return _mm_sra_epi32(_mm_sll_epi32(i, sh), sh);
+}
+
+inline float4x4 load_f(const int8_t *i8) {
```
mzient, line 104: Loads a vector of int8 and expands it to vectors of float.

dali/kernels/common/simd.h
```diff
@@ -0,0 +1,209 @@
...
+  __m128i i32_1 = _mm_unpackhi_epi8(lo16, zero);
+  __m128i i32_2 = _mm_unpacklo_epi8(hi16, zero);
+  __m128i i32_3 = _mm_unpackhi_epi8(hi16, zero);
+  return {{ _mm_cvtepi32_ps(sext8_32(i32_0)),
+            _mm_cvtepi32_ps(sext8_32(i32_1)),
+            _mm_cvtepi32_ps(sext8_32(i32_2)),
+            _mm_cvtepi32_ps(sext8_32(i32_3)) }};
+}
+
+inline __m128i sext16_32(__m128i i) {
```
mzient, line 119: Sign extend.

dali/kernels/common/simd.h
```diff
@@ -0,0 +1,209 @@
...
+  __m128i sh = _mm_set_epi32(0, 0, 0, 16);
+  return _mm_sra_epi32(_mm_sll_epi32(i, sh), sh);
+}
+
+inline float4x2 load_f(const uint16_t *u16) {
```
mzient, line 124: Loads a vector of uint16 and expands it to vectors of float.

dali/kernels/common/simd.h
```diff
@@ -0,0 +1,209 @@
...
+  __m128i in = _mm_loadu_si128(reinterpret_cast<const __m128i *>(u16));
+  __m128i zero = _mm_setzero_si128();
+  __m128i i32_0 = _mm_unpacklo_epi16(in, zero);
+  __m128i i32_1 = _mm_unpackhi_epi16(in, zero);
+  return {{ _mm_cvtepi32_ps(i32_0), _mm_cvtepi32_ps(i32_1) }};
+}
+
+inline float4x2 load_f(const int16_t *i16) {
```
mzient, line 132: Loads a vector of int16 and expands it to vectors of float.

dali/kernels/common/simd.h
```diff
@@ -0,0 +1,209 @@
...
+  return {{ _mm_cvtepi32_ps(sext16_32(i32_0)), _mm_cvtepi32_ps(sext16_32(i32_1)) }};
+}
+
+inline float4x1 load_f(const float *f) {
+  return {{ _mm_loadu_ps(f) }};
+}
+
+template <typename Out>
+inline std::enable_if_t<std::is_integral<Out>::value>
+store(Out *out, float4x<sizeof(float)/sizeof(Out)> f) {
```
mzient, line 146: Convert vectors of float to one vector of Out and store.

dali/kernels/common/simd.h
```diff
@@ -0,0 +1,209 @@
...
+
+inline float4x4 load_f(const int8_t *i8) {
+  __m128i in = _mm_loadu_si128(reinterpret_cast<const __m128i *>(i8));
+  __m128i zero = _mm_setzero_si128();
+  __m128i lo16 = _mm_unpacklo_epi8(in, zero);
+  __m128i hi16 = _mm_unpackhi_epi8(in, zero);
+  __m128i i32_0 = _mm_unpacklo_epi8(lo16, zero);
+  __m128i i32_1 = _mm_unpackhi_epi8(lo16, zero);
+  __m128i i32_2 = _mm_unpacklo_epi8(hi16, zero);
+  __m128i i32_3 = _mm_unpackhi_epi8(hi16, zero);
```
mzient, line 112: interleave with zeros until 32-bit wide

dali/kernels/common/simd.h
```diff
@@ -0,0 +1,209 @@
...
+  return {{ _mm_cvtepi32_ps(sext8_32(i32_0)),
+            _mm_cvtepi32_ps(sext8_32(i32_1)),
+            _mm_cvtepi32_ps(sext8_32(i32_2)),
+            _mm_cvtepi32_ps(sext8_32(i32_3)) }};
```
mzient, line 116: sign-extend and convert to float.

dali/kernels/common/simd.h
```diff
@@ -0,0 +1,209 @@
...
+  __m128i sv1 = _mm_packs_epi32(iv.v[2], iv.v[3]);
+  _mm_storeu_si128(reinterpret_cast<__m128i*>(u8), _mm_packus_epi16(sv0, sv1));
+}
+
+inline void store(int16_t *i16, i128x2 iv) {
+  _mm_storeu_si128(reinterpret_cast<__m128i*>(i16), _mm_packs_epi32(iv.v[0], iv.v[1]));
+}
+
+inline void store(uint16_t *u16, i128x2 iv) {
+  __m128i sh = _mm_set_epi32(0, 0, 0, 1);
```
mzient, line 71: There's no instruction that can pack 32-bit values to 16-bit without _signed_ saturation - neither is there a byte shuffle.

dali/kernels/common/simd.h
```diff
@@ -0,0 +1,209 @@
...
+using i128x1 = i128x<1>;
+using i128x2 = i128x<2>;
+using i128x4 = i128x<4>;
+
+inline __m128i clamp_round(__m128 f, float lo, float hi) {
+  f = _mm_min_ps(_mm_max_ps(f, _mm_set1_ps(lo)), _mm_set1_ps(hi));
+  return _mm_cvtps_epi32(f);  // round
+}
+
+inline void store(int8_t *i8, i128x4 iv) {
```
JanuszL, line 54:
```suggestion
inline void store_converted(int8_t *i8, i128x4 iv) {
```

dali/kernels/common/simd.h
```diff
@@ -0,0 +1,209 @@
...
+  __m128i sv0 = _mm_packs_epi32(iv.v[0], iv.v[1]);
+  __m128i sv1 = _mm_packs_epi32(iv.v[2], iv.v[3]);
+  _mm_storeu_si128(reinterpret_cast<__m128i*>(i8), _mm_packs_epi16(sv0, sv1));
+}
+
+inline void store(uint8_t *u8, i128x4 iv) {
```
JanuszL, line 60: Can't you have one store. As I see the destination type doesn't change the function behaviour.

dali/kernels/common/simd.h
mzient: True that. It should actually be store_i32.

dali/kernels/common/simd.h
mzient: There's a difference:
`_mm_packs_epi16` - signed
`_mm_packus_epi16` - unsigned

JanuszL: Oh. Ok. Missed that.

dali/kernels/common/simd.h
JanuszL: Can you add this as a comment to this function?

JanuszL: So this function just cuts the MSB?

dali/kernels/common/simd.h
```diff
@@ -0,0 +1,209 @@
...
+
+  DALI_FORCEINLINE static multivec load(const float *in) noexcept  {
+    multivec m;
+    for (int i = 0; i < num_vecs; i++)
+      m.v[i] = _mm_loadu_ps(in + 4*i);
+    return m;
+  }
+
+  template <typename In>
+  DALI_FORCEINLINE static multivec load(const In *in) noexcept  {
```
klecki, line 177: Don't you have to specify the correct `num_vecs` based on the input type?
Can't you compute it based on the input type instead? It seems error prone to me.

klecki (commented):
Comments to the first file.
It would be nice to get those comments you put in this PR as docstrings.

dali/kernels/common/simd.h
```diff
@@ -0,0 +1,209 @@
...
+  __m128i i32_1 = _mm_unpackhi_epi8(lo16, zero);
+  __m128i i32_2 = _mm_unpacklo_epi8(hi16, zero);
+  __m128i i32_3 = _mm_unpackhi_epi8(hi16, zero);
+  return {{ _mm_cvtepi32_ps(i32_0),
+            _mm_cvtepi32_ps(i32_1),
+            _mm_cvtepi32_ps(i32_2),
+            _mm_cvtepi32_ps(i32_3) }};
+}
+
+inline __m128i sext8_32(__m128i i) {
```
JanuszL, line 99: It extracts LS byte So maybe extract_lsbyte?

dali/kernels/common/simd.h
JanuszL: You can add a comment here about it.

dali/kernels/common/simd.h
JanuszL: I would write this down as well.

dali/kernels/common/simd.h
JanuszL: Oh, so sext is signed extend, I would use that name instead of this abbreviation.

dali/kernels/common/simd.h
```diff
@@ -0,0 +1,209 @@
...
+
+inline __m128i clamp_round(__m128 f, float lo, float hi) {
+  f = _mm_min_ps(_mm_max_ps(f, _mm_set1_ps(lo)), _mm_set1_ps(hi));
+  return _mm_cvtps_epi32(f);  // round
+}
+
+inline void store(int8_t *i8, i128x4 iv) {
+  __m128i sv0 = _mm_packs_epi32(iv.v[0], iv.v[1]);
+  __m128i sv1 = _mm_packs_epi32(iv.v[2], iv.v[3]);
+  _mm_storeu_si128(reinterpret_cast<__m128i*>(i8), _mm_packs_epi16(sv0, sv1));
```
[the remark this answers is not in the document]
mzient: We probably could shave a few cycles from aligning the temporary buffers and having an aligned store here - but that would require doubling this function as aligned/unaligned.
This function explicitly uses unaligned store*u*.

dali/kernels/common/simd.h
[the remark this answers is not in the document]
mzient: _mm_set_epi32 is not an instruction - compiler is at liberty to implement it as setzero/insert or however else it sees fit.

dali/kernels/common/simd.h
```diff
@@ -0,0 +1,209 @@
...
+}
+
+inline void store(uint16_t *u16, i128x2 iv) {
+  __m128i sh = _mm_set_epi32(0, 0, 0, 1);
+  __m128i one = _mm_set1_epi32(1);
+  // avoid signed saturation by shifting right and saving LSB, then reconsitituting it
+  __m128i hi0 = _mm_srl_epi32(iv.v[0], sh);
+  __m128i hi1 = _mm_srl_epi32(iv.v[1], sh);
+  __m128i lo0 = _mm_and_si128(iv.v[0], one);
+  __m128i lo1 = _mm_and_si128(iv.v[0], one);
```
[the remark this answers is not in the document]
mzient: Hmm.. apparently we don't test this code path.

dali/kernels/common/simd.h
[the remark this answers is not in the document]
klecki: Isn't the requirement to be aligned with sizeof of the type, not the whole vector? The docs for `_mm_store_si128` says:
```
must be aligned on a 16-byte boundary or a general-protection exception may be generated.
```

dali/kernels/common/simd.h
```diff
@@ -0,0 +1,209 @@
...
+  __m128i lo = _mm_packs_epi32(lo0, lo1);
+  __m128i hi = _mm_packs_epi32(hi0, hi1);
+  __m128i out = _mm_or_si128(_mm_sll_epi16(hi, sh), lo);
```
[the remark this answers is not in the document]
mzient: This function doesn't advertise that it can handle out-of-range arguments. The problem is that there's no easy way (in SSE2 at least) to just narrow 32 bits to 16 bits without _signed_ saturation. This will work for all 16-bit values, but no, it won't work for values outside 0..0xffff range.

dali/kernels/common/simd.h
```diff
@@ -0,0 +1,209 @@
...
+
+inline float4x4 load_f(const uint8_t *u8) {
+  __m128i in = _mm_loadu_si128(reinterpret_cast<const __m128i *>(u8));
+  __m128i zero = _mm_setzero_si128();
+  __m128i lo16 = _mm_unpacklo_epi8(in, zero);
+  __m128i hi16 = _mm_unpackhi_epi8(in, zero);
+  __m128i i32_0 = _mm_unpacklo_epi8(lo16, zero);
+  __m128i i32_1 = _mm_unpackhi_epi8(lo16, zero);
+  __m128i i32_2 = _mm_unpacklo_epi8(hi16, zero);
+  __m128i i32_3 = _mm_unpackhi_epi8(hi16, zero);
```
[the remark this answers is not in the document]
mzient: In general it should be, but it doesn't matter when we're inserting zeros.

dali/kernels/common/simd.h
mzient: That's quite common abbreviation (zext/sext) - I'll add a comment, though.

dali/kernels/common/simd.h
mzient: No. It will load as many values from input as is necessary to fill the multivec.
I can add
```C++
static_assert(load_vecs > 0, "Too few lanes to use vector load for this input type");
```

dali/kernels/common/simd.h
JanuszL: In the intrinsic enthusiast word I guess. And not necessarily everyone who may need to read that code belongs there.

dali/kernels/imgproc/resample/resampling_impl_cpu.h
```diff
@@ -83,6 +85,148 @@ void ResampleCol(Out *out, const In *in, int x, int w, const int32_t *in_columns
...
+            float tmp_coeffs[kNumLanes];
+            for (int l = 0; l < kNumLanes; l++)
+              tmp_coeffs[l] = coeffs[(x + l) * support + k];  // interleave per-column coefficients
+            vec_pack vcoeffs = vec_pack::load(tmp_coeffs);
+
+            for (int l = 0; l < kNumLanes; l++) {
+              int srcx = in_columns[x + l] + k;
+              if (clamp_left) if (srcx < 0) srcx = 0;
+              if (clamp_right) if (srcx > w-1) srcx = w-1;
+              tmpin[l] = in[srcx * channels + c];
```
JanuszL, line 165: Does it make sense to make it a funciton?

dali/kernels/imgproc/resample/resampling_impl_cpu.h
```diff
@@ -83,6 +85,148 @@ void ResampleCol(Out *out, const In *in, int x, int w, const int32_t *in_columns
...
+
+    int x = ox0;
+#ifdef __SSE2__
+    float tmpin[kNumLanes];
+    for (; x + kNumLanes <= ox1; x += kNumLanes) {
+      Out tmp_out[kNumLanes];
+
+      if (static_channels < 0) {
+        // we don't know how many channels we have at compile time - inner loop over filter
+        for (int c = 0; c < channels; c++) {
```
JanuszL, line 151: What difference it would make if we have a variants for predefined number of channels - 1 and 3 for example?

dali/kernels/imgproc/resample/resampling_impl_cpu.h
```diff
@@ -144,37 +288,30 @@ void ResamplHorzRow(Out *out_row, int out_width, const In *in_row, int in_width,
...
                     int first_regular_col, int last_regular_col, bool flipped) {
   int x = 0;
+  // if last_regular_col < first_regular_col, then we can only use one-sided clamp
+  // up to last_regular_col-1
+  int max_one_sided_clamp = std::min(first_regular_col, last_regular_col+1);
+
+  SIMD_horz_resample_impl<Out, In> impl;
   if (flipped) {
-    for (; x < first_regular_col && x <= last_regular_col; x++) {
-      ResampleCol<static_channels, false, true>(
```
JanuszL, line 149: Do we still need the regular `ResampleCol`?

mzient: Yes, it's used for processing SIMD tails.

dali/kernels/imgproc/resample/resampling_impl_cpu.h
mzient: That's... risky. I don't want to go looking for the lost performance for having done so.

JanuszL: Sure

dali/kernels/imgproc/resample/resampling_impl_cpu.h
JanuszL: Missed that

dali/kernels/imgproc/resample/resampling_impl_cpu.h
mzient: Then it goes for the branch below (static_channels) and then `channels` is effectively a constant expression. It's quite optimized. We _could_ use a specialized variant for 1 channel, 2, and 4 channels - 3 not so much, interleaving and deinterleaving triples on X86 is a pain.
We could use it on ARM NEON, though, with its beautiful interleaved load and store.

dali/kernels/common/simd.h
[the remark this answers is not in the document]
mzient: store is aligned
storeu is unaligned

dali/kernels/common/simd.h
mzient: Per-element, it does this:
a = (unsigned)x >> 1
b = x & 1
c = saturate<int16>(a)
out = c << 1 | b

dali/kernels/imgproc/resample/resampling_impl_cpu.h
JanuszL: So when the number of channels is not static?

dali/kernels/common/simd.h
klecki: I mean, if you try to load uint16 into the float4x1, it will just probably segfault. There is no check for that, IMO it should be
```
static_assert(num_vecs % load_vecs == 0);
```
or something similar.

klecki: Not the implementation of multivec is tied to it's usage that assumes the correct storage size used for loading.

dali/kernels/common/simd.h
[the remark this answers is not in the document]
klecki: I mean, this function also doesn't advertise that it doesn't work for those ranges. I assumed that if you used instructions with saturation (I know there is no much choice there), you meant saturation.

dali/kernels/common/simd.h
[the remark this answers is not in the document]
klecki: Yes, I was asking to use it here, now I see I misread bytes and bits as I was reading about the int16 types.

klecki (commented):
The resampling looks fine, I would gladly see docs in the simd.h if you want to treat it as generic utility.

dali/kernels/imgproc/resample/resampling_impl_cpu.h
klecki: I guess it's for some weird cases not covered by static switch for the number of channels, like 7 or 16.

dali/kernels/common/simd.h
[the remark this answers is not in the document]
mzient: I've rewritten it so it properly clamps.

dali/kernels/common/simd.h
mzient: Done in description.

dali/kernels/imgproc/resample/resampling_impl_cpu.h
```diff
@@ -83,6 +85,148 @@ void ResampleCol(Out *out, const In *in, int x, int w, const int32_t *in_columns
...
+        Out tmp_out[kNCh][kNumLanes];
+        float tmp_in[kNCh][kNumLanes];
+
+        for (int c = 0; c < channels; c++)
+          vout[c] = vec_pack::zero();
+
+        for (int k = 0; k < support; k++) {
+          float tmp_coeffs[kNumLanes];
+          for (int l = 0; l < kNumLanes; l++)
+            tmp_coeffs[l] = coeffs[(x + l) * support + k];  // interleave per-column coefficients
```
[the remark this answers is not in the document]
mzient: Let's wait for another perf. complaint ;)

dali/kernels/common/simd.h
mzient: Fixed?

mzient:
!build

dali/kernels/common/simd.h
JanuszL: Doxygen should do

JanuszL (approved):

klecki (approved):

PR #2549, opened 2020-12-14
Remove constraint on input type for RandomResizedCrop

mzient:

#### Why we need this PR?
*Pick one, remove the rest*
- It fixes a bug #2545 - error raised for non-uint8 input in GPU RandomResizedCrop

 - What solution was applied:
     * Assert removed
     * Extended tests
     * RandomResizedCrop
     * N/A
 - Validation and testing:
     * Python tests
     * N/A

**JIRA TASK**: N/A

### DIFF
--- a/dali/test/python/test_operator_random_resized_crop.py
+++ b/dali/test/python/test_operator_random_resized_crop.py
@@ -1,6 +1,10 @@
 import nvidia.dali as dali
 import nvidia.dali.fn as fn
 import numpy as np
+import test_utils
+
+def close(a, b):
+    return abs(a - b) < 1e-5 or abs(a - b) < abs(a) + abs(b) * 1e-6
 
 def analyze_frame(image, channel_dim):
     def pixel(x, y):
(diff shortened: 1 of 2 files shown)


dali/test/python/test_operator_random_resized_crop.py
```diff
@@ -117,4 +137,6 @@ def test_random_resized_crop():
                     ((1, 2), (0.4, 1.0)),
                     ((0.5, 1), (0.1, 0.5))]:
                     for size in [(100,100), (640,480)]:
-                        yield _test_rrc, device, max_frames, layout, aspect, area, size
+                        input_type = types[np.random.randint(0, len(types))]
```
JanuszL, line 140: Can we just create a separate, deterministic test that would just go though all data types, and set other test parameters to one, well defined value?
On the other hand the set of input types is rather very immitted, so I think we will always test it anyway.

JanuszL (approved):

dali/test/python/test_operator_random_resized_crop.py
mzient: I don't wont to make this test run any longer than it absolutely needs to. There are just 6 options (3 input types + output original vs float). The test is deterministic (fixed seed) and all combinations are tested.

JanuszL: OK

jantonguirao (approved):

PR #2555, opened 2020-12-16
Add exp and log math functions

azrael417:
#### Why we need this PR?
- It adds log and exp preprocessing ops, the former is needed for one mlperf-hpc benchmark

- added dali.math.log and dali.math.exp functions to arithmetic expressions for CPU and GPU
- if arguments are of non-FP type, arguments will be casted to FP32
- added python test. The log test passes with precision requirement of 1e-6, the exp tests overflows, needs some debugging. 
- maybe we should specialize the exp and log templates to use the matching std types as well, at the cost of speed. That would not fix the GPU issues though
Main files to look at (I am not listing files which just include some boilerplate handling of the new expressions):
- dali/test/python/test_operator_arithmetic_ops.py 
- dali/operators/math/expressions/arithmetic_meta.h

**JIRA TASK**: NA


JanuszL (approved):

klecki:
!build

klecki (approved):

JanuszL (approved):

PR #2609, opened 2021-01-12
Integrate RMM

mzient:
#### Why we need this PR?
*Pick one, remove the rest*
- Refactoring to use RAPIDS memory manager

 - What solution was applied:
     * Use RMM fork with interfaces we need
     * Replace memory_resource and stream_memory_resource with interfaces from RMM
     * Adjust code to acommodate the interfaces
     * Add RMM dependencies to build system
     * Add RMM and ints dependencies to include paths
     * build
     * core/memory manager
     * N/A
 - Validation and testing:
     * N/A
     * N/A

**JIRA TASK**: [DALI-1782] [DALI-1783]

### DIFF
--- a/cmake/Dependencies.common.cmake
+++ b/cmake/Dependencies.common.cmake
@@ -199,3 +199,19 @@ add_library(cocoapi STATIC ${SOURCE_FILES})
 set_target_properties(cocoapi PROPERTIES POSITION_INDEPENDENT_CODE ON)
 list(APPEND DALI_LIBS cocoapi)
 list(APPEND DALI_EXCLUDES libcocoapi.a)
+
+##################################################################
+# RAPIDS Memory Manager
+##################################################################
+include_directories(SYSTEM ${PROJECT_SOURCE_DIR}/third_party/rmm/include)
+
+##################################################################
+# spdlog and fmt
+##################################################################
+include_directories(SYSTEM ${PROJECT_SOURCE_DIR}/third_party/spdlog/include)
+include_directories(SYSTEM ${PROJECT_SOURCE_DIR}/third_party/fmt/include)
+
+##################################################################
+# thrust
+##################################################################
+include_directories(SYSTEM ${PROJECT_SOURCE_DIR}/third_party/thrust/)
--- a/dali/core/mm/mm_test_utils.h
+++ b/dali/core/mm/mm_test_utils.h
@@ -21,6 +21,8 @@
 #include <unordered_map>
 #include <unordered_set>
 #include <utility>
+#include <rmm/mr/device/pool_memory_resource.hpp>
+#include <rmm/mr/device/cuda_memory_resource.hpp>
 #include "dali/core/mm/memory_resource.h"
 #include "dali/core/mm/malloc_resource.h"
 #include "dali/core/mm/detail/util.h"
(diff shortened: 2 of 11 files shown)


mzient:
!build

cmake/Dependencies.common.cmake
```diff
@@ -199,3 +199,19 @@ add_library(cocoapi STATIC ${SOURCE_FILES})
 set_target_properties(cocoapi PROPERTIES POSITION_INDEPENDENT_CODE ON)
 list(APPEND DALI_LIBS cocoapi)
 list(APPEND DALI_EXCLUDES libcocoapi.a)
+
+##################################################################
+# RAPIDS Memory Manager
+##################################################################
+include_directories(SYSTEM ${PROJECT_SOURCE_DIR}/third_party/rmm/include)
```
JanuszL, line 206: Can you check if that compiler for QNX, aarch64 and SBSA?

mzient: QNX passed.

dali/core/mm/mm_test_utils.h
```diff
@@ -29,18 +30,24 @@ namespace mm {
...
       throw std::bad_alloc();
     void *ret;
     if (security_check) {
-      ret = upstream_->allocate(extra..., bytes + sizeof(sentinel_t), alignment);
-      detail::write_sentinel<sentinel_t>(ret, bytes);
+      ret = ua(bytes + sizeof(sentinel_t), alignment);
+      detail::write_sentinel<sentinel_t>(ret, bytes, std::forward<Extra>(extra)...);
     } else {
-      ret = upstream_->allocate(extra..., bytes, alignment);
+      ret = upstream_->allocate(bytes, alignment, std::forward<Extra>(extra)...);
```
JanuszL, line 50: Why use `ua` in one place and `upstream_` directly in the second?

dali/core/mm/mm_test_utils.h
```diff
@@ -29,18 +30,24 @@ namespace mm {
 namespace test {
 
 template <bool owning, bool security_check,
-          typename Base, typename Upstream, typename... ExtraParams>
-class test_resource_wrapper : public Base {
+          typename Base, typename Upstream>
+class test_resource_wrapper;
+
+template <bool owning, bool security_check, typename Upstream>
+class test_resource_wrapper_impl {
```
JanuszL, line 37: Why you need to split wrapper into impl and wrapper itself?

dali/core/mm/mm_test_utils.h
mzient: It's a bug. It should be `ua`.

dali/core/mm/monotonic_resource_test.cc
```diff
@@ -69,25 +69,26 @@ TEST(MMTest, MonotonicHostResource) {
 }
 
 TEST(MMTest, MonotonicDeviceResource) {
-  test_host_resource upstream;
+  test_device_resource upstream;
```
JanuszL, line 72: So RMM is not applicable to the host memory yet, or there is another reason for this change?

include/dali/core/mm/monotonic_resource.h
```diff
@@ -260,6 +272,10 @@ class monotonic_device_resource : public memory_resource {
   SmallVector<upstream_block, 8> blocks_;
 };
 
+using monotonic_host_resource = monotonic_memory_resource<memory_kind::host>;
```
JanuszL, line 275: Maybe I'm missing something, but order doesn't have any default value nor is provided in this using statement.

mzient: The default argument is defined in line 84.

dali/core/mm/monotonic_resource_test.cc
mzient: Previously the resources did not distinguish memory type.

dali/core/mm/mm_test_utils.h
mzient: The `impl` has template callable UpstreamAlloc and UpstreamDealloc which can resolve to `upstream_->(de)allocate` or `upstream_->(de)allocate_async` (with stream). Without that I'd need to have 2 or 3 near-copies of alloc/dealloc functions.

JanuszL: ok

dali/core/mm/monotonic_resource_test.cc
JanuszL: Understood

include/dali/core/mm/monotonic_resource.h
JanuszL: Missed that.

jantonguirao (approved):

JanuszL (approved):

PR #2634, opened 2021-01-24
Add handling of empty inputs and tiny outputs in Resize op and Resampling kernels.

mzient:

#### Why we need this PR?
*Pick one, remove the rest*
- It fixes a bug:
    * division by zero when empty output sample is encountered in GPU resampling setup
    * incorrect indexing resulting in spurious assertion in CPU resampling when empty output is encountered
    * incorrect DALI_ENFORCE when requested output size is nonzero before adjusting to empty input (with not_larger/resize_longer)

 - What solution was applied:
    * skip generating blocks for empty samples (in all passes)
    * early return in CPU resampling when empty output is requested
    * don't force output to at least 1 pixel when input sample is empty
    * check for non-empty output for empty input after adjusting the input size, not before (mitigating incorrect ENFORCE)
    * limit the ROI adjustment factor to 10
    * limit ROI to reasonable range
    * limit filter support to 8k samples (32kB) to avoid shared memory overrun
     * Resampling, resize
     * N/A
 - Validation and testing:
     * Python test with empty inputs
     * Python test with tiny output
     * N/A

**JIRA TASK**: N/A

### DIFF
--- a/dali/kernels/imgproc/resample/resampling_setup.cc
+++ b/dali/kernels/imgproc/resample/resampling_setup.cc
@@ -208,6 +208,11 @@ void SeparableResamplingSetup<2>::ComputeBlockLayout(SampleDesc &sample) const {
  */
 template <>
 void SeparableResamplingSetup<3>::ComputeBlockLayout(SampleDesc &sample) const {
+  if (volume(sample.out_shape()) == 0) {
+    for (int pass = 0; pass < 3; pass++)
+      sample.logical_block_shape[pass] = 0;
+    return;
+  }
   for (int pass = 0; pass < 3; pass++) {
     int axis = sample.order[pass];
 
@@ -352,7 +357,10 @@ void BatchResamplingSetup<spatial_ndim>::SetupBatch(
     auto ts_out = output_shape.tensor_shape_span(i);
     static_assert(channel_dim == spatial_ndim, "Shape calculation requires channel-last layout");
     auto sample_shape = shape_cat(vec2shape(desc.out_shape()), desc.channels);
+
     output_shape.set_tensor_shape(i, sample_shape);
+    if (volume(desc.out_shape()) == 0)
+      continue;  // this sample does not generate any blocks
 
     for (int pass = 0; pass < spatial_ndim; pass++) {
       ivec<spatial_ndim> blocks;
@@ -405,9 +413,11 @@ void BatchResamplingSetup<spatial_ndim>::InitializeSampleLookup(
   for (int pass = 0; pass < spatial_ndim; pass++) {
     for (int i = 0; i < N; i++) {
       auto &desc = sample_descs[i];
-      int sample_block_count = AddBlocks(sample_lookup.data + block, i,
-                                         desc.logical_block_shape[pass], desc.shapes[pass+1]);
-      block += sample_block_count;
+      if (volume(desc.out_shape()) > 0) {
+        int sample_block_count = AddBlocks(sample_lookup.data + block, i,
+                                          desc.logical_block_shape[pass], desc.shapes[pass+1]);
+        block += sample_block_count;
+      }
     }
   }
   assert(block == blocks_in_all_passes);
--- a/dali/kernels/imgproc/resample/resampling_setup.cc
+++ b/dali/kernels/imgproc/resample/resampling_setup.cc
@@ -22,6 +22,8 @@ namespace dali {
 namespace kernels {
 namespace resampling {
 
+constexpr int kMaxGPUFilterSupport = 8192;
+
 ResamplingFilter GetResamplingFilter(const ResamplingFilters *filters, const FilterDesc &params) {
   switch (params.type) {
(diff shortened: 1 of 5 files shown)


mzient:
!build

mzient:
!build

dali/kernels/imgproc/resample/resampling_setup.cc
```diff
@@ -208,6 +212,11 @@ void SeparableResamplingSetup<2>::ComputeBlockLayout(SampleDesc &sample) const {
  */
 template <>
 void SeparableResamplingSetup<3>::ComputeBlockLayout(SampleDesc &sample) const {
+  if (volume(sample.out_shape()) == 0) {
```
klecki, line 215: Why there is no similar change for `SeparableResamplingSetup<2>`?

dali/test/python/test_operator_resize.py
```diff
@@ -527,3 +527,72 @@ def test_stitching():
                 for channel_first in [False, True]:
                     for interp in [types.INTERP_LINEAR, types.INTERP_CUBIC, types.INTERP_TRIANGULAR, types.INTERP_LANCZOS3]:
                         yield _test_stitching, device, dim, channel_first, dtype, interp
+
+def _test_empty_input(dim, device):
```
klecki, line 531: You can use: 
`from nose.tools import nottest`
```suggestion
@nottest
def test_empty_input(dim, device):
```

to state the intention and not make one wander why it's private. Or I would just write `check_empty_input`.

dali/test/python/test_operator_resize.py
```diff
@@ -527,3 +527,72 @@ def test_stitching():
...
+    pipe.build()
+
+    for it in range(3):
+        out_no_empty, out_with_empty = pipe.run()
+        if device == "gpu":
+            out_no_empty = out_no_empty.as_cpu()
+            out_with_empty = out_with_empty.as_cpu()
+        for i in range(batch_size):
+            if i%2 != 0:
+                assert np.array_equal(out_no_empty.at(i), out_with_empty.at(i))
```
klecki, line 564: We deprecated `at` in favor of `__getitem__`

dali/test/python/test_operator_resize.py
```diff
@@ -527,3 +527,72 @@ def test_stitching():
...
+    images = images_cpu if device == "cpu" else images_cpu.gpu()
+
+    resize_tiny = fn.resize(images, size = 1e-10)
+
+    pipe.set_outputs(resize_tiny)
+    pipe.build()
+
+    for it in range(3):
+        out, = pipe.run()
+        ref_size = [1,1,1,1] if dim == 3 else [1,1,3]
```
klecki, line 591: Why for DHWC we end with `[1, 1, 1, 1]` and for HWC with `[1, 1, 3]`?

dali/test/python/test_operator_resize.py
JanuszL: For CPU inputs "at" returns numpy, while `[]` returns tensor....

klecki: But it will spew warnings in the tests.
```suggestion
                assert np.array_equal(np.array(out_no_empty[i]), np.array(out_with_empty[i]))
```

dali/kernels/imgproc/resample/resampling_setup.cc
mzient: because the 2D version does not contain division - it will end up with zeros in logical_block_shape anyway.

dali/test/python/test_operator_resize.py
mzient: It does not.

dali/test/python/test_operator_resize.py
mzient: Because the data is monochrome.

dali/test/python/test_operator_resize.py
mzient: > `@nottest`
I can add `@nottest`, but I'd keep the underscore - `check_` functions have a different purpose in our tests (_vide_ `check_batch`). Still, that's just more typing and the pattern of public enumerator and private test implementation is rather pervasive in our tests.

> `check_empty_input`
I'd call that confusing - sounds too much like our `check_batch`.

klecki: I guess it's anecdotal evidence, as I think it's first time I see functions starting with underscore and I rather see a lot of `check_something` to circumvent the nose running `test_something` functions. Whatever.

dali/kernels/imgproc/resample/resampling_setup.cc
klecki: :+1:

dali/test/python/test_operator_resize.py
klecki: That's consistency of our APIs for you.

klecki (approved):

awolant (approved):

PR #2636, opened 2021-01-25
Fix SeparateQueuePolicy handling of the CPU stage

JanuszL:
- when a CPU-only pipeline was introduced MakeContiguousCPU was created and CPU inputs to the GPU stage can be generated directly by the CPU stage as well, not only the Mixed one. SeparateQueuePolicy assumes otherwise and overwrites the CPU stage buffer as soon as it is consumed by the Mixed stage.  This PR makes the CPU stage a first-class citizen on SeparateQueuePolicy

#### Why we need this PR?
*Pick one, remove the rest*
- It fixes a bug in SeparateQueuePolicy handling of the CPU stage

 - What solution was applied:
     when a CPU-only pipeline was introduced MakeContiguousCPU was created and CPU inputs to the GPU stage can be generated directly by the CPU stage as well, not only the Mixed one. SeparateQueuePolicy assumes otherwise and overwrites the CPU stage buffer as soon as it is consumed by the Mixed stage
    makes the CPU stage a first-class citizen on SeparateQueuePolicy
     executor
     executor?
 - Validation and testing:
     added a test for CPU only SeparateQueuePolicy
     run `python test_RN50_data_pipeline.py --gpus 1 -b 10 --workers 3 --separate_queue --cpu_size 2 --gpu_size 2 --fp16 --nhwc -s` in a loop
     NA

**JIRA TASK**: *[DALI-1821]*

### DIFF
--- a/dali/pipeline/executor/executor.cc
+++ b/dali/pipeline/executor/executor.cc
@@ -57,7 +57,8 @@ void Executor<WorkspacePolicy, QueuePolicy>::RunCPU() {
   DeviceGuard g(device_id_);
 
   auto cpu_idxs = QueuePolicy::AcquireIdxs(OpType::CPU);
-  if (exec_error_ || QueuePolicy::IsStopSignaled() || !QueuePolicy::AreValid(cpu_idxs)) {
+  if (exec_error_ || QueuePolicy::IsStopSignaled() ||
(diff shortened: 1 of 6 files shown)


JanuszL:
!build

banasraf (approved):

klecki (approved):

PR #2637, opened 2021-01-26
Fix variable batch size for list of tensors. Make Constant op constant again

mzient:

#### Why we need this PR?
*Pick one, remove the rest*
- It fixes a bug: assertion when providing variable batch as a list of tensors.
- It doesn not recreate constants when batch size changes.

 - What solution was applied:
     * changed the condition in BackendImpl from `== p->batch_size()` to ` <= p->max_batch_size()`
     * max_batch_size TensorList/TensorVector is populated and then only trimmed to required length.
     * Python interface for Pipeline
     * Constant op
     * Constant op tests
     * N/A
 - Validation and testing:
     * Python test
     * documentation stated that Constants are never reallocated which is true... again

**JIRA TASK**: N/A

### DIFF
--- a/dali/test/python/test_operator_constant.py
+++ b/dali/test/python/test_operator_constant.py
@@ -14,6 +14,7 @@
 
 from nvidia.dali.pipeline import Pipeline
 import nvidia.dali.ops as ops
+import nvidia.dali.fn as fn
 import nvidia.dali.types as types
 import numpy as np
 import os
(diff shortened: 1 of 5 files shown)


mzient:
!build

dali/test/python/test_operator_constant.py
```diff
@@ -162,3 +163,23 @@ def test_constant_fn():
 def test_scalar_constant_promotion():
     yield _test_scalar_constant_promotion, "cpu"
     yield _test_scalar_constant_promotion, "gpu"
+
```
JanuszL, line 166: Maybe this should go to test_dali_variable_batch_size.py?

JanuszL (approved):

dali/test/python/test_operator_constant.py
mzient: It's just my opinion, but I don't like the whole idea of test_dali_variable_batch_size - it's a bag of very shallow tests. When we have variable batch size support across the board, we might just use variable batch size for new tests for other ops.

JanuszL: So do we want to extend every operator's test to cover batch size variability more thoroughly?

mzient: When it doesn't require jumping though too many hoops, then yes. Currently any test that uses readers is not exactly easily convertible to variable batch size.

szalpal (approved):

PR #2656, opened 2021-02-04
Workaround a compiler problem that caused Invalid device function error.

mzient:

#### Why we need this PR?
*Pick one, remove the rest*
- It fixes bugs:
    * Invalid device function when calling operator Slice on half precision data.
    * Unknown type: float16 in DLTensor

 - What solution was applied:
     * WAR Use a compiler-dependent macro instead of an identity typedef
     * use `DALI_TYPE_SWITCH_WITH_FP16` and `is_fp_or_half` in DLTensor
     * float16 header
     * Slice and SliceFlipNormalize kernels
     * Slice tests
     * DLTensor
     * N/A
 - Validation and testing:
     * Python tests
     * N/A

**JIRA TASK**: DALI-1831

### DIFF
--- a/dali/kernels/slice/slice_gpu.cuh
+++ b/dali/kernels/slice/slice_gpu.cuh
@@ -256,7 +256,7 @@ class SliceGPU {
     for (int i = 0; i < in.size(); i++) {
       if (default_fill_values_) {
         assert(nfill_values_ == 1);
-        fill_values_cpu[i] = static_cast<OutputType>(0.f);
+        fill_values_cpu[i] = OutputType{};
       } else {
         auto *fill_values = fill_values_cpu + i * nfill_values_;
         for (int d = 0; d < nfill_values_; d++)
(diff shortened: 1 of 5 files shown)


mzient:
!build

dali/pipeline/data/dltensor.cc
```diff
@@ -19,10 +19,10 @@ namespace dali {
 
 DLDataType GetDLType(const TypeInfo &type) {
   DLDataType dl_type{};
-  DALI_TYPE_SWITCH(type.id(), T,
+  DALI_TYPE_SWITCH_WITH_FP16(type.id(), T,
       dl_type.bits = sizeof(T) * 8;
       dl_type.lanes = 1;
-      if (std::is_floating_point<T>::value) {
+      if (dali::is_fp_or_half<T>::value) {
```
JanuszL, line 25: Bad idea. I'm not sure if fp16 is supported by dlpack at all. It supports bfloat but not fp16. Now you would put fp16 inside and claim it to be float.

JanuszL: Check https://github.com/dmlc/dlpack/blob/main/include/dlpack/dlpack.h#L84

JanuszL: Ok. Slice tests checks this indirectly.

JanuszL (approved):

dali/pipeline/data/dltensor.cc
mzient: For the record: DLPack capsules created this way are compatible with PyTorch. My guess is that it's a _de-facto_ standard for float16 DLPack tensors now.

dali/kernels/slice/slice_gpu.cuh
```diff
@@ -256,7 +256,7 @@ class SliceGPU {
     for (int i = 0; i < in.size(); i++) {
       if (default_fill_values_) {
         assert(nfill_values_ == 1);
-        fill_values_cpu[i] = static_cast<OutputType>(0.f);
+        fill_values_cpu[i] = OutputType{};
```
klecki, line 259: Can you check if this works as intended? When I was writing the GaussianBlur, the compiler did some weird things if the output was half, and I had to initialize the 0 by using the conversion from float (that one worked without problems).

klecki (commented):
Looks ok, but I had bad experience with the commented constructor being used.

klecki (approved):

PR #2671, opened 2021-02-10
Fix apex compilation on Ubuntu 20.04 in TL1_ssd_training

JanuszL:
- Ubuntu 20.04 has gcc-9 as a default compiler which is not supported by CUDA 10.0 used in the TL1_ssd_training test. Install gcc-7 in the test and set it as a one that should be used to build apex

#### Why we need this PR?
*Pick one, remove the rest*
- It fixes apex compilation on Ubuntu 20.04 in TL1_ssd_training

 - What solution was applied:
     Ubuntu 20.04 has gcc-9 as a default compiler which is not supported by CUDA 10.0 used in the TL1_ssd_training test. Install gcc-7 in the test and set it as a one that should be used to build apex
     TL1_ssd_training 
     NA
 - Validation and testing:
     present tests apply
     NA

**JIRA TASK**: *[NA]*

### DIFF
--- a/qa/TL1_ssd_training/test.sh
+++ b/qa/TL1_ssd_training/test.sh
@@ -6,6 +6,7 @@ target_dir=./docs/examples/use_cases/pytorch/single_stage_detector/
 test_body() {
     # workaround due to errors while pycocotools is int "pip_packages" above
     pip install -I pycocotools
+    apt update && apt install -y gcc-7 g++-7
 
     #install APEX
     git clone https://github.com/nvidia/apex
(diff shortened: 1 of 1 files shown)


klecki (commented):
Same as #2669, I guess you pushed wrong branch.

JanuszL:
> Same as #2669, I guess you pushed wrong branch.

My bad. Fixed.

klecki (approved):

awolant (approved):

PR #2676, opened 2021-02-15
Make DALI pipeline use default seed (-1) when None is set to seed

JanuszL:

#### Why we need this PR?
*Pick one, remove the rest*
- It makes DALI pipeline use default seed (-1) when None is set to seed

 - What solution was applied:
     allows passing None as a pipeline's seed and translate it internally to -1
     pipeline.py
     NA
 - Validation and testing:
     new test is added
     NA

**JIRA TASK**: *[NA]*

### DIFF
--- a/dali/test/python/test_pipeline.py
+++ b/dali/test/python/test_pipeline.py
@@ -424,6 +424,19 @@ def test_seed():
             img_chw = img_chw_test
         assert(np.sum(np.abs(img_chw - img_chw_test)) == 0)
 
+def test_none_seed():
+    batch_size = 10
+    shape = (120, 60, 3)
+    pipe = Pipeline(batch_size=batch_size, num_threads=4, device_id=0, seed=None)
+    data = RandomDataIterator(batch_size, shape=shape, dtype=np.uint8)
+    with pipe:
+        input = fn.external_source(data, layout="HWC")
+        coin = fn.random.coin_flip(seed=None)
+        pipe.set_outputs(input, coin)
+    pipe.build()
+
+    result = pipe.run()
+
 def test_as_array():
     batch_size = 64
     class HybridPipe(Pipeline):
--- a/dali/test/python/test_pipeline.py
+++ b/dali/test/python/test_pipeline.py
@@ -425,17 +425,20 @@ def test_seed():
         assert(np.sum(np.abs(img_chw - img_chw_test)) == 0)
 
 def test_none_seed():
-    batch_size = 10
-    shape = (120, 60, 3)
-    pipe = Pipeline(batch_size=batch_size, num_threads=4, device_id=0, seed=None)
-    data = RandomDataIterator(batch_size, shape=shape, dtype=np.uint8)
-    with pipe:
-        input = fn.external_source(data, layout="HWC")
-        coin = fn.random.coin_flip(seed=None)
-        pipe.set_outputs(input, coin)
-    pipe.build()
+    batch_size = 60
 
-    result = pipe.run()
+    for i in range(50):
+        pipe = Pipeline(batch_size=batch_size, num_threads=2, device_id=0, seed=None)
+        with pipe:
+            coin = fn.random.uniform(range = (0.0,1.0))
+        pipe.set_outputs(coin)
+        pipe.build()
+        pipe_out = pipe.run()[0]
+        img_chw_test = pipe_out.as_array()
+        if i == 0:
+            img_chw = img_chw_test
+        else:
+            assert(np.sum(np.abs(img_chw - img_chw_test)) != 0)
 
 def test_as_array():
     batch_size = 64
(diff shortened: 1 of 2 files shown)


JanuszL:
!build

jantonguirao (approved):

szalpal (dismissed):

docs/examples/use_cases/tensorflow/resnet-n/nvutils/image_processing.py
```diff
@@ -217,6 +217,9 @@ def __init__(self,
     device_id = hvd.local_rank()
     shard_id = hvd.rank()
     num_gpus = hvd.size()
+    kwargs = dict()
+    if deterministic:
+       kwargs['seed'] = 7 * (1 + hvd.rank())
```
szalpal, line 222: Correct me if I'm wrong, but If `deterministic == False`, `kwargs['seed']` wouldn't still be None?
Shouldn't it be like:
```python
kwargs['seed']=7 * (1 + hvd.rank()) if deterministic else -1
```

JanuszL: If deterministic is false then there is no such keyword arg and the default DALI value use used.

szalpal (approved):

JanuszL:
!build

dali/test/python/test_pipeline.py
```diff
@@ -424,6 +424,19 @@ def define_graph(self):
...
 
+def test_none_seed():
+    batch_size = 10
+    shape = (120, 60, 3)
+    pipe = Pipeline(batch_size=batch_size, num_threads=4, device_id=0, seed=None)
+    data = RandomDataIterator(batch_size, shape=shape, dtype=np.uint8)
+    with pipe:
+        input = fn.external_source(data, layout="HWC")
+        coin = fn.random.coin_flip(seed=None)
+        pipe.set_outputs(input, coin)
```
mzient, line 435:
```suggestion
    with pipe:
        coin = fn.random.coin_flip(seed=None)
        pipe.set_outputs(coin)
```
...unless RandomDataIterator is driving variable batch size

dali/test/python/test_pipeline.py
```diff
@@ -424,6 +424,19 @@ def define_graph(self):
             img_chw = img_chw_test
         assert(np.sum(np.abs(img_chw - img_chw_test)) == 0)
 
+def test_none_seed():
+    batch_size = 10
+    shape = (120, 60, 3)
+    pipe = Pipeline(batch_size=batch_size, num_threads=4, device_id=0, seed=None)
```
mzient, line 430: A useful test (if we indeed need any?) would be to build two pipelines and check that they have different output. This kind of test would possibly need a few re-runs in case we get extremely unlucky - however, if pipeline gives us exactly the same output, it would indicate a regression.

JanuszL: Done

dali/test/python/test_pipeline.py
JanuszL: Removed

JanuszL:
!build

dali/test/python/test_pipeline.py
```diff
@@ -424,6 +424,22 @@ def define_graph(self):
...
+    batch_size = 60
+
+    for i in range(50):
+        pipe = Pipeline(batch_size=batch_size, num_threads=2, device_id=0, seed=None)
+        with pipe:
+            coin = fn.random.uniform(range = (0.0,1.0))
+        pipe.set_outputs(coin)
+        pipe.build()
+        pipe_out = pipe.run()[0]
+        img_chw_test = pipe_out.as_array()
```
mzient, line 437: Nitpick:
```suggestion
        test_out = pipe_out.as_array()
```

dali/test/python/test_pipeline.py
```diff
@@ -424,6 +424,22 @@ def define_graph(self):
...
+        if i == 0:
+            img_chw = img_chw_test
+        else:
+            assert(np.sum(np.abs(img_chw - img_chw_test)) != 0)
```
mzient, line 441: This test will require that we don't get the same seed in any of the iterations - I thought about the opposite (check that in N pipelines there's at least one different seed). I don't require that, but in that case, running it just 2 times would be enough.

dali/test/python/test_pipeline.py
JanuszL: Done

JanuszL:
!build

dali/test/python/test_pipeline.py
JanuszL: We want to avoid that. I remember a case when two pipelines received the same seed because they were launched one after another and that can be bad (when you have a DDP strategy launching multiple pipelines in parallel).

mzient (approved):

PR #2681, opened 2021-02-16
GPU MultiPaste

TheTimemaster:
#### Why we need this PR?
- It adds a new feature needed to do MultiPaste on GPU

 - What solution was applied:
      The workflow is, that images are divided into an unequal grid where each cell contains data from only one paste, that one which happened last over this area.  This grid is counted in the operator. Then kernel just copies data, but before each pixel, it has to check if it should jump to the next input cell vertically or horizontally.
     GPU Implementation for Multipaste operator and kernel.
     MultiPaste GPU kernel + multipaste.cu operator
 - Validation and testing:
     For the grid sweeping algorithm, it was implemented and tested outside dali: https://pastebin.com/0QSnu3Bw. Now automatic tests are added that run the same cases but on GPU.

     No new documentation added

**JIRA TASK**: NA


TheTimemaster:
The GPU operator seems to be working now and passes the tests most of time.

Sometimes though, with a roughly 20% chance, there is a bug that produces random non-filled regions in the output:
[image](https://drive.google.com/file/d/1TsyE3kO6GXCSBRD1rnnNi6x0QcOAmcMv/view?usp=sharing)

I made non-pasted regions be gray, not black. That would produce the following output, that has a gray background, but still has black glitched regions:
[image](https://drive.google.com/file/d/1KtBWsnoO0YndxwrMaej8DYPEHuLNHhn0/view?usp=sharing)

Finally, I tried removing copying of the image completely, and instead, it would assign a gray shade depending on the grid cell. This should never output 0 to any pixel, also the values were not taken from any pointer. Yet the glitched black regions remained:
[image](https://drive.google.com/file/d/1h2fjIwYmfYCNnT8mKYLSaYKKKvVz3hbP/view?usp=sharing)

Also, the operation did not crash, so each thread should have gone through the whole for loop, yet it seems some of them sometimes do not.

dali/operators/image/paste/multipaste.cu
```diff
@@ -0,0 +1,210 @@
...
+  const auto &images = ws.template Input<GPUBackend>(0);
+  auto &output = ws.template Output<GPUBackend>(0);
+
+  output.SetLayout(images.GetLayout());
+  auto out_shape = output.shape();
+  using Kernel = kernels::PasteGPU<OutputType, InputType, 3>;
+  auto in_view = view<const InputType, 3>(images);
+  auto out_view = view<OutputType, 3>(output);
+
+  kernels::KernelContext ctx;
```
mzient, line 196:
```suggestion
  kernels::KernelContext ctx;
  ctx.gpu.stream = ws.stream();
```

dali/kernels/imgproc/paste/paste_gpu.h
```diff
@@ -0,0 +1,247 @@
...
+ */
+template <int ndims>
+TensorListShape<ndims - 1> FlattenChannels(const TensorListShape<ndims> &shape) {
+  static_assert(ndims >= 2, "If there are less than 2 dims, there's nothing to flatten...");
+  TensorListShape<ndims - 1> ret(shape.size());
+  for (int i = 0; i < shape.size(); i++) {
+    ret.set_tensor_shape(i, FlattenChannels<ndims>(shape[i]));
+  }
+  return ret;
+}
```
mzient, line 81: Just use `collapse_dim` for that
```suggestion
```

dali/kernels/imgproc/paste/paste_gpu.h
```diff
@@ -0,0 +1,247 @@
...
+        if (in.shape[i][ndims - 1] != ref_nchannels) {
+          return false;
+        }
+      }
+      return true;
+    }(), "Number of channels for every image in batch must be equal");
+
+    KernelRequirements req;
+    ScratchpadEstimator se;
+    TensorListShape<spatial_dims> flattened_shape(paste::FlattenChannels<ndims>(out_shape));
```
mzient, line 208:
```suggestion
    auto flattened_shape = collapse_dim(out_shape, 1);  // merge dim 1 (width) with 2 (channels)
```

dali/kernels/imgproc/paste/paste_gpu_input.h
```diff
@@ -0,0 +1,38 @@
...
+struct GridCellInput {
+    int input_idx;
+    ivec<ndims> cell_start, cell_end, in_anchor;
+};
+
+template <int ndims>
+struct MultiPasteSampleInput {
+    int grid_cell_start_idx;
+    ivec<ndims> cell_counts;
+};
```
mzient, line 32: Honestly, I have hard time deciphering what this "grid cell" is and what it's for.

dali/kernels/imgproc/paste/paste_gpu.h
```diff
@@ -0,0 +1,247 @@
...
+
+namespace dali {
+namespace kernels {
+namespace paste {
+
+template<class InputType, int ndims>
+struct GridCellGPU {
+    const InputType *in;
+    ivec<ndims> cell_start, cell_end, in_anchor, in_pitch;
+};
```
mzient, line 36: Grid is, by definition, something regular, sostart/end/anchor and pitch dont look right. Or is it more like `PatchDesc`?

dali/kernels/imgproc/paste/paste_gpu.h
```diff
@@ -0,0 +1,247 @@
...
+      if (cell->in == nullptr) {
+          out[y * sample.out_pitch[1] + x] = 0;
+      } else {
+        out[y * sample.out_pitch[1] + x] = ConvertSat<OutputType>(
+                cell->in[(y - cell->cell_start[0] + cell->in_anchor[0]) * cell->in_pitch[1]
+                         + (x - cell->cell_start[1] + cell->in_anchor[1])]);
+      }
+    }
+  }
+}
```
mzient, line 172: This doesn't scale well - you could have some kind of block descriptor that would contain the range of contributing grid cells (possibly empty, if there are no data-containing cells). With that approach, most of the blocks would have much fewer contributing cells (typically 1).

mzient: On the other hand, assuming that the logical block is not too large compared to inputs, you could also have just a list of contributing _inputs_ for each block - then you'd just traverse it backwards in each thread, yielding the topmost one. This is much simpler to implement and should be reasonably fast.

dali/kernels/imgproc/paste/paste_gpu.h
```diff
@@ -0,0 +1,247 @@
...
+    req.scratch_sizes = se.sizes;
+    return req;
+  }
+
+
+  void Run(
+      KernelContext &context,
+      const OutListGPU<OutputType, ndims> &out, const InListGPU<InputType, ndims> &in,
+      const std::vector<paste::MultiPasteSampleInput<spatial_dims>> &samples,
+      const std::vector<paste::GridCellInput<spatial_dims>> &grid) {
```
mzient, line 225: This isn't a good design - preparing the grid cells is by far the most complicated part of the whole affair - and it's left for the caller to provide. This is tied closely to the particular implementation and should either be a part of this kernel or a utility function shipped along the kernel - in any case, the algorithm of preparing the grid cells should be opaque - but see my other comments if that's indeed required.

dali/kernels/imgproc/paste/paste_gpu.h
```diff
@@ -0,0 +1,247 @@
...
+    // req.output_shapes = {in.shape}; Once again, this is determined by operator
+    req.scratch_sizes = se.sizes;
+    return req;
+  }
+
+
+  void Run(
+      KernelContext &context,
+      const OutListGPU<OutputType, ndims> &out, const InListGPU<InputType, ndims> &in,
+      const std::vector<paste::MultiPasteSampleInput<spatial_dims>> &samples,
```
mzient, line 224:
```suggestion
      span<const paste::MultiPasteSampleInput<spatial_dims>> samples,
```

dali/test/python/test_operator_multipaste.py
```diff
@@ -220,19 +224,34 @@ def test_operator_multipaste():
...
+        [4, 2, (128, 128), (128, 256), True, True, False, False, False, types.UINT8, True],
+
+        [4, 2, (64, 64), (128, 128), False, False, True, False, False, types.UINT8, True],
+        [4, 2, (64, 64), (128, 128), False, False, False, True, False, types.UINT8, True],
+        [4, 2, (64, 64), (128, 128), False, False, False, False, True, types.UINT8, True],
+
+        [4, 2, (128, 128), (128, 128), False, False, False, False, False, types.UINT8, True],
+        [4, 2, (128, 128), (128, 128), False, False, False, False, False, types.INT16, True],
+        [4, 2, (128, 128), (128, 128), False, False, False, False, False, types.INT32, True],
+        [4, 2, (128, 128), (128, 128), False, False, False, False, False, types.FLOAT, True]
```
mzient, line 254: This doubling is not necessary...

dali/test/python/test_operator_multipaste.py
```diff
@@ -220,19 +224,34 @@ def test_operator_multipaste():
...
     ]
     for t in tests:
         yield (check_operator_multipaste, *t)
```
mzient, line 257: Instead of copy-pasting the table you could do:
```suggestion
        yield (check_operator_multipaste, *t, "cpu")
        yield (check_operator_multipaste, *t, "gpu")
```

dali/test/python/test_operator_multipaste.py
```diff
@@ -187,7 +189,7 @@ def show_images(batch_size, image_batch):
 
 
 def check_operator_multipaste(bs, pastes, in_size, out_size, even_paste_count, no_intersections, full_input, in_anchor_top_left,
-                  out_anchor_top_left, out_dtype):
+                              out_anchor_top_left, out_dtype, use_gpu):
```
mzient, line 192:
```suggestion
                              out_anchor_top_left, out_dtype, device):
```

dali/kernels/imgproc/paste/paste_gpu.h
TheTimemaster: Assume that the grid size is GridX by GridY cells and the current block intersects with a subgrid of size BlockGridX by BlockGridY.

In each thread the while loop for Y will run at most GridY times. 
I changed the kernel a bit. Now each thread finds minimum_grid_x before it runs, which takes GridX iterations of the loop.

Thanks to that change, though, when looping over pixels themself, `grid_x++` in the while will run `BlockGridX` times for each `y` level, instead of `GridX`, as it was in the previous solution. With this change, I think the performance should be about equal to your first solution.

Of course it has to check if it should increment `grid_x` before filling every pixel, but if I understand correctly that would also happen in your first solution, even if the contributing cell count was 1.

We can get rid of (GridX + GridY) constant by assigning min_grid_x and min_grid_y for each block, but I do not have any idea how to precompute something for each block.

dali/kernels/imgproc/paste/paste_gpu.h
TheTimemaster: True, changed name to PatchDesc

dali/kernels/imgproc/paste/paste_gpu.h
TheTimemaster: Moved patch generation to the kernel

dali/kernels/imgproc/paste/paste_gpu.h
```diff
@@ -0,0 +1,293 @@
...
+  }
+}
+
+
+
+template <class OutputType, class InputType, int ndims>
+__global__ void
+PasteKernel(const SampleDescriptorGPU<OutputType, InputType, ndims> *samples,
+                  const PatchDesc<InputType, ndims> *patchs,
+                  const BlockDesc<ndims> *blocks) {
```
mzient, line 202: Nitpick:
```suggestion
            const PatchDesc<InputType, ndims> *patchs,
            const BlockDesc<ndims> *blocks) {
```

dali/kernels/imgproc/paste/paste_gpu.h
```diff
@@ -0,0 +1,293 @@
...
+  std::vector<PatchDesc> patch_descriptors_;
+
+ public:
+  BlockSetup<spatial_dims, -1 /* No channel dimension, only spatial */> block_setup_;
+
+  KernelRequirements Setup(
+      KernelContext &context,
+      const InListGPU<InputType, ndims> &in,
+      span<paste::MultiPasteSampleInput<spatial_dims>> samples,
+      TensorListShape<ndims> out_shape) {
```
mzient, line 253: It's a relatively heavy object.
```suggestion
      const TensorListShape<ndims> &out_shape) {
```

dali/kernels/imgproc/paste/paste_gpu.h
```diff
@@ -0,0 +1,293 @@
...
+    paste::CreateSampleDescriptors(sample_descriptors_, patch_descriptors_, in, samples);
+
+    KernelRequirements req;
+    ScratchpadEstimator se;
+    auto flattened_shape = collapse_dim(out_shape, 1);
+    block_setup_.SetupBlocks(flattened_shape, true);
+    se.add<SampleDesc>(AllocType::GPU, sample_descriptors_.size());
+    se.add<PatchDesc>(AllocType::GPU, patch_descriptors_.size());
+    se.add<BlockDesc>(AllocType::GPU, block_setup_.Blocks().size());
+    // req.output_shapes = {in.shape}; Once again, this is determined by operator
```
mzient, line 263: If we still use KernelRequirements, I guess we should output the shape - even though here it doesn't make much sense (and it's quite heavy).
```suggestion
    req.output_shapes = { out_shape };
```

dali/kernels/imgproc/paste/paste_gpu_input.h
```diff
@@ -0,0 +1,37 @@
...
+
+namespace dali {
+namespace kernels {
+namespace paste {
+
+
+template <int ndims>
+struct MultiPasteSampleInput {
+    vector<ivec<ndims>> in_anchors, out_anchors, sizes;
+    vector<int> in_idx;
```
mzient, line 28: It would be cheaper and more reliable to have an array of structures - single allocation, one size that's always consistent.
```suggestion
    struct InputPatch {
      ivec<ndims> out_anchor, in_anchor, size;
      int in_idx;
    };
    vector<InputPatch> inputs;
```

dali/test/python/test_operator_multipaste.py
```diff
@@ -220,6 +224,7 @@ def test_operator_multipaste():
         # - should "in_anchors" parameter be omitted
         # - should "out_anchors" parameter be omitted
         # - output dtype
+        # - should use GPU operator
```
mzient, line 227:
```suggestion
        # - backend
```

dali/kernels/imgproc/paste/paste_gpu.h
```diff
@@ -0,0 +1,293 @@
...
+ * @see FlattenChannels
+ */
+template <class OutputType, class InputType, int ndims>
+void CreateSampleDescriptors(
+    vector<SampleDescriptorGPU<OutputType, InputType, ndims - 1>> &out_descs,
+    vector<PatchDesc<InputType, ndims - 1>> &out_patchs,
+    const InListGPU<InputType, ndims> &in,
+    span<paste::MultiPasteSampleInput<ndims - 1>> samples
+) {
+  assert(ndims == 3);
```
mzient, line 73: How does it work with rectangles that are out of output bounds?
For example:
output is 640x480
input starts at 600, 400 and is specified as 100x100.

We should either clip (preferable) or throw, but I don't see it happening here.

dali/kernels/imgproc/paste/paste_gpu.h
```diff
@@ -0,0 +1,293 @@
...
+/**
+ * Note: Since the operation we perform is channel-agnostic (it is performed in the same
+ * way for every channel), SampleDescriptor assumes, that sample is channel-agnostic. Therefore it
+ * needs to be flattened
+ * @see FlattenChannels
+ */
+template <class OutputType, class InputType, int ndims>
+void CreateSampleDescriptors(
+    vector<SampleDescriptorGPU<OutputType, InputType, ndims - 1>> &out_descs,
+    vector<PatchDesc<InputType, ndims - 1>> &out_patchs,
```
mzient, line 69:
```suggestion
    vector<PatchDesc<InputType, ndims - 1>> &out_patches,
```

dali/test/python/test_operator_multipaste.py
```diff
@@ -232,7 +237,8 @@ def test_operator_multipaste():
         [4, 2, (128, 128), (128, 128), False, False, False, False, False, types.UINT8],
         [4, 2, (128, 128), (128, 128), False, False, False, False, False, types.INT16],
         [4, 2, (128, 128), (128, 128), False, False, False, False, False, types.INT32],
-        [4, 2, (128, 128), (128, 128), False, False, False, False, False, types.FLOAT]
+        [4, 2, (128, 128), (128, 128), False, False, False, False, False, types.FLOAT],
```
mzient, line 240: Do we have tests with out-of-bounds/clipping?

dali/test/python/test_operator_multipaste.py
```diff
@@ -148,7 +147,10 @@ def get_pipeline(
         if not out_anchor_top_left:
             kwargs["out_anchors"] = out_anchors
 
-        pasted = fn.multi_paste(resized, **kwargs)
+        if use_gpu:
+            kwargs["device"] = "gpu"
```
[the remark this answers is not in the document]
mzient: Yes, it will.

dali/test/python/test_operator_multipaste.py
```diff
@@ -148,7 +147,10 @@ def get_pipeline(
...
+
+        pasted = fn.multi_paste(resized.gpu() if use_gpu else resized, **kwargs)
```
mzient, line 153:
```suggestion
        pasted = fn.multi_paste(resized.gpu() if use_gpu else resized)
```

TheTimemaster: Will remove kwargs["device"], but there are other kwargs that still have to be passed

dali/kernels/imgproc/paste/paste_gpu.h
TheTimemaster: For now, I added throwing exceptions from the operator if some box is outside in/out bounds.

dali/test/python/test_operator_multipaste.py
TheTimemaster: Added those tests now

mzient (approved):

dali/test/python/test_operator_multipaste.py
```diff
@@ -201,15 +235,22 @@ def check_operator_multipaste(bs, pastes, in_size, out_size, even_paste_count, n
...
         use_gpu=device == 'gpu'
     )
     pipe.build()
-    result, input = pipe.run()
-    r = result.as_cpu() if device == 'gpu' else result
-    if SHOW_IMAGES:
-        show_images(bs, r)
-    manual_verify(bs, input, r, in_idx_l, in_anchors_l, shapes_l, out_anchors_l, [out_size + (3,)] * bs, out_dtype)
-
+    try:
```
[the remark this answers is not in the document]
timemaster-airev: I checked the 'assert_raises` method, but it only asserts the type (RuntimeError) of the exception and any `DALI_FAIL` results in this type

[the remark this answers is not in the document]
TheTimemaster: I checked the `assert_raises` method, but it only asserts the type (`RuntimeError`) of the exception, and any `DALI_FAIL` results in this type

awolant (approved):

awolant:
!build

JanuszL:
Clang complains about:
```
              ^
In file included from /opt/dali/dali/operators/image/paste/multipaste.cu:17:
/opt/dali/dali/kernels/imgproc/paste/paste_gpu.h:213:38: error: comparison of integers of different signs: 'unsigned int' and 'const int' [-Werror,-Wsign-compare]
  while (threadIdx.x + block.start.x >= my_patches[min_patch_x].patch_end[1]) min_patch_x++;
         ~~~~~~~~~~~~~~~~~~~~~~~~~~~ ^  ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/opt/dali/dali/kernels/imgproc/paste/paste_gpu.h:286:12: note: in instantiation of function template specialization 'dali::kernels::paste::PasteKernel<unsigned char, int, 2>' requested here
    paste::PasteKernel<<<patch_dim, block_dim, 0, stream>>>(
           ^
/opt/dali/dali/kernels/kernel_manager.h:249:24: note: in instantiation of member function 'dali::kernels::PasteGPU<unsigned char, int, 3>::Run' requested here
    inst.get<Kernel>().Run(context, std::forward<OutInArgs>(out_in_args)...);
                       ^
/opt/dali/dali/kernels/kernel_manager.h:221:5: note: in instantiation of function template specialization 'dali::kernels::KernelManager::Run<dali::kernels::PasteGPU<unsigned char, int, 3>, dali::TensorListView<dali::StorageGPU, unsigned char, 3> &, dali::TensorListView<dali::StorageGPU, const int, 3> &, dali::span<dali::kernels::paste::MultiPasteSampleInput<2>, -1> >' requested here
    Run<Kernel>(sa, instance_idx, context, std::forward<OutInArgs>(out_in_args)...);
    ^
/opt/dali/dali/operators/image/paste/multipaste.cu:102:19: note: in instantiation of function template specialization 'dali::kernels::KernelManager::Run<dali::kernels::PasteGPU<unsigned char, int, 3>, dali::TensorListView<dali::StorageGPU, unsigned char, 3> &, dali::TensorListView<dali::StorageGPU, const int, 3> &, dali::span<dali::kernels::paste::MultiPasteSampleInput<2>, -1> >' requested here
  kernel_manager_.Run<Kernel>(ws.thread_idx(), 0, ctx, out_view, in_view, make_span(samples));
                  ^
/opt/dali/dali/operators/image/paste/multipaste.cu:110:15: note: in instantiation of function template specialization 'dali::MultiPasteGPU::RunImplExplicitlyTyped<int, unsigned char>' requested here
              RunImplExplicitlyTyped<InputType, OutputType>(ws);
```
@TheTimemaster - can you check it?

JanuszL:
You can try `EXTRA_CMAKE_OPTIONS="-DCMAKE_CXX_COMPILER=clang++ -DCMAKE_C_COMPILER=clang -DDALI_CLANG_ONLY=ON" CUDA_VERSION=112 ./build.sh` to run clang build inside the container.

JanuszL:
!build

awolant (approved):

Issue #2648, opened 2021-02-01
Should ops.WarpAffine be equal to cv2.warpAffine?


zhiheng-ldj:
Hi,

Thanks for your help.

I discovered that the output of ops.WarpAffine is not equal to cv2.warpAffine with the same image decode.

Specifically:

~~~
image = cv2.warpAffine(
            image, mat_input, (self.input_size, self.input_size)
        )

self.wa = ops.WarpAffine(device="gpu",
                                size=(96,96))
~~~

Should they be the same?

The affine matrix and outputs of decoder is same as expected.

When I print the output after affine:

~~~
cv2 version:

[[187 189 178]
 [187 188 177]
 [187 187 177]
 [194 194 184]
 [203 203 193]
 [201 201 191]
 [185 182 173]
 [158 155 146]
 [177 173 164]
 [161 157 148]
 [180 176 167]
 [190 186 176]
 [200 191 182]
 [204 197 186]
 [178 172 160]
 [168 164 152]
 [156 152 140]
 [168 166 153]
 [182 182 172]
 [177 178 169]
 [169 170 165]
 [143 145 142]
 [160 161 162]
 [159 159 162]
 [149 147 150]
 [132 128 129]
 [108 104 105]
 [ 83  81  84]
 [ 69  68  71]
 [ 62  61  66]
 [ 65  66  71]
 [ 64  65  70]
 [ 62  63  68]
 [ 66  69  76]
 [ 91  94 101]
 [118 122 130]
 [167 177 189]
 [190 201 216]
 [209 219 232]
 [213 221 234]
 [214 222 232]
 [217 224 232]
 [188 191 200]
 [226 228 231]
 [245 246 247]
 [255 255 255]
 [253 253 253]
 [253 253 253]
 [255 255 253]
 [255 255 254]
 [254 252 253]
 [250 248 253]
 [247 245 250]
 [244 241 248]
 [241 238 245]
 [243 241 247]
 [248 246 250]
 [255 254 255]
 [252 251 251]
 [252 252 250]
 [255 255 255]
 [255 255 255]
 [255 255 255]
 [255 255 255]
 [255 255 255]
 [255 255 255]
 [255 255 255]
 [255 255 255]
 [255 255 255]
 [255 255 255]
 [255 255 255]
 [255 255 255]
 [250 250 250]
 [252 252 252]
 [248 248 248]
 [237 237 237]
 [244 244 244]
 [239 239 239]
 [243 243 243]
 [239 239 239]
[...]

PR #2693, opened 2021-02-18
Fix layout length check in Transpose.

mzient:

#### Why we need this PR?
*Pick one, remove the rest*
- It fixes a bug: transpose checks _input_ layout length instead of _output_ layout length when explicit layout is provided

Mentioned in #2648

 - What solution was applied:
     * Changed output_layout_ (not assigned yet) to output_layout_arg_ (what is to be assigned).
     * Transpose op with explicit layout provided
     * N/A
 - Validation and testing:
     * Python test added (it failed before the fix)
     * N/A

**JIRA TASK**: N/A


mzient:
!build

JanuszL (approved):

awolant (approved):

PR #2704, opened 2021-02-22
Update TensorFlow ReseNet50 example to work with the latest TF 2.4.x version

JanuszL:
- makes test runnable for TensorFlow test for TF 2.3.x for CUDA 10.x version and 2.4.x for CUDA 11.x

#### Why we need this PR?
*Pick one, remove the rest*
- It updates TensorFlow ReseNet50 example to work with the latest TF 2.4.x version

 - What solution was applied:
     updates TensorFlow ReseNet50 example to work with the latest TF 2.4.x version
     makes test runnable for TensorFlow test for TF 2.3.x for CUDA 10.x version and 2.4.x for CUDA 11.x
     TensorFlow ReseNet50 example
     TL1_tensorflow-dali_test
     NA
 - Validation and testing:
     current tests apply
     TensorFlow ReseNet50 example has been updated

**JIRA TASK**: *[DALI-1864]*

### DIFF
--- a/qa/TL3_RN50_convergence/test_tensorflow.sh
+++ b/qa/TL3_RN50_convergence/test_tensorflow.sh
@@ -29,7 +29,7 @@ mpiexec --allow-run-as-root --bind-to socket -np ${NUM_GPUS} \
     python -u resnet.py \
     --data_dir=$DATA_SET_DIR --data_idx_dir=idx-files/ \
     --precision=fp16 --num_iter=90 --iter_unit=epoch --display_every=50 \
-    --batch=192 --use_xla --log_dir=$OUT \
+    --batch=128 --use_xla --log_dir=$OUT \
     --dali_mode="GPU" 2>&1 | tee $LOG
 
 RET=${PIPESTATUS[0]}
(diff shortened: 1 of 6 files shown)


qa/TL3_RN50_convergence/test_tensorflow.sh
```diff
@@ -29,7 +29,7 @@ mpiexec --allow-run-as-root --bind-to socket -np ${NUM_GPUS} \
     python -u resnet.py \
     --data_dir=$DATA_SET_DIR --data_idx_dir=idx-files/ \
     --precision=fp16 --num_iter=90 --iter_unit=epoch --display_every=50 \
-    --batch=192 --use_xla --log_dir=$OUT \
+    --batch=128 --use_xla --log_dir=$OUT \
```
JanuszL, line 32: It looks that the newest TF need even more GPU memory.

awolant (approved):
Too bad we need to keep the TF 2.3.x support here, but nothing we can do.

JanuszL:
> Too bad we need to keep the TF 2.3.x support here, but nothing we can do.

2.3.x is the last supporting CUDA 10.x. I don't think we can do much about it.

banasraf (approved):

PR #2722, opened 2021-02-24
Move tfrecord reader to readers module

klecki:
#### Why we need this PR?
For completeness with readers module

 - What solution was applied:
     TFRecordReader is special. 
     Added placeholders/alias schemas in the same fashion as other ops (to facilitate grepping with `readers__`).
     Reworked the wrapper class to be internal only. Use it to create two proper wrapper classes inserted into appropriate modules and wrapped into corresponding fn api functions.

      TFRecord Reader and related tests/notebooks
     If the doc is rendered and the op is accessible?
 - Validation and testing:
      
     Some changes in docs.


klecki:
!build

dali/operators/reader/tfrecord_reader_op.cc
```diff
@@ -36,24 +47,22 @@ The index files can be obtained from TFRecord files by using the ``tfrecord2idx`
...
       DALI_TF_FEATURE_VEC)
-  .AddParent("_TFRecordReaderBase")
+  .AddParent("readers___TFRecordBase")
   .AddParent("LoaderBase")
   .MakeInternal();
 
-// Schema for the actual TFRecordReader op exposed
-// in Python. It is here for proper docstring generation
-DALI_SCHEMA(TFRecordReader)
+// Schema for the actual readers.tfrecord op expose in Python.
```
JanuszL, line 63:
```suggestion
// Schema for the actual readers.tfrecord op exposed in Python.
```

jantonguirao (approved):

JanuszL (approved):

dali/operators/reader/tfrecord_reader_op.cc
klecki: As this is a comment, and the CI appears to be slow, I will fix it in a separate one.

PR #2726, opened 2021-02-25
Fix constant input promotion for mixed backend.

mzient:

#### Why we need this PR?
*Pick one, remove the rest*
- It fixes a bug: attempt to create constants op with "mixed" backend when a constant is passed as a regular input to a mixed-backend operator.

 - What solution was applied:
     * Use default_input_device (which is "cpu" for mixed op) instead of op.device
     * Constant promotion / Python API
     * N/A
 - Validation and testing:
     * Python unit test in test_operator_constant
     * N/A

**JIRA TASK**: N/A


mzient:
!build

jantonguirao (approved):

PR #2739, opened 2021-03-01
Remove dynlink CUDA libs from the build image

JanuszL:
- as DALI doesn't link any CUDA libraries dynamically but just dlopen then there is no need to have dynlink stubs in the LIBRARY_PATH in the docker build image. Thanks to that builder image can be used to build and run DALI as well (during the development)

#### Why we need this PR?
*Pick one, remove the rest*
- It removes dynlink CUDA libs from the build image

 - What solution was applied:
     as DALI doesn't link any CUDA libraries dynamically but just dlopen then there is no need to have dynlink stubs in the LIBRARY_PATH in the docker build image. Thanks to that builder image can be used to build and run DALI as well (during the development)
     main Dockerfile
     NA
 - Validation and testing:
     just build in CI
     NA

**JIRA TASK**: *[ NA]*


JanuszL:
!build

mzient (approved):

jantonguirao (approved):

PR #2742, opened 2021-03-01
Update memory resource interfaces.

mzient:

#### Why we need this PR?
*Pick one, remove the rest*
- Refactoring to match proposed libcu++ memory resource interfaces.

 - What solution was applied:
     * Some renaming
     * Change allocation_order enum to Context type parameter
     * Memory manager, third_party
     * N/A
 - Validation and testing:
     * Existing tests apply
     * N/A

**JIRA TASK**: N/A

### DIFF
--- a/dali/core/mm/mm_test_utils.h
+++ b/dali/core/mm/mm_test_utils.h
@@ -163,6 +163,10 @@ class test_resource_wrapper<owning, security_check, memory_resource<kind, Contex
       return this->upstream_->deallocate(p, b, a);
     }, ptr, bytes, alignment);
   }
+
+  virtual Context do_get_context() const noexcept {
+    return this->upstream_->get_context();
+  }
 };
 
 
--- a/dali/core/mm/mm_test_utils.h
+++ b/dali/core/mm/mm_test_utils.h
@@ -136,16 +136,16 @@ class test_resource_wrapper_impl {
   }
 };
 
-template <memory_kind kind, allocation_order order, bool owning, bool security_check,
+template <memory_kind kind, typename Context, bool owning, bool security_check,
           typename Upstream>
-class test_resource_wrapper<owning, security_check, memory_resource<kind, order>, Upstream>
-: public memory_resource<kind, order>
+class test_resource_wrapper<owning, security_check, memory_resource<kind, Context>, Upstream>
+: public memory_resource<kind, Context>
 , public test_resource_wrapper_impl<owning, security_check, Upstream> {
   static_assert(!security_check || kind != memory_kind::device,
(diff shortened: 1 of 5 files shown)


mzient:
!build

JanuszL (approved):

dali/core/mm/mm_test_utils.h
```diff
@@ -163,6 +163,10 @@ class test_resource_wrapper<owning, security_check, memory_resource<kind, order>
       return this->upstream_->deallocate(p, b, a);
     }, ptr, bytes, alignment);
   }
+
+  virtual Context do_get_context() const noexcept {
```
mzient, line 167: This cannot be marked `override` because the base class for `any_context` does not contain a virtual function of this name (for any context this function is never used, but it's here regardless to avoid excessive complexity).

include/dali/core/mm/monotonic_resource.h
```diff
@@ -163,11 +163,15 @@ class monotonic_memory_resource<kind, order, true> : public memory_resource<kind
   void do_deallocate(void *data, size_t bytes, size_t alignment) override {
   }
 
+  virtual Context do_get_context() const noexcept {
```
mzient, line 166: Cannot mark override due to variable interface - see previous comment.

include/dali/core/mm/monotonic_resource.h
```diff
@@ -254,9 +258,13 @@ class monotonic_memory_resource<kind, order, false> : public memory_resource<kin
...
```
mzient, line 261: Cannot mark override due to variable interface - see previous comment.

include/dali/core/mm/pool_resource.h
```diff
@@ -127,14 +127,18 @@ class pool_resource_base : public memory_resource<kind, order> {
     }
   }
 
+  virtual Context do_get_context() const noexcept {
```
mzient, line 130: Cannot mark override due to variable interface - see previous comment.

dali/core/mm/mm_test_utils.h
klecki: Every time I see a comment in review that justifies how the code is written I think it should probably be put into the code itself.

JanuszL: Cannot agree more.

mzient: Explain - yes.
Justifies - not really. That would be just noise _and_ it's likely to go stale should this function make its way back to the interface (which I'm arguing for).

jantonguirao (approved):

PR #2750, opened 2021-03-02
Shutdown executor when an error occurs in the executor itself, not in one of operators.

mzient:

#### Why we need this PR?
*Pick one, remove the rest*
- It fixes a bug: exceptions in RunXXX (but not in the loop over operators) cause the executor to hang

 - What solution was applied:
     * Handle errors and shutdown the queue
     * Executor
     * N/A
 - Validation and testing:
     * Python test with a trigger for previously erroneous behavior
     * N/A

**JIRA TASK**: N/A

### DIFF
--- a/dali/pipeline/executor/executor.cc
+++ b/dali/pipeline/executor/executor.cc
@@ -40,7 +40,7 @@ void Executor<WorkspacePolicy, QueuePolicy>::PreRun() {
 }
 
 template <typename WorkspacePolicy, typename QueuePolicy>
-void Executor<WorkspacePolicy, QueuePolicy>::RunCPU() {
+void Executor<WorkspacePolicy, QueuePolicy>::RunCPUImpl() {
   PreRun();
 
   if (device_id_ < 0) {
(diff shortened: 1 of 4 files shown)


mzient:
!build

dali/pipeline/executor/executor.cc
```diff
@@ -243,6 +243,44 @@ void Executor<WorkspacePolicy, QueuePolicy>::RunGPU() {
   QueuePolicy::QueueOutputIdxs(gpu_idxs, gpu_op_stream_);
 }
 
+template <typename WorkspacePolicy, typename QueuePolicy>
+void Executor<WorkspacePolicy, QueuePolicy>::RunCPU() {
```
mzient, line 247: This is a bit repetitive; I tried some wrapper with a callable objects, but stack traces looked horrific.

dali/pipeline/executor/executor.cc
```diff
@@ -243,6 +243,44 @@ void Executor<WorkspacePolicy, QueuePolicy>::RunGPU() {
...
+  try {
+    RunCPUImpl();
+  } catch (std::exception &e) {
+    HandleError(make_string("Exception in CPU stage: ", e.what()));
+    throw;
```
mzient, line 252: ....should we rethrow? Perhaps only in synchronous executor?

klecki: No, we shouldn't. The whole point is executor collects the errors and we throw them from main thread, as otherwise it kills Python (or hangs).

mzient:
!build

jantonguirao (approved):

mzient:
!build

klecki (approved):

PR #2756, opened 2021-03-04
Move all deps into subrepos

JanuszL:
- by moving add dependencies to subrepositories hosted in DALI_deps it is easier to track as they are in one place, upgrade for all platforms and conduct CVE scans

#### Why we need this PR?
*Pick one, remove the rest*
- It move all deps into subrepos

 - What solution was applied:
     by moving add dependencies to DALI_deps subrepositories it is easier to track as they are in one place, upgrade for all platforms and conduct CVE scans
     Dockerfile.build.aarch64-linux
     Dockerfile.build.aarch64-qnx
     Dockerfile.deps
     build_deps.sh
     NA
 - Validation and testing:
     run in CI
     NA

**JIRA TASK**: *[NA]*


szalpal:
I'd have 2 questions about this:
1. Do we know, how does adding, e.g. FFmpeg, (LGPL content) impact the licensing for whole repo? Can we still re-license it on Apache2.0?
2. What's the rough number of files that will be added to recursively cloned repo? I mean, by how much files DALI repo will grow after this change?

.gitmodules
```diff
@@ -38,3 +38,45 @@
 [submodule "third_party/thrust"]
 	path = third_party/thrust
 	url = https://github.com/NVIDIA/thrust
+[submodule "third_party/libsndfile"]
```
jantonguirao, line 41: I am missing information about a tag for a particular release.

klecki:
> I'd have 2 questions about this:
> 
>     1. Do we know, how does adding, e.g. FFmpeg, (LGPL content) impact the licensing for whole repo? Can we still re-license it on Apache2.0?
> 

This is fine, as the submodule is only a link, the source code is not part of our repo. @JanuszL confirmed that.

.gitmodules
JanuszL: This is tracked by git. At the very bottom you have a list of relevant sha (git gibmodules don't track tags, just sha). You need to check tags on your own in each repo (git describe --tags in each subrepo).

klecki:
Stats:
* 327M for current master (340K in third_party, 202M in .git)
* 653M for current master + deps: (127M in third_party and 383M in .git)
* 2.3G for this PR: (589M in third_party, 1.6G in .git)

JanuszL:
> @szalpal
> 
> > I'd have 2 questions about this:
> > ```
> > 1. Do we know, how does adding, e.g. FFmpeg, (LGPL content) impact the licensing for whole repo? Can we still re-license it on Apache2.0?
> > ```
> 
> This is fine, as the submodule is only a link, the source code is not part of our repo. @JanuszL confirmed that.

Yes. I double checked with relevant people. Subrepo is not a part of repo, it is just a link.

JanuszL:
> Stats:
> 
> * 327M for current master (340K in third_party, 202M in .git)
> * 653M for current master + deps: (127M in third_party and 383M in .git)
> * 2.3G for this PR with deps: (589M in third_party, 1.6G in .git)

You can always - https://stackoverflow.com/questions/2144406/how-to-make-shallow-git-submodules, `git submodule update --depth 1 --recursive`.

JanuszL:
We can always create third_party_installable to keep it separated from the other.

klecki:
> We can always create third_party_installable to keep it separated from the other.

but I don't see a way to update/init submodules from particular subdirectory. I'm fine with this change, just posted the stats as someone @szalpal asked.

JanuszL:
> but I don't see a way to update/init submodules from particular subdirectory

You can sync dir only with a submodule, but this dir needs to be repo. My idea was to highlight the distinction between the deps we use during the build and one we just expect to be installed.

docker/Dockerfile.deps
```diff
@@ -24,48 +25,39 @@ RUN yum remove -y devtoolset* && \
...
 
 #zlib
-RUN LIBZ_VERSION=1.2.11 && \
-    cd /tmp && \
-    curl -L https://github.com/madler/zlib/archive/v${LIBZ_VERSION}.tar.gz | tar -xzf - && \
-    cd zlib-${LIBZ_VERSION} && \
-      CFLAGS="-fPIC" \
-      CXXFLAGS="-fPIC" \
+COPY third_party/zlib/ /tmp/zlib
+RUN cd /tmp/zlib && \
```
szalpal, line 29: `COPY` directives will invalidate the cache for **all** the subsequent layers, if the `<src>` part of the `COPY` changes. I believe, that every `git submodule update` will actually change the `<src>`, since the `.git` directory might change. That might impose the need of frequent rebuilding these changed Dockerfiles.

Maybe we could leverage multi-stage build to tackle this problem? Building every dependency in a separate stage, like:
```dockerfile
FROM ${FROM_IMAGE_NAME} as deps_builder
# ...

FROM deps_builder as dep_zlib
COPY third_part/zlib/ /tmp/zlib
RUN cd /tmp/zlib...

FROM deps_builder as dep_cmake
# etc... for all the dependencies

FROM ${FROM_IMAGE_NAME}
COPY --from=dep_zlib /usr/local/zlib /usr/local/zlib
COPY --from=dep_cmake ...
# etc...
```
This ensures, only the changed submodules will be rebuilt

JanuszL: > that every git submodule update will actually change the <src>, since the .git directory might change.

The submodule itself doesn't have a `.git.` folder. Submodules history is kept in the main `.git/modules/PATH_TO_THE_MODULE/` folder. So Docker copy should work as it used to when we did the source downloads inside the docker.

`COPY --from=dep_zlib /usr/local/zlib /usr/local/zlib` is a terrible idea because of many side effects when you install given lib we may not be aware of (it installs headers, libs, pkg files, and other things). So there are multiple paths to track for every lib and it would be easy to miss things when we update dependencies.

mzient:
I think that dependencies which are not built in the CMake build step should not be a part of this repository.
The general rule which I'd like to follow is:
- the main source repository and its submodules are part of CMake build
- the build environment (tools and libraries which we expect to be installed in the system or otherwise pointed to by hand) should not be a part of the main repository.

docker/Dockerfile.deps
```diff
@@ -21,119 +22,24 @@ RUN yum remove -y devtoolset* && \
...
+    cd /tmp/dali_deps                                            && \
+    git submodule init                                           && \
+    git checkout ${DALI_DEPS_VERSION}                            && \
+    git submodule update --depth 1 --recursive                   && \
+    export CC_COMP=${CC}                                         && \
+    export CXX_COMP=${CXX}                                       && \
+    export CMAKE_TARGET_ARCH="$(uname -m)"                       && \
+    export OPENCV_TOOLCHAIN_FILE="linux/gnu.toolchain.cmake"     && \
+    export WITH_FFMPEG=1                                         && \
+    ../build_deps.sh && rm -rf /tmp/dali_deps && rm -rf /tmp/build_deps.sh && rm -rf /tmp/DALI_DEPS_VERSION
```
jantonguirao, line 42:
```suggestion
    /tmp/build_deps.sh && rm -rf /tmp/dali_deps && rm -rf /tmp/build_deps.sh && rm -rf /tmp/DALI_DEPS_VERSION
```
?

third_party/build_deps.sh
```diff
@@ -0,0 +1,271 @@
+#!/bin/bash -xe
+
+ROOT_DIR=$(pwd)
+export INSTALL_PREFIX=${INSTALL_PREFIX:-/usr/local}
```
jantonguirao, line 4: I have general comments about this file:

1. Shouldn't this belong to DALI_deps?
2. You already have -e in the options. You don't need to chain every line with && since the script will be aborted on first error already
3. I'd probably create a single script for every dependency and this script will invoke each of them (probably all of that could live in DALI_deps)

I am thinking about the scenario that a dependency update changes the way to install it. I believe it'd make more sense if the installation script is also part of the DALI_DEPS version

klecki: Bit of agree, we already moved patches there, which in fact are code written by us, on the other hand that repo is supposed to hold only references to deps.

third_party/build_deps.sh
```diff
@@ -0,0 +1,271 @@
...
+echo ${HOST_ARCH_OPTION}
+echo ${SYSROOT_ARG}
+echo ${WITH_FFMPEG}
+echo ${EXTRA_PROTOBUF_FLAGS}
+echo ${OPENCV_TOOLCHAIN_FILE}
+echo ${EXTRA_FLAC_FLAGS}
+echo ${EXTRA_LIBSND_FLAGS}
+
+#zlib
+pushd third_party/zlib                                                            && \
```
klecki, line 21: I agree that excessive && are, well, excessive.
Or the problem is with setting the variables for the following commands?

third_party/build_deps.sh
JanuszL: > Shouldn't this belong to DALI_deps?

I need to sort out the licensing internally first.

> You already have -e in the options. You don't need to chain every line with && since the script will be aborted on first error already

Done

> I'd probably create a single script for every dependency and this script will invoke each of them (probably all of that could live in DALI_deps)

Won't it introduce too much fragmentation? I would check if we can move the main script to the deps repo first.

third_party/build_deps.sh
JanuszL: Done

third_party/build_deps.sh
```diff
@@ -0,0 +1,271 @@
...
+popd
+
+# protobuf, make two steps for cross compilation
+pushd third_party/protobuf                                                        && \
+./autogen.sh                                                                      && \
+./configure CXXFLAGS="-fPIC" --prefix=/usr/local --disable-shared 2>&1 > /dev/null && \
+make -j"$(grep ^processor /proc/cpuinfo | wc -l)" 2>&1 > /dev/null                && \
+make install 2>&1 > /dev/null                                                     && \
+make clean                                                                        && \
+    ./autogen.sh && ./configure \
```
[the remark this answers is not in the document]
JanuszL: Done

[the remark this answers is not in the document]
JanuszL: Done

third_party/build_deps.sh
```diff
@@ -0,0 +1,283 @@
...
+# protobuf, make two steps for cross compilation
+pushd third_party/protobuf
+./autogen.sh
+./configure CXXFLAGS="-fPIC" --prefix=/usr/local --disable-shared 2>&1 > /dev/null
+make -j"$(grep ^processor /proc/cpuinfo | wc -l)" 2>&1 > /dev/null
+make install 2>&1 > /dev/null
+# only when cross compiling
+if [ "${CC_COMP}" != "gcc" ]; then
+  make clean
+      ./autogen.sh
```
jantonguirao, line 47: this line should be indented at the same level as `make clean`

third_party/build_deps.sh
```diff
@@ -0,0 +1,283 @@
...
+
+# libjpeg-turbo
+pushd third_party/libjpeg-turbo/
+echo "set(CMAKE_SYSTEM_NAME Linux)" > toolchain.cmake
+echo "set(CMAKE_SYSTEM_PROCESSOR ${CMAKE_TARGET_ARCH})" >> toolchain.cmake
+echo "set(CMAKE_C_COMPILER ${CC_COMP})" >> toolchain.cmake
+  CFLAGS="-fPIC" \
+  CXXFLAGS="-fPIC" \
+  CC=${CC_COMP} \
+  CXX=${CXX_COMP} \
```
jantonguirao, line 77:
```suggestion
  CFLAGS="-fPIC" \
  CXXFLAGS="-fPIC" \
  CC=${CC_COMP} \
  CXX=${CXX_COMP} \
```
```suggestion
CFLAGS="-fPIC" \
CXXFLAGS="-fPIC" \
CC=${CC_COMP} \
CXX=${CXX_COMP} \
```
nitpick: weird indent

third_party/build_deps.sh
```diff
@@ -0,0 +1,283 @@
...
+cmake -G"Unix Makefiles" -DCMAKE_TOOLCHAIN_FILE=toolchain.cmake -DENABLE_SHARED=TRUE -DCMAKE_INSTALL_PREFIX=${INSTALL_PREFIX} . 2>&1 >/dev/null
+  CFLAGS="-fPIC" \
+  CXXFLAGS="-fPIC" \
+  CC=${CC_COMP} \
+  CXX=${CXX_COMP} \
```
jantonguirao, line 82:
```suggestion
  CFLAGS="-fPIC" \
  CXXFLAGS="-fPIC" \
  CC=${CC_COMP} \
  CXX=${CXX_COMP} \
```
```suggestion
CFLAGS="-fPIC" \
CXXFLAGS="-fPIC" \
CC=${CC_COMP} \
CXX=${CXX_COMP} \
```
nitpick: weird indent

third_party/build_deps.sh
```diff
@@ -0,0 +1,283 @@
...
+pushd third_party/openjpeg
+mkdir build
+cd build
+echo "set(CMAKE_SYSTEM_NAME Linux)" > toolchain.cmake
+echo "set(CMAKE_SYSTEM_PROCESSOR ${CMAKE_TARGET_ARCH})" >> toolchain.cmake
+echo "set(CMAKE_C_COMPILER ${CC_COMP})" >> toolchain.cmake
+  CFLAGS="-fPIC" \
+  CXXFLAGS="-fPIC" \
+  CC=${CC_COMP} \
+  CXX=${CXX_COMP} \
```
jantonguirao, line 121:
```suggestion
  CFLAGS="-fPIC" \
  CXXFLAGS="-fPIC" \
  CC=${CC_COMP} \
  CXX=${CXX_COMP} \
```
```suggestion
CFLAGS="-fPIC" \
CXXFLAGS="-fPIC" \
CC=${CC_COMP} \
CXX=${CXX_COMP} \
```

third_party/build_deps.sh
```diff
@@ -0,0 +1,283 @@
...
+./configure \
+  CFLAGS="-fPIC ${EXTRA_LIBSND_FLAGS}" \
+  CXXFLAGS="-fPIC ${EXTRA_LIBSND_FLAGS}" \
+  CC=${CC_COMP} \
+  CXX=${CXX_COMP} \
+  ${HOST_ARCH_OPTION} \
+  --prefix=${INSTALL_PREFIX}
+make -j"$(grep ^processor /proc/cpuinfo | wc -l)"
+make install
+popd
```
jantonguirao, line 283: nitpick: empty line missing

jantonguirao (approved):

third_party/build_deps.sh
JanuszL: The is will be indistinctable from other lines, the point here is to show that it a part of the following command.

jantonguirao (approved):

awolant (approved):

PR #2765, opened 2021-03-06
Fix paddle ssd

awolant:
#### Why we need this PR?

Paddle SSD test failes when there are more GPUs in the test environment. This might casue L1 to fail, when runner with more than one GPUs is selected to run this job.

 - What solution was applied:
     *Make Paddle see only one GPU during the test, regardless on the actual runner, by using `CUDA_VISIBLE_DEVICES` in this test*
     *Paddle SSD example*
 - Validation and testing:
     *CI L1 tests*

**JIRA TASK**: *[NA]*


awolant:
!build

jantonguirao (approved):

JanuszL (approved):

PR #2773, opened 2021-03-09
Fix a race when loader starts reading even the metadata is not ready yet

JanuszL:
- PrepareMetadata sets that the reader is ready to read first before the metadata is prepared. In effect when the user queries loader for any meta, like size, it will start prefetching even the metadata itself is not ready yet. Fix the race by preparing the meta first and then setting the ready variable
- fixes a deadlock caused by a sequence of calls PrepareMetadata->PrepareMetadataImpl->Reset->Size->PrepareMetadata, by replacing the call to Size by SizeImpl which provides the same functionality with the used set of arguments, but it can be done only when the loader has the necessary data fields initialized

#### Why we need this PR?
*Pick one, remove the rest*
- It fixes a race when loader starts reading even the metadata is not ready yet
- It fixes a deadlock caused by a sequence of calls PrepareMetadata->PrepareMetadataImpl->Reset->Size->PrepareMetadata

 - What solution was applied:
     Fix the race by preparing the meta first and then setting the ready variable
     Fixes a deadlock caused by a sequence of calls PrepareMetadata->PrepareMetadataImpl->Reset->Size->PrepareMetadata
     loader
     loader subclasses
     check if there is no more inversed order leading to a race
 - Validation and testing:
     current tests applies
     NA

**JIRA TASK**: *[DALI-1909]*

### DIFF
--- a/dali/operators/reader/loader/loader.h
+++ b/dali/operators/reader/loader/loader.h
@@ -217,8 +217,8 @@ class Loader {
   void PrepareMetadata() {
     std::lock_guard<std::mutex> l(prepare_metadata_mutex_);
     if (!loading_flag_) {
-      loading_flag_ = true;
       PrepareMetadataImpl();
+      loading_flag_ = true;
     }
   }
 
--- a/dali/operators/reader/loader/loader.h
+++ b/dali/operators/reader/loader/loader.h
@@ -25,6 +25,7 @@
 #include <utility>
 #include <vector>
 #include <deque>
+#include <atomic>
 
 #include "dali/core/nvtx.h"
 #include "dali/core/common.h"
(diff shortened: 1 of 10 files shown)


mzient (approved):

dali/operators/reader/loader/loader.h
```diff
@@ -217,8 +217,8 @@ class Loader {
   void PrepareMetadata() {
     std::lock_guard<std::mutex> l(prepare_metadata_mutex_);
```
mzient, line 218: Perhaps instead of checking this flag outside, it should be checked here and the `if (!loading_flag)` removed from all usages.
```C++
if (loading_flag_)
  return;
```

dali/operators/reader/loader/loader.h
```diff
@@ -217,8 +217,8 @@ class Loader {
...
     if (!loading_flag_) {
-      loading_flag_ = true;
       PrepareMetadataImpl();
+      loading_flag_ = true;
```
mzient, line 221: If we're paranoid, we could insert a memory fence before setting this flag, so we're 100% sure that other threads will see all the changes before this flag is set.

dali/operators/reader/loader/loader.h
JanuszL: Done

dali/operators/reader/loader/loader.h
JanuszL: Done

mzient (approved):
Looks like it could work.

PR #2774, opened 2021-03-09
Require uniform StopIteration for parallel External Source

klecki:
Fix potential mismatch between worker processes, when
StopIteration is raised for only part of the calculated
batch.

#### Why we need this PR?
- Fix a mismatch in parallel External Source test and add an error when it happens. 

 - What solution was applied:
     StopIteration is treated under the hood as different entity and included in calculation of processed tasks.
     Batch is completed when all tasks from workers are returned back (either processed of signalling Stop).
     If there are StopIterations, they must come from all workers at the same time (cover all samples).
     Raise an error if there is a mismatch, adjust tests' epoch sizes.
     Adjust docs
     Parallel External Source and docs.
     
 - Validation and testing:
     CI.
     Warning in docs added.

**JIRA TASK**: *[DALI-1910]*

### DIFF
--- a/dali/python/nvidia/dali/_multiproc/pool.py
+++ b/dali/python/nvidia/dali/_multiproc/pool.py
@@ -107,46 +107,63 @@ received so far.
     def reset(self):
         self.partially_received = {}
         self.scheduled = OrderedDict()
+        # Mapping from batch_i to (excpetion, traceback_str) for failed batches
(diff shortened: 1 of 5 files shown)


JanuszL (approved):

klecki:
#2777 is more complete solution for the problem.

PR #2777, opened 2021-03-10
Fix iteration number tracking with pipeline.reset

mzient:
#### Why we need this PR?
*Pick one, remove the rest*
- It fixes bugs:
   - reset schedules additional processing
   - data race when Reset is called
   - batch index is ambiguous for small batches

 - What solution was applied:
     * Batches are scheduled to be placed in a specific queue entry
     * Entries are bound to a particular global iteration number (not withing epoch and not a job subscription number)
     * Jobs are scheduled with a unique ID instead of batch index
     * External source
     * N/A
 - Validation and testing:
     * Existing tests apply, added a test with large data
     * N/A

**JIRA TASK**: DALI-1910

Co-authored-by: Krzysztof Lecki <klecki@nvidia.com>


dali/python/nvidia/dali/external_source.py
```diff
@@ -97,32 +100,47 @@ def __init__(
...
         return (arg,)
 
     def reset_indices(self):
         self.current_iter = 0
         self.current_sample = 0
+        self.cancel_prefetch()
 
-    def schedule_batch(self, pool, context_i, batch_size):
+    def cancel_prefetch(self):
+        self.scheduled_ahead = 0
```
JanuszL, line 128: Does that cancel prefetch (I don't see how `self.scheduled_ahead = 0` does that)?

JanuszL (approved):

dali/test/python/test_external_source_parallel_utils.py
```diff
@@ -142,6 +142,7 @@ def check_stop_iteration_resume(pipe, batch_size, layout):
         try:
             while True:
                 (r, ) = pipe.run()
+                r = [r.at(i) for i in range(len(r))]
                 output.append(r)
```
klecki, line 146: I think we should do a copy here, for example:
```suggestion
                output.append(np.copy(r.as_array()))
```

mzient:
!build

klecki (approved):

dali/python/nvidia/dali/external_source.py
mzient: This is sort of spread between pool.reset and this place. Pool.reset marks the outputs as ignored, whereas this makes it possible to reschedule the batches.

JanuszL: What if someone calls `cancel_prefetch` while `schedule_and_feed` is running? You can get `self.scheduled_ahead==-1`

mzient:
!build

PR #2784, opened 2021-03-11
ExternalSource - discard data from all callbacks when one raises StopIteration

mzient:
#### Why we need this PR?
*Pick one, remove the rest*
- It fixes bugs: putting batches that should be discarded in ExternalSourceQueue
- Bug results in:
    * phase shift
    * potentially wrong data in input buffers
    * resource leak
- Also fixed: copying 0D data from parallel external source workers

 - What solution was applied:
     * Gather results from all callbacks and feed them after we've collected all
     * Python modules: ExternalSource, Pipleine
     * N/A
 - Validation and testing:
     * Python test + existing tests for regressions
     * N/A

**JIRA TASK**: DALI-1915

### DIFF
--- a/dali/python/nvidia/dali/pipeline.py
+++ b/dali/python/nvidia/dali/pipeline.py
@@ -1154,19 +1154,24 @@ Parameters
         if self._input_callbacks is None:
             return
 
+        batches = []   # data from external source callbacks is gathered here
         stop_iter = False
         for i, group in enumerate(self._parallel_input_callbacks):
             try:
-                group.schedule_and_feed(self, self._py_pool, i, self._max_batch_size)
+                batches.append(group.schedule_and_receive(self, self._py_pool, i, self._max_batch_size))
             except StopIteration:
                 stop_iter = True
         for group in self._seq_input_callbacks:
             try:
-                group.call_and_feed(self, self._max_batch_size)
+                batches.append(group.get_batch(self, self._max_batch_size))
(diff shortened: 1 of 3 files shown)


dali/python/nvidia/dali/pipeline.py
```diff
@@ -1154,20 +1154,25 @@ def _run_input_callbacks(self):
...
         for group in self._seq_input_callbacks:
             try:
-                group.call_and_feed(self, self._max_batch_size)
+                batches.append(group.get_batch(self, self._max_batch_size))
             except StopIteration:
                 stop_iter = True
         if stop_iter:
             raise StopIteration()
 
+        # we only fill external source queues when we know that all callbacks succeeded
```
JanuszL, line 1172: What happens in case of failure? Are we loosing the data from all but failed callbacks?

JanuszL (approved):

dali/python/nvidia/dali/pipeline.py
mzient: The whole iteration is discarded - that's the point of this PR.

JanuszL: Maybe we should add this to our docs. Imagine that multiple ESes consume data from the network, I wonder if the user wants to lose data from them if any raises StopIteration. Maybe when we restart we should not run ES that has data stored already?

PR #2812, opened 2021-03-23
Fix coverity issues

awolant:

#### Why we need this PR?
*Pick one, remove the rest*

 - What solution was applied:
 - Validation and testing:


dali/kernels/imgproc/jpeg/jpeg_distortion_gpu_test.cu
```diff
@@ -37,7 +37,7 @@ class JpegDistortionTestGPU : public ::testing::Test {
   static constexpr bool horz_subsample = GTestParams::horz_subsample;
 
  public:
-  JpegDistortionTestGPU() {
+  JpegDistortionTestGPU() : input_device_(nullptr), output_(nullptr) {
```
mzient, line 40: As a rule, it's preferable to initialize members in declaration instead of constructor - this way every new constructor doesn't have to repeat the initialization.
Also, this file is under active development by @jantonguirao, so changing it here is a disfavor.

awolant: This is [WIP] to test Coverity and see what it reports and how. Please wait with the reviews.

dali/core/mm/mm_test_utils.h
```diff
@@ -198,16 +198,17 @@ class test_resource_wrapper<owning, security_check, stream_aware_memory_resource
     }, ptr, bytes, alignment);
   }
 
-  void *do_allocate_async(size_t bytes, size_t alignment, stream_view sv) override {
+  void *do_allocate_async(size_t bytes, size_t alignment, stream_view strm_vw) override {
```
mzient, line 201: Es ist kein Stormvolkswagen ;)
Perhaps just
```suggestion
  void *do_allocate_async(size_t bytes, size_t alignment, stream_view stream) override {
```

dali/kernels/imgproc/structure/label_bbox.h
```diff
@@ -224,7 +224,7 @@ void GetLabelBoundingBoxes(span<Box<ndim, Coord>> boxes,
   int max_d = -1;
   int64_t max_result = 0;
   for (int d = 0; d < simplified_ndim; d++) {
-    int mul = d == simplified_ndim - 1 ? 1 : 2;  // it's less efficient to slice last dim
+    int mul = (d == (simplified_ndim - 1) ? 1 : 2);  // it's less efficient to slice last dim
```
mzient, line 227: Is that supposed to be easier to read?!?!?!?!?

dali/kernels/signal/window/extract_windows_gpu_test.cu
```diff
@@ -34,7 +34,7 @@ TEST(ExtractWindowsGPU, NonBatchedKernel) {
   int step = 10;
   int length = windows * step - 100;;
   int center = 5;
-  bool reflect = false;
+  const bool reflect = false;
```
mzient, line 37: Actually, this makes me wonder why this is fixed at `false` - I'd give it a short at `true` and if it happens to work, have a loop:
```suggestion
  for (bool reflect : { false, true }) {
```

include/dali/core/geom/box.h
```diff
@@ -158,8 +158,8 @@ operator!=(const Box<ndims, CoordinateType> &lhs, const Box<ndims, CoordinateTyp
 template <int ndims, typename CoordinateType>
 std::ostream &operator<<(std::ostream &os, const Box<ndims, CoordinateType> &box) {
   auto print_corner = [&os](const typename Box<ndims, CoordinateType>::corner_t &c) {
-      for (size_t i = 0; i < ndims; i++)
-        os << (i ? ", " : "(") << c[i];
+      for (int i = 0; i < ndims; i++)
+        os << (i == 0 ? ", " : "(") << c[i];
```
mzient, line 162: Now it's broken - the condition is reversed.

dali/kernels/signal/window/extract_windows_gpu_test.cu
awolant: Done

include/dali/core/geom/box.h
awolant: Fixed. I meant `i != 0`.

dali/kernels/imgproc/structure/label_bbox.h
awolant: For me adding additional parenthesis to  ternary operator was always more readable. Since this is matter of preference, I'm leaving it as it is.

dali/core/mm/mm_test_utils.h
awolant: It follows general convention and indicates type, so I'm leaving it as it is.

mzient (approved):
I'd really prefer to get rid of the Hungarian Volkswagen knock-off. Otherwise OK.

JanuszL (approved):

PR #2814, opened 2021-03-23
Asynchronous pool memory resource

mzient:
#### Why we need this PR?
*Pick one, remove the rest*
- It adds new feature: Async pool required for stream-aware pinned memory pool

 - What solution was applied:
     * Pending free list + non-coalescing free tree for managing pending per-stream deallocations
     * Use existing pool for non-per-stream pool
     * MM
     * N/A
 - Validation and testing:
     * Unit tests
     * Doxygen

**JIRA TASK**: DALI-1755


mzient:
!build

mzient:
!build

mzient:
!build

include/dali/core/mm/async_pool.h
```diff
@@ -0,0 +1,428 @@
...
+   * @brief Waits until all pending frees are finished.
+   */
+  void synchronize() {
+    {
+      std::lock_guard<LockType> guard(lock_);
+      for (auto &kv : stream_free_) {
+        if (!kv.second.free_list.head)
+          continue;
+        if (!sync_stream_)
+          sync_stream_ = CUDAStream::Create(true);
```
JanuszL, line 56: Why not create it in the constructor?

include/dali/core/mm/async_pool.h
```diff
@@ -0,0 +1,428 @@
...
+        CUDA_DTOR_CALL(cudaStreamWaitEvent(sync_stream_, kv.second.free_list.head->event, 0));
+      }
+    }
+    CUDA_DTOR_CALL(cudaStreamSynchronize(sync_stream_));
```
JanuszL, line 60: What other work can be scheduled in `sync_stream_` we can wait here for?

include/dali/core/mm/async_pool.h
```diff
@@ -0,0 +1,428 @@
...
+
+  /**
+   * @brief Restores original block parameters (address, size and alignment) of a block
+   *        that was used for allocating a smaller block, possibly with stricter alignment.
+   */
+  void restore_original_params(char *&p, size_t &bytes, size_t &alignment) {
+    auto it = padded_.find(p);
+    if (it != padded_.end()) {
+      if (it->second.front_padding + bytes > it->second.bytes) {
+        throw std::invalid_argument("The deallocated memory points to a block that's smaller than "
```
JanuszL, line 327: This error message clearly states that this function can be used only for deallocation, while the name of it is very generic.
I would rename the function somehow.

include/dali/core/mm/async_pool.h
```diff
@@ -0,0 +1,428 @@
...
+   * There are four cases:
+   * (1) Attempt to allocate from stream-specific free list - a smallest suitable block is used,
+   *     padding is applied, if necessary, and recorded.
+   * (2) If (1) fails and there are no pending per-stream free operations, the requested memory
+   *     is allocated from the global pool.
+   * (3) If (1) fails and there are free pending per-stream operations, an attempt is made to get
+   *     memory from the free list of the underlying global pool, without using upstream resource.
+   *     If this allocation succeeds, the memory allocated this way is returned.
+   * (4) If (3) fails, the pending per-stream frees are collected to the global pool and then
+   *     an allocation is made from the global pool.
```
JanuszL, line 90: Do I map the cases correctly:
1) it != stream_free_.end() && ptr
2) it == stream_free_.end()
3) it != stream_free_.end() && !ptr
4) ?

JanuszL: Or the description includes `allocate_from_global_pool`?
If so I would clarify it a bit.

include/dali/core/mm/async_pool.h
```diff
@@ -0,0 +1,428 @@
...
+   * If the alignment requirements are low compared to the size, the alignment requirement is
+   * increased, which improves reusability of the block. Further, the block size is aligned
+   * to a next multiple of the (new) alignment, reducing the need for block splitting in case
+   * of multiple allocations of similarly-sized blocks.
+   */
+  static void adjust_size_and_alignment(size_t &size, size_t &alignment) {
+    if (size == 0)
+      return;
+    int l = ilog2(size);
+    size_t min_align = (1 << (l >> 1));  // 2^(l/2)
```
JanuszL, line 152: `l` looks very alike to `1`, can you rename it to increase readibility?

include/dali/core/mm/async_pool.h
```diff
@@ -0,0 +1,428 @@
...
+    auto *prev = f->prev;
+    auto *next = f->next;
+    if (free.head == f)
+      free.head = next;
+    if (free.tail == f)
+      free.tail = prev;
+    if (prev) prev->next = next;
+    if (next) next->prev = prev;
+    *f = {};
+#if 0  // check list consistency
```
JanuszL, line 374: I think we can enable it. This is for debug builds anyway, right?

include/dali/core/mm/async_pool.h
mzient: If there's nothing to clean, we shouldn't even create this stream.

include/dali/core/mm/async_pool.h
mzient: None. Actually, it looks like a bug - there should be an "if" there to see that the stream is indeed initialized.

include/dali/core/mm/async_pool.h
mzient: Yes, it includes `allocate_from_global_pool` - it was factored out after this comment had been written.

include/dali/core/mm/async_pool.h
mzient: sure

include/dali/core/mm/async_pool.h
mzient: There are more problems with this function - the biggest being that it also removes the pointer from `padded_`. I have no idea what better name could there be without getting ridiculously long. The real meaning is:
`restore_and_erase_original_allocation_params_before_deallocation` which is ludicrous.

include/dali/core/mm/async_pool.h
mzient: No, it's waaay too slow even for debug.
It was useful when I moved stuff around and had some problems with the list, but I do hope noone will ever need to re-enable it - but it's there so that when we see something fail we can reenable it quickly.

include/dali/core/mm/async_pool.h
mzient: Also, this should probably be just an assert.

mzient:
!build

mzient:
!build

mzient:
!build

include/dali/core/mm/async_pool.h
JanuszL: What is the cost of the creating stream?
What is the cost of having one created ahead?
Maybe we can create it to have the `synchronize()` execution time more predictable?

include/dali/core/mm/async_pool.h
JanuszL: Can you update it then?

include/dali/core/mm/async_pool.h
JanuszL: How about `#ifdef DEBUG_LIST_CONSISTENCY` and add on top of the file `//#define DEBUG_LIST_CONSISTENCY` with an appropriate comment?

include/dali/core/mm/async_pool.h
```diff
@@ -0,0 +1,431 @@
...
+    PendingFreeList free_list;
+  };
+
+  /**
+   * @brief Try to allocate a block from per-stream free memory
+   *
+   * If the allocation fails, the function returns nullptr.
+   */
+  void *try_allocate(PerStreamFreeBlocks &from, size_t bytes, size_t alignment) {
+    unsigned max_padding = std::min<size_t>(std::max<size_t>(bytes / 16, 16), (1 << 20));
```
JanuszL, line 220: How these values have been selected? Runtime experiments?

include/dali/core/mm/async_pool.h
```diff
@@ -0,0 +1,431 @@
...
+      char *aligned = detail::align_ptr(base, alignment);
+      size_t front_padding = aligned - base;
+      assert(static_cast<ptrdiff_t>(front_padding) >= 0);
+      // NOTE: block_size - front_padding >= size  can overflow and fail - meh, unsigned size_t
+      if (block_size >= bytes + front_padding) {
+        if (!supports_splitting && block_size - bytes - front_padding > max_padding)
+          return nullptr;  // no point in continuing - the map is sorted
+
+        from.by_size.erase(it);
+        char *block_end = base + block_size;
```
JanuszL, line 236:
```suggestion
        char *block_end = aligned + block_size;
```
?

include/dali/core/mm/async_pool.h
```diff
@@ -0,0 +1,431 @@
...
+        char *remainder = detail::align_ptr(aligned + bytes, remainder_alignment);
+        bool split = supports_splitting && remainder + min_split_remainder < block_end;
```
JanuszL, line 238:
```suggestion
        bool split = supports_splitting && min_split_remainder  < block_end - remainder ;
```
It is the same but is is cleared that the remaining part should be greater than `min_split_remainder`.

include/dali/core/mm/async_pool.h
```diff
@@ -0,0 +1,431 @@
...
+    }
+    return nullptr;
+  }
+
+  /**
+   * @brief Searches per-stream free blocks to find the most recently freed one.
+   */
+  pending_free *find_first_ready(PerStreamFreeBlocks &free) {
+    SmallVector<pending_free *, 128> pending;
+    int step = 1;
```
JanuszL, line 269: Do you plan to adjust this value? If not then the below loop can be simplified:
```
for (int i = 0; i < step; i++) {
        f = f->next;
        if (!f)
          break;
        pending.push_back(f);
      }
```

include/dali/core/mm/async_pool.h
```diff
@@ -0,0 +1,431 @@
...
+        f = f->next;
+        if (!f)
+          break;
+        pending.push_back(f);
+      }
+    }
+    if (pending.empty()) {
+      if (f) {
+        assert(f->ready());
+        assert(!f->prev || !f->ready());
```
JanuszL, line 285:
```suggestion
        assert(!f->prev);
```
I think the preceding assert makes check for `!f->ready()` redundant.

include/dali/core/mm/async_pool.h
```diff
@@ -0,0 +1,431 @@
...
+      }
+      return f;
+    }
+    auto it = std::partition_point(pending.begin(), pending.end(), [&](pending_free *f) {
+      return !f->ready();
+    });
```
JanuszL, line 291: With `int step = 1;` it doesn't do much, right?

include/dali/core/mm/async_pool.h
```diff
@@ -0,0 +1,431 @@
...
+  void *allocate_from_global_pool(size_t bytes, size_t alignment) {
+    void *ptr;
+    if (num_pending_frees_ == 0) {
+      ptr = global_pool_.allocate(bytes, alignment);
+      return ptr;
+    }
+    ptr = global_pool_.try_allocate_from_free(bytes, alignment);
+    if (ptr)
+      return ptr;
+    for (auto &kv : stream_free_)
```
JanuszL, line 126: Maybe there should be 5), where you synchronize to make sure that all `PerStreamFreeBlocks` are ready and then allocate try again global_pool_?

dali/core/mm/async_pool_test.cu
```diff
@@ -0,0 +1,325 @@
...
+  failure_buf.from_host(&failures, 1, stream);
+  GPUHog hog;
+  if (use_hog)
+    hog.init();
+  int hogs = 0;
+  int max_hogs = sync_dist(rng);
+  CUDAEvent event = CUDAEvent::Create();
+  for (int i = 0; i < max_iters; i++) {
+    if (use_hog && hog_dist(rng)) {
+      if (hogs++ > max_hogs) {
```
JanuszL, line 108: So the bigger the `hogs` value is the less likely the code here is happen to be executed, right?

dali/core/mm/async_pool_test.cu
```diff
@@ -0,0 +1,325 @@
...
+      {
+        std::lock_guard<Mutex> guard(mtx);
+        blocks.push_back({ ptr, size, fill, stream });
+      }
+    } else {
+      block blk;
+      {
+        std::lock_guard<Mutex> guard(mtx);
+        if (blocks.empty())
+          continue;
```
JanuszL, line 131: According to L114 it can't happen, right?

dali/core/mm/async_pool_test.cu
```diff
@@ -0,0 +1,325 @@
...
+    for (auto &t : threads)
+      t.join();
+  }
+  std::cerr << "Peak consumption:     " << upstream.get_peak_size() << " bytes\n";
+  std::cerr << "Upstream allocations: " << upstream.get_num_allocs() << std::endl;
+  upstream.check_leaks();
+}
+
+
+TEST(MMAsyncPool, TwoStream) {
```
JanuszL, line 229: I would move this to the end as it doesn't use `AsyncPoolTest` while one before and after use it.

dali/core/mm/async_pool_test.cu
```diff
@@ -0,0 +1,325 @@
...
+    for (auto &t : threads)
+      t.join();
+  }
+  std::cerr << "Peak consumption:     " << upstream.get_peak_size() << " bytes\n";
+  std::cerr << "Upstream allocations: " << upstream.get_num_allocs() << std::endl;
+  upstream.check_leaks();
+}
+
+
+TEST(MMAsyncPool, CrossStream) {
```
JanuszL, line 295: Does it make any sense to add `CrossStream` test with Hogs ?

dali/core/mm/async_pool_test.cu
mzient: Actually, it can - line 114 is outside mutex. It kept failing in a test against RMM to the point that I filed a bug for RMM (by luck it didn't happen with my resource), but then I caught it in the debugger :)

dali/core/mm/async_pool_test.cu
mzient: I could.

include/dali/core/mm/async_pool.h
mzient: It would bind the constructor to a particular device that was set as current during construction. Having that said, perhaps I should store the current device at 1st allocation and restore it in the destructor with some `deviceguard`.

include/dali/core/mm/async_pool.h
mzient: It would make sense if `upstream` is known to be slow (cudaMalloc, cudaHostAlloc). If `upstream` is not that slow (i.e. another pool/slab/) the performance would suffer. The idea itself is nice, but I'd make it configurable.

JanuszL: It is just an idea.

include/dali/core/mm/async_pool.h
mzient: Added DeviceGuard.

cmake/CUDA_utils.cmake
```diff
@@ -174,3 +174,22 @@ function(CUDA_find_library_stub out_path lib_name)
...
+  endif()
+  list(REMOVE_ITEM ${include_dirs} "${CMAKE_CUDA_TOOLKIT_INCLUDE_DIRECTORIES}" "${CUDA_TOOLKIT_PATH_VAR}/include")
+  set(${include_dirs} ${${include_dirs}} PARENT_SCOPE)
+endfunction()
+
+function(CUDA_move_toolkit_include_dirs_to_end)
+  get_property(tmp_include_dirs DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR} PROPERTY INCLUDE_DIRECTORIES)
+  CUDA_remove_toolkit_include_dirs(tmp_include_dirs)
+  set_property(DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR} PROPERTY INCLUDE_DIRECTORIES ${tmp_include_dirs})
+  include_directories(SYSTEM "${CMAKE_CUDA_TOOLKIT_INCLUDE_DIRECTORIES}")
```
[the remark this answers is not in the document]
mzient: Yes. I could detect that something was removed but I don't think that's necessary.

dali/core/mm/async_pool_test.cu
mzient: max_hogs are max scheduled hogs - when we've scheduled max_hogs, a synchronization is done to avoid scheduling too much work ahead.

dali/core/mm/async_pool_test.cu
mzient: Done.

include/dali/core/mm/async_pool.h
mzient: Done.

include/dali/core/mm/async_pool.h
mzient: No, that's original block's end.

include/dali/core/mm/async_pool.h
mzient: It is not the same because size_t is unsigned and it would wrap around into huge regions. Been there, debugged that.

include/dali/core/mm/async_pool.h
mzient: It was doubled. I must have accidentally removed it. Thanks!

mzient:
!build

include/dali/core/mm/async_pool.h
mzient: Doubling restored (see below).

include/dali/core/mm/async_pool.h
mzient: It's actually a race condition. Thanks!

include/dali/core/mm/async_pool.h
mzient: Fixed.

include/dali/core/mm/async_pool.h
mzient: Well, perhaps... if there are more things to fix.

dali/core/mm/async_pool_test.cu
mzient: I moved it to the front.

include/dali/core/mm/async_pool.h
mzient: ...I can still use `is_ready` flag - calling ready() on something that's not ready will call cudaEventQuery again and may yield different result if the job on the GPU has finished between calls.

include/dali/core/mm/async_pool.h
```diff
@@ -0,0 +1,457 @@
...
+      free_pending(kv.second);
+    if (avoid_upstream_ && num_pending_frees_ > 0) {
+      ptr = global_pool_.try_allocate_from_free(bytes, alignment);
```
JanuszL, line 149:
```suggestion
      ptr = global_pool_.allocate(bytes, alignment);
```
As I understand `free_pending` doesn't affect free pool, but the global one, right?

include/dali/core/mm/async_pool.h
```diff
@@ -0,0 +1,457 @@
...
+    DeviceGuard dg(device_id_);
+    synchronize_impl(true);
+  }
+
+ private:
+  void synchronize_impl(bool lock) {
+    {
+      std::unique_lock<std::mutex> ulock(lock_, std::defer_lock);
+      if (lock)
+        ulock.lock();
```
JanuszL, line 70: Maybe you can move this lock to `synchronize()`?

include/dali/core/mm/async_pool.h
```diff
@@ -0,0 +1,457 @@
...
+    if (num_pending_frees_ == 0) {
+      ptr = global_pool_.allocate(bytes, alignment);
+      return ptr;
+    }
+    ptr = global_pool_.try_allocate_from_free(bytes, alignment);
+    if (ptr)
+      return ptr;
+    for (auto &kv : stream_free_)
+      free_pending(kv.second);
+    if (avoid_upstream_ && num_pending_frees_ > 0) {
```
JanuszL, line 148:
```suggestion
    if (avoid_upstream_) {
```
`free_pending` just decreases `num_pending_frees_`.

include/dali/core/mm/async_pool.h
```diff
@@ -0,0 +1,457 @@
...
+      assert(static_cast<ptrdiff_t>(front_padding) >= 0);
+      // NOTE: block_size - front_padding >= size  can overflow and fail - meh, unsigned size_t
+      if (block_size >= bytes + front_padding) {
+        if (!supports_splitting && block_size - bytes - front_padding > max_padding)
+          return nullptr;  // no point in continuing - the map is sorted
+
+        from.by_size.erase(it);
+        char *block_end = base + block_size;
+        char *remainder = detail::align_ptr(aligned + bytes, remainder_alignment);
+        bool split = supports_splitting && remainder + min_split_remainder < block_end;
```
JanuszL, line 269: I don't get this check `remainder + min_split_remainder < block_end`
`remainder` is always >= `block_end` as reminder is aligned on both ends. Do I miss anything?

mzient: No. block_end is the end of the block _before splitting_.
It goes like this:
```
|<-------------------------- block_size ------------------------------->|
|<------------ split_size --------------------->|                       |
v                                               v                       v
^                   ^                           ^                       ^
|<- front_padding ->|<- bytes -> <- rem_align-> |<-- remainder_size --> |
|                   |                           |                       |
|___ base           |__aligned                  |__remainder            |__ block_end
```

JanuszL: Now it is clear. I confused `block_size` with `size`.

include/dali/core/mm/async_pool.h
mzient: `free_pending` returns memory from per-stream pending free list to the global pool.

JanuszL (approved):

include/dali/core/mm/async_pool.h
mzient: There would be too many layers to pass the lock through when calling synchronize from `allocate_from_global_pool`

awolant (approved):

mzient:
!build

PR #2817, opened 2021-03-25
Add Numba func example notebook

majra20:
#### Why we need this PR?
- It adds python notebook example for numba_func

 - What solution was applied:
     *NA*
     *NA*
     *Clear explanation of numba_func in notebook*
 - Validation and testing:
     *NA*
     *NA*

**JIRA TASK**: *DALI-1935*


docs/examples/numba_func/Numba_func.ipynb
```diff
@@ -0,0 +1,242 @@
+{
```
mzient, line 1: <blockquote>operators written as a Numba compiled C callback.</blockquote><p> </p><p>Don't mix singular and plural - "run <strong>an operator</strong> written as a Numba...." or "run operators written as Numba compiled C callback<strong>s</strong>".</p><p> </p><blockquote>The operator is used together with<strong><s> a</s></strong> Numba <code>@cfunc</code> annotated function</blockquote><p>Remove "a"</p><p> </p><blockquote>translated into a compiled function foreign C code, allowing to eliminate completely or reduce the python interpreter overhead.</blockquote><p> </p><p><strong>compiled to a function callable from external C/C++</strong> code, <strong>eliminating</strong> <strong>P</strong>ython interpreter overhead.</p><p> </p><p>^^^ If it's an externally callable, the interpreter is removed completely from the equation, isn't it?</p><p> </p><blockquote>1. Let's start by importing DALI and a handful of utils.</blockquote><p> </p><p>I don't think that numbering works well with "let's" tone. I'd remove the numbers or put all instructions in one place and then elaborate on particular steps.</p><p> </p><p> </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/NVIDIA/DALI/pull/2817/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='0'/>

mzient, line 1: <blockquote>As you can see we don't specify number of samples in batch. That's because this function is run on a per-sample basis. </blockquote><p>As you can see we don't specify <strong>the </strong>number of samples in <strong>the</strong> batch. That's because this function is run on a per-sample basis. </p><p> </p><blockquote>Note: Please remember passing incorrect outputs and inputs types to <code>run_fn_sig</code> will result in unexpected behaviour. </blockquote><p>Note: Passing incorrect outputs and inputs types to <code>run_fn_sig</code> will result in undefined behaviour. </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/NVIDIA/DALI/pull/2817/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='6'/>

PR #2847, opened 2021-04-07
Split Python tests into smaller chunks

klecki:
#### Why we need this PR?
Balance CI load by splitting long test

 - What solution was applied:
     Test was split into `-operators` (all the test_python_operator_*.py files except readers and decoders), and `-core`: all other files and custom not-nosetests tests.
     Fallback directory was added, that should allow to run those tests as one job instead of 2 separate for L1 test level.
     Test. CI config also adjusted.
     Typos in file filters
 - Validation and testing:
     CI
     Nope

**JIRA TASK**: *[DALI-1901]*


qa/TL0_python-self-test/test.sh
```diff
@@ -1,2 +1,9 @@
 #!/bin/bash -e
-./test_nofw.sh
\ No newline at end of file
+
```
JanuszL, line 2: Do we still need need `TL0_python-self-test`?

klecki: So my idea is to keep it and run all the split test in there, otherwise we would have new jobs: 3 * `number of supported architectures` in the CI. And this way we can run it in one go when doing a build with L1.

JanuszL: So please remove other files from qa/TL0_python-self-test, like `test_body.sh`, `test_no_setup.sh` or `test_nofw.sh`.

qa/TL0_python-self-test-readers-decoders/test_nofw.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash -e
+# used pip packages
+pip_packages="nose numpy>=1.17 opencv-python pillow librosa scipy nvidia-ml-py==11.450.51 psutil numba"
```
JanuszL, line 3: Do we need all of this?

qa/TL0_python-self-test-operators/test_nofw.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash -e
+# used pip packages
+pip_packages="nose numpy>=1.17 opencv-python pillow librosa scipy nvidia-ml-py==11.450.51 psutil numba"
```
JanuszL, line 3: Do we need all of this here?

qa/TL0_python-self-test/test.sh
klecki: I already did.

JanuszL: Ok. Git shows them as moved. Missed that.

qa/TL0_python-self-test-readers-decoders/test_nofw.sh
klecki: This is good question, that I would prefer not to answer here. :V

klecki: I hope now it runs.

JanuszL (approved):

dali/CMakeLists.txt
```diff
@@ -143,8 +143,20 @@ if (BUILD_PYTHON)
...
+  add_custom_command(
+    TARGET check-python
+    WORKING_DIRECTORY ${PROJECT_SOURCE_DIR}/qa/TL0_python-self-test-readers-decoders/
+    COMMAND PYTHONPATH=${PYTHONPATH} ${PROJECT_SOURCE_DIR}/qa/TL0_python-self-test-readers-decoders/test_no_setup.sh
+  )
+
+  add_custom_command(
+    TARGET check-python
+    WORKING_DIRECTORY ${PROJECT_SOURCE_DIR}/qa/TL0_python-self-test-operators/
+    COMMAND PYTHONPATH=${PYTHONPATH} ${PROJECT_SOURCE_DIR}/qa/TL0_python-self-test-operators/test_no_setup.sh
```
szalpal, line 159: Maybe you could use `qa/TL0_python-self-test/test.sh`? Since it's a gather-all test?

qa/TL0_python-self-test-operators/test_body.sh
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash -e
+
+test_nose() {
+    # readers and decoders are testend in the TL0_python-self-test-core
```
szalpal, line 4:
```suggestion
    # readers and decoders are tested in the TL0_python-self-test-core
```

szalpal (approved):

dali/CMakeLists.txt
klecki: No, cause it involves running the pip installs, and here we run the scripts without using pip, and assuming that you already have the deps.

PR #2862, opened 2021-04-13
Updates MXNet plugin documentation regarding ``last_batch_policy``

JanuszL:
- in the case of MXNet both `FILL` and `PARTIAL` policies return a full batch
  every time, the only difference is a value of the `pad` property in the
  returned MXNet array. This PR adds proper documentation of that

#### Why we need this PR?
*Pick one, remove the rest*
- It updates MXNet plugin documentation regarding ``last_batch_policy``
 - What solution was applied:
     in the case of MXNet both `FILL` and `PARTIAL` policies return a full batch every time, the only difference is a value of the `pad` property in the returned MXNet array. This PR adds proper documentation of that
     mxnet.py
     wording
 - Validation and testing:
     NA
     Only docs are updated

**JIRA TASK**: *[NA]*


JanuszL:
!build

awolant (approved):

jantonguirao (approved):

PR #2864, opened 2021-04-15
Better error handling in pipeline.py

szalpal:
#### Why we need this PR?
*Pick one, remove the rest*
- Refactoring to improve error handling. This is especially useful in dali_backend pipelines, where the common pattern used is:

```python
@piepline_def(...)
def pipe():
    ...
    return x

def main(filename):
    pipe().serialize(filename)
```
Without the change, the code above would explode with meaningless error, saying that you can't use `define_graph` and `set_outputs` at the same time...

 - What solution was applied:
     Added `if` statement
     `pipeline.py`
     NA
 - Validation and testing:
     NA
     NA

### DIFF
--- a/dali/python/nvidia/dali/pipeline.py
+++ b/dali/python/nvidia/dali/pipeline.py
@@ -1021,8 +1021,8 @@ Parameters
         if define_graph is not None and not callable(define_graph):
             raise TypeError(
                 "Provided `define_graph` argument is not callable." +
-                " Didn't you want to write `.serialize(filename=...)`?"
-                if isinstance(define_graph, str) else "")
+                (" Didn't you want to write `.serialize(filename=...)`?"
+                if isinstance(define_graph, str) else ""))
         if not self._py_graph_built:
             self._build_graph(define_graph)
         if not self._backend_prepared:
--- a/dali/python/nvidia/dali/pipeline.py
+++ b/dali/python/nvidia/dali/pipeline.py
@@ -1018,6 +1018,11 @@ Parameters
         kwargs : dict
                 Refer to Pipeline constructor for full list of arguments.
         """
+        if define_graph is not None and not callable(define_graph):
+            raise TypeError(
+                "Provided `define_graph` argument is not callable." +
+                " Didn't you want to write `.serialize(filename=...)`?"
+                if isinstance(define_graph, str) else "")
         if not self._py_graph_built:
             self._build_graph(define_graph)
         if not self._backend_prepared:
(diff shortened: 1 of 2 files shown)


szalpal:
!build

dali/python/nvidia/dali/pipeline.py
```diff
@@ -1018,6 +1018,11 @@ def serialize(self, define_graph=None, filename=None):
         kwargs : dict
                 Refer to Pipeline constructor for full list of arguments.
         """
+        if define_graph is not None and not callable(define_graph):
```
JanuszL, line 1021: Could you add a test for that?
I'm not sure what is the exact use case where it should help.

dali/python/nvidia/dali/pipeline.py
```diff
@@ -1018,6 +1018,11 @@ def serialize(self, define_graph=None, filename=None):
...
+            raise TypeError(
+                "Provided `define_graph` argument is not callable." +
+                " Didn't you want to write `.serialize(filename=...)`?"
+                if isinstance(define_graph, str) else "")
```
mzient, line 1025:
```suggestion
                (" Didn't you want to write `.serialize(filename=...)`?"
                if isinstance(define_graph, str) else ""))
```

mzient (approved):

dali/python/nvidia/dali/pipeline.py
JanuszL: Also another question is if we should add similar guarding code to other methods?

szalpal: Sure, added a test.

As for the use case, I've explained it in the PR description, here's a copy-paste:

This is especially useful in dali_backend pipelines, where the common pattern used is:
```python
@piepline_def(...)
def pipe():
    ...
    return x

def main(filename):
    pipe().serialize(filename)
```
Without the change, the code above would explode with meaningless error, saying that you can't use define_graph and set_outputs at the same time...

dali/python/nvidia/dali/pipeline.py
szalpal: Done.

dali/python/nvidia/dali/pipeline.py
JanuszL: I think you should write it guards from the case where the filename is passed as the first positional argument, while it is the second.
This description tells nothing and I have spend a couple of minutes trying to understand the nature of the problem.

JanuszL (approved):

szalpal:
!build

PR #2879, opened 2021-04-19
Remove workaround for the problem with patchelf changing TLS alignment for CUDA < 10.2 and > 11.1

JanuszL:
- the workaround for the problem with patchelf changing TLS is not needed
  for > CUDA 11.1 and < 10.2 as CUDA RT handles TLS differently in that
  versions

#### Why we need this PR?
*Pick one, remove the rest*
- It removes workaround for the problem with patchelf changing TLS alignment for CUDA < 10.2 and > 11.1

 - What solution was applied:
     the workaround for the problem with patchelf changing TLS is not needed for > CUDA 11.1 and < 10.2 as CUDA RT handles TLS differently in that versions
     all dali .so libraries and runables
     NA
 - Validation and testing:
     CI build for CUDA 11.3
     NA

**JIRA TASK**: *[NA]*

### DIFF
--- a/dali/benchmark/dali_bench.cc
+++ b/dali/benchmark/dali_bench.cc
@@ -13,6 +13,7 @@
 // limitations under the License.
 
 #include "dali/benchmark/dali_bench.h"
+#include <cuda_runtime_api.h>
 
 #include <benchmark/benchmark.h>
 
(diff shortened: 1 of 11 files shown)


klecki:
Question: how does it work if I install CUDA 11.2 build in 11.1 environment? As we use static CUDA RT it shouldn't be a problem, is there any non-static part that may still cause it? At first I thought you removed the WAR completely, but ifdefing it should be fine.

JanuszL:
> Question: how does it work if I install CUDA 11.2 build in 11.1 environment? As we use static CUDA RT it shouldn't be a problem, is there any non-static part that may still cause it? At first I thought you removed the WAR completely, but ifdefing it should be fine.

No, this is strictly related to the static part we link to. That is why patchelf used to break it.

klecki (approved):

mzient (approved):

PR #2902, opened 2021-04-26
Use stream 0 in VideoDecoder when running driver >460 / CUDA >= 11.3.

mzient:

#### Why we need this PR?
*Pick one, remove the rest*
- It fixes a bug: crash when running 2 video decoders in parallel.

 - What solution was applied:
     * Detect driver version >460 or driver CUDA version >= 11.3 and use default stream
     * NVDEC
     * dali_operators linkage (added dependency on dynlink_nvml)
     * N/A
 - Validation and testing:
     * Existing tests cover it
     * N/A

**JIRA TASK**: [DALI-1973]

### DIFF
--- a/dali/operators/reader/nvdecoder/nvdecoder.cc
+++ b/dali/operators/reader/nvdecoder/nvdecoder.cc
@@ -32,6 +32,7 @@
 #include "dali/core/error_handling.h"
 #include "dali/operators/reader/nvdecoder/imgproc.h"
 #include "dali/core/device_guard.h"
+#include "dali/util/nvml.h"
 
 namespace dali {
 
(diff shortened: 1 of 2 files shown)


mzient:
!build

dali/operators/reader/nvdecoder/nvdecoder.cc
```diff
@@ -57,6 +58,37 @@ NvDecoder::NvDecoder(int device_id,
...
+
+      return std::stof(version);
+    }();
+    if (driver_version > 460)
+      use_default_stream = true;
+  }
+#else
+  {
+    int driver_cuda_version = 0;
+    CUDA_CALL(cuDriverGetVersion(&driver_cuda_version));
```
JanuszL, line 81: Why can't use this in all cases?

mzient: Maybe we'll be able to cut it off somewhere between versions that differ in supported CUDA version. If it coincides, we can drop NVML part.

JanuszL (approved):

awolant (approved):
Maybe it would be good to add runtime warning, when this is triggered?

mzient:
> Maybe it would be good to add runtime warning, when this is triggered?

Done.

awolant (approved):

mzient:
!build

PR #2924, opened 2021-05-05
Better error message when insufficient data in cache

szalpal:

#### Why we need this PR?
*Pick one, remove the rest*
This PR adds better error message, when DALI fails on the situation, that there's not enough data in external_source's cache

 - What solution was applied:
     try-catch block
     executor
     NA
 - Validation and testing:
    NA
     NA

### DIFF
--- a/dali/pipeline/executor/executor.cc
+++ b/dali/pipeline/executor/executor.cc
@@ -362,7 +362,7 @@ int Executor<WorkspacePolicy, QueuePolicy>::InferBatchSize(
     for (auto &bsp : bsps) {
       bsp->Advance();
     }
-  } catch (const std::exception &e) {
+  } catch (const std::out_of_range &e) {
     DALI_FAIL(
         make_string("Failed to acquire the next batch. Make sure, that DALI Pipeline is fed "
                     "with sufficient amount of data. ", e.what()));
--- a/dali/pipeline/executor/executor.cc
+++ b/dali/pipeline/executor/executor.cc
@@ -356,8 +356,9 @@ int Executor<WorkspacePolicy, QueuePolicy>::InferBatchSize(
     DALI_ENFORCE(bsps[0]->NextBatchSize() == bsps[i]->NextBatchSize(),
                  "Batch size must be uniform across an iteration");
   }
+  int batch_size;
   try {
-    auto batch_size = bsps[0]->NextBatchSize();
+    batch_size = bsps[0]->NextBatchSize();
     assert(batch_size > 0);
     for (auto &bsp : bsps) {
       bsp->Advance();
--- a/dali/pipeline/executor/executor.cc
+++ b/dali/pipeline/executor/executor.cc
@@ -356,10 +356,20 @@ int Executor<WorkspacePolicy, QueuePolicy>::InferBatchSize(
     DALI_ENFORCE(bsps[0]->NextBatchSize() == bsps[i]->NextBatchSize(),
                  "Batch size must be uniform across an iteration");
   }
-  auto batch_size = bsps[0]->NextBatchSize();
-  assert(batch_size > 0);
-  for (auto &bsp : bsps) {
-    bsp->Advance();
+  try {
+    auto batch_size = bsps[0]->NextBatchSize();
+    assert(batch_size > 0);
+    for (auto &bsp : bsps) {
+      bsp->Advance();
+    }
+  } catch (const std::runtime_error &e) {
+    DALI_FAIL(
+        make_string("Failed to acquire the next batch size. Make sure, that DALI Pipeline is fed "
+                    "with sufficient amount of data. ", e.what()));
+  } catch (const std::out_of_range &e) {
+    DALI_FAIL(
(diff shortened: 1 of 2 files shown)


szalpal:
!build

dali/pipeline/executor/executor.cc
```diff
@@ -356,10 +356,20 @@ int Executor<WorkspacePolicy, QueuePolicy>::InferBatchSize(
...
+      bsp->Advance();
+    }
+  } catch (const std::runtime_error &e) {
+    DALI_FAIL(
+        make_string("Failed to acquire the next batch size. Make sure, that DALI Pipeline is fed "
+                    "with sufficient amount of data. ", e.what()));
+  } catch (const std::out_of_range &e) {
+    DALI_FAIL(
+        make_string("No more batches in cache left. Make sure, that DALI Pipeline is fed with "
+                    "sufficient amount of data. ", e.what()));
```
JanuszL, line 372: What is the cache? I think both errors should go to the same bag - the user hasn't provided enough data.

szalpal:
!build

dali/pipeline/executor/executor.cc
```diff
@@ -356,10 +356,16 @@ int Executor<WorkspacePolicy, QueuePolicy>::InferBatchSize(
     DALI_ENFORCE(bsps[0]->NextBatchSize() == bsps[i]->NextBatchSize(),
                  "Batch size must be uniform across an iteration");
   }
-  auto batch_size = bsps[0]->NextBatchSize();
-  assert(batch_size > 0);
-  for (auto &bsp : bsps) {
-    bsp->Advance();
+  try {
+    auto batch_size = bsps[0]->NextBatchSize();
+    assert(batch_size > 0);
```
mzient, line 361: I wouldn't enforce that - and definitely not with an assert (only debug). Once we have full batch variability, batches of 0 will be perfectly acceptable (imagine conditional execution where the predicate matches for all samples).

JanuszL (approved):

dali/pipeline/executor/executor.cc
```diff
@@ -356,10 +356,16 @@ int Executor<WorkspacePolicy, QueuePolicy>::InferBatchSize(
...
+    for (auto &bsp : bsps) {
+      bsp->Advance();
+    }
+  } catch (const std::exception &e) {
```
mzient, line 365: I think it's dirty to catch a generic exception and assume its meaning. There should be a better error resolution upstream (preferred) or better inspection here (passable, but still ugly). The way it is, we can produce truly ridiculous errors here, like:
```
"Failed to acquire the next batch. Make sure, that DALI Pipeline is fed with sufficient amount of data. Bad alloc."
```

dali/pipeline/executor/executor.cc
```diff
@@ -356,10 +356,16 @@ int Executor<WorkspacePolicy, QueuePolicy>::InferBatchSize(
...
+    DALI_FAIL(
+        make_string("Failed to acquire the next batch. Make sure, that DALI Pipeline is fed "
+                    "with sufficient amount of data. ", e.what()));
```
mzient, line 368: Can we have a name of the offending node here? If the node was populated with `feed_input`, we'll have the name of the very source the user failed to feed.

JanuszL: Usually the name is generated and doesn't tell much.

dali/pipeline/executor/executor.cc
szalpal: How about adding specialized exception to `batch_size_provider.h`? `NextBatchSize` and `Advance` could throw this one then
```c++
struct cache_empty : std::out_of_range {};
```
?

dali/pipeline/executor/executor.cc
szalpal: Generally, I believe, that when we do have full batch variability, this function would need to be rewritten from scratch. It's because this assumes, that every `BatchSizeProvider` would return the same batch size. I've put this condition in the `assert`, cause if the batch size is nonpositive, we rather have a coding error somewhere. User can't fill the cache with `batch_size=0` data...

mzient: Are you sure you can't just provide an empty batch in external source? Are we guarded against that?

dali/pipeline/executor/executor.cc
mzient: I don't think that's necessary. However, we already throw `out_of_range` and we could catch that. This at least eliminates everything else we could possibly catch here.

dali/pipeline/executor/executor.cc
mzient: Not for external_source that's manually fed with `feed_input`, usually _by name_ (like in Triton) - and for others it would at least be "ExternalSource_<index>" and that index would most likely correspond to the ordinal of the external source in order of operator instantiation.

mzient (approved):

szalpal:
!build

szalpal:
!build

PR #2951, opened 2021-05-12
Enable pool to return memory to the upstream upon OOM.

mzient:

#### Why we need this PR?
*Pick one, remove the rest*
- It adds new feature: **returning memory to upstream** needed because we allocate huge memory chunks for tensor lists.

 - What solution was applied:
     * Added an option to get a specific address range from a free list and free tree
     * Upon OOM, run over upstream blocks and return the completely free blocks to the upstream, removing them from free list
     * If the upstream uses paging, it can rearrange physical storage and assign a new continuous virtual address space to the released fragments
     * Pooling resource
     * Free lists
     * Test tracking resource
 - Validation and testing:
     * Unit tests for free_list
     * Unit test for pooling resource that checks that the deallocation actually happens
     * Doxygen docstrings, when necessary

**JIRA TASK**: DALI-2027 (needed as a fix)

### DIFF
--- a/dali/core/mm/free_list_test.cc
+++ b/dali/core/mm/free_list_test.cc
@@ -83,7 +83,8 @@ TEST(MMBestFitFreeList, PutGetMoveGet) {
 template <typename FreeList>
 void TestCoalescingRemoveIf() {
   FreeList fl;
-  char a alignas(16)[1000];
+  char storage alignas(16)[4000];
+  char *a = storage + 128;   // thank you, overzealous compiler!
   ASSERT_FALSE(fl.remove_if_in_list(a, 1));  // total removed: 250..750
   fl.put(a, 500);
   fl.put(a + 500, 500);
@@ -188,7 +189,8 @@ TEST(MMFreeTree, PutGet) {
 
 TEST(MMBestFitFreeList, RemoveIf) {
   best_fit_free_list fl;
-  char a alignas(16)[1000];
+  char storage alignas(16)[4000];
+  char *a = storage + 128;   // thank you, overzealous compiler!
   fl.put(a, 100);
   fl.put(a + 100, 200);
   fl.put(a + 300, 700);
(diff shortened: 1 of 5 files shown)


include/dali/core/mm/detail/free_list.h
```diff
@@ -278,6 +276,43 @@ class best_fit_free_list {
     unused_blocks_ = b;
   }
 
+  /**
+   * @brief Gets a specific chunk of memory from a specific block
+   *
+   * The start and end parameters must lie inside the block.
+   */
+  void *get_from_block(block **pblock, char *start, char *end) {
```
mzient, line 284: This function is factored out of the old `get` - it's not new code.

dali/core/mm/free_list_test.cc
```diff
@@ -148,6 +170,14 @@ TEST(MMFreeTree, PutGet) {
   TestCoalescingPutGet<free_tree>();
 }
 
+TEST(MMCoalescingFreeList, RemoveIf) {
+  TestCoalescingRemoveIf<coalescing_free_list>();
```
JanuszL, line 174: Shouldn't you test `best_fit_free_list` as well?

mzient: Well, that would require a different test. The implementation is almost trivial, though, and I don't plan to use it - it's there more or less for symmetry.

include/dali/core/mm/detail/free_list.h
```diff
@@ -478,6 +539,62 @@ class free_tree {
...
+   */
+  void *get_specific_block(char *start, char *end) {
+    assert(end > start);
+    if (by_addr_.empty())
+      return nullptr;
+    size_t size = end - start;
+
+    auto it = by_addr_.lower_bound(start);
+    if (it == by_addr_.end() || it->first > start)
+      it--;
```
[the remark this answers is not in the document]
mzient: Good point.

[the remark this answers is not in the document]
mzient: Fixed.

dali/core/mm/free_list_test.cc
mzient: Done.

JanuszL (approved):

banasraf (approved):

PR #2952, opened 2021-05-12
Fix handling of non_existing entries in TFRecord reader

JanuszL:
- DALI assumes that all features provided in the TFRecord reader
  exists in every entry, otherwise protobuf will throw.
  This PR lifts that assumption makes the reader return
  empty tensor when the desired entry has not been found.
  It should be in pair what TensorFlow reader does.

#### Why we need this PR?
*Pick one, remove the rest*
- It fixes handling of non_existing entries in TFRecord reader

 - What solution was applied:
     DALI assumes that all features provided in the TFRecord reader exists in every entry, otherwise protobuf will throw.
     This PR lifts that assumption makes the reader return empty tensor when the desired entry has not been found.
     It should be in pair what TensorFlow reader does.
     tfrecord_parser.h
     should we return empty tensor and/or warn/enforce that all entries exits
 - Validation and testing:
     test is added
     TFRecord Reader documentation is updated

**JIRA TASK**: *[NA]*

### DIFF
--- a/dali/operators/reader/parser/tfrecord_parser.h
+++ b/dali/operators/reader/parser/tfrecord_parser.h
@@ -66,7 +66,26 @@ class TFRecordParser : public Parser<Tensor<CPUBackend>> {
       auto& output = ws->Output<CPUBackend>(i);
       Feature& f = features_[i];
       std::string& name = feature_names_[i];
-      auto& encoded_feature = example.features().feature().at(name);
+      auto& feature = example.features().feature();
+      auto it = feature.find(name);
+      if (it == feature.end()) {
+        output.Resize({0});
+        // set type
+        switch (f.GetType()) {
+          case FeatureType::int64:
+            (void) output.mutable_data<int64_t>();
+            break;
+          case FeatureType::string:
+            (void) output.mutable_data<uint8_t>();
+            break;
+          case FeatureType::float32:
+            (void) output.mutable_data<float>();
+            break;
+        }
+        output.SetSourceInfo(data.GetSourceInfo());
+        continue;
+      }
+      auto& encoded_feature = it->second;
       if (f.HasShape() && f.GetType() != FeatureType::string) {
         if (f.Shape().empty()) {
           output.Resize({1});
--- a/dali/operators/reader/parser/tfrecord_parser.h
+++ b/dali/operators/reader/parser/tfrecord_parser.h
@@ -73,13 +73,13 @@ class TFRecordParser : public Parser<Tensor<CPUBackend>> {
         // set type
         switch (f.GetType()) {
           case FeatureType::int64:
-            (void) output.mutable_data<int64_t>();
+             output.Resize({0}, TypeTable::GetTypeInfo(DALI_INT64));
             break;
           case FeatureType::string:
-            (void) output.mutable_data<uint8_t>();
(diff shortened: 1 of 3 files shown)


JanuszL:
!build

JanuszL:
!build

dali/operators/reader/parser/tfrecord_parser.h
```diff
@@ -66,7 +66,26 @@ class TFRecordParser : public Parser<Tensor<CPUBackend>> {
...
+        switch (f.GetType()) {
+          case FeatureType::int64:
+            (void) output.mutable_data<int64_t>();
+            break;
+          case FeatureType::string:
+            (void) output.mutable_data<uint8_t>();
+            break;
+          case FeatureType::float32:
+            (void) output.mutable_data<float>();
+            break;
```
mzient, line 83: Setting the type as a side effect of calling `mutable_data` is ugly.
```suggestion
            output.Resize({0}, TypeTable::GetTypeInfo(DALI_INT64));
            break;
          case FeatureType::string:
            output.Resize({0}, TypeTable::GetTypeInfo(DALI_UINT8));
            break;
          case FeatureType::float32:
            output.Resize({0}, TypeTable::GetTypeInfo(DALI_FLOAT));
            break;
```

dali/test/python/test_operator_readers_index.py
```diff
@@ -146,3 +146,25 @@ def test_tfrecord_reader_alias():
...
+def test_tfrecord_reader_alias():
+    tfrecord = os.path.join(get_dali_extra_path(), 'db', 'tfrecord', 'train')
+    tfrecord_idx = os.path.join(get_dali_extra_path(), 'db', 'tfrecord', 'train.idx')
+    pipe = tfrecord_pipe_empty_fields(tfrecord, tfrecord_idx)
+    pipe.build()
+    out = pipe.run()
+    for tensor in out[0]:
+        assert len(np.array(tensor)) != 0
+    for tensor in out[1]:
+        assert len(np.array(tensor)) == 0
```
mzient, line 168:
```suggestion
        assert len(np.array(tensor)) == 0
        assert np.dtype == np.int64
```

dali/test/python/test_operator_readers_index.py
```diff
@@ -146,3 +146,25 @@ def test_tfrecord_reader_alias():
...
+    for tensor in out[2]:
+        assert len(np.array(tensor)) == 0
```
mzient, line 170:
```suggestion
        assert len(np.array(tensor)) == 0
        assert np.dtype == np.uint8        
```

JanuszL: Done

dali/test/python/test_operator_readers_index.py
JanuszL: Done

dali/operators/reader/parser/tfrecord_parser.h
JanuszL: Done

JanuszL:
!build

mzient (approved):

awolant (approved):

JanuszL:
@Riyria-was-taken - this should fix the issue with the TFRecird you mentioned some time ago.

PR #2955, opened 2021-05-13
Guard changes to default resources with a mutex.

mzient:

#### Why we need this PR?
*Pick one, remove the rest*
- It fixes a race condition when multiple threads try to initialize the default resource - it seems that it fixes the segmentation fault seen in #2948

 - What solution was applied:
     * Guard changes to default resources with a mutex.
     * Guard initialization of default resources with a mutex.
     * Default memory resources
     * N/A
 - Validation and testing:
     * N/A
     * N/A

**JIRA TASK**: N/A

### DIFF
--- a/dali/core/mm/default_resources.cc
+++ b/dali/core/mm/default_resources.cc
@@ -255,6 +255,13 @@ void SetDefaultDeviceResource(int device_id, device_async_resource *resource, bo
   SetDefaultDeviceResource(device_id, wrap(resource, own));
 }
 
+// This function is for testing purposes only - it must be visible
+DLL_PUBLIC void _Test_FreeDeviceResources() {
+  // clear does not deallocate - we need something stronger
+  decltype(g_resources.device) empty;
+  g_resources.device.swap(empty);
+}
+
 template <> DLL_PUBLIC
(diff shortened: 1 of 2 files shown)


dali/core/mm/default_resources.cc
```diff
@@ -174,6 +177,7 @@ const std::shared_ptr<device_async_resource> &ShareDefaultDeviceResourceImpl(int
     CUDA_CALL(cudaGetDevice(&device_id));
   }
   if (g_resources.device.empty()) {
+    std::lock_guard<std::mutex> lock(g_resources.mtx);
     int ndevs = 0;
```
JanuszL, line 181: Shouldn't you check `g_resources.device.empty()` again after taking the lock or we are fine with multiple `g_resources.device.resize(ndevs);` calls?

mzient: Not really - cudaGetDeviceCount is a cheap call and if the array it's not empty, resize will be a no-op.

mzient:
!build

JanuszL (approved):

szalpal (approved):

PR #2961, opened 2021-05-14
Fix returning memory to upstream from pool resource

mzient:

#### Why we need this PR?
*Pick one, remove the rest*
- It fixes a bug: incomplete return to upstream from pool

 - What solution was applied:
     * Don't decrement the index of deleted element when iterating backwards.
     * Pool resource
     * N/A
 - Validation and testing:
     * Existing tests apply
     * N/A

**JIRA TASK**: N/A

### DIFF
--- a/include/dali/core/mm/pool_resource.h
+++ b/include/dali/core/mm/pool_resource.h
@@ -167,7 +167,7 @@ class pool_resource_base : public memory_resource<kind, Context> {
             // - we can safely free it to the upstream.
             if (free_list_.remove_if_in_list(blk.ptr, blk.bytes)) {
               upstream_->deallocate(blk.ptr, blk.bytes, blk.alignment);
-              blocks_.erase_at(i--);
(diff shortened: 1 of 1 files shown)


include/dali/core/mm/pool_resource.h
```diff
@@ -167,7 +167,7 @@ class pool_resource_base : public memory_resource<kind, Context> {
             // - we can safely free it to the upstream.
             if (free_list_.remove_if_in_list(blk.ptr, blk.bytes)) {
               upstream_->deallocate(blk.ptr, blk.bytes, blk.alignment);
-              blocks_.erase_at(i--);
+              blocks_.erase_at(i);
```
mzient, line 170: `i--` would be necessary if we were iterating forward - but we're iterating backward, so we'd skip the previous element.

JanuszL (approved):

awolant (approved):

PR #2962, opened 2021-05-14
Add coverage test for CPU only and variable batch size test

JanuszL:
- adds a test that checks if all available DALI operators are covered
  by CPU only and variable batch size test. The test lists all operators
  and checks if the list of tested operators matches it. The list is
  manually maintained and need to be updated when a new test is added

#### Why we need this PR?
*Pick one, remove the rest*
- It adds coverage test for CPU only and variable batch size test

 - What solution was applied:
     adds a test that checks if all available DALI operators are covered by CPU only and variable batch size test. The test lists all operators and checks if the list of tested operators matches it. The list is manually maintained and need to be updated when a new test is added
     test_dali_variable_batch_size.py
     test_dali_cpu_only.py
     check list correctness
 - Validation and testing:
     CI run
     NA

**JIRA TASK**: *[DALI-2032]*

### DIFF
--- a/dali/test/python/test_dali_cpu_only.py
+++ b/dali/test/python/test_dali_cpu_only.py
@@ -27,6 +27,7 @@ import os
 import glob
 from math import ceil, sqrt
 import tempfile
+import sys
 
 data_root = get_dali_extra_path()
 images_dir = os.path.join(data_root, 'db', 'single', 'jpeg')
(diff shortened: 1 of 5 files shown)


JanuszL:
!build

dali/test/python/test_dali_cpu_only.py
```diff
@@ -817,4 +818,159 @@ def test_separated_exec_setup():
...
+    "dl_tensor_python_function",
+    "hidden.arithmetic_generic_op", #internal
+    "hidden.transform_translation", #internal
+    "video_reader",         # not supported for CPU
+    "video_reader_resize",  # not supported for CPU
+    "readers.video",        # not supported for CPU
+    "readers.video_resize", # not supported for CPU
+    "optical_flow",         # not supported for CPU
+]
+def test_coverage():
```
klecki, line 957: Please put it in some test utils and reuse in both cases.

dali/test/python/test_dali_cpu_only.py
```diff
@@ -817,4 +818,159 @@ def test_separated_exec_setup():
...
+            if len(prefix):
+                prefix += '.'
+            else:
+                prefix = ""
+            res.append(prefix + cls.__name__)
+        else:
+            for c in cls.__dict__.keys():
+                if not c.startswith("_") and c not in sys.builtin_module_names:
+                    c = cls.__dict__[c]
+                    res += get_functions(c, cls.__name__)
```
klecki, line 972: Dunno how it works, but when I imported the `fn` module, and did this:
`fn.__dict__['readers'].__dict__['file'].__name__`, it is `'file'` and not `'nvidia.dali.fn.readers.file'` or `'readers.file'`. So I'm thinking there might be a problem with the prefix handling and not reporting submodules of `fn` as `readers.file` for example.

klecki (commented):
As a general remark:
we are not able to test plugin operators that we include by default (pytorch, numba?).

There arises a question if we should test all the arithmetic/mathematical ops. (And the `math` module). Now it's only a subset, with no detection.

dali/test/python/test_dali_cpu_only.py
JanuszL: That is why you have prefix argument. In your case you get `fn.__dict__['readers'].__name__` as prefix, then you strip `nvidia.dali.fn` and prepend to `fn.__dict__['readers'].__dict__['file'].__name__`.

dali/test/python/test_dali_cpu_only.py
JanuszL: Done

JanuszL:
> As a general remark:
> we are not able to test plugin operators that we include by default (pytorch, numba?).

Not automatically. At least I don't have a good idea how to extract only relevant info from plugins modules where we have much more than we need.

> There arises a question if we should test all the arithmetic/mathematical ops. (And the `math` module). Now it's only a subset, with no detection.

Added math module.

klecki (approved):

dali/test/python/test_dali_variable_batch_size.py
```diff
@@ -692,3 +697,187 @@ def pipe(max_batch_size, input_data, device, input_layout):
...
+    "math.fpow",
+    "math.acosh",
+    "math.min",
+    "pytorch.TorchPythonFunction",
+    "numba.fn.experimental.numba_function",
+    "hidden.transform_translation", # intentional
+    "hidden.arithmetic_generic_op", # intentional
+    "coco_reader",              # readers are do not support variable batch size yet
+    "sequence_reader",          # readers are do not support variable batch size yet
+    "numpy_reader",             # readers are do not support variable batch size yet
```
szalpal, line 854:
```suggestion
    "numpy_reader",             # readers do not support variable batch size yet
```

szalpal (approved):

JanuszL:
!build

dali/test/python/test_dali_variable_batch_size.py
JanuszL: Done

PR #2983, opened 2021-05-22
Add synchronization and deferred deallocation to pool_resource

mzient:
* Add synchronization
* Add bulk deallocation
* Add deferred bulk deallocation utility
* Use deferred deallocation in async_pool

#### Why we need this PR?
*Pick one, remove the rest*
- It fixes a bug in #2948 - memory reused before usage is complete
- It adds new feature: deferred deallocation, which improves performance

 - What solution was applied:
     * `pool_resource` now synchronizes when calling `deallocate`. The synchronization scope is configurable.
     * `pool_resource` exposes `deallocate_no_sync` in which the caller guarantees that the memory can be reused immediately
     * `pool_resource` exposes `bulk_deallocate` which allows to free multiple blocks with just one synchronization and lock
     * `deferred_deallocator` collects deallocation requests and issues them in bulk in a worker thread
     * `deferred_dealloc_pool` uses deferred deallocator to implement `do_deallocate`
     * `async_pool` now uses `deferred_dealloc_pool`
     * `pool_options` have configurable synchronization scope and limit max. outstanding deallocations
     * Memory manager: pool, async pool
     * Synchronization issues
 - Validation and testing:
     * Unit tests:
         * check bulk deallocation
         * check deferred deallocator with a test CRTP class
         * modify `async_pool` test to use one thread with non-stream allocation/deallocation
      * NOTE: I've confirmed experimentally under stress conditions that bulk asyncrhonous deallocation actually occurs and that this code path is actually used. Unit tests can't check that.
     * Doxygen

**JIRA TASK**: DALI-2095 DALI-2096

### DIFF
--- a/include/dali/core/mm/pool_resource.h
+++ b/include/dali/core/mm/pool_resource.h
@@ -16,13 +16,21 @@
 #define DALI_CORE_MM_POOL_RESOURCE_H_
 
 #include <mutex>
+#include <condition_variable>
 #include "dali/core/mm/memory_resource.h"
 #include "dali/core/mm/detail/free_list.h"
 #include "dali/core/small_vector.h"
+#include "dali/core/device_guard.h"
 #include "dali/core/util.h"
 
 namespace dali {
 namespace mm {
+
+enum class sync_scope {
+  none  = 0,   ///< no synchronization required
+  device = 1,  ///< synchronize with the current device
+  system = 2   ///< synchronize with all devices in the system
+};
 
 struct pool_options {
   /**
@@ -48,6 +56,25 @@ struct pool_options {
    * @remarks This option is ignored when `try_smaller_on_failure` is set to `false`.
    */
   bool return_to_upstream_on_failure = true;
+
+  /**
+   * @brief To what extent should `deallocate` synchronize before making the memory available
+   */
+  sync_scope sync = sync_scope::none;
+
+  /**
+   * @brief Enables deferred deallocation if the pool supports it (otherwise ignored)
+   */
+  bool enable_deferred_deallocation = false;
+
+  /**
+   * @brief Maximum number of outstanding deferred deallocations
+   *
+   * If there are more outstanding deferred deallocations than this number,
+   * the subsequent allocation blocks.
+   */
+  int max_outstanding_deallocations = 16;
+
   size_t upstream_alignment = 256;
 };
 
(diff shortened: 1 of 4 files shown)


include/dali/core/mm/pool_resource.h
```diff
@@ -100,6 +322,18 @@ class pool_resource_base : public memory_resource<kind, Context> {
...
   }
 
+  /**
+   * @brief Deallocates a block memory without synchronization
+   *
+   * Places a block of memory in the free list for immediate reuse.
+   * The caller must guarantee, that the memory is available without
+   * any additional synchronization in the execution context for this resource.
+   */
+  void deallocate_no_sync(void *ptr, size_t bytes, size_t alignment = alignof(std::max_align_t)) {
```
mzient, line 332: I have to admit that I don't like this function very much. I just couldn't find a better alternative if I want to have both synchronized and unsynchronized deallocation in one resource.

include/dali/core/mm/pool_resource.h
```diff
@@ -48,6 +56,25 @@ struct pool_options {
...
+   */
+  bool enable_deferred_deallocation = false;
+
+  /**
+   * @brief Maximum number of outstanding deferred deallocations
+   *
+   * If there are more outstanding deferred deallocations than this number,
+   * the subsequent allocation blocks.
+   */
+  int max_outstanding_deallocations = 16;
```
JanuszL, line 76: How you selected this number?

include/dali/core/mm/pool_resource.h
```diff
@@ -59,11 +86,159 @@ constexpr pool_options default_device_pool_opts() noexcept {
...
+constexpr sync_scope default_sync_scope() {
+  return kind == memory_kind::device ? sync_scope::device
+                                     : kind == memory_kind::host ? sync_scope::none
+                                                                 : sync_scope::system;
+}
+
+template <memory_kind kind>
+constexpr pool_options default_pool_opts() noexcept {
+  if (kind == memory_kind::host) {
+    return default_host_pool_opts();
```
JanuszL, line 99: Why `opt.sync` is not set for symmetry?

include/dali/core/mm/pool_resource.h
```diff
@@ -59,11 +86,159 @@ constexpr pool_options default_device_pool_opts() noexcept {
...
+  int outstanding_dealloc_count() const {
+    return deallocs_[0].size() + deallocs_[1].size();
+  }
+
+  /**
+   * @brief Waits until currently scheduled deallocations are flushed.
+   *
+   * This function waits until the worker notifies that it's completed flushing
+   * current queue (there are two queues). It doesn't wait for the other queue
+   * nor prevent new deallocatiosn from being scheduled.
```
JanuszL, line 157:
```suggestion
   * nor prevent new deallocations from being scheduled.
```

include/dali/core/mm/pool_resource.h
```diff
@@ -59,11 +86,159 @@ constexpr pool_options default_device_pool_opts() noexcept {
...
+      return;  // nothing to do
+    if (device_id < 0) {
+      CUDA_CALL(cudaGetDevice(&device_id));
+    }
+
+    {
+      std::lock_guard<std::mutex> g(mtx_);
+      deallocs_[queue_idx_].push_back({device_id, ptr, bytes, alignment});
+
+      if (!started_)
```
JanuszL, line 141: Maybe `start_worker` should check this condition?

include/dali/core/mm/pool_resource.h
mzient: Nothing special, really. In fact, it's quite conservative. I've tried many times to even observe bulk deallocation and only saw it once until I ran `stress - c7` on another console (8-thread machine). I considered specifying it in outstanding deallocated bytes (instead of number of allocations) - but then decided that small deallocations also constribute to fragmentation and, pessimistiaclly, this could lead to indefinite growth of the outstanding allocation queue.

include/dali/core/mm/pool_resource.h
mzient: Because there's only one `host` kind and the default is already set by default_**host**_pool_opts - for device, it depends on the specific kind, so we have to change it.

include/dali/core/mm/pool_resource.h
mzient: I found it more human readable this way "if not already started then start" but I can move it if you insist.

include/dali/core/mm/pool_resource.h
mzient: Will fix.

include/dali/core/mm/pool_resource.h
JanuszL: Understood

include/dali/core/mm/pool_resource.h
JanuszL: Just asking. No strong opinion.

JanuszL (approved):

banasraf (approved):

PR #2994, opened 2021-05-26
Lifts the sm limitation that is tested in the numpy reader test

JanuszL:
- the first version didn't support GPUs with sm < 6.0 and the GPU
  test variant was disabled in such case. The current version do
  work so the limitation can be lifted

#### Why we need this PR?
*Pick one, remove the rest*
- It lifts the sm limitation that is tested in the numpy reader test

 - What solution was applied:
     the first version didn't support GPUs with sm < 6.0 and the GPU test variant was disabled in such case. The current version do work so the limitation can be lifted
     test_operator_readers_numpy.py
     NA
 - Validation and testing:
     CI run, local testing with sm 3.5
     NA

**JIRA TASK**: *[NA]*

### DIFF
--- a/dali/test/python/test_operator_readers_numpy.py
+++ b/dali/test/python/test_operator_readers_numpy.py
@@ -36,22 +36,24 @@ def is_gds_supported(device_id=0):
     if is_gds_supported_var is not None:
         return is_gds_supported_var
 
-    compute_cap = 0
-    try:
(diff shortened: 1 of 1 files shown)


JanuszL:
!build

jantonguirao (approved):

awolant (approved):

PR #3019, opened 2021-06-07
Extend `pip install` command timeout to 100s

JanuszL:
- attempts to reduce the number of timeout pip install attempts by extending the command timeout from default 60s to 100s

#### Why we need this PR?
*Pick one, remove the rest*
- It reduces the number of timeout pip install attempts by extending the command timeout from default 60s to 100s

 - What solution was applied:
     attempts to reduce the number of timeout pip install attempts by extending the command timeout from default 60s to 100s
     QA scripts
     NA
 - Validation and testing:
     CI run
     NA

**JIRA TASK**: *[NA]*

### DIFF
--- a/qa/test_template.sh
+++ b/qa/test_template.sh
@@ -56,7 +56,7 @@ do
         # install packages
         inst=$($topdir/qa/setup_packages.py -i $i -u $pip_packages --cuda ${CUDA_VERSION})
         if [ -n "$inst" ]; then
-            pip install $inst -f /pip-packages
+            pip install --default-timeout=100 $inst -f /pip-packages
 
             # If we just installed tensorflow, we need to reinstall DALI TF plugin
             if [[ "$inst" == *tensorflow* ]]; then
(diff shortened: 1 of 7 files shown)


qa/test_template.sh
```diff
@@ -56,7 +56,7 @@ do
         # install packages
         inst=$($topdir/qa/setup_packages.py -i $i -u $pip_packages --cuda ${CUDA_VERSION})
         if [ -n "$inst" ]; then
-            pip install $inst -f /pip-packages
+            pip install --default-timeout=100 $inst -f /pip-packages
```
klecki, line 59: I'm thinking, maybe it would be better to hit everything with 
`export PIP_DEFAULT_TIMEOUT=100` in the setup script?

But I see that the L3 jobs do not necessarily use it.

klecki (approved):
I tried to search for all of the `pip install` commands, looks like this covers all that happen in .sh

I don't remember this happening in docker.

jantonguirao (approved):

PR #3026, opened 2021-06-08
Fix batch size handling in PermuteBatch.

mzient:

#### Why we need this PR?
*Pick one, remove the rest*
- It fixes a bug: output batch size taken from input instead of RequestedBatchSize

 - What solution was applied:
     * Use RequestedBatchSize
     * PermuteBatch
     * N/A
 - Validation and testing:
     * Existing tests apply
     * Variable batch size tested in a debugger before being temporarily disallowed
     * N/A

**JIRA TASK**: N/A

### DIFF
--- a/dali/operators/generic/permute_batch.h
+++ b/dali/operators/generic/permute_batch.h
@@ -29,19 +29,25 @@ class PermuteBatchBase : public Operator<Backend> {
   }
 
   bool SetupImpl(vector<OutputDesc> &outputs, const workspace_t<Backend> &ws) override {
-    int curr_batch_size = ws.GetRequestedBatchSize(0);
     outputs.resize(1);
     auto &input = ws.template InputRef<Backend>(0);
     const auto &in_shape = input.shape();
     outputs[0].type = input.type();
 
     if (has_indices_input_) {
-      GetPerSampleArgument<int>(indices_, "indices", this->spec_, ws, curr_batch_size);
+      auto &idx_in = ws.ArgumentInput("indices");
+      auto idx_view = view<const int, 0>(idx_in);
+      indices_.resize(idx_view.num_samples());
+      for (int i = 0; i < idx_view.num_samples(); i++) {
+        indices_[i] = *idx_view.data[i];
+      }
     } else {
       this->spec_.TryGetRepeatedArgument(indices_, "indices");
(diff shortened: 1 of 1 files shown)


mzient:
!build

mzient:
!build

dali/operators/generic/permute_batch.h
```diff
@@ -29,17 +29,25 @@ class PermuteBatchBase : public Operator<Backend> {
...
   bool SetupImpl(vector<OutputDesc> &outputs, const workspace_t<Backend> &ws) override {
-    auto curr_batch_size = ws.GetInputBatchSize(0);
     outputs.resize(1);
     auto &input = ws.template InputRef<Backend>(0);
     const auto &in_shape = input.shape();
     outputs[0].type = input.type();
 
     if (has_indices_input_) {
-      GetPerSampleArgument<int>(indices_, "indices", this->spec_, ws, curr_batch_size);
+      auto &idx_in = ws.ArgumentInput("indices");
```
mzient, line 38: We don't know the batch size - it should be taken _from_ the argument, not imposed on it.

dali/operators/generic/permute_batch.h
```diff
@@ -29,17 +29,25 @@ class PermuteBatchBase : public Operator<Backend> {
...
+      auto idx_view = view<const int, 0>(idx_in);
+      indices_.resize(idx_view.num_samples());
+      for (int i = 0; i < idx_view.num_samples(); i++) {
+        indices_[i] = *idx_view.data[i];
+      }
     } else {
       this->spec_.TryGetRepeatedArgument(indices_, "indices");
     }
+    // TODO(michalz): Remove when fully variable batch size is supported
```
banasraf, line 47: I see that `GetRequestedBatchSzie(...)` returns the expected batch size of the output. Shouldn't number of elements in `indices` always match with the expected output batch size, even with fully variable BS?

mzient: Not really. We'll need to have some flags for operators which can drive batch size change and these operators will not have an "expected batch size" at all. For example, when we have batch splitting operators for conditional application then there's no "expected" batch size - their batch size will be propagated down the graph and will drive the expected batch size for other operators (inputless).

banasraf (approved):

szalpal (approved):

PR #3031, opened 2021-06-09
Add C API call to check backend of an operator

banasraf:
#### Why we need this PR?
- It adds new feature needed in DALI Triton backend.

We need to check the backend of an ExternalSource in DALI Triton backend. 

 - What solution was applied:
     *Added an C API call to check backend of an operator.*
     *c_api.h/cc*
     *api design*
 - Validation and testing:
     *added case to C API tests*
     *API call documentation*

**JIRA TASK**: *DALI-2000*

### DIFF
--- a/dali/c_api/c_api.cc
+++ b/dali/c_api/c_api.cc
@@ -1,4 +1,4 @@
-// Copyright (c) 2017-2020, NVIDIA CORPORATION. All rights reserved.
+// Copyright (c) 2017-2021, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
--- a/dali/c_api/c_api.cc
+++ b/dali/c_api/c_api.cc
@@ -547,10 +547,10 @@ dali_backend_t daliGetOperatorBackend(daliPipelineHandle* pipe_handle, const cha
   dali::Pipeline* pipeline = reinterpret_cast<dali::Pipeline*>(pipe_handle->pipe);
   auto *node = pipeline->GetOperatorNode(name);
   switch (node->op_type) {
-    case dali::OpType::GPU:
-      return dali_backend_t::DALI_BACKEND_GPU;
     case dali::OpType::CPU:
       return dali_backend_t::DALI_BACKEND_CPU;
+    case dali::OpType::GPU:
+      return dali_backend_t::DALI_BACKEND_GPU;
     case dali::OpType::MIXED:
       return dali_backend_t::DALI_BACKEND_MIXED;
     default:
--- a/dali/c_api/c_api.cc
+++ b/dali/c_api/c_api.cc
@@ -545,7 +545,7 @@ void daliGetReaderMetadata(daliPipelineHandle* pipe_handle, const char *reader_n
 
 dali_backend_t daliGetOperatorBackend(daliPipelineHandle* pipe_handle, const char *operator_name) {
   dali::Pipeline* pipeline = reinterpret_cast<dali::Pipeline*>(pipe_handle->pipe);
-  auto *node = pipeline->GetOperatorNode(name);
+  auto *node = pipeline->GetOperatorNode(operator_name);
   switch (node->op_type) {
     case dali::OpType::CPU:
       return dali_backend_t::DALI_BACKEND_CPU;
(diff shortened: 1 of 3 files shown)


klecki (approved):

dali/c_api/c_api.cc
```diff
@@ -543,6 +543,21 @@ void daliGetReaderMetadata(daliPipelineHandle* pipe_handle, const char *reader_n
   meta->stick_to_shard = returned_meta.stick_to_shard;
 }
 
+dali_backend_t daliGetOperatorBackend(daliPipelineHandle* pipe_handle, const char *name) {
```
szalpal, line 546: How about `operator_name`? Or `op_name`? Won't hurt us ;)

include/dali/c_api.h
```diff
@@ -407,6 +413,14 @@ DLL_PUBLIC void daliLoadLibrary(const char *lib_path);
  */
 DLL_PUBLIC void daliGetReaderMetadata(daliPipelineHandle* pipe_handle, const char *reader_name,
                                       daliReaderMetadata* meta);
+
+/**
+ * @brief Returns the backend of the operator with a given \p name
+ * @param name Name of the operator to query
+ */
+DLL_PUBLIC dali_backend_t daliGetOperatorBackend(daliPipelineHandle* pipe_handle,
+                                                 const char *name);
```
szalpal, line 422: How about `operator_name`? Or `op_name`? Won't hurt us ;)

include/dali/c_api.h
```diff
@@ -42,6 +42,12 @@ typedef enum {
   GPU = 1
 } device_type_t;
 
+typedef enum {
+  DALI_BACKEND_GPU = 0,
+  DALI_BACKEND_CPU = 1,
+  DALI_BACKEND_MIXED = 2
+} dali_backend_t;
```
szalpal, line 49: Couple things here. 

In `device_type_t` we assign `CPU=0` and `GPU=1` and here it's reversed. How about making this consistent?
```C++
CPU=0, GPU=1, MIXED=2
```

Secondly, in `device_type_t` we don't use `DALI` prefix for the enum name. Maybe we should keep this consistent also? After all, there's `dali` already in type name, so it's still readable:
```c++
typedef enum {
  CPU = 0,
  GPU = 1,
  MIXED = 2
} dali_backend_t;

dali_backend_t::CPU;
dali_backend_t::GPU;
dali_backend_t::MIXED;
```

banasraf: It's device type that is inconsistent with the rest of the enums. Also, C enums without a prefix is a bad idea, because you can reference them without device_type_t::*, so our namespace is polluted with `CPU` and `GPU`.

banasraf: I can't even name those `CPU` and `GPU`. The names are already taken by the device_type_t

include/dali/c_api.h
banasraf: Done

dali/c_api/c_api.cc
banasraf: Done

include/dali/c_api.h
szalpal: > It's device type that is inconsistent with the rest of the enums

In C API I see only `device_type_t` enum so far, so maybe it's good idea to keep them consistent within single API?

banasraf: About the reverse order - it's taken from OpType

banasraf: > I'm wondering, that in C API I see only `device_type_t` enum so far, so maybe it's good idea to keep them consistent within single API?

```c++
typedef enum {
  DALI_NO_TYPE  = -1,
  DALI_UINT8    =  0,
  DALI_UINT16   =  1,
  DALI_UINT32   =  2,
  DALI_UINT64   =  3,
  DALI_INT8     =  4,
  DALI_INT16    =  5,
  DALI_INT32    =  6,
  DALI_INT64    =  7,
  DALI_FLOAT16  =  8,
  DALI_FLOAT    =  9,
  DALI_FLOAT64  =  10,
  DALI_BOOL     =  11,
} dali_data_type_t;
```

banasraf: I changed the order of the GPU and CPU

szalpal (approved):

banasraf:
!build

banasraf:
!build

PR #3041, opened 2021-06-10
Add override flags for no_copy option of External Source

klecki:

#### Why we need this PR?
- It's needed for better TF integration 

 - What solution was applied:
     The External Source no_copy option is a "static" parameter, that is fixed after the pipeline build (and serialization). We need a way to control it from TF integration, so we added override flags for C API to control the External Source behaviour.
   This solution fits in with the other flags that we already have.   
     External Source & C API.
     Check if the flags are not swapped, or smth.
 - Validation and testing:
     C API test added.
     Yes.

**JIRA TASK**: *[Use DALI-2141 or NA]*

### DIFF
--- a/dali/c_api/c_api.cc
+++ b/dali/c_api/c_api.cc
@@ -1,4 +1,4 @@
-// Copyright (c) 2017-2020, NVIDIA CORPORATION. All rights reserved.
+// Copyright (c) 2017-2021, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -20,6 +20,7 @@
 #include <vector>
 #include <map>
 
+#include "dali/core/common.h"
 #include "dali/core/cuda_stream.h"
 #include "dali/core/format.h"
 #include "dali/core/tensor_shape.h"
@@ -57,6 +58,18 @@ int PopCurrBatchSize(batch_size_map_t *batch_size_map, int max_batch_size,
 }
 
 
+dali::DALIExtNoCopyMode GetExternalSourceCopyMode(unsigned int flags) {
+  dali::DALIExtNoCopyMode no_copy_mode = dali::DALIExtNoCopyMode::DEFAULT;
+  DALI_ENFORCE(!((flags & DALI_ext_force_copy) && (flags & DALI_ext_force_no_copy)),
+               "External Source cannot be forced to use copy and no copy at the same time.");
+  if (flags & DALI_ext_force_copy) {
+    no_copy_mode = dali::DALIExtNoCopyMode::FORCE_COPY;
+  } else if (flags & DALI_ext_force_no_copy) {
+    no_copy_mode = dali::DALIExtNoCopyMode::FORCE_NO_COPY;
+  }
+  return no_copy_mode;
+}
+
 template <typename Backend>
 void SetExternalInput(daliPipelineHandle *pipe_handle, const char *name, const void *data_ptr,
                       dali_data_type_t data_type, const int64_t *shapes, int sample_dim,
@@ -82,7 +95,8 @@ void SetExternalInput(daliPipelineHandle *pipe_handle, const char *name, const v
   data.SetLayout(layout);
   pipeline->SetExternalInput(name, data, stream,
                              flags & DALI_ext_force_sync,
-                             flags & DALI_use_copy_kernel);
+                             flags & DALI_use_copy_kernel,
+                             GetExternalSourceCopyMode(flags));
 }
 
 
@@ -114,7 +128,8 @@ void SetExternalInputTensors(daliPipelineHandle *pipe_handle, const char *name,
   }
   pipeline->SetExternalInput(name, data, stream,
                              flags & DALI_ext_force_sync,
-                             flags & DALI_use_copy_kernel);
+                             flags & DALI_use_copy_kernel,
+                             GetExternalSourceCopyMode(flags));
 }
 
 dali::kernels::AllocType GetAllocType(device_type_t device_type, bool is_pinned) {
--- a/dali/c_api/c_api.cc
+++ b/dali/c_api/c_api.cc
@@ -57,15 +57,20 @@ int PopCurrBatchSize(batch_size_map_t *batch_size_map, int max_batch_size,
   return ret;
 }
 
-
-dali::DALIExtNoCopyMode GetExternalSourceCopyMode(unsigned int flags) {
-  dali::DALIExtNoCopyMode no_copy_mode = dali::DALIExtNoCopyMode::DEFAULT;
+/**
+ * @brief Extract ExtSrcNoCopyMode based on the DALI_ext_force_copy and DALI_ext_force_no_copy
+ *
+ * @param flags Flags typically specified in daliSetExternalInput* functions.
+ */
+dali::ExtSrcNoCopyMode GetExternalSourceCopyMode(unsigned int flags) {
+  dali::ExtSrcNoCopyMode no_copy_mode = dali::ExtSrcNoCopyMode::DEFAULT;
   DALI_ENFORCE(!((flags & DALI_ext_force_copy) && (flags & DALI_ext_force_no_copy)),
-               "External Source cannot be forced to use copy and no copy at the same time.");
+               "External Source cannot be forced to use DALI_ext_force_copy and "
+               "DALI_ext_force_no_copy at the same time.");
   if (flags & DALI_ext_force_copy) {
-    no_copy_mode = dali::DALIExtNoCopyMode::FORCE_COPY;
+    no_copy_mode = dali::ExtSrcNoCopyMode::FORCE_COPY;
   } else if (flags & DALI_ext_force_no_copy) {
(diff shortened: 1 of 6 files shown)


klecki:
!build

klecki:
!build

dali/c_api/c_api_test.cc
```diff
@@ -610,6 +636,100 @@ TYPED_TEST(CApiTest, UseCopyKernel) {
...
+  TensorListShape<> input_shape = {{37, 23, 3}, {12, 22, 3}, {42, 42, 3}, {8, 8, 3},
+                                   {64, 32, 3}, {32, 64, 3}, {20, 20, 3}, {64, 64, 3},
+                                   {10, 10, 3}, {60, 50, 3}, {10, 15, 3}, {48, 48, 3}};
+  TensorList<CPUBackend> input_cpu;
+  input_cpu.Resize(input_shape, TypeInfo::Create<uint8_t>());
+
+  TensorList<TypeParam> input;
+
+  auto device = backend_to_device_type<TypeParam>::value;
+  std::string device_str = device == CPU ? "cpu" : "gpu";
```
JanuszL, line 650: I would create a function for that.
```suggestion
  std::string device_str = GetDeviceStr(device);
```

dali/c_api/c_api_test.cc
```diff
@@ -610,6 +636,100 @@ TYPED_TEST(CApiTest, UseCopyKernel) {
   }
 }
 
+
+TYPED_TEST(CApiTest, NoCopyForce) {
```
JanuszL, line 640:
```suggestion
TYPED_TEST(CApiTest, NoCopyForceFail) {
```

dali/pipeline/pipeline.h
```diff
@@ -195,12 +194,14 @@ class DLL_PUBLIC Pipeline {
...
-  SetExternalInput(const string &name, const TensorList<Backend> &tl, cudaStream_t stream = 0,
-                   bool sync = false, bool use_copy_kernel = false) {
-    SetExternalInputHelper(name, tl, stream, sync, use_copy_kernel);
+   * @param no_copy_mode Select whether to use the parameter defined in the External Source or
+   *                     override the mode of operation forcing the copy or no-copy
+   */
+  template <typename Backend>
+  DLL_PUBLIC inline void SetExternalInput(
+      const string &name, const TensorList<Backend> &tl, cudaStream_t stream = 0, bool sync = false,
+      bool use_copy_kernel = false, DALIExtNoCopyMode no_copy_mode = DALIExtNoCopyMode::DEFAULT) {
```
JanuszL, line 203: Why not to use `SetDataConfig` in the `SetExternalInput` as well?

include/dali/core/common.h
```diff
@@ -111,6 +111,12 @@ enum DALIImageType {
   DALI_ANY_DATA     = 4
 };
 
+enum class DALIExtNoCopyMode {
```
JanuszL, line 114: I wonder if it deserves to be placed in common.h.

dali/pipeline/pipeline.h
klecki: Cause it's the most external part of the API, so I decided not to make a breaking change. If we find out that the list of additional flags is growing, we can make a break then.

I don't think I can make an overload with it as optional and those two would clash.

include/dali/core/common.h
klecki: I can move it somewhere to external source or pipeline.h

dali/pipeline/pipeline.h
JanuszL: OK

include/dali/core/common.h
JanuszL: Maybe external_source.h. Pipeline.h includes it anyway.

dali/pipeline/operator/builtin/external_source.h
```diff
@@ -161,6 +162,22 @@ auto get_batch_size(const BatchContainer<Backend>& container) {
...
+ * @brief Options that can be configured when setting data for the External Source
+ *
+ * @param sync If SetExternalInputHelper should be blocking - waits until provided data is copied
+ *             to the internal buffer
+ * @param use_copy_kernel If true, a copy kernel will be used to make a
+ *                        contiguous buffer instead of cudaMemcpyAsync.
+ * @param no_copy_mode Select whether to use the parameter defined in the External Source or
+ *                     override the mode of operation forcing the copy or no-copy
+ */
+struct SetDataConfig {
```
szalpal, line 175: How about changing the name to `SourceDataConfig` or `ExternalSourceDataConfig`? I feel, that `Set` here doesn't really fit.

dali/pipeline/operator/builtin/external_source.h
```diff
@@ -469,10 +485,24 @@ class ExternalSource : public Operator<Backend>, virtual public BatchSizeProvide
...
+    switch (set_data_config.no_copy_mode) {
+      case DALIExtNoCopyMode::FORCE_COPY:
+        actual_no_copy = false;
+        break;
+      case DALIExtNoCopyMode::FORCE_NO_COPY:
+        actual_no_copy = true;
+        break;
+      default:
+        actual_no_copy = no_copy_;
+        break;
```
szalpal, line 499:
```suggestion
      default:
        break;
```

dali/pipeline/operator/builtin/external_source.h
```diff
@@ -197,39 +214,38 @@ class ExternalSource : public Operator<Backend>, virtual public BatchSizeProvide
   /**
    * @brief Sets the data that should be passed out of the op on the next iteration.
    */
-  template<typename SrcBackend>
+  template <typename SrcBackend>
   inline void SetDataSource(const TensorList<SrcBackend> &tl, cudaStream_t stream = 0,
-                            bool sync = false, bool use_copy_kernel = false) {
+                            SetDataConfig set_data_config = {}) {
```
szalpal, line 219: Naming this var `set_data_config` kind of suggests, that it's a function - we sometimes use snake_case for functions and the `set` prefix really indicates a function. Maybe renaming it, as I suggested above?

dali/c_api/c_api.cc
```diff
@@ -57,6 +58,18 @@ int PopCurrBatchSize(batch_size_map_t *batch_size_map, int max_batch_size,
 }
 
 
+dali::DALIExtNoCopyMode GetExternalSourceCopyMode(unsigned int flags) {
+  dali::DALIExtNoCopyMode no_copy_mode = dali::DALIExtNoCopyMode::DEFAULT;
+  DALI_ENFORCE(!((flags & DALI_ext_force_copy) && (flags & DALI_ext_force_no_copy)),
```
szalpal, line 63: Please add doc explaining `flags`

dali/c_api/c_api.cc
```diff
@@ -57,6 +58,18 @@ int PopCurrBatchSize(batch_size_map_t *batch_size_map, int max_batch_size,
...
+               "External Source cannot be forced to use copy and no copy at the same time.");
```
szalpal, line 64: I feel, that this err msg would be more meaningful, if it would explicitly say, that this is because of `DALI_ext_force_copy` and `DALI_ext_force_no_copy` flags

include/dali/c_api.h
```diff
@@ -137,6 +138,17 @@ enum {
...
+  /**
+   * Override the `no_copy` specified for given External Source and force the data to be copied.
+   */
+  DALI_ext_force_copy = (1 << 3),
+
+  /**
+   * Override the `no_copy` specified for given External Source and pass the data directly to the
+   * Pipeline.
+   */
+  DALI_ext_force_no_copy = (1 << 4),
```
szalpal, line 151: What lacks for me in this docs is information, how does it different from specifying `no_copy` in "regular" way (by "regular" I mean whatever that must be overwritten)

szalpal (commented):
Please consider changing the `SetDataConfig` struct name

dali/c_api/c_api_test.cc
klecki: done

dali/c_api/c_api_test.cc
klecki: Done something similar. I think we can start with Force...

include/dali/core/common.h
klecki: Done.

dali/pipeline/operator/builtin/external_source.h
klecki: We also use *only* snake case for variable names, especially function parameters.

dali/pipeline/operator/builtin/external_source.h
klecki: Done with ExtSrcSettingMode

dali/pipeline/operator/builtin/external_source.h
klecki: I think I prefer to write it twice rather than miss it once.

dali/c_api/c_api.cc
klecki: Done

dali/c_api/c_api.cc
klecki: done

klecki:
!build

szalpal (approved):

JanuszL (approved):

PR #3050, opened 2021-06-14
Remove mixed image decoder from GetBackendTest

JanuszL:
- Jetson platform doesn't support image decoder for the mixed
  backend so replace it with MakeContiguous in GetBackendTest
  test as it needs only any mixed operator

#### Why we need this PR?
*Pick one, remove the rest*
- It removes mixed image decoder from GetBackendTest

 - What solution was applied:
     Jetson platform doesn't support image decoder for the mixed backend so replace it with MakeContiguous in GetBackendTest test as it needs only any mixed operator
     c_api_test
     NA
 - Validation and testing:
     test on Xavier platform
     NA

**JIRA TASK**: *[NA]*

### DIFF
--- a/dali/c_api/c_api_test.cc
+++ b/dali/c_api/c_api_test.cc
@@ -757,21 +757,21 @@ TEST(CApiTest, GetBackendTest) {
                           .AddArg("device", "cpu")
                           .AddArg("name", es_cpu_name)
                           .AddOutput(es_cpu_name, "cpu"), es_cpu_name);
-  std::string cnt_name = "decoder";
+  std::string cont_name = "contiguous";
   pipe.AddOperator(OpSpec("MakeContiguous")
                           .AddArg("device", "mixed")
-                          .AddArg("name", cnt_name)
+                          .AddArg("name", cont_name)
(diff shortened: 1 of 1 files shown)


banasraf (approved):

dali/c_api/c_api_test.cc
```diff
@@ -757,21 +757,21 @@ TEST(CApiTest, GetBackendTest) {
                           .AddArg("device", "cpu")
                           .AddArg("name", es_cpu_name)
                           .AddOutput(es_cpu_name, "cpu"), es_cpu_name);
-  std::string decoder_name = "decoder";
-  pipe.AddOperator(OpSpec("decoders__Image")
+  std::string cnt_name = "decoder";
```
mzient, line 760: Nitpick:
```suggestion
  std::string cont_name = "contiguous";
```
It's no longer a "decoder" and `cnt` sounds a bit too much like `count`.

JanuszL: Done

mzient (approved):

PR #3055, opened 2021-06-15
Make the Python test utilities have local random state

klecki:
Python random data generators in test utils
utilized global random state and even influenced the seed.

Make them not interfere with global state.

#### Why we need this PR?
Fix tests accessing global seed.

 - What solution was applied:
     Use local rng.
     Python test utils, most of the Python tests
     Is this ok?
 - Validation and testing:
     Let's see in CI
     No.


klecki:
!build

mzient (approved):
As long as it doesn't break tests...

JanuszL (approved):

PR #3057, opened 2021-06-16
Add warning about not using external_inputs in proto

klecki:
The use of this field is deprecated.

#### Why we need this PR?
Add a comment about no longer used filed in proto

 - What solution was applied:
     Comment added
     DALI proto (but not really)
     Comment
 - Validation and testing:
     CI
     Sort of

### DIFF
--- a/dali/pipeline/proto/dali.proto
+++ b/dali/pipeline/proto/dali.proto
@@ -1,4 +1,4 @@
-// Copyright (c) 2017-2018, NVIDIA CORPORATION. All rights reserved.
+// Copyright (c) 2017-2021, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


dali/pipeline/proto/dali.proto
```diff
@@ -1,4 +1,4 @@
-// Copyright (c) 2017-2018, NVIDIA CORPORATION. All rights reserved.
+// Copyright (c) 2017-2021, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
banasraf, line 1: I think it's rather `2017, 2021`

szalpal (approved):
👍🏻

dali/pipeline/proto/dali.proto
klecki: If anything it should be 2017-2018, 2021, but my script doesn't handle that.

banasraf: Okay, let's not complicate it. 👍

banasraf (approved):

klecki:
!build

JanuszL (approved):

klecki:
!build

klecki:
!build

PR #3058, opened 2021-06-16
Enable no_copy mode handling in TF DALI Dataset

klecki:

#### Why we need this PR?
Improves the TF DALI Dataset

 - What solution was applied:
     Detect device of external source node
and force copy or no-copy when feeding
that external source with input.
It can only be done for matching input
and external source placement.

Remove unused pipeline_handle_ member
     TF DALI Dataset
     Not sure about the docstring - is it needed?
 - Validation and testing:
     Tests extended, CI
     I described the behaviour in the docstring, I'm not sure if the user needs that much information.

**JIRA TASK**: *[Use DALI-2151 or NA]*

### DIFF
--- a/dali/python/nvidia/dali/plugin/tf.py
+++ b/dali/python/nvidia/dali/plugin/tf.py
@@ -54,6 +54,11 @@ but not a different number of dimensions.
 .. warning::
     This class is experimental and its API might change without notice.
 
+.. note::
+    Setting ``no_copy`` on the external source nodes when defining the pipeline is considered
+    a no-op when used with DALI Dataset. The ``no_copy`` option is handled internally
+    and enabled automatically if possible.
+
 The operator adds additional parameters to the ones supported by the
 :class:`~nvidia.dali.plugin.tf.DALIDataset`:
 
--- a/dali/python/nvidia/dali/external_source.py
+++ b/dali/python/nvidia/dali/external_source.py
@@ -288,8 +288,10 @@ from Python by several methods.
 
 The simplest and preferred way is to specify a ``source``, which can be a callable or iterable.
 
-.. warning::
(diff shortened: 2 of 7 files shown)


dali/python/nvidia/dali/plugin/tf.py
```diff
@@ -54,6 +54,11 @@
 .. warning::
     This class is experimental and its API might change without notice.
 
+.. note::
+    Setting ``no_copy`` on the external source nodes when defining the pipeline is considered
```
klecki, line 58: Is this level of detail OK or too much?

klecki:
!build

dali/python/nvidia/dali/plugin/tf.py
JanuszL: Do we need this option if it is `enabled automatically if possible`?

klecki:
!build

dali/python/nvidia/dali/plugin/tf.py
klecki: This option is present in the external_source operator. I really can't prohibit user from setting it before he passes the Pipeline to the Dataset, so they might think, that setting it will do something (make stuff faster?).

JanuszL: Make sense.

JanuszL (approved):

dali/python/nvidia/dali/plugin/tf.py
mzient: I think this information should be added to the documentation of `no_copy` argument in `ExternalSource`.

mzient (approved):
I'd prefer to see some notices about no_copy having no effect on TF Dataset in the documentation of `no_copy` in ExternalSource, but otherwise OK.

klecki:
!build

PR #3064, opened 2021-06-16
CUDA Virtual Memory API wrappers.

mzient:

#### Why we need this PR?
*Pick one, remove the rest*
- It adds new feature needed for device memory resources that can avoid VA fragmentation

 - What solution was applied:
     * Added C++ wrappers for VM handles and functions
     * Added unit test
     * Add cuMem* functions to dynlink_cuda
     * dynlink_cuda
     * N/A
 - Validation and testing:
     * GTest
     * Doxygen

**JIRA TASK**: DALI-2033


include/dali/core/mm/cu_vm.h
```diff
@@ -0,0 +1,166 @@
...
+};
+
+/**
+ * @brief Manages a backing storage block for a virtual address range
+ */
+class CUMem : public UniqueHandle<CUMemAllocation, CUMem> {
+ public:
+  DALI_INHERIT_UNIQUE_HANDLE(CUMemAllocation, CUMem);
+
+  static void DestroyHandle(CUMemAllocation handle) {
```
JanuszL, line 114: I would move it to the end of the class, to keep it consistent with `CUMemAddressRange`.

include/dali/core/mm/cu_vm.h
```diff
@@ -0,0 +1,166 @@
...
+    CUDA_DTOR_CALL(cuMemRelease(handle.first));
+  }
+
+  static CUMem Create(size_t size, int device_id = -1) {
+    return Create(size, DeviceMemProp(device_id));
+  }
+
+  static CUMem Create(size_t size, const CUmemAllocationProp &prop) {
+    size = align_up(size, GetAddressGranularity());
```
JanuszL, line 123:
```suggestion
    size_t grain = GetAddressGranularity();
    size = align_up(size, grain);
```
Maybe you can even extract it to a function for that as the same pater happens in L66.

dali/core/mm/cu_vm_test.cc
```diff
@@ -0,0 +1,82 @@
...
+  CUpointer_attribute attrs[3] = {
+    CU_POINTER_ATTRIBUTE_RANGE_START_ADDR,
+    CU_POINTER_ATTRIBUTE_RANGE_SIZE,
+    CU_POINTER_ATTRIBUTE_MAPPED
+  };
+  CUdeviceptr start = 0;
+  size_t size = 0;
+  bool mapped = false;
+  void *data[3] = { &start, &size, &mapped };
+  CUDA_CALL(cuPointerGetAttributes(3, attrs, data, range.ptr() + 100));
```
JanuszL, line 38: Why you take the address from somewhere inside the range?

dali/core/mm/cu_vm_test.cc
```diff
@@ -0,0 +1,82 @@
...
+#include <gtest/gtest.h>
+
+#if DALI_USE_CUDA_VM_MAP
+
+namespace dali {
+namespace mm {
+namespace test {
+
+TEST(CUMemAddressRange, Reserve) {
+  ASSERT_TRUE(cuInitChecked());
```
JanuszL, line 25: Do you need to call it? Wouldn't other functions do that?

dali/core/mm/cu_vm_test.cc
```diff
@@ -0,0 +1,82 @@
...
+  void *data[3] = { &start, &size, &mapped };
+  CUDA_CALL(cuPointerGetAttributes(3, attrs, data, range.ptr() + 100));
+  EXPECT_EQ(start, range.ptr());
+  EXPECT_EQ(size, range.size());
+  EXPECT_FALSE(mapped);
+  EXPECT_NO_THROW(range.reset());
+}
+
+TEST(CUMemAddressRange, ReserveAndMap) {
+  ASSERT_TRUE(cuInitChecked());
```
JanuszL, line 46: Do you need to call it? Wouldn't other functions do that?

dali/core/mm/cu_vm_test.cc
```diff
@@ -0,0 +1,82 @@
...
+  CUpointer_attribute attrs[3] = {
+    CU_POINTER_ATTRIBUTE_RANGE_START_ADDR,
+    CU_POINTER_ATTRIBUTE_RANGE_SIZE,
+    CU_POINTER_ATTRIBUTE_MAPPED
+  };
+  CUdeviceptr start = 0;
+  size_t size = 0;
+  bool mapped = false;
+  void *data[3] = { &start, &size, &mapped };
+  void *ptr = cuvm::Map(base, mem);
```
JanuszL, line 62: DO you expect to pass any address from the base to base+virt_size or `cuvm::Map` should just accept `cuvm::CUMemAddressRange` instead of a raw pointer?

dali/core/mm/cu_vm_test.cc
```diff
@@ -0,0 +1,82 @@
...
+  EXPECT_EQ(ptr, reinterpret_cast<void*>(base));
+  CUDA_CALL(cuPointerGetAttributes(3, attrs, data, base + 100));
+  EXPECT_EQ(start, range.ptr());
+  EXPECT_EQ(size, range.size());
+  EXPECT_TRUE(mapped);
```
JanuszL, line 67: In this case virtual size is bigger than the physical one? Is that possible when one is mapped to the other?

include/dali/core/mm/cu_vm.h
```diff
@@ -0,0 +1,166 @@
...
+#include "dali/core/mm/detail/align.h"
+
+namespace dali {
+namespace mm {
+namespace cuvm {
+
+/**
+ * @brief A virtual address range
+ */
+struct CUAddressRange : std::pair<CUdeviceptr, size_t> {
```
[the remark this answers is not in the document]
klecki: Do you use anything out of the pair to inherit from it? or coud this just be a
`struct { CUdeviceptr ptr; size_t size;};`

include/dali/core/mm/cu_vm.h
```diff
@@ -0,0 +1,166 @@
...
+  if (device_id < 0) {
+    CUDA_CALL(cudaGetDevice(&device_id));
+  }
+  prop.type = CU_MEM_ALLOCATION_TYPE_PINNED;
+  prop.location.type = CU_MEM_LOCATION_TYPE_DEVICE;
+  prop.location.id = device_id;
+  return prop;
+}
+
+struct CUMemAllocation : std::pair<CUmemGenericAllocationHandle, size_t> {
```
klecki, line 99: same question about pair here.

dali/core/mm/cu_vm_test.cc
mzient: Yes, that's the whole point. You allocate a huge swathe of virtual memory and map only parts of it as you see fit.

dali/core/mm/cu_vm_test.cc
mzient: No, you can map any (properly aligned) part of the VA range.

dali/core/mm/cu_vm_test.cc
mzient: Because I want to check that the VA machinery worked properly and I'm not getting the base address by sheer coincidence.

include/dali/core/mm/cu_vm.h
mzient: I'll consider that when there are at least 3 usages ;) Besides, according to the discussion on CUDA channel, this is a bit coincidental - the two granularities don't need to be equal.

include/dali/core/mm/cu_vm.h
```diff
@@ -0,0 +1,166 @@
...
+    return CUMem({ handle, size });
+  }
+
+  CUmemGenericAllocationHandle handle() const noexcept { return handle_.first; }
+
+  size_t size() const noexcept { return handle_.second; }
+};
+
+template <typename T = void>
+T *Map(CUdeviceptr virt_addr, CUmemGenericAllocationHandle mem, size_t size, int device_id = -1) {
```
klecki, line 135: As we already have the wrapper types with sizes, do you think a check for that could be in place?

include/dali/core/mm/cu_vm.h
```diff
@@ -0,0 +1,166 @@
...
+  CUmemAccessDesc access = {};
+  access.location.type = CU_MEM_LOCATION_TYPE_DEVICE;
+  access.location.id = device_id;
+  access.flags = CU_MEM_ACCESS_FLAGS_PROT_READWRITE;
+  CUDA_CALL(cuMemSetAccess(virt_addr, size, &access, 1));
+  return reinterpret_cast<T*>(virt_addr);
+}
+
+template <typename T = void>
+T *Map(CUdeviceptr virt_addr, CUMemAllocation mem, int device_id = -1) {
```
klecki, line 149: Why not the CUMemAddressRange instead of CUdeviceptr, and the size from the address range instead of the `mem`? The docs says that the size is taken from the address range. Can we map it not fully?

dali/core/mm/cu_vm_test.cc
```diff
@@ -0,0 +1,82 @@
...
+  CUDA_CALL(cuPointerGetAttributes(3, attrs, data, range.ptr() + 100));
+  EXPECT_EQ(start, range.ptr());
+  EXPECT_EQ(size, range.size());
+  EXPECT_FALSE(mapped);
+  EXPECT_NO_THROW(range.reset());
+}
+
+TEST(CUMemAddressRange, ReserveAndMap) {
+  ASSERT_TRUE(cuInitChecked());
+  int64_t virt_size = 10<<20;
```
klecki, line 47: Asking this again, but looking at the doc of cuMemMap, it seems like the mapped range `[addr, addr + size]` should come from the cuMemAddressReserve (I mean the virt range and should match what we are mapping, but the physicall mapping can be bigger), but we are using a bigger range to map it to a smaller physical one (so the opposite). Is it ok to do it?

klecki:
Some comments and questions.

klecki (commented):
Didn't publish the first time.

include/dali/core/mm/cu_vm.h
mzient: The documentation is unclear - and it seems that  you misunderstood it.
The range being mapped must be _within_ a range reserved with `cuMemAddressReserve`. The size must be equal to the size of the _physical allocation_ (the `CUmemGenericAllocationHandle`), not the virtual address range. The whole point of this API is that the VA range can be much larger than physical backing storage - perhaps many times larger.

dali/core/mm/cu_vm_test.cc
mzient: It's OK and it's the very purpose of this API.

include/dali/core/mm/cu_vm.h
mzient: We can't check it _here_, can we? This wrapper merely combines cuMemMap with cuMemSetAccess and adds error handling.

include/dali/core/mm/cu_vm.h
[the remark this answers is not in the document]
mzient: comparison

dali/core/mm/cu_vm_test.cc
mzient: Well, I can remove it to check that the functions actually load - but it wouldn't matter unless running a single test.

klecki (approved):
Please add some docstrings for the key functions:
* Reserve
* Create
* Map
* Unmap

include/dali/core/mm/cu_vm.h
```diff
@@ -61,13 +61,40 @@ class CUMemAddressRange : public UniqueHandle<CUAddressRange, CUMemAddressRange>
...
+   * it is guaranteed that physical allocations can be mapped to addresses from this range
+   * and that no other allocation functions (like cudaMalloc) will return addresses from this range.
+   *
+   * The addresses can be made usable by mapping phycial storage to them with cuvm::Map or
+   * cuMemMap + cuMemSetAccess.
+   *
+   * @param size      Size, in bytes - should be a multiple of GetAddressGranularity - if it's not
+   *                  it is rounded up to the next multiple
+   * @param alignment Mostly ignored
+   * @param start     The requested starting address - a hint for a starting dddress of the range.
```
klecki, line 93: dddress -> address

klecki (approved):

dali/core/mm/cu_vm_test.cc
JanuszL: ok

dali/core/mm/cu_vm_test.cc
JanuszL: ok

dali/core/mm/cu_vm_test.cc
JanuszL: ok

include/dali/core/mm/cu_vm.h
JanuszL: Amway I would make the code match L66.

JanuszL (approved):

PR #3069, opened 2021-06-18
Fix error handling in Executor

mzient:

#### Why we need this PR?
*Pick one, remove the rest*
- It fixes a bug: race condition causing a spurious "unknown error" message when a known error occurs.
 
 - What solution was applied:
      * swap error message insertion and error notification
      * produce different message for pure shutdown and uknown error - Affected modules and functionalities:
     * N/A
 - Validation and testing:
     * Multipaste test covers it; no reliable test possible
     * N/A

**JIRA TASK**: DALI-2156

### DIFF
--- a/dali/pipeline/executor/executor.h
+++ b/dali/pipeline/executor/executor.h
@@ -1,4 +1,4 @@
-// Copyright (c) 2017-2018, NVIDIA CORPORATION. All rights reserved.
+// Copyright (c) 2017-2021, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -236,10 +236,12 @@ class DLL_PUBLIC Executor : public ExecutorBase, public WorkspacePolicy, public 
   }
 
   void HandleError(const std::string& message = "Unknown exception") {
+    {
+      std::lock_guard<std::mutex> errors_lock(errors_mutex_);
+      errors_.push_back(message);
+    }
     exec_error_ = true;
     ShutdownQueue();
-    std::lock_guard<std::mutex> errors_lock(errors_mutex_);
-    errors_.push_back(message);
   }
 
   void PruneUnusedGraphNodes() override;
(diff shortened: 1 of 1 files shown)


mzient:
!build

dali/pipeline/executor/executor.h
```diff
@@ -236,10 +236,12 @@ class DLL_PUBLIC Executor : public ExecutorBase, public WorkspacePolicy, public
   }
 
   void HandleError(const std::string& message = "Unknown exception") {
+    {
+      std::lock_guard<std::mutex> errors_lock(errors_mutex_);
+      errors_.push_back(message);
```
mzient, line 241: This is the fix - set message first, then notify.

dali/pipeline/executor/executor.h
```diff
@@ -478,15 +480,19 @@ void Executor<WorkspacePolicy, QueuePolicy>::ShareOutputs(DeviceWorkspace *ws) {
   if (exec_error_ || QueuePolicy::IsStopSignaled()) {
     // TODO(klecki) collect all errors
     std::lock_guard<std::mutex> errors_lock(errors_mutex_);
-    std::string error = errors_.empty() ? "Unknown error" : errors_.front();
+    std::string error = errors_.empty() ? QueuePolicy::IsStopSignaled() && !exec_error_
+                                            ? "Stop signaled" : "Unknown error"
+                                        : errors_.front();
```
klecki, line 485: I was never a fan of nested ternary operators, I would pack this up and use twice:

```
std::string GetErrorMessageHelper() {
  if (!errors_.empty()) { 
    return errors_.front();
  }
  if (QueuePolicy::IsStopSignaled() && !exec_error_) {
    return "Stop signaled";
  }
  return "Unknown error";
}
```

mzient: I'm all for extracting it to a function, but ternary operators vs if/else is a matter of personal taste.

JanuszL (approved):

mzient:
!build

klecki (approved):

PR #3083, opened 2021-06-23
Bit mask utility

mzient:
#### Why we need this PR?
*Pick one, remove the rest*
- It adds new feature: a bit mask for extremely fast searching for set/cleared bits.
     * the search is 2 orders of magnitude faster than std::find on a `vector<bool>` or `vector<char>`

 - What solution was applied:
     * Added a `bitmask` collection, somewhat similar to vector<bool>
     * Use a bulk comparison for consecutive bits that fit into a single machine word
     * core/util
     * ?
 - Validation and testing:
     * Unit tests
     * Doxygen

**JIRA TASK**: N/A

### DIFF
--- a/include/dali/core/bitmask.h
+++ b/include/dali/core/bitmask.h
@@ -31,6 +31,11 @@ struct bitmask {
   static constexpr const int storage_bits = sizeof(bit_storage_t) * 8;
   static constexpr const int storage_bits_log = ilog2(storage_bits);
 
+  /**
+   * @brief Find the first bit with given value, starting at start
+   *
+   * @return The index of the first bit found of the size of the bit mask if not found.
+   */
   ptrdiff_t find(bool value, ptrdiff_t start = 0) const {
     assert(start >= 0);
     if (start >= size_)
@@ -61,6 +66,9 @@ struct bitmask {
     return size_;
   }
 
+  /**
+   * @brief Fill a range of bits with given value
+   */
   void fill(ptrdiff_t start, ptrdiff_t end, bool value) {
     if (start >= end)
       return;
@@ -98,12 +106,18 @@ struct bitmask {
     }
   }
 
+  /**
+   * @brief Fill the entire mask with given value.
+   */
   void fill(bool value) {
     size_t s = size();
     clear();
     resize(s, value);
   }
 
+  /**
+   * @brief A pseudoreference to a single bit in the mask
+   */
   struct bitref {
     bitref &operator=(bool value) {
       if (value)
@@ -208,6 +222,7 @@ struct bitmask {
  private:
   // vector of bit storage words - the bits that are outside of the mask are 0
   std::vector<bit_storage_t> storage_;
+  // number of bits in the mask
   ptrdiff_t size_ = 0;
 };
 
--- a/include/dali/core/bitmask.h
+++ b/include/dali/core/bitmask.h
@@ -24,7 +24,7 @@ namespace dali {
 /**
  * @brief A vector of bits with a utility for quickly searching for set/cleared bits.
  */
-struct bitmask {
+class bitmask {
  public:
   using bit_storage_t = uint64_t;
 
--- a/include/dali/core/bitmask.h
+++ b/include/dali/core/bitmask.h
@@ -31,17 +31,25 @@ class bitmask {
   static constexpr const int storage_bits = sizeof(bit_storage_t) * 8;
   static constexpr const int storage_bits_log = ilog2(storage_bits);
 
+  static constexpr int bit_idx(ptrdiff_t idx) {
+    return idx & (storage_bits - 1);
+  }
+
+  static constexpr int word_idx(ptrdiff_t idx) {
+    return idx >> storage_bits_log;
+  }
+
   /**
    * @brief Find the first bit with given value, starting at start
    *
-   * @return The index of the first bit found of the size of the bit mask if not found.
+   * @return The index of the first bit found or the size of the bit mask if not found.
    */
   ptrdiff_t find(bool value, ptrdiff_t start = 0) const {
     assert(start >= 0);
     if (start >= size_)
       return size_;
-    auto index = start >> storage_bits_log;
-    int bit = start & (storage_bits - 1);
+    auto index = word_idx(start);
+    int bit = bit_idx(start);
     bit_storage_t flip = 0;
     if (!value)
       flip = ~flip;
@@ -72,16 +80,16 @@ class bitmask {
   void fill(ptrdiff_t start, ptrdiff_t end, bool value) {
     if (start >= end)
       return;
-    ptrdiff_t start_idx = start >> storage_bits_log;
-    ptrdiff_t end_idx = end >> storage_bits_log;
+    ptrdiff_t start_idx = word_idx(start);
+    ptrdiff_t end_idx = word_idx(end);
     bit_storage_t ones = ~bit_storage_t(0);
     bit_storage_t start_mask = ones;
     bit_storage_t end_mask = 0;
-    if (start & (storage_bits - 1)) {
-        start_mask <<= (start & (storage_bits - 1));
+    if (bit_idx(start)) {
+        start_mask <<= bit_idx(start);
     }
-    if (end & (storage_bits - 1)) {
-        end_mask = ones >> (storage_bits - (end & (storage_bits - 1)));
+    if (bit_idx(end)) {
+        end_mask = ones >> (storage_bits - bit_idx(end));
     }
     if (start_idx == end_idx) {
         start_mask &= end_mask;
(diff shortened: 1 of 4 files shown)


include/dali/core/bitmask.h
```diff
@@ -0,0 +1,231 @@
...
+
+  bool operator[](const ptrdiff_t index) const {
+    bit_storage_t storage = storage_[index >> storage_bits_log];
+    bit_storage_t mask = bit_storage_t(1) << (index & (storage_bits - 1));
+    return storage & mask;
+  }
+
+ private:
+  // vector of bit storage words - the bits that are outside of the mask are 0
+  std::vector<bit_storage_t> storage_;
```
mzient, line 224: To consider: use SmallVector?

include/dali/core/bitmask.h
```diff
@@ -0,0 +1,231 @@
...
+  void fill(bool value) {
+    size_t s = size();
+    clear();
+    resize(s, value);
+  }
+
+  /**
+   * @brief A pseudoreference to a single bit in the mask
+   */
+  struct bitref {
```
mzient, line 121: Note to reviewers: there are no iterators and that's intentional. There are few algorithms that can be used on a collection of booleans and those taken from STL would perform poorly.

mzient:
!build

include/dali/core/util.h
```diff
@@ -143,6 +143,52 @@ constexpr std::enable_if_t<std::is_integral<T>::value, int> ilog2(T x) {
...
+std::enable_if_t<std::is_integral<Integer>::value, int> ctz(Integer word) {
+  std::make_unsigned_t<Integer> uword = word;
+  if (uword == 0)
+    return sizeof(uword) * 8;
+  // The hardware can check for multiple bits at a time, so we don't have to scan sequentially
+  int bit = 0;
+  if (sizeof(uword) > 4 && (uword & 0xffffffffu) == 0) {
+    bit += 32;
+    const int shift32 = sizeof(uword) > 4 ? 32 : 0;  // workaround undefined shift warning
+    uword >>= shift32;
```
JanuszL, line 164:
```suggestion
    const int shift32 = 32;  // workaround undefined shift warning
    uword >>= shift32;
```
The above if already guarantees that.

JanuszL: Or this is just for the warning?

mzient: Yes, it's for the compiler warning.

include/dali/core/bitmask.h
```diff
@@ -0,0 +1,231 @@
...
+ public:
+  using bit_storage_t = uint64_t;
+
+  static constexpr const int storage_bits = sizeof(bit_storage_t) * 8;
+  static constexpr const int storage_bits_log = ilog2(storage_bits);
+
+  /**
+   * @brief Find the first bit with given value, starting at start
+   *
+   * @return The index of the first bit found of the size of the bit mask if not found.
```
JanuszL, line 37:
```suggestion
   * @return The index of the first bit found of the size or the bit mask if not found.
```

mzient: Actually, 
```suggestion
   * @return The index of the first bit found or the size of the bit mask if not found.
```

include/dali/core/bitmask.h
```diff
@@ -0,0 +1,231 @@
...
+    }
+
+    return size_;
+  }
+
+  /**
+   * @brief Fill a range of bits with given value
+   */
+  void fill(ptrdiff_t start, ptrdiff_t end, bool value) {
+    if (start >= end)
```
JanuszL, line 73: Maybe it should assert?

include/dali/core/bitmask.h
```diff
@@ -0,0 +1,231 @@
...
+   * @brief Find the first bit with given value, starting at start
+   *
+   * @return The index of the first bit found of the size of the bit mask if not found.
+   */
+  ptrdiff_t find(bool value, ptrdiff_t start = 0) const {
+    assert(start >= 0);
+    if (start >= size_)
+      return size_;
+    auto index = start >> storage_bits_log;
+    int bit = start & (storage_bits - 1);
```
JanuszL, line 44: I wonder if moving `start & (storage_bits - 1)` to an utility function would make the code clearer (it repeats in multiple places)

mzient: Perhaps.

include/dali/core/bitmask.h
```diff
@@ -0,0 +1,231 @@
...
+  }
+
+  bitref operator[](ptrdiff_t index) {
+    return { storage_.data(), index };
+  }
+
+  bool operator[](const ptrdiff_t index) const {
+    bit_storage_t storage = storage_[index >> storage_bits_log];
+    bit_storage_t mask = bit_storage_t(1) << (index & (storage_bits - 1));
+    return storage & mask;
```
JanuszL, line 219: Can't you express this using bitref as well?

include/dali/core/bitmask.h
mzient: Done

include/dali/core/bitmask.h
mzient: I think that keeping swapped boundaries as no-op has some value. But it should assert that start and end are between 0 and size.

include/dali/core/bitmask.h
JanuszL: Similar for `something >> storage_bits_log`.

include/dali/core/bitmask.h
```diff
@@ -0,0 +1,231 @@
...
+  }
+
+  void resize(size_t new_size, bool value = false) {
+    if (new_size == size())
+      return;
+    bit_storage_t ones = ~bit_storage_t(0);
+    bit_storage_t fill = value ? ones : 0;
+    size_t prev_words = storage_.size();
+    storage_.resize((new_size + storage_bits - 1) >> storage_bits_log, fill);
+    if (storage_.size() == prev_words && new_size > size()) {
```
JanuszL, line 204: Is it possible that `storage_.size()` before and after resize is the same?

include/dali/core/bitmask.h
```diff
@@ -0,0 +1,231 @@
...
+    } else {
+        for (; idx < end_idx; idx++)
+            storage_[idx] = 0;
+    }
+    if (end_mask && start_idx != end_idx) {
+      if (value)
+        storage_[end_idx] |= end_mask;
+      else
+        storage_[end_idx] &= ~end_mask;
+    }
```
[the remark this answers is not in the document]
mzient: I've been there, it doens't work :) It's not true when `start_idx == end_idx`, because there's idx++.

include/dali/core/bitmask.h
mzient: It is - when resizing, say, from 13 to 17 bits - it's still just one word.

include/dali/core/bitmask.h
mzient: No, because bitref requires non-const pointer.

include/dali/core/bitmask.h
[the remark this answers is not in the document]
JanuszL: But you check `start_idx != end_idx` in the if.

include/dali/core/bitmask.h
```diff
@@ -0,0 +1,257 @@
...
+
+  static constexpr const int storage_bits = sizeof(bit_storage_t) * 8;
+  static constexpr const int storage_bits_log = ilog2(storage_bits);
+
+  static constexpr int bit_idx(ptrdiff_t idx) {
+    return idx & (storage_bits - 1);
+  }
+
+  static constexpr int word_idx(ptrdiff_t idx) {
+    return idx >> storage_bits_log;
```
JanuszL, line 39: 👍

include/dali/core/bitmask.h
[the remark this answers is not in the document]
mzient: At some point it didn't work. I could check it but I see little value in it.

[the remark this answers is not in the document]
JanuszL: Up to you.

include/dali/core/bitmask.h
JanuszL: Right

include/dali/core/bitmask.h
```diff
@@ -126,14 +134,30 @@ class bitmask {
...
+    bitref &operator&=(bool value) {
+      if (!value)
+        *storage &= ~mask;
+      return *this;
+    }
+    bitref &operator^=(bool value) {
+      if (value)
+        *storage ^= mask;
+      return *this;
+    }
```
JanuszL, line 153: Shouldn't you add a test for it?

JanuszL (approved):

include/dali/core/bitmask.h
mzient: Meh. I could... that's rather trivial, though.

mzient: Done in index test.

JanuszL (approved):

szalpal (approved):

PR #3096, opened 2021-06-28
Bug fixes

cyyever:
#### Why we need this PR?
*Pick one, remove the rest*

- Refactoring to improve code quality and fix some pointer bugs.

 - What solution was applied:
     *Fix some problems found by code analyzer*

**JIRA TASK**: *NA*

### DIFF
--- a/dali/kernels/signal/moving_mean_square_test.cc
+++ b/dali/kernels/signal/moving_mean_square_test.cc
@@ -51,8 +51,8 @@ class MovingMeanSquareCpuTest : public ::testing::Test {
 
  private:
   void calc_output() {
-    ref_output_.resize(buffer_length_ - window_size_);
-    for (int i = 0; i <= buffer_length_ - window_size_; i++) {
+    ref_output_.resize(buffer_length_ - window_size_ + 1);
+    for (int i = 0; i < buffer_length_ - window_size_+ 1; i++) {
       float sumsq = 0;
       for (int j = 0; j < window_size_; j++) {
         auto val = static_cast<float>(input_[i + j]);
--- a/dali/kernels/signal/moving_mean_square_test.cc
+++ b/dali/kernels/signal/moving_mean_square_test.cc
@@ -52,7 +52,7 @@ class MovingMeanSquareCpuTest : public ::testing::Test {
  private:
   void calc_output() {
     ref_output_.resize(buffer_length_ - window_size_ + 1);
-    for (int i = 0; i < buffer_length_ - window_size_+ 1; i++) {
+    for (int i = 0; i < buffer_length_ - window_size_ + 1; i++) {
       float sumsq = 0;
       for (int j = 0; j < window_size_; j++) {
         auto val = static_cast<float>(input_[i + j]);
@@ -113,4 +113,3 @@ TYPED_TEST(MovingMeanSquareCpuTest, RunTest) {
 }  // namespace test
 }  // namespace kernels
 }  // namespace dali
-
--- a/dali/kernels/test/simd_test.cc
+++ b/dali/kernels/test/simd_test.cc
@@ -49,8 +49,9 @@ template <typename In, int nvec = sizeof(float)/sizeof(In)>
 void TestConvertLoad(In lo, In hi) {
   int lanes = 16 / sizeof(In);
   In in[lanes];  // NOLINT
-  for (int i = 0; i < lanes; i++) {
-    in[i] = int64_t(lo) + (int64_t(hi) - int64_t(lo)) * i / (lanes - 1);
+  // promote to int64_t computation to avoid integer overflow
+  for (int64_t i = 0; i < lanes; i++) {
+    in[i] = lo + (hi - lo) * i / (lanes - 1);
   }
   float4x<nvec> v = load_f(in);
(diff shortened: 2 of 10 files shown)


dali/kernels/signal/moving_mean_square_test.cc
```diff
@@ -51,8 +51,8 @@ class MovingMeanSquareCpuTest : public ::testing::Test {
 
  private:
   void calc_output() {
-    ref_output_.resize(buffer_length_ - window_size_);
-    for (int i = 0; i <= buffer_length_ - window_size_; i++) {
+    ref_output_.resize(buffer_length_ - window_size_ + 1);
+    for (int i = 0; i < buffer_length_ - window_size_+ 1; i++) {
```
mzient, line 55: Nitpick: 
```suggestion
    for (int i = 0; i < buffer_length_ - window_size_ + 1; i++) {
```

mzient:
!build

mzient (approved):

include/dali/core/small_vector.h
```diff
@@ -248,6 +248,10 @@ class SmallVector : SmallVectorAlloc<T, allocator>, SmallVectorBase<T> {
   DALI_NO_EXEC_CHECK
   DALI_HOST_DEV void copy_assign(const T *data, size_t count) {
     clear();
+    if (!data) {
+      assert(count == 0);
+      return;
+    }
```
mzient, line 254: This is not necessary and `assert` doesn't come for free in CUDA code (it significantly increases register pressure) and SmallVector can (at least theoretically) be used in device-side code.
Also, the logic is somewhat broken in non-debug builds - there's no assert and the function will quit silently instead of reproducibly causing an access violation and terminating. This might expose the calling code to security threats by hiding the situation where the vector contents were expected to be replaced but instead remained unchanged due to null source pointer.
Please revert this change.

mzient (changes requested):
Please revert the change in `small_vector.h` - it constitutes a security threat.

dali/kernels/test/simd_test.cc
```diff
@@ -50,7 +50,7 @@ void TestConvertLoad(In lo, In hi) {
   int lanes = 16 / sizeof(In);
   In in[lanes];  // NOLINT
   for (int i = 0; i < lanes; i++) {
-    in[i] = lo + (hi - lo) * i / (lanes - 1);
+    in[i] = int64_t(lo) + (int64_t(hi) - int64_t(lo)) * i / (lanes - 1);
```
mzient, line 53: This function can (at least theoretically) be used with floating point input. It's better to use `int64_t i` which will promote the multiplication to int64 for smaller types but maintain original floating point and signedness.
```suggestion
  for (int64_t i = 0; i < lanes; i++) {
    in[i] = lo + (hi - lo) * i / (lanes - 1);
```

include/dali/core/small_vector.h
cyyever: A test passes null  begin and end pointers to SmallVector. So we should check it to avoid calling memcpy with null pointers.

include/dali/core/small_vector.h
```diff
@@ -248,6 +248,9 @@ class SmallVector : SmallVectorAlloc<T, allocator>, SmallVectorBase<T> {
   DALI_NO_EXEC_CHECK
   DALI_HOST_DEV void copy_assign(const T *data, size_t count) {
     clear();
+    if (!data) {
+      return;
+    }
```
mzient, line 253: This is still broken. If the caller passes a null pointer and nonzero count, then the function is expected to reliably crash (however that sounds) rather than return quitely dismiss the error. This causes unexpected internal state:
- the data buffer is not changed
- the size is 0 instead of expected `count`
The subsequent code might attempt to access up to `count` elements which have unexpected values. This is far worse than a null pointer dereference which would simply kill the program.

mzient: If you insist on having that check (which is superfluous), a viable alternative would be:
```suggestion
    if (!count)
      return;
#ifndef __CUDA_ARCH
      assert(data != nullptr);
#endif
```
(and remember to add `#include <cassert>` before `<utility>`)

cyyever: In face, I think this is a bad interface by design, we should accept a reference  and a size_t. We shouldn't care a null pointer in such a low level method. If the constructor or any other caller  called it, than it would have to throw an exception or simply crash.

mzient: Let's not go this far in this PR. Also, it mimics what you can do in `std::string` and discussing whether or not std::string is a good design is a bit out of scope here.

cyyever: Ok, let's revert it.

mzient (approved):

mzient:
!build

klecki (approved):
Hi @cyyever, thanks for the contribution. Would you mind sharing what code analyzer did you use to find those?

cyyever:
> Hi @cyyever, thanks for the contribution. Would you mind sharing what code analyzer did you use to find those?

I mainly use [pvs-studio](https://pvs-studio.com/en/) and clang-tidy, and since it core dumped when I run the test, I re-compiled the source code with ASAN and UBSAN to find  the bugs.

PR #3097, opened 2021-06-28
Fix override no copy flags for External Source C API

klecki:
#### Why we need this PR?
Fix a bug

 - What solution was applied:

GPU op used only the static flag when accessing the data.

External Source state was extended with flag
indicating no_copy value for given batch.

Test was added to run both override flags.

External Source

N/A

 - Validation and testing:

Test was extended to cover this codepath.

A bit of docstrings.


klecki:
!build

JanuszL (approved):

mzient (approved):

PR #3108, opened 2021-06-30
minor  fixes

cyyever:
#### Why we need this PR?

- Refactoring to improve *performance*

 - What solution was applied:
     *refactor code to improve performance*
 

**JIRA TASK**: *NA*

### DIFF
--- a/dali/kernels/scratch.h
+++ b/dali/kernels/scratch.h
@@ -167,7 +167,7 @@ class ScratchpadAllocator {
    *
    * See `Reserve(AllocType, size_t)` for details.
    */
-  void Reserve(std::array<size_t, NumAllocTypes> sizes) {
+  void Reserve(const std::array<size_t, NumAllocTypes> &sizes) {
     for (size_t idx = 0; idx < NumAllocTypes; idx++) {
       Reserve(AllocType(idx), sizes[idx]);
     }
--- a/dali/kernels/scratch.h
+++ b/dali/kernels/scratch.h
@@ -167,7 +167,7 @@ class ScratchpadAllocator {
    *
    * See `Reserve(AllocType, size_t)` for details.
    */
-  void Reserve(const std::array<size_t, NumAllocTypes> &sizes) {
+  void Reserve(std::array<size_t, NumAllocTypes> sizes) {
     for (size_t idx = 0; idx < NumAllocTypes; idx++) {
       Reserve(AllocType(idx), sizes[idx]);
     }
--- a/dali/operators/generic/pad.h
+++ b/dali/operators/generic/pad.h
@@ -100,7 +100,7 @@ class Pad : public Operator<Backend> {
         }
       }
       auto shape = shape_.tensor_shape_span(i);
-      DALI_ENFORCE(axes_.size() == shape.size(),
+      DALI_ENFORCE(static_cast<int>(axes_.size()) == shape.size(),
           make_string(
               "If explicit shape is provided, there should be a value per axis to be padded. "
               "Expected ",
--- a/dali/operators/generic/pad.h
+++ b/dali/operators/generic/pad.h
@@ -100,7 +100,7 @@ class Pad : public Operator<Backend> {
         }
       }
       auto shape = shape_.tensor_shape_span(i);
-      DALI_ENFORCE(static_cast<int>(axes_.size()) == shape.size(),
+      DALI_ENFORCE(axes_.size() == shape.size(),
           make_string(
               "If explicit shape is provided, there should be a value per axis to be padded. "
               "Expected ",
--- a/dali/operators/math/expressions/arithmetic.h
+++ b/dali/operators/math/expressions/arithmetic.h
@@ -155,7 +155,7 @@ inline std::vector<ExprImplTask> CreateExecutionTasks(const ExprNode &expr, Expr
   return result;
 }
 
-inline TensorListShape<> ShapePromotion(std::string op, span<const TensorListShape<> *> shapes,
+inline TensorListShape<> ShapePromotion(const std::string &op, span<const TensorListShape<> *> shapes,
                                         int batch_size) {
   const TensorListShape<> *out_shape = nullptr;
   bool only_scalars = true;
--- a/dali/operators/math/expressions/arithmetic.h
+++ b/dali/operators/math/expressions/arithmetic.h
@@ -155,7 +155,8 @@ inline std::vector<ExprImplTask> CreateExecutionTasks(const ExprNode &expr, Expr
   return result;
 }
 
-inline TensorListShape<> ShapePromotion(const std::string &op, span<const TensorListShape<> *> shapes,
+inline TensorListShape<> ShapePromotion(const std::string &op,
+                                        span<const TensorListShape<> *> shapes,
                                         int batch_size) {
   const TensorListShape<> *out_shape = nullptr;
   bool only_scalars = true;
--- a/dali/operators/reader/loader/recordio_loader.h
+++ b/dali/operators/reader/loader/recordio_loader.h
@@ -73,7 +73,7 @@ class RecordIOLoader : public IndexedFileLoader {
     // skip 0 sized images
     if (size) {
       indices_.emplace_back(temp.back() - file_offsets[file_offset_index],
-          size, file_offset_index);
+                            size, file_offset_index);
     }
     index_file.close();
   }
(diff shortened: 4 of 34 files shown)


dali/kernels/scratch.h
```diff
@@ -167,7 +167,7 @@ class ScratchpadAllocator {
    *
    * See `Reserve(AllocType, size_t)` for details.
    */
-  void Reserve(std::array<size_t, NumAllocTypes> sizes) {
+  void Reserve(const std::array<size_t, NumAllocTypes> &sizes) {
```
mzient, line 170: This array is very small. Passing it by value was intentional.

dali/operators/generic/pad.h
```diff
@@ -100,7 +100,7 @@ class Pad : public Operator<Backend> {
         }
       }
       auto shape = shape_.tensor_shape_span(i);
-      DALI_ENFORCE(static_cast<int>(axes_.size()) == shape.size(),
+      DALI_ENFORCE(axes_.size() == shape.size(),
```
mzient, line 103: This will result in signed/unsigned comparison warning and will break our build. Please revert.

dali/operators/reader/loader/recordio_loader.h
```diff
@@ -65,15 +65,15 @@ class RecordIOLoader : public IndexedFileLoader {
       int64 size = temp[i + 1] - temp[i];
       // skip 0 sized images
       if (size) {
-        indices_.push_back(std::make_tuple(temp[i] - file_offsets[file_offset_index],
-                                          size, file_offset_index));
+        indices_.emplace_back(temp[i] - file_offsets[file_offset_index],
+            size, file_offset_index);
```
mzient, line 69: Nitpick:
```suggestion
        indices_.emplace_back(temp[i] - file_offsets[file_offset_index],
                                    size, file_offset_index);
```

dali/operators/reader/loader/recordio_loader.h
```diff
@@ -65,15 +65,15 @@ class RecordIOLoader : public IndexedFileLoader {
...
       }
     }
     int64 size = file_offsets.back() - temp.back();
     // skip 0 sized images
     if (size) {
-      indices_.push_back(std::make_tuple(temp.back() - file_offsets[file_offset_index],
-                                        size, file_offset_index));
+      indices_.emplace_back(temp.back() - file_offsets[file_offset_index],
+          size, file_offset_index);
```
mzient, line 76:
```suggestion
      indices_.emplace_back(temp.back() - file_offsets[file_offset_index],
                              size, file_offset_index);
```

dali/operators/ssd/random_crop.h
```diff
@@ -38,13 +38,13 @@ class SSDRandomCrop : public Operator<Backend> {
...
-    sample_options_.push_back(SampleOption{false, 0.7});
-    sample_options_.push_back(SampleOption{false, 0.9});
-    sample_options_.push_back(SampleOption{true, 0});
+    sample_options_.emplace_back(false, -1.f);
+    sample_options_.emplace_back(false, 0.1);
+    sample_options_.emplace_back(false, 0.3);
+    sample_options_.emplace_back(false, 0.5);
+    sample_options_.emplace_back(false, 0.7);
+    sample_options_.emplace_back(false, 0.9);
+    sample_options_.emplace_back(true, 0);
```
mzient, line 47: I think the code is equivalent (technically there's an additional move, but the object is small and trivially copyable/movable) and with the old code it was obvious to the reader that what kind of object is being added - now you have to look up the definition of `sample_options_` to see that these are `SampleOption` and not `std::pair<bool , float>`, for example.

include/dali/core/tensor_view.h
```diff
@@ -605,10 +605,10 @@ struct TensorListView : TensorListViewBase<Backend, DataType, sample_ndim> {
   //@{
   /** @brief Implicit conversion */
 
-  template <typename U>
-  TensorListView(const TensorListView<Backend, U, sample_ndim> &other)
+  template <typename U, int other_sample_ndim>
```
mzient, line 608: Nice catch! :+1:
I wonder if it would break something... CI will find out.

dali/pipeline/executor/workspace_policy.h
```diff
@@ -112,7 +112,7 @@ void add_output(op_type_to_workspace_t<op_type> &ws, const tensor_data_store_que
 template <OpType op_type>
 void SetupInputOutput(op_type_to_workspace_t<op_type> &ws, const OpGraph &graph, const OpNode &node,
                       const std::vector<tensor_data_store_queue_t> &tensor_to_store_queue,
-                      const QueueIdxs idxs) {
+                      const QueueIdxs &idxs) {
```
mzient, line 115: These are merely three integers. If anything, I'd remove the `const` but pass it by value.
@klecki What do you think?

klecki: The `const` is here to indicate that it's an input parameter, and the lack of reference was most probably intentional due to the size of the object.

dali/util/random_crop_generator.cc
```diff
@@ -114,7 +114,7 @@ RandomCropGenerator::GenerateCropWindows(const TensorShape<>& shape,
   std::vector<CropWindow> crop_windows;
   for (std::size_t i = 0; i < N; i++) {
     rand_gen_.seed(seeds[i]);
-    crop_windows.push_back(GenerateCropWindowImpl(shape));
+    crop_windows.emplace_back(GenerateCropWindowImpl(shape));
```
mzient, line 117: Please revert. The function returns exactly the element type and the push_back overload used is `push_back(T &&)`. The switch to `emplace_back` doesn't change anything and looks a bit foreign.

dali/pipeline/operator/operator.cc
```diff
@@ -23,7 +23,7 @@ void OperatorBase::EnforceUniformInputBatchSize(const workspace_t<Backend> &ws)
     DALI_ENFORCE(curr_batch_size == ws.GetInputBatchSize(i),
                  "Batch size has to be uniform across one iteration.");
   }
-  auto argument_ws = dynamic_cast<const ArgumentWorkspace &>(ws);
+  const auto &argument_ws = dynamic_cast<const ArgumentWorkspace &>(ws);
```
mzient, line 26: Nice catch. Actually, this is a superclass, so this should work, too (and be faster):
```suggestion
  const ArgumentWorkspace &argument_ws = ws;
```

dali/operators/math/expressions/arithmetic.h
```diff
@@ -155,7 +155,7 @@ inline std::vector<ExprImplTask> CreateExecutionTasks(const ExprNode &expr, Expr
   return result;
 }
 
-inline TensorListShape<> ShapePromotion(std::string op, span<const TensorListShape<> *> shapes,
+inline TensorListShape<> ShapePromotion(const std::string &op, span<const TensorListShape<> *> shapes,
```
mzient, line 158:
```suggestion
inline TensorListShape<> ShapePromotion(const std::string &op,
                                        span<const TensorListShape<> *> shapes,
```
The line length limit of 100 was exceeded.

dali/pipeline/executor/workspace_policy.h
mzient: It stands out, though. We don't use that convention anywhere else.

dali/pipeline/executor/workspace_policy.h
```diff
@@ -171,21 +171,21 @@ void SetupInputOutput(op_type_to_workspace_t<op_type> &ws, const OpGraph &graph,
...
 inline void SetupThreadPool(op_type_to_workspace_t<op_type> &ws, const OpGraph &, const OpNode &,
-                            ThreadPool *, const QueueIdxs) {
+                            ThreadPool *, const QueueIdxs &) {
   /* No-op if we are not CPU */
 }
 
 template <>
 inline void SetupThreadPool<OpType::CPU>(op_type_to_workspace_t<OpType::CPU> &ws, const OpGraph &,
-                                         const OpNode &, ThreadPool *thread_pool, const QueueIdxs) {
+                                         const OpNode &, ThreadPool *thread_pool, const QueueIdxs &) {
```
mzient, line 180:
```suggestion
                                         const OpNode &,
                                         ThreadPool *thread_pool, const QueueIdxs &) {
```
Line length limit of 100 exceeded.

mzient:
!build

mzient (approved):

banasraf (approved):

Issue #3084, opened 2021-06-24
 NVJPEG error "4" : NVJPEG_STATUS_JPEG_NOT_SUPPORTED on A100, but works on V100


melgor:
Hi, I encounter the error while reading my dataset after switching GPU from **V100** to **A100** on exactly the same software stack (GCP instance):
OS- Ubuntu
CUDA: NVIDIA-SMI 450.80.02    Driver Version: 450.80.02    CUDA Version: 11.0
DALI: 1.1
Pytorch: 1.7.1

I upload the minimal example to reproduce (use old API, but on a new one I also face the same problem). Last time you did advise me that I should upgrade to CUDA 11.1 what is the option for me, but need additional work. Also, I check that when I read and write the images using OpenCV, then those images works also at A100 (so look like it the JPEG coding problem?).

But my main concern is that the decoding does not work in the same way in V100 and A100 what does not give as-needed flexibility.

Example image:
![6e83485f4aad83bab23eba8b2c7d_TH](https://user-images.githubusercontent.com/2587438/123219781-92ba5000-d4cd-11eb-9ab7-fd90678fd698.jpeg)

[minimal_example_SIGSEGV.zip](https://github.com/NVIDIA/DALI/files/6707130/minimal_example_SIGSEGV.zip)

JanuszL:
Hi,
I can confirm the bug. https://github.com/NVIDIA/DALI/pull/3086 should fix that.
Please check the nightly build that follows the merge of it or the next release (1.4 most probably).

melgor:
I have a question about HW decoder. So when I would set 
```
ops.ImageDecoder(device = "mixed", output_type = types.RGB, hw_decoder_load=0.0)
```
then above error should not occur as HW decoder would not get any images to decode, right? As the above error is caused by HW.

Because in my case, in the above settings, the problem still occurs.

JanuszL:
Hi @melgor,

The issue happens before the `hw_decoder_load` value is applied. First DALI parses all images in the batch to know how many of them can be offloaded to HW decoder, or hybrid decoder, or need to be processed by CPU. After then numbers are adjusted depending on the load. The problem happens during the parsing.
I don't think there is much you can do besides waiting for the nightly build with the fix.

melgor:
Thanks, I was just thinking that in case when I turn off HW decoder, the error should not occur, what would mean another bug. But it is not the case. Thanks for helpl

melgor:
[...]

PR #3113, opened 2021-07-01
Make nvJPEG detect corrupted stream before offloading to HW decoder

JanuszL:
- there are cases when the JPEG stream can be corrupted and HW decoder
  will fail. Replace check nvjpegJpegStreamParseHeader with nvjpegJpegStreamParse
  to catch this corruption earlier

#### Why we need this PR?
*Pick one, remove the rest*
- It fixes a HW decoder error when JPEG stream is corrupted

 - What solution was applied:
     there are cases when the JPEG stream can be corrupted and HW decoder will fail. Replace check nvjpegJpegStreamParseHeader with nvjpegJpegStreamParse to catch this corruption earlier
     mixed backed of image decoder
      NA
 - Validation and testing:
     current test applies
     NA

Relates to https://github.com/NVIDIA/DALI/issues/3084#issuecomment-872108185
**JIRA TASK**: *[NA]*

### DIFF
--- a/dali/operators/decoder/nvjpeg/nvjpeg_decoder_decoupled_api.h
+++ b/dali/operators/decoder/nvjpeg/nvjpeg_decoder_decoupled_api.h
@@ -62,6 +62,10 @@ class nvJPEGDecoder : public Operator<MixedBackend>, CachedDecoderImpl {
     output_shape_(max_batch_size_, kOutputDim),
     pinned_buffers_(num_threads_*2),
     jpeg_streams_(num_threads_*2),
+#if IS_HW_DECODER_COMPATIBLE
+    // only create one stream per thread - needed only for parsing
+    hw_decoder_jpeg_streams_(num_threads_),
+#endif
 #if NVJPEG2K_ENABLED
     nvjpeg2k_intermediate_buffer_(),
     nvjpeg2k_dev_alloc_(nvjpeg_memory::GetDeviceAllocatorNvJpeg2k()),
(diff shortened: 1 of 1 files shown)


jantonguirao (approved):

mzient (approved):
Looks OK - provided that performance is not negatively affected.

dali/operators/decoder/nvjpeg/nvjpeg_decoder_decoupled_api.h
```diff
@@ -521,100 +527,125 @@ class nvJPEGDecoder : public Operator<MixedBackend>, CachedDecoderImpl {
...
+          if (sz > hybrid_huffman_threshold_ && !data.is_progressive) {
+            data.selected_decoder = &data.decoders[NVJPEG_BACKEND_GPU_HYBRID];
+          } else {
+            data.selected_decoder = &data.decoders[NVJPEG_BACKEND_HYBRID];
+          }
+        }
+      }, in_size);
+    }
+    thread_pool_.RunAll();
+    thread_pool_.WaitForWork();
```
mzient, line 626:
```suggestion
    thread_pool_.RunAll();
```
See:
```C++
DLL_PUBLIC void RunAll(bool wait = true);
```

JanuszL: Done

dali/operators/decoder/nvjpeg/nvjpeg_decoder_decoupled_api.h
```diff
@@ -521,100 +527,124 @@ class nvJPEGDecoder : public Operator<MixedBackend>, CachedDecoderImpl {
...
-          data.selected_decoder = &data.decoders[NVJPEG_BACKEND_GPU_HYBRID];
-        } else {
-          data.selected_decoder = &data.decoders[NVJPEG_BACKEND_HYBRID];
+        // only when we have ROI info check if nvjpegDecodeBatchedSupportedEx supports it
+        if (nvjpeg_decode) {
+  #if IS_HW_DECODER_COMPATIBLE
+          if (state_hw_batched_ != nullptr) {
+            // in some cases hybrid decoder can handle the image but HW decoder can't, we should not
+            // error in that case
+            ret = nvjpegJpegStreamParse(handle_, input_data, in_size, 0, 0,
```
mzient, line 589: These stream handles are per-sample and don't appear to be special or different from those used by the hybrid decoder. Can we reuse these already parsed streams for the regular decoder? In this case we'd need an array of booleans which we would zero-fill at the beginning of this function and here we would store:
`jpeg_stream_ready_[i] = (ret == NVJPEG_STATUS_SUCCESS)`.
Then in the ProcessSample we could just use the already parsed stream.

mzient: OK, now they're per-thread - nvm.

mzient (approved):
If perf is ok, then ship it.

PR #3117, opened 2021-07-02
VM Async Pool - refactoring & tests

mzient:
#### Why we need this PR?
*Pick one, remove the rest*
- Refactoring to allow aync_pool to be used with `cuda_vm_resource`

 - What solution was applied:
     * Make global_pool type a template argument of the async resource
     * Conditionally enable the constructor with upstream only if the pool can be constructed with Upstream*
     * Add a constructor with arbitrary global pool arguments forwarded.
     * Specialize `can_merge` to pools, not just free lists
     * Memory pool
     * N/A
 - Validation and testing:
     * TODO
     * N/A

**JIRA TASK**: DALI-2178

### DIFF
--- a/dali/core/mm/async_pool_test.cu
+++ b/dali/core/mm/async_pool_test.cu
@@ -1,4 +1,4 @@
-// Copyright (c) 2021, NVIDIA CORPORATION. All rights reserved.
+// Copyright (c) 2021, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -55,7 +55,7 @@ TEST(MMAsyncPool, SingleStreamReuse) {
   CUDAStream stream = CUDAStream::Create(true);
   test::test_device_resource upstream;
 
-  async_pool_resource<memory_kind::device, coalescing_free_tree, std::mutex> pool(&upstream);
+  async_pool_resource<memory_kind::device> pool(&upstream);
   stream_view sv(stream);
   int size1 = 1<<20;
   void *ptr = pool.allocate_async(size1, sv);
@@ -80,7 +80,7 @@ TEST(MMAsyncPool, TwoStream) {
   int stream_not_busy = 0;
   int success = 0;
   while (success < min_success) {
-    async_pool_resource<memory_kind::device, coalescing_free_tree, std::mutex> pool(&upstream);
+    async_pool_resource<memory_kind::device> pool(&upstream);
     void *p1 = pool.allocate_async(1000, sv1);
     hog.run(s1);
     pool.deallocate_async(p1, 1000, sv1);
@@ -204,7 +204,7 @@ TEST(MMAsyncPool, SingleStreamRandom) {
   test::test_device_resource upstream;
 
   {
-    async_pool_resource<memory_kind::device, coalescing_free_tree, std::mutex> pool(&upstream);
+    async_pool_resource<memory_kind::device> pool(&upstream);
     vector<block> blocks;
     detail::dummy_lock mtx;
     AsyncPoolTest(pool, blocks, mtx, stream);
@@ -223,7 +223,7 @@ TEST(MMAsyncPool, MultiThreadedSingleStreamRandom) {
     vector<block> blocks;
     std::mutex mtx;
 
-    async_pool_resource<memory_kind::device, coalescing_free_tree, std::mutex> pool(&upstream);
+    async_pool_resource<memory_kind::device> pool(&upstream);
 
     vector<std::thread> threads;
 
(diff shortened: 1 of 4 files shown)


include/dali/core/mm/async_pool.h
```diff
@@ -34,18 +34,35 @@
 namespace dali {
 namespace mm {
 
-template <memory_kind kind, typename FreeList = coalescing_free_tree,
-          typename LockType = std::mutex, typename Upstream = memory_resource<kind>>
+template <memory_kind kind,
+    typename GlobalPool = deferred_dealloc_pool<kind, any_context, coalescing_free_tree, spinlock>,
+    typename LockType = std::mutex,
+    typename Upstream = memory_resource<kind>>
```
mzient, line 40: Previously only the parameters of the pool were here - now the entire pool class is specified.

include/dali/core/mm/cuda_vm_resource.h
```diff
@@ -265,8 +265,8 @@ class cuda_vm_resource : public memory_resource<memory_kind::device> {
     size_t                  block_size = 0;
   };
 
-  SmallVector<va_region, 8> va_regions_;
-  SmallVector<cuvm::CUMemAddressRange, 8> va_ranges_;
+  std::vector<va_region> va_regions_;
+  std::vector<cuvm::CUMemAddressRange> va_ranges_;
```
mzient, line 269: These are not allocated frequently and SmallVector caused warnings in when compiled with NVCC.

JanuszL, line 269: Is there any perf difference?

dali/core/mm/async_pool_test.cu
```diff
@@ -351,5 +350,103 @@ TEST(MMAsyncPool, CrossStreamWithHogs) {
...
+    vector<std::thread> threads;
+
+    for (int t = 0; t < 10; t++) {
+      threads.push_back(std::thread([&]() {
+        AsyncPoolTest(pool, blocks, mtx, stream);
+      }));
+    }
+    for (auto &t : threads)
+      t.join();
+  }
```
JanuszL, line 375: Is there any value in showing stats like in `TEST(MMAsyncPool, MultiThreadedSingleStreamRandom) {`?

dali/core/mm/async_pool_test.cu
```diff
@@ -351,5 +350,103 @@ TEST(MMAsyncPool, CrossStreamWithHogs) {
...
+    threads.push_back(std::thread([&]() {
+      CUDAStream stream = CUDAStream::Create(true);
+      vector<block> blocks;
+      detail::dummy_lock mtx;
+      AsyncPoolTest(pool, blocks, mtx, stream);
+      CUDA_CALL(cudaStreamSynchronize(stream));
+    }));
+  }
+  for (auto &t : threads)
+    t.join();
```
JanuszL, line 396: Why not checking leaks ` upstream.check_leaks();` ?

JanuszL, line 396: Is it possible to check leaks?

dali/core/mm/async_pool_test.cu
```diff
@@ -351,5 +350,103 @@ TEST(MMAsyncPool, CrossStreamWithHogs) {
...
+}
+
+
+
+TEST(MM_VMAsyncPool, CrossStreamWithHogs) {
```
JanuszL, line 423: Why not add `TEST(MM_VMAsyncPool, CrossStream) {` as well?

dali/core/mm/async_pool_test.cu
mzient: I'd have to add them to async_pool (it currently doesn't have any). The global_pool_ member is not accessible here.

include/dali/core/mm/cuda_vm_resource.h
JanuszL: No I see https://github.com/NVIDIA/DALI/pull/3117#discussion_r665267246. Ok

dali/core/mm/async_pool_test.cu
mzient: Not really. If there was an access to the global pool itself (but it's private) I could check the data returned by `get_stat().`
That said, `async_resource` behaves the same way with both pool types, so if there are no leaks with default pool, there would be no leaks from `cuda_vm_resource`.

dali/core/mm/async_pool_test.cu
mzient: I can add it.

include/dali/core/mm/cuda_vm_resource.h
mzient: There shouldn't be any. Increasing the VA size is very rare and it's a big event in itself, taking much more time than simple heap reallocation.

JanuszL (approved):

banasraf (approved):

PR #3118, opened 2021-07-02
Add DALI to EfficientDet example

awolant:
Signed-off-by: Jagoda Kamińska <jk406179@students.mimuw.edu.pl>

#### Why we need this PR?
- It adds DALI to EfficientDet example

 - What solution was applied:
     *Add DALI pipeline to EfficientDet example*
     *EfficientDet example*
     *Is DALI pipeline correct and the same as TF pipeline*
 - Validation and testing:
     *L3 test is added, runs on NGC*
     *Includes documentation*

**JIRA TASK**: *DALI-2120*


docs/examples/use_cases/tensorflow/efficientdet/dataset/create_coco_tfrecord.py
```diff
@@ -160,37 +153,19 @@ def create_tf_example(
...
+            "image/object/bbox/xmin": tfrecord_util.float_list_feature(xmin),
+            "image/object/bbox/xmax": tfrecord_util.float_list_feature(xmax),
+            "image/object/bbox/ymin": tfrecord_util.float_list_feature(ymin),
+            "image/object/bbox/ymax": tfrecord_util.float_list_feature(ymax),
+            "image/object/class/text": tfrecord_util.bytes_list_feature(category_names),
+            "image/object/class/label": tfrecord_util.int64_list_feature(category_ids),
+            "image/object/is_crowd": tfrecord_util.int64_list_feature(is_crowd),
+            "image/object/area": tfrecord_util.float_list_feature(area),
+        }
+    )
```
mzient, line 167: Why has this been moved outside `if bbox_annotations:`?

docs/examples/use_cases/tensorflow/efficientdet/dataset/create_tfrecord_indexes.py
```diff
@@ -15,7 +15,7 @@
 r"""Generate TFRecord index files necessary when using DALI preprocessing.
 
 Example usage:
-    python create_tfrecord_indexes.py  --dali_dir=~/DALI  \
+    python create_tfrecord_indexes.py  --tfrecord2idx_script=~/DALI/tools/tfrecord2idx  \
```
mzient, line 18: What's the benefit of this change?

docs/examples/use_cases/tensorflow/efficientdet/docker/docker_build.py
```diff
@@ -0,0 +1,15 @@
...
+
+import argparse
+import os
+
+parser = argparse.ArgumentParser()
+parser.add_argument("--tag", default="dali-effdet", help="Tag for ")
+args = parser.parse_args()
+
+# We copy tfrecord2idx to
+os.system(" cp ../../../../../../tools/tfrecord2idx ../__docker_tfrecord2idx")
```
mzient, line 11: Hmm.... This would fail miserably when launched from a different path:
```python  docs/examples/use_cases/tensorflow/efficientdet/docker/docker_build.py```
Can we use a path relative to the script source?

File: docs/examples/use_cases/tensorflow/efficientdet/dataset/create_coco_tfrecord.py [hunk 4]
```diff
@@ -160,37 +153,19 @@ def create_tf_example(
...
-                run_len_encoding = mask.frPyObjects(
-                    object_annotations["segmentation"], image_height, image_width
-                )
-                binary_mask = mask.decode(run_len_encoding)
-                if not object_annotations["iscrowd"]:
-                    binary_mask = np.amax(binary_mask, axis=2)
-                pil_image = PIL.Image.fromarray(binary_mask)
-                output_io = io.BytesIO()
-                pil_image.save(output_io, format="PNG")
-                encoded_mask_png.append(output_io.getvalue())
```
mzient, line 173: Do we envision extending EfficientDet support to segmentation? If so, perhaps commenting out the mask support instead of outright deleting it would be sufficient?

docs/examples/use_cases/tensorflow/efficientdet/pipeline/dali/fn_pipeline.py
```diff
@@ -0,0 +1,254 @@
+# Copyright 2021 Kacper Kluk, Paweł Anikiel, Jagoda Kamińska. All Rights Reserved.
```
mzient, line 1: General comment about this file:
why the name `fn_pipeline`? The functional API is now the main API. A more adequate name would be `efficientdet_pipeline` or `ed_pipeline` or maybe just `data_pipeline`.

docs/examples/use_cases/tensorflow/efficientdet/pipeline/dali/ops.py
```diff
@@ -0,0 +1,216 @@
+# Copyright 2021 Kacper Kluk, Paweł Anikiel, Jagoda Kamińska. All Rights Reserved.
```
mzient, line 1: This file name is really confusing - not that I have a much better alternative at hand, but being inside a directory called `dali` it is a bit too similar to `dali.ops`.

docs/examples/use_cases/tensorflow/efficientdet/pipeline/dali/ops.py
```diff
@@ -0,0 +1,216 @@
...
+    l = dali.math.min(0.5 * heights, 0.3 * widths, device=device)
+    r = dali.math.max(0.5 * heights, 0.3 * widths, device=device)
+    tile = dali.fn.cast(
+        (dali.fn.random.uniform(range=[0.0, 1.0], device=device) * (r - l) + l),
+        dtype=dali.types.INT32,
+        device=device,
+    )
+
+    gridmask = dali.fn.grid_mask(
+        images, ratio=ratio, angle=angle, tile=tile, device=device
```
mzient, line 142: This can't work for `device="gpu"`. Keyword arguments must be on CPU, regardless of the operation's backend.
This is a major issue.

docs/examples/use_cases/tensorflow/efficientdet/pipeline/dali/ops.py
```diff
@@ -0,0 +1,216 @@
...
+    scaled_heights = heights * image_scale
+
+    images = dali.fn.resize(
+        images,
+        resize_x=scaled_widths.gpu() if device == "gpu" else scaled_widths,
+        resize_y=scaled_heights.gpu() if device == "gpu" else scaled_heights,
+        device=device,
+    )
+
+    crop_shape = dali.fn.constant(idata=output_size, device="cpu")
```
mzient, line 175: Don't use `fn.constant` directly. Just pass the constant to a keyword argument - it will be wrapped in a constant node _if needed_ - if `output_size` is a plain list or tuple, such promotion is not necessary.

docs/examples/use_cases/tensorflow/efficientdet/pipeline/dali/ops.py
```diff
@@ -0,0 +1,216 @@
...
+
+    scaled_widths = widths * image_scale
+    scaled_heights = heights * image_scale
+
+    images = dali.fn.resize(
+        images,
+        resize_x=scaled_widths.gpu() if device == "gpu" else scaled_widths,
+        resize_y=scaled_heights.gpu() if device == "gpu" else scaled_heights,
+        device=device,
+    )
```
mzient, line 173: This is very inefficient - see below.

docs/examples/use_cases/tensorflow/efficientdet/pipeline/dali/ops.py
```diff
@@ -0,0 +1,216 @@
...
+
+    image_scale = dali.math.min(scaled_x / widths, scaled_y / heights)
+
+    scaled_widths = widths * image_scale
+    scaled_heights = heights * image_scale
+
+    images = dali.fn.resize(
+        images,
+        resize_x=scaled_widths.gpu() if device == "gpu" else scaled_widths,
+        resize_y=scaled_heights.gpu() if device == "gpu" else scaled_heights,
```
mzient, line 171:
```suggestion
        resize_x= scaled_widths,
        resize_y= scaled_heights,
```
BTW - it would be more efficient (2x fewer nodes) when operating on `sizes` instead of separate widths and heights.

docs/examples/use_cases/tensorflow/efficientdet/pipeline/dali/ops.py
```diff
@@ -0,0 +1,216 @@
...
+    else:
+        scale_factor = dali.fn.random.uniform(range=scaling, device="cpu")
+
+    scaled_x = scale_factor * output_size[0]
+    scaled_y = scale_factor * output_size[1]
+
+    image_scale = dali.math.min(scaled_x / widths, scaled_y / heights)
+
+    scaled_widths = widths * image_scale
+    scaled_heights = heights * image_scale
```
mzient, line 166: It would be more efficient (2x fewer nodes) when operating on `sizes` instead of separate widths and heights.

docs/examples/use_cases/tensorflow/efficientdet/pipeline/dali/ops.py
```diff
@@ -0,0 +1,216 @@
...
+
+    if scaling is None:
+        scale_factor = 1.0
+    else:
+        scale_factor = dali.fn.random.uniform(range=scaling, device="cpu")
+
+    scaled_x = scale_factor * output_size[0]
+    scaled_y = scale_factor * output_size[1]
+
+    image_scale = dali.math.min(scaled_x / widths, scaled_y / heights)
```
mzient, line 163: Couldn't this be achieved with resize `mode="not_larger"`?

docs/examples/use_cases/tensorflow/efficientdet/pipeline/dali/ops.py
```diff
@@ -0,0 +1,216 @@
...
+
+    anchors, shapes, bboxes, classes = dali.fn.random_bbox_crop(
+        bboxes,
+        classes,
+        crop_shape=crop_shape,
+        input_shape=dali.fn.cast(
+            dali.fn.cat(scaled_widths, scaled_heights, device="cpu"),
+            dtype=dali.types.INT32,
+            device="cpu",
+        ),
```
mzient, line 185:
```suggestion
        input_shape=dali.fn.cast(dali.fn.cat(scaled_widths, scaled_heights), dtype=dali.types.INT32),```
...or better still - have `sizes`, as noted above.

File: docs/examples/use_cases/tensorflow/efficientdet/pipeline/dali/ops.py [hunk 14]
```diff
@@ -0,0 +1,216 @@
...
+
+    images = dali.fn.slice(
+        images,
+        anchors,
+        shapes,
+        normalized_anchor=False,
+        normalized_shape=False,
+        out_of_bounds_policy="pad",
+        device=device,
+    )
```
mzient, line 199: It's a pity that there's padding - otherwise we could use just `resize` with ROI here. Perhaps we still could, although calculating the ROI would be a bit difficult.

docs/examples/use_cases/tensorflow/efficientdet/pipeline/dali/ops.py
```diff
@@ -0,0 +1,216 @@
...
+
+    return (
+        images,
+        bboxes.gpu() if device == "gpu" else bboxes,
+        classes.gpu() if device == "gpu" else classes,
```
mzient, line 204: Be cautious with moving stuff to the GPU - we can't move it back. Are we sure that this won't be needed in a CPU-only context (keyword arguments, CPU ops?)

docs/examples/use_cases/tensorflow/efficientdet/pipeline/dali/ops.py
```diff
@@ -0,0 +1,216 @@
...
+    )
+
+
+def bbox_to_effdet_format(device, bboxes, image_size):
+    w = image_size[0]
+    h = image_size[1]
+    M = [0.0, h, 0.0, 0.0,
+         w, 0.0, 0.0, 0.0,
+         0.0, 0.0, 0.0, h,
+         0.0, 0.0, w, 0.0]
```
mzient, line 214: Nitpick:
```suggestion
    M = [0.0,   h, 0.0, 0.0,
           w, 0.0, 0.0, 0.0,
         0.0, 0.0, 0.0,   h,
         0.0, 0.0,   w, 0.0]
```

docs/examples/use_cases/tensorflow/efficientdet/pipeline/dali/ops.py
```diff
@@ -0,0 +1,216 @@
...
+        dtype=dali.types.INT32,
+        device=device,
+    )
+
+    gridmask = dali.fn.grid_mask(
+        images, ratio=ratio, angle=angle, tile=tile, device=device
+    )
+
+    p = dali.fn.random.coin_flip(device=device)
+    images = images * (1 - p) + gridmask * p
```
mzient, line 146: Perhaps instead of masking entire image (a lot of memory bandwidth) just mask `ratio`? With ratio being 0 (or 1, I don't remember), gridmask should be a no-op.

docs/examples/use_cases/tensorflow/efficientdet/utils.py
```diff
@@ -30,6 +28,94 @@
 # pylint: disable=logging-format-interpolation
 
 
+class PipelineType(Enum):
+    syntetic = 0
```
mzient, line 32:
```suggestion
    synthetic = 0
```

docs/examples/use_cases/tensorflow/efficientdet/utils.py
```diff
@@ -30,6 +28,94 @@
...
+
+
+class InputType(Enum):
+    tfrecord = 0
+    coco = 1
+
+
+def dict_to_namedtuple(字典):
+    NamedTuple = namedtuple("NamedTuple", 字典.keys())
+    return NamedTuple._make(字典.values())
```
mzient, line 45: Google Translate to the rescue!
```suggestion
def dict_to_namedtuple(dictionary):
    NamedTuple = namedtuple("NamedTuple", dictionary.keys())
    return NamedTuple._make(dictionary.values())
```

mzient (changes requested):
GPU DataNodes passed to keyword arguments - the code apparently wasn't tested.
Fix and at least try to run it with both backends.

awolant:
> GPU DataNodes passed to keyword arguments - the code apparently wasn't tested.
> Fix and at least try to run it with both backends.

Which path was not tested based on your review?

docs/examples/use_cases/tensorflow/efficientdet/docker/docker_build.py
awolant: I think we can remove this whole file. It's not used by the example or the QA test.

mzient:
@awolant The code works by accident - there's a bug somewhere in handling of DataNodes and a DataNode which has both CPU and GPU storage can be passed to keyword arguments. This is misleading, contrary to what our documentation says and subject to change at any moment when we fix it.
Consider the code:
```Python
resized = fn.resize(input, size=size.gpu())
```
This works only because the node `size` is a CPU node and the `.gpu()` is effectively discarded. This invocation of `resize` is still a CPU node and the transfer of `size` to the GPU will never happen - it will be pruned in graph optimization.
Note that the following code will not work:
```Python
resized = fn.resize(input, size=size.gpu() + 0)
```
because now the argument passed to `size` is a result of a GPU-based arithmetic expression and therefore doesn't have a CPU backing storage. The pipeline with such a construct will fail to build.

docs/examples/use_cases/tensorflow/efficientdet/dataset/create_coco_tfrecord.py
szalpal: I believe commented out code works against readability, especially in some examples. I'd go for either keeping it or removing.

docs/examples/use_cases/tensorflow/efficientdet/dataset/create_tfrecord_indexes.py
szalpal: I'd see a benefit when the path to `tfrecord2idx` script changes its path inside DALI repo. We'd get a clear error, saying `"{FLAGS.tfrecord2idx_script} does not lead to valid tfrecord2idx script."` instead a vague `"{FLAGS.dali_dir} does not lead to valid DALI installation."`

docs/examples/use_cases/tensorflow/efficientdet/docker/Dockerfile
```diff
@@ -0,0 +1,11 @@
+FROM nvcr.io/nvidia/tensorflow:21.05-tf2-py3
```
szalpal, line 1:
```suggestion
ARG BASE_VERSION=21.05
FROM nvcr.io/nvidia/tensorflow:${BASE_VERSION}-tf2-py3
```

docs/examples/use_cases/tensorflow/efficientdet/docker/SAMPLE_COMMAND
```diff
@@ -0,0 +1,4 @@
+./docker_run.py --coco_eval_dataset train2017 --cmd '--pipeline dali_gpu --train_batch_size 16 --epochs 50 --output /data/out_weights_1.h5 --eval_steps 500 --train_steps 100 --multi_gpu'
+
+./prepare_coco_dataset_docker.py --coco ~/COCO 
+
```
szalpal, line 4: I'd vote for removing this file and extending `README.rst` with such information

qa/TL3_EfficientDet_convergence/test_tensorflow.sh
```diff
@@ -0,0 +1,21 @@
+#!/bin/bash -e
+
+function CLEAN_AND_EXIT {
+    exit $1
+}
+
+cd /opt/dali/docs/examples/use_cases/tensorflow/efficientdet
```
szalpal, line 7: I suggest using `pushd` and `popd`. `cd` might fail when we decide to run this script from another script

qa/TL3_EfficientDet_convergence/test_tensorflow.sh
```diff
@@ -0,0 +1,21 @@
...
+
+python train.py \
+ --epochs 1 --multi_gpu --seed 1234 --pipeline dali_gpu --input coco \
+ --images_path /data/coco/coco-2017/coco2017/train2017 \
+ --annotations_path /data/coco/coco-2017/coco2017/annotations/instances_train2017.json \
+ --train_batch_size 4 --train_steps 4000 --eval_after_training --eval_during_training \
+ --eval_steps 1000 --eval_freq 1 2>&1 | tee $LOG
+
+
+CLEAN_AND_EXIT 0
```
szalpal, line 21: Do we need this? I though, that when the control flow reaches the end of file, the exit code is always 0?

docs/examples/use_cases/tensorflow/efficientdet/README.rst
```diff
@@ -0,0 +1,124 @@
+EfficientDet with TensorFlow and DALI
+=================================
```
szalpal, line 2: Please make this `README` better. Let's be more user-friendly and when user navigates here, he should get some useful information. I don't expect this README to be like the ones in [Deep Learning Examples](https://github.com/NVIDIA/DeepLearningExamples/blob/master/PyTorch/SpeechRecognition/Jasper/README.md), but I do believe README is one of the most important parts of the example.

docs/examples/use_cases/tensorflow/efficientdet/docker/docker_run.py
```diff
@@ -0,0 +1,56 @@
+#!/usr/bin/env python
+
+import argparse
+import os
```
szalpal, line 4: Similarly to `SAMPLE_COMMAND` file and `docker_build.py` file, I'm having hard time to grasp, why is it here in the first place? If the only purpose would be to encapsulate `docker run` command, maybe we can include this command in the `README`?

`prepare_coco_dataset_docker.py` also falls into this category.

docs/examples/use_cases/tensorflow/efficientdet/train.py
```diff
@@ -0,0 +1,232 @@
...
+    parser = argparse.ArgumentParser()
+    parser.add_argument(
+        "--initial_epoch",
+        type=int,
+        default=0,
+        help="epoch from which to start training",
+    )
+    parser.add_argument(
+        "--epochs", type=int, default=300, help="epoch on which training should finish"
+    )
```
szalpal, line 160: Naming this argument `epochs` suggest, that it is a number of epochs, not the index of final epoch. May we consider changing either the behaviour or the name?

docs/examples/use_cases/tensorflow/efficientdet/train.py
```diff
@@ -0,0 +1,232 @@
...
+    parser.add_argument(
+        "--input",
```
szalpal, line 162: Consider changing to `input_type` to reflect the description better

docs/examples/use_cases/tensorflow/efficientdet/train.py
```diff
@@ -0,0 +1,232 @@
...
+            model.evaluate(eval_dataset, steps=args.eval_steps)
+
+        model.save_weights(args.output)
+
+
+if __name__ == "__main__":
+    import argparse
+    from argparse_utils import enum_action
+
+    parser = argparse.ArgumentParser()
```
szalpal, line 151: General comment to these argument definitions. I believe it would be easier to read, if the descriptions (`help` args) would be unified to start from a capital letter and end with a dot.

docs/examples/use_cases/tensorflow/efficientdet/train.py
```diff
@@ -0,0 +1,232 @@
...
+        "--annotations_path",
+        required=False,
+        help="Path to COCO annotations",
+    )
+    parser.add_argument(
+        "--train_file_pattern",
+        required=False,
+        help="glob pattern for TFrecord files with training data",
+    )
+    parser.add_argument("--train_batch_size", type=int, default=64)
```
szalpal, line 182: How about having just a `batch_size`?

docs/examples/use_cases/tensorflow/efficientdet/train.py
```diff
@@ -0,0 +1,232 @@
...
+    parser.add_argument(
+        "--train_steps", type=int, default=2000, help="number of steps in each epoch"
```
szalpal, line 184: What are the `steps` here? If these are meant to be `iterations`, shouldn't it be inferred from the batch size and dataset size?

docs/examples/use_cases/tensorflow/efficientdet/train.py
```diff
@@ -0,0 +1,232 @@
...
+    parser.add_argument(
+        "--eval_steps", type=int, default=5000, help="number of examples to evaluate"
+    )
+    parser.add_argument(
+        "--eval_freq", type=int, default=1, help="during training evalutaion frequency"
+    )
+    parser.add_argument(
+        "--eval_during_training",
+        action="store_true",
+        help="whether to run evaluation every EVAL_FREQ epochs",
```
szalpal, line 200: I believe this formatting is how we should refer to other arguments
```python
"whether to run evaluation every `eval_freq` epochs"
```

docs/examples/use_cases/tensorflow/efficientdet/train.py
```diff
@@ -0,0 +1,232 @@
...
+        "--eval_after_training",
+        action="store_true",
+        help="whether to run evaluation after finished training",
+    )
+    parser.add_argument(
+        "--pipeline",
+        action=enum_action(utils.PipelineType),
+        required=True,
+        help="pipeline type",
+    )
```
szalpal, line 212: What is the `pipeline` in this context? I feel this arguments needs some explanation

docs/examples/use_cases/tensorflow/efficientdet/train.py
```diff
@@ -0,0 +1,232 @@
...
+        type=int,
+        help="list of GPUs to use, defaults to all visible GPUs",
+    )
+    parser.add_argument("--seed", type=int)
+    parser.add_argument(
+        "--hparams", default="", help="string or filename with parameters"
+    )
+    parser.add_argument("--model_name", default="efficientdet-d1")
+    parser.add_argument(
+        "--output", default="output.h5", help="filename for final weights to save"
```
szalpal, line 225: How about `output_filename`?

docs/examples/use_cases/tensorflow/efficientdet/train.py
```diff
@@ -0,0 +1,232 @@
...
+            tf.io.gfile.makedirs(ckpt_dir)
+        config_file = os.path.join(ckpt_dir, "config.yaml")
+        if not tf.io.gfile.exists(config_file):
+            tf.io.gfile.GFile(config_file, "w").write(str(config))
+
+    if params["seed"]:
+        seed = params["seed"]
+        os.environ["PYTHONHASHSEED"] = str(seed)
+        tf.random.set_seed(seed)
+        np.random.seed(seed)
```
szalpal, line 44: How hard would it be to migrate from the legacy global random state in numpy and go with object-based?
```python
rng = np.random.default_rng(seed)
```

docs/examples/use_cases/tensorflow/efficientdet/eval.py
```diff
@@ -0,0 +1,75 @@
...
+        image_size = params["image_size"]
+        model.predict(np.zeros((1, image_size[0], image_size[1], 3)))
+        model.load_weights(args.weights)
+
+    model.evaluate(dataset, steps=args.eval_steps)
+
+
+if __name__ == "__main__":
+    import argparse
+    from argparse_utils import enum_action
```
szalpal, line 49: The comments I had to the corresponding code in `train.py` apply here also

docs/examples/use_cases/tensorflow/efficientdet/eval.py
```diff
@@ -0,0 +1,75 @@
...
+
+import hparams_config
+import utils
+from model import efficientdet_net
+from pipeline.dali import fn_pipeline
+
+
+def run_eval(args):
+    logging.set_verbosity(logging.WARNING)
+
```
szalpal, line 16: Don't `eval.py` and `train.py` have a few lines of duplicated code? Would it make sense to extract it?

docs/examples/use_cases/tensorflow/efficientdet/utils.py
szalpal: I agree here. We shouldn't get outside of ascii.

docs/examples/use_cases/tensorflow/efficientdet/dataset/create_coco_tfrecord.py
Riyria-was-taken: That's because in COCO there are pictures without any bounding boxes and without that change, DALI's tfrecord reader was failing as it was expecting to find some labels and they weren't added. With that change, we add labels with empty lists even if a picture doesn't have bounding boxes.

docs/examples/use_cases/tensorflow/efficientdet/dataset/create_tfrecord_indexes.py
Riyria-was-taken: Also, with that change users can use that script even if they installed DALI via pip and don't have DALI source code locally.

docs/examples/use_cases/tensorflow/efficientdet/dataset/create_coco_tfrecord.py
Riyria-was-taken: In order to make EfficientDet example as small as possible, we removed everything connected to segmentation from the original code, not only from that part. If we wanted to support segmentation it would be too problematic at that point.

docs/examples/use_cases/tensorflow/efficientdet/pipeline/dali/fn_pipeline.py
Riyria-was-taken: Good point, it's a remnant from the time when we had pipelines in both APIs.

docs/examples/use_cases/tensorflow/efficientdet/pipeline/dali/ops.py
Riyria-was-taken: I have no good idea for a name too, would one from these: utils, helpers, helper_ops work?

Riyria-was-taken: I decided to use ops_util.

docs/examples/use_cases/tensorflow/efficientdet/train.py
Riyria-was-taken: Actually, NumPy's randomness isn't used in code explicitly but in case there is NumPy used in imported modules in such a way we need to set seed in the legacy way.

docs/examples/use_cases/tensorflow/efficientdet/docker/docker_build.py
Riyria-was-taken: I removed that file.

docs/examples/use_cases/tensorflow/efficientdet/pipeline/dali/ops.py
Riyria-was-taken: Fixed.

docs/examples/use_cases/tensorflow/efficientdet/pipeline/dali/ops.py
Riyria-was-taken: Fixed.

docs/examples/use_cases/tensorflow/efficientdet/pipeline/dali/ops.py
Riyria-was-taken: I guess so, I changed it :)

docs/examples/use_cases/tensorflow/efficientdet/pipeline/dali/ops.py
Riyria-was-taken: Changed to the first version.

docs/examples/use_cases/tensorflow/efficientdet/docker/Dockerfile
Riyria-was-taken: Changed.

qa/TL3_EfficientDet_convergence/test_tensorflow.sh
Riyria-was-taken: Changed.

docs/examples/use_cases/tensorflow/efficientdet/train.py
Riyria-was-taken: Steps are indeed iterations, I'm using Keras naming convention here.
In order to infer iterations number from batch size and dataset size, we'd have to ask the user for dataset size (num_examples_per_epoch argument in original EfficientDet implementation) and I think that asking here for the number of steps is better as it is easier to run training with a small number of steps for example for testing purposes.

docs/examples/use_cases/tensorflow/efficientdet/train.py
Riyria-was-taken: Fixed.

docs/examples/use_cases/tensorflow/efficientdet/train.py
Riyria-was-taken: Changed.

docs/examples/use_cases/tensorflow/efficientdet/utils.py
Riyria-was-taken: Fixed.

docs/examples/use_cases/tensorflow/efficientdet/utils.py
Riyria-was-taken: Fixed ;)

qa/TL3_EfficientDet_convergence/test_tensorflow.sh
Riyria-was-taken: I cannot say for sure if we need this but I wrote it based on other tests and it's done in the same way in the YOLOv4 example so I think it is fine.

docs/examples/use_cases/tensorflow/efficientdet/train.py
Riyria-was-taken: I'm following here a naming convention used in TensorFlow's Keras API's fit function so I think that name is fine and should be familiar to TensorFlow users. I'd prefer not to change it.

docs/examples/use_cases/tensorflow/efficientdet/train.py
Riyria-was-taken: Done.

docs/examples/use_cases/tensorflow/efficientdet/train.py
Riyria-was-taken: Fixed.

docs/examples/use_cases/tensorflow/efficientdet/train.py
Riyria-was-taken: Changed.

docs/examples/use_cases/tensorflow/efficientdet/eval.py
Riyria-was-taken: Fixed.

docs/examples/use_cases/tensorflow/efficientdet/eval.py
Riyria-was-taken: I extracted duplicated code where we prepare GPUs (set up memory_growth and soft_device_placement).
When it comes to the marked part of the code the only part I think is possible to extract is
```
    config = hparams_config.get_efficientdet_config(args.model_name)
    config.override(args.hparams, allow_new_keys=True)
    config.image_size = utils.parse_image_size(config.image_size)
```
but I don't think it makes sense to do that (it'd be something like get_config and we have already get_efficientdet_config and I think that would be too confusing).

docs/examples/use_cases/tensorflow/efficientdet/train.py
Riyria-was-taken: Done.

docs/examples/use_cases/tensorflow/efficientdet/pipeline/dali/ops.py
Riyria-was-taken: Changed.

docs/examples/use_cases/tensorflow/efficientdet/README.rst
awolant: This was based on other README pages we have for our other use cases. How about we figure out some concrete ideas on how to make docs better for this whole section and do it as a separate effort?

docs/examples/use_cases/tensorflow/efficientdet/dataset/create_tfrecord_indexes.py
mzient: Ooookaaaay. It's a minor thing and I will not stop this PR over this, but... I still think that we can combine better error message with ease of use (like `Cannot find the tfrecord2idx tool in DALI installation directory. The expected path "{FLAGS.dali_dir}/tools/tfrecord2idx" is invalid. Please check your DALI installation.`). The argument about the file being moved is invalid, because we ship _this_ file with DALI and we still need to keep it up to date (at least to have a proper default). On the other hand, if the user updates DALI _and_ we move the file, then the usage of this script changes - while it wouldn't change if the user updates DALI _and_ examples - the script would pick up the tool from where it should.

docs/examples/use_cases/tensorflow/efficientdet/pipeline/dali/efficientdet_pipeline.py
```diff
@@ -0,0 +1,262 @@
...
+        self._max_instances_per_image = params["max_instances_per_image"] or 100
+        seed = params["seed"] or -1
+
+        self._pipe = dali.pipeline.Pipeline(
+            batch_size=self._batch_size,
+            num_threads=self._num_shards,
+            device_id=device_id,
+            seed=seed,
+        )
+        self._define_pipeline()
```
mzient, line 76: Can we apply our decorator to members? If so, then how about decorating _define_pipeline with `@pipeline_def` and having
```suggestion
        self._pipe = self._define_pipeline(
            batch_size=self._batch_size,
            num_threads=self._num_shards,
            device_id=device_id,
            seed=seed,
        )
```

docs/examples/use_cases/tensorflow/efficientdet/pipeline/dali/efficientdet_pipeline.py
```diff
@@ -0,0 +1,262 @@
...
+                file_pattern = args.eval_file_pattern or args.train_file_pattern
+
+            self._tfrecord_files = glob(file_pattern)
+            self._tfrecord_idxs = [
+                filename + "_idx" for filename in self._tfrecord_files
+            ]
+        elif self._input_type == InputType.coco:
+            self._images_path = args.images_path
+            self._annotations_path = args.annotations_path
+
```
mzient, line 57: Hmm... if there's no default input type, then I guess we should error-out?
```suggestion
            else:
              raise ValueError("Invalid input type: ", self._input_type)
```

docs/examples/use_cases/tensorflow/efficientdet/pipeline/dali/efficientdet_pipeline.py
```diff
@@ -0,0 +1,262 @@
...
+            if self._input_type == InputType.tfrecord:
+                images, bboxes, classes, widths, heights = ops_util.input_tfrecord(
+                    self._tfrecord_files,
+                    self._tfrecord_idxs,
+                    device=self._device,
+                    shard_id=self._shard_id,
+                    num_shards=self._num_shards,
+                    random_shuffle=self._is_training,
+                )
+            elif self._input_type == InputType.coco:
```
mzient, line 97: How about:
```suggestion
            else:
                assert self._input_type == InputType.coco:
```
?

docs/examples/use_cases/tensorflow/efficientdet/pipeline/dali/efficientdet_pipeline.py
```diff
@@ -0,0 +1,262 @@
...
+                self._image_size,
+                [0.1, 2.0] if self._is_training else None,
+            )
+
+            if self._device == "gpu":
+                bboxes = bboxes.gpu()
+                classes = classes.gpu()
+
+            enc_bboxes, enc_classes = dali.fn.box_encoder(
+                bboxes, classes, anchors=self._boxes, offset=True, device=self._device
```
mzient, line 130: In functional API the device is inferred.
```suggestion
                bboxes, classes, anchors=self._boxes, offset=True
```

docs/examples/use_cases/tensorflow/efficientdet/pipeline/dali/efficientdet_pipeline.py
```diff
@@ -0,0 +1,262 @@
...
+            enc_classes -= 1
+
+            # convert to tlbr
+            enc_bboxes = dali.fn.coord_transform(
+                enc_bboxes,
+                M=[0, 1, 0, 0,
+                   1, 0, 0, 0,
+                   0, 0, 0, 1,
+                   0, 0, 1, 0],
+                device=self._device,
```
mzient, line 145: Drop device?

docs/examples/use_cases/tensorflow/efficientdet/pipeline/dali/ops_util.py
```diff
@@ -0,0 +1,195 @@
...
+    )
+
+    return (
+        images,
+        bboxes,
+        classes,
+    )
+
+
+def bbox_to_effdet_format(device, bboxes, image_size):
```
mzient, line 187:
```suggestion
def bbox_to_effdet_format(bboxes, image_size):
```

docs/examples/use_cases/tensorflow/efficientdet/pipeline/dali/ops_util.py
```diff
@@ -0,0 +1,195 @@
...
+    w = image_size[0]
+    h = image_size[1]
+    M = [0.0,   h, 0.0, 0.0,
+           w, 0.0, 0.0, 0.0,
+         0.0, 0.0, 0.0,   h,
+         0.0, 0.0,   w, 0.0]
+
+    return dali.fn.coord_transform(bboxes, M=M, device=device)
```
mzient, line 195:
```suggestion
    return dali.fn.coord_transform(bboxes, M=M)
```

docs/examples/use_cases/tensorflow/efficientdet/pipeline/dali/ops_util.py
```diff
@@ -0,0 +1,195 @@
...
+        scale_factor = 1.0
+    else:
+        scale_factor = dali.fn.random.uniform(range=scaling)
+
+    sizes = dali.fn.cat(widths, heights)
+    image_scale = dali.math.min(
+        scale_factor * output_size[0] / widths,
+        scale_factor * output_size[1] / heights,
+    )
+    scaled_sizes = sizes * image_scale
```
mzient, line 152: Note: you're about to pass floating-point output size here. Of course, the output has to have an integer size and subpixel scaling will kick in - namely, the output size will be rounded and the input ROI adjusted so that the input and output centers are aligned and the scaling factor maintained exactly. The boxes, however, are obtained by calling `random_bbox_crop` with integer input_shape. This will cause subtle misplacement (up to half a pixel) of bounding boxes near the edges.
If a slight loss of aspect ratio (up to 1/output_size) is acceptable, round the sizes here.
```suggestion
    scaled_sizes = fn.cast(sizes * image_scale, dtype=dalit.types.INT32)
```

docs/examples/use_cases/tensorflow/efficientdet/pipeline/dali/ops_util.py
```diff
@@ -0,0 +1,195 @@
...
+        images,
+        size=scaled_sizes,
+        device=device,
+    )
+
+    anchors, shapes, bboxes, classes = dali.fn.random_bbox_crop(
+        bboxes,
+        classes,
+        crop_shape=output_size,
+        input_shape=dali.fn.cast(scaled_sizes, dtype=dali.types.INT32),
```
mzient, line 164: (if you follow the advice above, remove the cast here)

docs/examples/use_cases/tensorflow/efficientdet/pipeline/dali/ops_util.py
```diff
@@ -0,0 +1,195 @@
...
+    )
+
+    images = dali.fn.slice(
+        images,
+        anchors,
+        shapes,
+        normalized_anchor=False,
+        normalized_shape=False,
+        out_of_bounds_policy="pad",
+        device=device,
```
mzient, line 177: Device is not necessary.

docs/examples/use_cases/tensorflow/efficientdet/pipeline/dali/efficientdet_pipeline.py
```diff
@@ -0,0 +1,262 @@
...
+            if self._device == "gpu":
+                bboxes = bboxes.gpu()
+                classes = classes.gpu()
+
+            enc_bboxes, enc_classes = dali.fn.box_encoder(
+                bboxes, classes, anchors=self._boxes, offset=True, device=self._device
+            )
+            num_positives = dali.fn.reductions.sum(
+                dali.fn.cast(enc_classes != 0, dtype=dali.types.FLOAT),
+                device=self._device,
```
mzient, line 134: device is not necessary

docs/examples/use_cases/tensorflow/efficientdet/pipeline/dali/efficientdet_pipeline.py
```diff
@@ -0,0 +1,262 @@
...
+            ]
+
+            bboxes = ops_util.bbox_to_effdet_format(
+                self._device, bboxes, self._image_size
+            )
+            bboxes = dali.fn.pad(
+                bboxes,
+                fill_value=-1,
+                shape=(self._max_instances_per_image, 4),
+                device=self._device,
```
mzient, line 167: not necessary

docs/examples/use_cases/tensorflow/efficientdet/pipeline/dali/efficientdet_pipeline.py
```diff
@@ -0,0 +1,262 @@
...
+            )
+            classes = dali.fn.pad(
+                classes,
+                fill_value=-1,
+                shape=(self._max_instances_per_image,),
+                device=self._device,
```
mzient, line 173: not necessary

docs/examples/use_cases/tensorflow/efficientdet/pipeline/dali/efficientdet_pipeline.py
```diff
@@ -0,0 +1,262 @@
...
+                        enc_bboxes,
+                        (count, 0),
+                        (steps, 4),
+                        axes=[0, 1],
+                        device=self._device,
+                    ),
+                    shape=[feat_size["height"], feat_size["width"], -1],
+                    device=self._device,
+                )
+            )
```
mzient, line 205: I'm not sure I properly understand this line - what are the anchor and shape arguments?
```suggestion
            enc_bboxes_layers.append(enc_boxes[count:count+steps,:, dali.newaxis])
```
?

docs/examples/use_cases/tensorflow/efficientdet/pipeline/dali/efficientdet_pipeline.py
```diff
@@ -0,0 +1,262 @@
...
+            enc_classes_layers.append(
+                dali.fn.reshape(
+                    dali.fn.slice(
+                        enc_classes, count, steps, axes=[0], device=self._device
+                    ),
+                    shape=[feat_size["height"], feat_size["width"], -1],
+                    device=self._device,
+                )
+            )
```
mzient, line 214: ...and here:
```suggestion
            enc_classes_layers.append(enc_classes[count:count+steps,:,dali.newaxis])
```
?

docs/examples/use_cases/tensorflow/efficientdet/pipeline/dali/ops_util.py
```diff
@@ -0,0 +1,195 @@
...
+
+    images = dali.fn.decoders.image(
+        encoded,
+        device="mixed" if device == "gpu" else "cpu",
+        output_type=dali.types.RGB,
+    )
+
+    shape = dali.fn.peek_image_shape(encoded, type=dali.types.FLOAT)
+    heights = dali.fn.slice(shape, 0, 1, axes=(0,))
+    widths = dali.fn.slice(shape, 1, 1, axes=(0,))
```
mzient, line 93:
```suggestion
    heights = shape[0]
    widths = shape[1]
```

docs/examples/use_cases/tensorflow/efficientdet/pipeline/dali/ops_util.py
```diff
@@ -0,0 +1,195 @@
...
+def random_crop_resize(
+    device, images, bboxes, classes, widths, heights, output_size, scaling=[0.1, 2.0]
+):
+
+    if scaling is None:
+        scale_factor = 1.0
+    else:
+        scale_factor = dali.fn.random.uniform(range=scaling)
+
+    sizes = dali.fn.cat(widths, heights)
```
mzient, line 147: The size for `resize` is in height, width order
```suggestion
    sizes = dali.fn.cat(heights, widths)
```

docs/examples/use_cases/tensorflow/efficientdet/pipeline/dali/ops_util.py
awolant: Didn't work.

docs/examples/use_cases/tensorflow/efficientdet/pipeline/dali/ops_util.py
awolant: Done

docs/examples/use_cases/tensorflow/efficientdet/pipeline/dali/efficientdet_pipeline.py
awolant: It can't happen. Input pipeline type is defined as an enum. Argparse will handle invalid type and return error message like: 
`train.py: error: argument --input_type: invalid choice: 'abcd' (choose from tfrecord, coco)`

docs/examples/use_cases/tensorflow/efficientdet/pipeline/dali/efficientdet_pipeline.py
awolant: Done

docs/examples/use_cases/tensorflow/efficientdet/pipeline/dali/efficientdet_pipeline.py
awolant: Actually, `offset` parameter controls the mode of operation of box encoder. It either returns offsets between matches or the actual matches, so it is important regardless of whether you provide other values from the formula in the documentation.

Removed `device`.

docs/examples/use_cases/tensorflow/efficientdet/pipeline/dali/ops_util.py
awolant: Done

docs/examples/use_cases/tensorflow/efficientdet/pipeline/dali/ops_util.py
awolant: Removed device. 

As to the order, it is a helper function, so there is no strict rule about the order of arguments AFAIK.

docs/examples/use_cases/tensorflow/efficientdet/pipeline/dali/efficientdet_pipeline.py
awolant: Done

docs/examples/use_cases/tensorflow/efficientdet/pipeline/dali/ops_util.py
awolant: Done

docs/examples/use_cases/tensorflow/efficientdet/pipeline/dali/efficientdet_pipeline.py
awolant: Done

docs/examples/use_cases/tensorflow/efficientdet/pipeline/dali/efficientdet_pipeline.py
awolant: Done

docs/examples/use_cases/tensorflow/efficientdet/pipeline/dali/efficientdet_pipeline.py
awolant: Done

docs/examples/use_cases/tensorflow/efficientdet/pipeline/dali/efficientdet_pipeline.py
awolant: Didn't work

docs/examples/use_cases/tensorflow/efficientdet/pipeline/dali/efficientdet_pipeline.py
awolant: Didn't work

docs/examples/use_cases/tensorflow/efficientdet/pipeline/dali/ops_util.py
awolant: Didn't work.

docs/examples/use_cases/tensorflow/efficientdet/pipeline/dali/ops_util.py
awolant: Didn't work.

docs/examples/use_cases/tensorflow/efficientdet/pipeline/dali/efficientdet_pipeline.py
awolant: Done

docs/examples/use_cases/tensorflow/efficientdet/pipeline/dali/ops_util.py
```diff
@@ -0,0 +1,192 @@
...
+def random_crop_resize(
+    images, bboxes, classes, widths, heights, output_size, scaling=[0.1, 2.0]
+):
+
+    if scaling is None:
+        scale_factor = 1.0
+    else:
+        scale_factor = dali.fn.random.uniform(range=scaling)
+
+    sizes = dali.fn.cat(heights, widths)
```
mzient, line 146: When using indexing, widths and heights are 0D - use `stack` instead of `cat`.

awolant: Done

mzient (approved):

docs/examples/use_cases/tensorflow/efficientdet/README.rst
szalpal: Fine by me. Just pointing out review issues I see ;)

docs/examples/use_cases/tensorflow/efficientdet/train.py
```diff
@@ -0,0 +1,239 @@
...
+    )
+    parser.add_argument(
+        "--pipeline_type",
+        action=enum_action(utils.PipelineType),
+        required=True,
+        help="R|Pipeline type used while loading and preprocessing data. One of:\n"
+        "tensorflow – pipeline used in original EfficientDet implementation on https://github.com/google/automl/tree/master/efficientdet;\n"
+        "synthetic – like `tensorflow` pipeline type but repeats one batch endlessly;\n"
+        "dali_gpu – pipeline which uses Nvidia Data Loading Library (DALI) to run part of data preprocessing on GPUs to improve efficiency;\n"
+        "dalu_cpu – like `dali_gpu` pipeline type but restricted to run only on CPU.",
```
szalpal, line 216:
```suggestion
        "dali_cpu – like `dali_gpu` pipeline type but restricted to run only on CPU.",
```

Please verify, that the argument is actually named `dali_cpu` not `dalu_cpu`

docs/examples/use_cases/tensorflow/efficientdet/train.py
```diff
@@ -0,0 +1,239 @@
...
+        help="Whether to run evaluation after finished training.",
+    )
+    parser.add_argument(
+        "--pipeline_type",
+        action=enum_action(utils.PipelineType),
+        required=True,
+        help="R|Pipeline type used while loading and preprocessing data. One of:\n"
+        "tensorflow – pipeline used in original EfficientDet implementation on https://github.com/google/automl/tree/master/efficientdet;\n"
+        "synthetic – like `tensorflow` pipeline type but repeats one batch endlessly;\n"
+        "dali_gpu – pipeline which uses Nvidia Data Loading Library (DALI) to run part of data preprocessing on GPUs to improve efficiency;\n"
```
szalpal, line 215:
```suggestion
        "dali_gpu – pipeline which uses NVIDIA DALI to run part of data preprocessing on GPUs to improve efficiency;\n"
```
I'd go with this, but it's a matter of taste. Anyway, `NVIDIA` should be capitalised

docs/examples/use_cases/tensorflow/efficientdet/train.py
```diff
@@ -0,0 +1,239 @@
...
+    parser.add_argument(
+        "--eval_file_pattern",
+        help="TFrecord files glob pattern for files with evaluation data, "
+        "defaults to `train_file_pattern` if not given.",
+    )
+    parser.add_argument(
+        "--eval_steps",
+        type=int,
+        default=5000,
+        help="Number of examples to evaluate during each evaluation.",
```
szalpal, line 193:
```suggestion
        help="Number of examples to process during each evaluation.",
```

docs/examples/use_cases/tensorflow/efficientdet/eval.py
```diff
@@ -0,0 +1,86 @@
...
+    )
+    parser.add_argument(
+        "--pipeline_type",
+        action=enum_action(utils.PipelineType),
+        required=True,
+        help="R|Pipeline type used while loading and preprocessing data. One of:\n"
+        "tensorflow – pipeline used in original EfficientDet implementation on https://github.com/google/automl/tree/master/efficientdet;\n"
+        "synthetic – like `tensorflow` pipeline type but repeats one batch endlessly;\n"
+        "dali_gpu – pipeline which uses Nvidia Data Loading Library (DALI) to run part of data preprocessing on GPUs to improve efficiency;\n"
+        "dalu_cpu – like `dali_gpu` pipeline type but restricted to run only on CPU.",
```
szalpal, line 75:
```suggestion
        "dali_cpu – like `dali_gpu` pipeline type but restricted to run only on CPU.",
```

docs/examples/use_cases/tensorflow/efficientdet/train.py
```diff
@@ -0,0 +1,239 @@
...
+    )
+    parser.add_argument(
+        "--eval_freq", type=int, default=1, help="During training evaluation frequency."
```
szalpal, line 196: This sentence looks strange. Please consider my suggestion.
```suggestion
        "--eval_freq", type=int, default=1, help="Run the evaluation every `eval_freq` iterations."
```

szalpal (approved):
Please consider applying some minor issues I've found. Otherwise LGTM.

docs/examples/use_cases/tensorflow/efficientdet/train.py
awolant: Done

docs/examples/use_cases/tensorflow/efficientdet/train.py
awolant: Done

docs/examples/use_cases/tensorflow/efficientdet/train.py
awolant: Done

docs/examples/use_cases/tensorflow/efficientdet/eval.py
awolant: Done

docs/examples/use_cases/tensorflow/efficientdet/train.py
awolant: Done

PR #3122, opened 2021-07-05
Integer literal suffixes

mzient:
* Add literal suffixes for size_t, ssize_t and fixed-size integer types.
* Rework usage of (unsigned) long or long long literals to proper 64-bit
integers or size_t.

#### Why we need this PR?
*Pick one, remove the rest*
- Refactoring to improve usage of properly-sized integers.

 - What solution was applied:
     * Added literal operators with suffixes:
     * `_z` for `ssize_t`
     * `_uz` and `_zu` for `size_t`
     * `_iXX and `_uXX` for fixed-size integers of 8, 16, 32, and 64 bits
     * Application code: reduce kernel
     * Tests
     * Build: almost everything
     * N/A
 - Validation and testing:
     * Existing tests apply
     * N/A

**JIRA TASK**: related to DALI-2166

### DIFF
--- a/dali/core/dev_buffer_test.cc
+++ b/dali/core/dev_buffer_test.cc
@@ -1,4 +1,4 @@
-// Copyright (c) 2020, NVIDIA CORPORATION. All rights reserved.
+// Copyright (c) 2020-2021, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -104,7 +104,7 @@ TEST(DeviceBuffer, FromDevice) {
 
 TEST(DeviceBufferFail, Resize) {
   DeviceBuffer<int> db;
-  size_t size = static_cast<size_t>(-1);
+  size_t size = -1_uz;
   EXPECT_THROW(db.resize(size), CUDABadAlloc);
 }
 
--- a/dali/core/error_test.cc
+++ b/dali/core/error_test.cc
@@ -1,4 +1,4 @@
-// Copyright (c) 2018, NVIDIA CORPORATION. All rights reserved.
+// Copyright (c) 2018, 2021, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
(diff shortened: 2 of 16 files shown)


jantonguirao (approved):

cliffwoolley:
I see what you did there with the constexpr operators (but only because they're all here in the same PR).  But it's not super obvious on inspection that an operator function is actually being called there, and if someone hasn't seen the operators, then this would read like as if it's some new language built-ins.  Is there a way to achieve this that's less magical-looking?

mzient:
> I see what you did there with the constexpr operators (but only because they're all here in the same PR). But it's not super obvious on inspection that an operator function is actually being called there, and if someone hasn't seen the operators, then this would read like as if it's some new language built-ins. Is there a way to achieve this that's less magical-looking?

Well, this magic was the intention - it reads like a normal literal which it pretty much is.
Regarding the `size_t` and `ssize_t` literal suffixes - there's a proposal to add them to C++23 (there won't be an underscore then - it will be `123z` (signed) or `123uz` (unsigned).

cliffwoolley:
> Well, this magic was the intention - it reads like a normal literal which it pretty much is.

Pretty much, but not in actuality.

Anyway you all own the code and if you collectively think this is a good thing, then so be it.  My advice from the outside looking in would be to think twice before implicitly extending the language.

mzient:
> > Well, this magic was the intention - it reads like a normal literal which it pretty much is.
> 
> Pretty much, but not in actuality.
> 
> Anyway you all own the code and if you collectively think this is a good thing, then so be it. My advice from the outside looking in would be to think twice before implicitly extending the language.

Well, we're not really extending the language - not much more than overloading other operators. Also, it's hardly new - it's a C++11 feature and the pattern of underscore after a number should be fairly recognizable by now. Also, these literals are placed in namespace dali, so this extension doesn't leak beyond our code (although it can be imported with `using namespace dali::literal`).

We keep receiving some requests for Windows support every now and then. If we are to approach native build, then we need to adjust the type system. Windows has `32-bit long` so much of the code that assumes 64-bit long would fail. While we don't use long/long long types (not frequently, at least), this PR is evidence that we would cut corners by using long constants. It's hard to blame the authors - putting casts everywhere would make the code unreadable.

JanuszL (approved):

PR #3124, opened 2021-07-05
Coverity  issue fixes - bulk drop, July 2021

mzient:
#### Why we need this PR?
*Pick one, remove the rest*
- It fixes bugs in the size of bit masks
- It fixes uninitialized but readable member in AsrSample
- It makes `half` type trivially copyable

 - What solution was applied:
     * Use proper literals `1_u64` or `1_uz` or `1u` when constructing bit masks
     * Make `half` copy constructor defaulted.
     * Initialize the index field in AsrSample with 0.
     * Reduction
     * Normalization
     * Nemo ASR reader
     * `half`
     * N/A
 - Validation and testing:
     * Existing tests + coverity
     * N/A

**JIRA TASK**: DALI-2166

### DIFF
--- a/dali/operators/reader/loader/nemo_asr_loader.h
+++ b/dali/operators/reader/loader/nemo_asr_loader.h
@@ -81,7 +81,7 @@ class AsrSample {
     return *decoder_;
   }
 
-  int64_t index_;
+  int64_t index_ = 0;
   std::string text_;
   AudioMetadata audio_meta_;
   std::string audio_filepath_;  // for tensor metadata purposes
--- a/include/dali/util/half.hpp
+++ b/include/dali/util/half.hpp
@@ -1016,12 +1016,10 @@ namespace half_float
       assign(static_cast<float>(rhs));
     }
 
-    // Copy constructor overriding the one above
-    // It avoids half->float->half path
+    /// Trivial copy constructor.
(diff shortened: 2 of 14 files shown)


JanuszL (approved):

dali/operators/reader/loader/nemo_asr_loader.h
```diff
@@ -81,7 +81,7 @@ class AsrSample {
     return *decoder_;
   }
 
-  int64_t index_;
+  int64_t index_ = 0;
```
mzient, line 84: This is a private field with a public getter - it should be initialized.

include/dali/util/half.hpp
```diff
@@ -1016,12 +1016,10 @@ namespace half_float
...
     }
 
-    // Copy constructor overriding the one above
-    // It avoids half->float->half path
+    /// Trivial copy constructor.
 	HALF_CONSTEXPR
     CAFFE_UTIL_HD
-    half(const half& rhs) :
-      data_(rhs.data_) {}
+    half(const half &) = default;
```
mzient, line 1022: GCC 9 protests when this type is copied via `memcpy` - it's not trivially copyable.

banasraf (approved):

PR #3128, opened 2021-07-06
Fix allocation hint handling in CUDA VM resource

mzient:

#### Why we need this PR?
*Pick one, remove the rest*
- It fixes a bug: allocation with a hint fails on some drivers

 - What solution was applied:
     * Try more hints
     * Fall back to allocation without a hint when allocation with a hint fails
     * CUDA VM resource
     * N/A
 - Validation and testing:
     * Existing tests apply
     * N/A

**JIRA TASK**: DALI-2033 (follow-up)

### DIFF
--- a/include/dali/core/mm/cuda_vm_resource.h
+++ b/include/dali/core/mm/cuda_vm_resource.h
@@ -331,22 +331,43 @@ class cuda_vm_resource : public memory_resource<memory_kind::device> {
   void va_allocate(size_t min_size) {
     size_t va_size = std::max(next_pow2(min_size), initial_va_size_);
 
-    CUdeviceptr hint = {};
-    va_region *last_va = nullptr;
+    cuvm::CUMemAddressRange va;
+
+    SmallVector<CUdeviceptr, 3> hints;
+
     if (va_regions_.empty()) {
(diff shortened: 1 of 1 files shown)


include/dali/core/mm/cuda_vm_resource.h
```diff
@@ -331,22 +331,43 @@ class cuda_vm_resource : public memory_resource<memory_kind::device> {
...
+      };
     } else {
-      // Try to allocate after the last VA for this device
-      last_va = &va_regions_.back();
-      hint = last_va->address_range.end();
+      // Try to allocate after the last VA for this device or before the first
+      auto &first_va = va_regions_.front();
+      auto &last_va = va_regions_.back();
+      assert(!va_ranges_.empty());
+      va_size = 2 * va_ranges_.back().size();
```
JanuszL, line 352: Where this formula came from?
Why you need to recalculate `va_size` in this case?

mzient: Oh, it's a bug - it should be `max`.

JanuszL (approved):

jantonguirao (approved):

PR #3130, opened 2021-07-06
Add operator `select` which selects a sample from one of the inputs

mzient:

#### Why we need this PR?
*Pick one, remove the rest*
- It adds new feature needed for sample masking/selection/conditional execution

 - What solution was applied:
     * A simple operator which just copies one of the inputs to the output
     * None (it's a new code)
     * N/A
 - Validation and testing:
     * Python unit tests for success
     * Python unit tests for errors
     * Local tests with Valgrind
     * Docstrings

**JIRA TASK**: N/A

### DIFF
--- a/dali/operators/generic/select.cc
+++ b/dali/operators/generic/select.cc
@@ -1,0 +1,99 @@
+// Copyright (c) 2021, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include <algorithm>
+#include "dali/operators/generic/select.h"
+
+namespace dali {
+
+DALI_SCHEMA(Select)
+  .DocStr(R"(Builds a batch by selecting each sample from one of the input batches.
+
+This operator is useful for conditionally selecting results of different operations.
+The shapes of the corresponding samples in the inputs may differ, but the number of dimensions
+and data type of all the inputs must be the same.)")
+  .NumInput(1, 99)
+  .NumOutput(1)
+  .AllowSequences()
+  .SupportVolumetric()
+  .AddArg("input_idx", R"(Index of the input to take the sample from.
+
+This argument contains (per-sample) indices that indicate from which input each
+sample is taken.
+Using a negative index will produce an empty tensor with the same number of dimensions as
+the inputs. Negative indices cannot be used with batches of scalars (0D tensors) since they
+can never be empty.
+)", DALI_INT32, true)
+  .AddOptionalArg<TensorLayout>("layout", R"(Layot string for the output.
+
+If not specified, the input layouts are checked and the first non-empty is used.)", nullptr);
+
+template <>
+void Select<CPUBackend>::RunImpl(HostWorkspace &ws) {
+  SetOutputLayout(ws);
+  TensorVector<CPUBackend> &out = ws.OutputRef<CPUBackend>(0);
+  const auto &out_shape = out.shape();
+  int N = out_shape.num_samples();
+  int64_t element_size = out.type().size();
+  int64_t total_size = out_shape.num_elements() * element_size;
+  const int64_t min_size = 16 << 10;
+
+  ThreadPool &tp = ws.GetThreadPool();
+  int min_blocks = tp.NumThreads() * 10;
+  int64_t block_size = std::max(total_size / min_size, min_size);
+  for (int i = 0; i < N; i++) {
+    auto sample_size = out_shape.tensor_size(i) * element_size;
+    if (sample_size == 0)
+      continue;
+    int idx = *input_idx_[i].data;
+    auto &inp = ws.InputRef<CPUBackend>(idx);
+    assert(idx >= 0 && idx < spec_.NumRegularInput());
+    int blocks = div_ceil(sample_size, block_size);
+    char *out_ptr = static_cast<char*>(out.raw_mutable_tensor(i));
+    const char *in_ptr = static_cast<const char*>(inp.raw_tensor(i));
+    ptrdiff_t start = 0;
+    for (int block = 0; block < blocks; block++) {
+      ptrdiff_t end = sample_size * (block + 1) / blocks;
+      tp.AddWork([in_ptr, out_ptr, start, end](int) {
+        memcpy(out_ptr + start, in_ptr + start, end - start);
+      }, end - start);
+      start = end;
+    }
+  }
+  tp.RunAll();
+}
+
+template <>
+void Select<GPUBackend>::RunImpl(DeviceWorkspace &ws) {
+  SetOutputLayout(ws);
+  TensorList<GPUBackend> &out = ws.OutputRef<GPUBackend>(0);
+  const auto &out_shape = out.shape();
+  int N = out_shape.num_samples();
+  int64_t element_size = out.type().size();
+  for (int i = 0; i < N; i++) {
+    auto sample_size = out_shape.tensor_size(i) * element_size;
+    if (sample_size == 0)
+      continue;
+    int idx = *input_idx_[i].data;
+    auto &inp = ws.InputRef<GPUBackend>(idx);
+    assert(idx >= 0 && idx < spec_.NumRegularInput());
+    sg_.AddCopy(out.raw_mutable_tensor(i), inp.raw_tensor(i), sample_size);
+  }
+  sg_.Run(ws.stream());
+}
+
+DALI_REGISTER_OPERATOR(Select, Select<CPUBackend>, CPU);
+DALI_REGISTER_OPERATOR(Select, Select<GPUBackend>, GPU);
+
+}  // namespace namespace dali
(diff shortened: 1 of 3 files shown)


dali/operators/generic/select.cc
```diff
@@ -0,0 +1,99 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include <algorithm>
+#include "dali/operators/generic/select.h"
+
+namespace dali {
+
+DALI_SCHEMA(Select)
+  .DocStr(R"(Builds a batch by selecting each sample from one of the input batches.
```
JanuszL, line 21:
```suggestion
  .DocStr(R"(Builds an output by selecting each sample from one of the inputs.
```

I don't think we need to stress out that the input/output is batch of samples.

dali/operators/generic/select.cc
```diff
@@ -0,0 +1,99 @@
...
+
+This operator is useful for conditionally selecting results of different operations.
+The shapes of the corresponding samples in the inputs may differ, but the number of dimensions
```
JanuszL, line 24: How about layouts?

dali/operators/generic/select.cc
```diff
@@ -0,0 +1,99 @@
...
+and data type of all the inputs must be the same.)")
+  .NumInput(1, 99)
+  .NumOutput(1)
+  .AllowSequences()
+  .SupportVolumetric()
+  .AddArg("input_idx", R"(Index of the input to take the sample from.
+
+This argument contains (per-sample) indices that point to the input from which each
+sample is taken.
+Providing a negative index will produce an empty tensor with the same number of dimensions as
```
JanuszL, line 34:
```suggestion
sample is taken.

Providing a negative index will produce an empty tensor with the same number of dimensions as
```
If you want to have a newline.

dali/operators/generic/select.cc
```diff
@@ -0,0 +1,99 @@
...
+    char *out_ptr = static_cast<char*>(out.raw_mutable_tensor(i));
+    const char *in_ptr = static_cast<const char*>(inp.raw_tensor(i));
+    ptrdiff_t start = 0;
+    for (int block = 0; block < blocks; block++) {
+      ptrdiff_t end = sample_size * (block + 1) / blocks;
+      tp.AddWork([in_ptr, out_ptr, start, end](int) {
+        memcpy(out_ptr + start, in_ptr + start, end - start);
+      }, end - start);
+      start = end;
+    }
```
JanuszL, line 72: I think this is a similar pattern used numpy reader. Maybe we can have a fucntion for this (even, make contiguous can use that in future).

dali/operators/generic/select.h
```diff
@@ -0,0 +1,107 @@
...
+
+  bool SetupImpl(vector<OutputDesc> &outputs, const workspace_t<Backend> &ws) override {
+    auto &inp0 = ws.template InputRef<Backend>(0);
+    int num_inp = spec_.NumRegularInput();
+    int sample_dim = inp0.sample_dim();
+    for (int i = 1; i < spec_.NumRegularInput(); i++) {
+      auto &inp = ws.template InputRef<Backend>(i);
+      DALI_ENFORCE(inp.type() == inp0.type(), make_string(
+        "All inputs must have the same type. "
+        "Got: ", inp0.type().id(), " and ", inp.type().id()));
```
JanuszL, line 44:
```suggestion
        "Got: ", inp0.type().id(), " and ", inp.type().id()));
```
Is `id()` returns a human readable name, or `name()` should be used.

dali/operators/generic/select.h
```diff
@@ -0,0 +1,107 @@
...
+
+      DALI_ENFORCE(inp.sample_dim() == sample_dim, make_string(
```
JanuszL, line 46: Shouldn't we check layouts as well?

dali/test/python/test_dali_cpu_only.py
```diff
@@ -891,6 +891,17 @@ def get_data():
         return out
     check_single_input(fn.squeeze, axis_names="YZ", get_data=get_data, input_layout="HWCYZ")
 
+def test_select_cpu():
```
JanuszL, line 894: I think you need to update test_dali_variable_batch_size.py as well.

dali/test/python/test_dali_cpu_only.py
```diff
@@ -891,6 +891,17 @@ def get_data():
...
+    pipe = Pipeline(batch_size=batch_size, num_threads=3, device_id=None)
+    data = fn.external_source(source=get_data, layout="HWC")
+    data2 = fn.external_source(source=get_data, layout="HWC")
+    data3 = fn.external_source(source=get_data, layout="HWC")
+    idx = fn.random.uniform(range=[0, 3], dtype=types.INT32)
```
JanuszL, line 899:
```suggestion
    idx = fn.random.uniform(range=[0, 2], dtype=types.INT32)
```

dali/test/python/test_operator_select.py
```diff
@@ -0,0 +1,157 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from numpy import random
+from numpy.core.fromnumeric import shape
+from nvidia.dali import Pipeline
```
JanuszL, line 17: Why not use decorator?

dali/test/python/test_operator_select.py
```diff
@@ -0,0 +1,157 @@
...
+            for ninp in (1, 2, 3, 4):
+                type = random.choice([dali.types.FLOAT, dali.types.UINT8, dali.types.INT32])
+                layout = make_layout(ndim, 'p') if np.random.randint(0, 2) else None
+                yield _test_select, ndim, ninp, type, 10, device, layout
+
+
+def test_error_inconsistent_ndim():
+    def check(device):
+        pipe = Pipeline(1, 3, 0)
+        pipe.set_outputs(fn.select(np.float32([0,1]), np.float32([[2,3,4]]), input_idx=2))
```
JanuszL, line 91:
```suggestion
        pipe.set_outputs(fn.select(np.float32([0,1]), np.float32([[2,3,4]]), input_idx=1))
```
To make sure that dimension problem is the only one and the order of error reporting doesn't matter here.

dali/test/python/test_operator_select.py
```diff
@@ -0,0 +1,157 @@
...
+        except RuntimeError as e:
+            assert "same number of dimensions" in str(e), "Unexpected exception"
+
+    for device in ["gpu", "cpu"]:
+        yield check, device
+
+def test_error_inconsistent_type():
+    def check(device):
+        pipe = Pipeline(1, 3, 0)
+        pipe.set_outputs(fn.select(np.float32([0,1]), np.int32([2,3,4]), input_idx=2))
```
JanuszL, line 105:
```suggestion
        pipe.set_outputs(fn.select(np.float32([0,1]), np.float32([[2,3,4]]), input_idx=1))
```
To make sure that dimension problem is the only one and the order of error reporting doesn't matter here.

dali/operators/generic/select.h
mzient: No; the semantics are that we pick the first non-empty layout - this sort-of implies that they can differ - and there are checks in the executor that require the non-empty input layouts to be of correct length.

dali/operators/generic/select.h
mzient: DALITypeId is formatted properly - I think it even uses the name consistent with PythonAPI whereas name() returns something else.

dali/test/python/test_dali_cpu_only.py
mzient: The regular test uses variable batch size.

dali/test/python/test_dali_cpu_only.py
mzient: Yup.

dali/test/python/test_operator_select.py
mzient: The decorator makes it easier to have a function that returns a pipeline (it does away with forwarding arguments, `with pipe`, `set_outputs`, etc - here there's no such function and forwarding arguments would likely make the code _more_ repetitive.

dali/test/python/test_operator_select.py
mzient: True.

dali/test/python/test_operator_select.py
mzient: It's about type:
```suggestion
        pipe.set_outputs(fn.select(np.float32([0,1]), np.int32([2,3,4]), input_idx=1))
```

dali/operators/generic/select.cc
mzient: Done. See below.

dali/operators/generic/select.cc
mzient: I guess we can factor it out then. NumpyReader is a bit more involved, though.

dali/test/python/test_operator_select.py
```diff
@@ -0,0 +1,157 @@
...
+import nvidia.dali.fn as fn
+import numpy as np
+import random
+import test_utils
+
+random.seed(1234)
+np.random.seed(1234)
+
+def generate_data(ndim, ninp, type, max_batch_size):
+    batch_size = np.random.randint(1, max_batch_size+1)
```
mzient, line 29: Variable batch size - voila!

dali/test/python/test_dali_cpu_only.py
JanuszL: We need as the test_dali_variable_batch_size.py checks if all operators are tested. There is no other way to enforce this kind of test.

dali/test/python/test_operator_select.py
JanuszL: Right. Just a copy paste from the previous example.

PR #3131, opened 2021-07-07
Throw std::bad_alloc when ordinary host memory runs out + tests for xxx_malloc resources.

mzient:

#### Why we need this PR?
*Pick one, remove the rest*
- It fixes a bug: no exception throw and nullptr returned from `malloc_memory_resource`

 - What solution was applied:
     * Short-circuit the case when no memory is requested
     * Otherwise check the returned pointer and throw if it's nullptr.
     * `cuda_malloc_resource` and its clients
     * N/A
 - Validation and testing:
     * N/A 
     * N/A

**JIRA TASK**: N/A

### DIFF
--- a/dali/core/mm/basic_resource_test.cc
+++ b/dali/core/mm/basic_resource_test.cc
@@ -1,0 +1,83 @@
+// Copyright (c) 2021, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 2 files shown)


dali/core/mm/basic_resource_test.cc
```diff
@@ -0,0 +1,83 @@
...
+};
+
+using BasicResourceTypes = ::testing::Types<
+  malloc_memory_resource,
+  cuda_malloc_memory_resource,
+  pinned_malloc_memory_resource>;
+TYPED_TEST_SUITE(MMBasicResourceTest, BasicResourceTypes);
+
+
+TYPED_TEST(MMBasicResourceTest, Allocate) {
```
JanuszL, line 69: How about testing `0` size allocation?

JanuszL (approved):

dali/core/mm/basic_resource_test.cc
mzient: Meh... That's a theoretical use case. The only place I ever saw 0-sized allocation were usages of `realloc`, which can be used as `free`, but `memory_resource` doesn't have such concept.

banasraf (approved):

PR #3137, opened 2021-07-08
Add ability to pass arguments to generator function in external_source op

szalpal:

#### Why we need this PR?
*Pick one, remove the rest*

- It adds new feature: Ability to provide argument list to a generator function in external_source op.

The need for this feature emerged, when I was migrating from old object-API to the new fn-API. Previously,
since the Pipeline class had a state, it was possible to use member variables as "arguments" in the generator function.
With the fn API we'd need to use global variables. However, using an additional argument to `external_source` operator, we can pass the arguments to the generator function within the `external_source`.

 - What solution was applied:
     Added an argument
     Python layer in the external_source operator
     NA
 - Validation and testing:
     Yes
     Yes

### DIFF
--- a/dali/python/nvidia/dali/external_source.py
+++ b/dali/python/nvidia/dali/external_source.py
@@ -69,20 +69,20 @@ class _CycleIter:
 
 
 class _CycleGenFunc:
-    def __init__(self, gen_func, mode, source_args={}):
+    def __init__(self, gen_func, mode, gen_func_args={}):
         self.source = gen_func
-        self.source_args = source_args
+        self.gen_func_args = gen_func_args
         self.signaling = (mode == "raise")
 
     def __iter__(self):
-        self.it = iter(self.source(**self.source_args))
+        self.it = iter(self.source(**self.gen_func_args))
         return self
 
     def __next__(self):
         try:
             return next(self.it)
         except StopIteration:
-            self.it = iter(self.source(**self.source_args))
+            self.it = iter(self.source(**self.gen_func_args))
             if self.signaling:
                 raise
             else:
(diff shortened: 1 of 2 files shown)


szalpal:
!build

dali/python/nvidia/dali/external_source.py
```diff
@@ -364,6 +371,11 @@ class ExternalSource():
     ``iter(source)`` returns a fresh iterator on each call, or a generator function.
     In the latter case, the generator function is called again when more data than was
     yielded by the function is requested.
+    
+    Please note: when the ``source`` is a generator function, and the cycling is turned on (either
```
mzient, line 375:
```suggestion
    Please note that when the ``source`` is a generator function, and the cycling is turned on (either
```

dali/python/nvidia/dali/external_source.py
```diff
@@ -67,25 +67,28 @@ def __next__(self):
...
+
+class _CycleGenFunc:
+    def __init__(self, gen_func, mode, gen_func_args={}):
         self.source = gen_func
+        self.gen_func_args = gen_func_args
         self.signaling = (mode == "raise")
 
     def __iter__(self):
-        self.it = iter(self.source())
+        self.it = iter(self.source(**self.gen_func_args))
```
mzient, line 78: This will work only with keyword arguments.

mzient (changes requested):
I think that this feature is introduced to satisfy a particular need without considering its rather broad impact on External Source API.
1. It's incomplete - it works only with keyword arguments.
2. It gives a bit too much of special treatment for generator functions
3. It's redundant. This can be achieved by simply doing this:
```Python
inp = fn.external_source(source=lambda: generator(arg1, arg2, kwarg=...))
```

PR #3143, opened 2021-07-08
Add support for functools.partial in ExternalSource.

mzient:

#### Why we need this PR?
*Pick one, remove the rest*
- It fixes a bug: infinite recursion when `functools.partial` is passed to ExternalSource
- It adds new feature: recognize partial invocations of generator functions

 - What solution was applied:
     * Check whether a callable is `functools.partial` and if so, use its `func` field instead to check if it's a generator function
     * Prevent recursion when `__call__` field points to self
     * External source Python wrapper
     * N/A
 - Validation and testing:
     * Python test
     * N/A

**JIRA TASK**: N/A

### DIFF
--- a/dali/python/nvidia/dali/external_source.py
+++ b/dali/python/nvidia/dali/external_source.py
@@ -15,8 +15,8 @@
 # custom wrappers around ops
 from nvidia.dali import backend as _b
 import inspect
+import functools
 import nvidia.dali.types
-
 
 def _get_batch_shape(data):
     if isinstance(data, (list, tuple, _b.TensorListCPU, _b.TensorListGPU)):
@@ -222,9 +222,14 @@ def _is_generator_function(x):
     where __call__ is a generator function"""
     if inspect.isgeneratorfunction(x):
         return True
+    if isinstance(x, functools.partial):
+        return _is_generator_function(x.func)
+
     if x is None or inspect.isfunction(x) or inspect.ismethod(x):
         return False
     call = getattr(x, "__call__", None)
+    if call == x:
+        return False
     return _is_generator_function(call)
 
 def _cycle_enabled(cycle):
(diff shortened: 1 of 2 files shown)


mzient:
!build

dali/python/nvidia/dali/external_source.py
```diff
@@ -222,9 +222,14 @@ def _is_generator_function(x):
...
     if inspect.isgeneratorfunction(x):
         return True
+    if isinstance(x, functools.partial):
+        return _is_generator_function(x.func)
+
     if x is None or inspect.isfunction(x) or inspect.ismethod(x):
         return False
     call = getattr(x, "__call__", None)
+    if call == x:
+        return False
```
mzient, line 232: Without the lines 225-227 we got here. Then the functools wrapper ended up having `__call__` point to the enclosing object which triggered infinite recursion. This condition prevents this from happening for in other cases. Note that the worst thing that can happen here is that we miss a generator function - much less problematic than crashing on an otherwise good callback.

dali/python/nvidia/dali/external_source.py
```diff
@@ -222,9 +222,14 @@ def _is_generator_function(x):
     where __call__ is a generator function"""
     if inspect.isgeneratorfunction(x):
         return True
+    if isinstance(x, functools.partial):
+        return _is_generator_function(x.func)
```
mzient, line 226: The field/property `func` comes from:
https://docs.python.org/3/library/functools.html#functools.partial

dali/python/nvidia/dali/external_source.py
JanuszL: What happens if we miss generator function here?

JanuszL: Are there any other cases where call==x?

mzient: > What happens if we miss generator function here?

It won't work as a generator function - so the particular use case is not handled


> Are there any other cases where call==x?

None so far, but I'd venture a guess that there might be something implemented in a similar way that `functools.partial` was.

JanuszL (approved):

banasraf (approved):

PR #3147, opened 2021-07-10
Fix some errors

cyyever:
#### Why we need this PR?
*Pick one, remove the rest*
- It fixes a bug

 - What solution was applied:
     *[ Fix some errors found by analyzer. ]*

**JIRA TASK**: *[NA]*

### DIFF
--- a/dali/operators/reader/loader/loader.h
+++ b/dali/operators/reader/loader/loader.h
@@ -69,7 +69,7 @@ class Loader {
       device_id_(options.GetArgument<int>("device_id")),
       skip_cached_images_(options.GetArgument<bool>("skip_cached_images")),
       lazy_init_(options.GetArgument<bool>("lazy_init")),
-      loading_flag_(ATOMIC_FLAG_INIT),
+      loading_flag_(false),
       read_sample_counter_(0),
       returned_sample_counter_(0),
       pad_last_batch_(options.GetArgument<bool>("pad_last_batch")),
@@ -213,8 +213,13 @@ class Loader {
   virtual void ReadSample(LoadTarget& tensor) = 0;
 
   void PrepareMetadata() {
-    if (!loading_flag_.test_and_set()) {
+    if (!loading_flag_) {
+      std::lock_guard<std::mutex> l(prepare_metadata_mutex_);
+      if (!loading_flag_) {
         PrepareMetadataImpl();
+        std::atomic_thread_fence(std::memory_order_release);
+        loading_flag_ = true;
+      }
     }
   }
 
@@ -342,8 +347,9 @@ class Loader {
 
   // Indicate whether the dataset preparation has to be done in the constructor or during the
   // first run
+  std::mutex prepare_metadata_mutex_;
   bool lazy_init_;
-  std::atomic_flag loading_flag_;
+  bool loading_flag_;
 
   // Image cache
   std::once_flag fetch_cache_;
--- a/dali/operators/reader/loader/loader.h
+++ b/dali/operators/reader/loader/loader.h
@@ -69,6 +69,7 @@ class Loader {
       device_id_(options.GetArgument<int>("device_id")),
       skip_cached_images_(options.GetArgument<bool>("skip_cached_images")),
       lazy_init_(options.GetArgument<bool>("lazy_init")),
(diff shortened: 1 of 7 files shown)


dali/operators/reader/loader/numpy_loader.cc
```diff
@@ -237,7 +237,7 @@ void NumpyLoader::ReadSample(NumpyFileWrapper& target) {
     target.data.SetMeta(meta);
     target.data.set_type(TypeInfo::Create<uint8_t>());
     target.data.Resize({0});
-    target.filename = "";
+    target.filename.clear();
```
mzient, line 240: A personal opinion: hile possibly clear is faster, I think that `= ""`.

dali/operators/reader/nvdecoder/nvdecoder.cc
```diff
@@ -244,15 +244,15 @@ int NvDecoder::handle_decode_(CUVIDPICPARAMS* pic_params) {
...
-}
-
 NvDecoder::MappedFrame::MappedFrame(CUVIDPARSERDISPINFO* disp_info,
                                     CUvideodecoder decoder,
                                     CUstream stream)
     : disp_info{disp_info}, valid_{false}, decoder_(decoder), params_{0} {
 
+  if (!disp_info) {
+    DALI_FAIL("disp_info is nullptr.");
+  }
```
mzient, line 254: That's a coding error, not a run-time error. If anything, it should be:
```suggestion
  assert(disp_info_);
```

dali/pipeline/operator/argument.h
```diff
@@ -122,7 +122,7 @@ class Argument {
 
   inline void clear_name() {
     has_name_ = false;
-    name_ = "";
+    name_.clear();
```
mzient, line 125: Same here - I think that `= ""` reads better.

include/dali/core/fast_div.h
```diff
@@ -212,7 +212,7 @@ __host__ __device__ __forceinline__ uint32_t operator/(uint32_t x, fast_div<uint
   return x >> y.shift;
 #else
   if (y.mul) {
-    uint32_t hi = static_cast<uint64_t>(x + y.add) * y.mul >> 32;
+    uint32_t hi = (static_cast<uint64_t>(x) + y.add) * y.mul >> 32;
```
mzient, line 215: It's intentional overflow. Please revert.

dali/operators/reader/loader/loader.h
```diff
@@ -213,14 +212,7 @@ class Loader {
...
   void PrepareMetadata() {
-    if (!loading_flag_) {
-      std::lock_guard<std::mutex> l(prepare_metadata_mutex_);
-      if (!loading_flag_) {
-        PrepareMetadataImpl();
-        std::atomic_thread_fence(std::memory_order_release);
-        loading_flag_ = true;
-      }
-    }
+    std::call_once(loading_flag_, &Loader::PrepareMetadataImpl, this);
```
mzient, line 215: It might be elegant, but the glibc implementation of call_once is
a) bugged - it fails when used before threading library is initialized
b) slow
This function is invoked for each sample. The check-lock-check idiom is 4x faster - but what's even more damning, is the fact that call_once uses atomics at the first line - this is not necessary and it causes a system-wide bus lock to be issued. That's an overkill on multi-socket systems which we target.
I've written a toy example where these two methods are used a million times and a background thread is incrementing a volatile (but not atomic) variable. The background thread slows down around 12x due to bus locks when using call_once, but is not affected by double-checking a boolean.

cyyever: The old code had used  ```std::atomic_thread_fence(std::memory_order_release);```  Should we use ```         std::atomic_thread_fence(std::memory_order_acquire );``` before first checking loading_flag_?

mzient (approved):

klecki (approved):

dali/operators/reader/loader/numpy_loader.cc
mzient: FYI: It's the custom in this project that the it's the reviewer that marks a conversation/issue as resolved.

PR #3155, opened 2021-07-16
Error in self.feed

AyanKumarBhunia:
untimeError: [/opt/dali/dali/python/backend_impl.cc:147] Assert on "strides[i] == stride_from_shape" failed: Strided data not supported. Dimension 0 has stride 7680 whereas densely packed data of this shape would have a stride 768
Stacktrace (77 entries)

```
        (lq, gt) = self.data_iterator.next()
        self.feed_input(self.input_lq_data, lq)
        self.feed_input(self.input_gt_data, gt)
```

lq => list of numpy arrays.


klecki:
Hi @AyanKumarBhunia,
looking at the error you try to pass strided data to DALI which, as the error states, is not supported.

If you want to discuss the issue further, please create one in [Issues](https://github.com/NVIDIA/DALI/issues) tab. The PR are meant for code changes only.

Also, in this case we won't be able to answer more without better, self-contained repro code.

PR #3198, opened 2021-07-29
Libtar api implementation

barci2:
## Description
- [X] **New feature** (*non-breaking change which adds functionality*)

### What happened in this PR

Adds a c++ interface for interacting with tar files
#### Additional information

- Key points relevant for the review: Partially uses libtar library and thus requires adding linking settings for it

#### Merge prerequisites
https://github.com/NVIDIA/DALI_extra/pull/61
https://github.com/NVIDIA/DALI_deps/pull/15

### Tests
- [ ] Existing tests apply
- [ ] New tests added
  - [ ] Python tests
  - [X] GTests
  - [ ] Benchmark
  - [ ] Other
- [ ] N/A

### Documentation
- [ ] Documentation updated
  - [ ] Docstring
  - [ ] Doxygen
  - [ ] RST
  - [ ] Jupyter
  - [ ] Other
- [X] N/A

#### Requirements
- [X] N/A

**REQ IDs**: N/A

**JIRA TASK**: DALI-2220


dali/operators/reader/loader/CMakeLists.txt
```diff
@@ -40,16 +40,23 @@ if (BUILD_NVDEC)
...
-
 set(DALI_OPERATOR_TEST_SRCS ${DALI_OPERATOR_TEST_SRCS}
   "${CMAKE_CURRENT_SOURCE_DIR}/loader_test.cc"
   "${CMAKE_CURRENT_SOURCE_DIR}/sequence_loader_test.cc"
   "${CMAKE_CURRENT_SOURCE_DIR}/numpy_loader_test.cc")
 
 if (BUILD_LIBSND)
   set(DALI_OPERATOR_TEST_SRCS ${DALI_OPERATOR_TEST_SRCS}
-    "${CMAKE_CURRENT_SOURCE_DIR}/nemo_asr_loader_test.cc")
+   "${CMAKE_CURRENT_SOURCE_DIR}/nemo_asr_loader_test.cc")
```
mzient, line 50: ?

dali/operators/reader/loader/tar_utils.cc
```diff
@@ -0,0 +1,174 @@
...
+#include <algorithm>
+#include <cstdarg>
+#include <cstdlib>
+#include <string>
+#include <tuple>
+#include <utility>
+#include "dali/core/error_handling.h"
+
+namespace dali {
+namespace detail {
```
mzient, line 26: If these are not used outside of this translation unit, then use an anonymous namespace - it will guarantee that the symbols won't leak to other translation units.

dali/operators/reader/loader/tar_utils.cc
```diff
@@ -0,0 +1,174 @@
...
+  for (auto& instances_entry : instances) {
+    if (instances_entry != nullptr) {
+      continue;
+    }
+    instances_entry = archive;
+    instance_handle = &instances_entry - instances.data();
+  }
+  if (instance_handle == -1) {
+    instance_handle = instances.size();
+    instances.push_back(archive);
```
mzient, line 46: This makes accessing the array for reading non thread-safe. The vector should be allocated once and "big enough" or you could even use a plain array - growing while it's not empty it will invalidate the pointer and a race condition might occur between `push_back` and unguarded `operator[]`.

dali/operators/reader/loader/tar_utils.cc
```diff
@@ -0,0 +1,174 @@
...
+
+  std::lock_guard<std::mutex> instances_lock(instances_mutex);
+  instances[instance_handle] = nullptr;
+  instance_handle = other.instance_handle;
+  return *this;
+}
+
+TarArchive::~TarArchive() {
+  std::lock_guard<std::mutex> instances_lock(instances_mutex);
+  instances[instance_handle] = nullptr;
```
mzient, line 80: I think that just for symmetry there should be an `Unregister` (inline) function.

dali/operators/reader/loader/tar_utils.cc
```diff
@@ -0,0 +1,174 @@
...
+  }
+  if (instance_handle == -1) {
+    instance_handle = instances.size();
+    instances.push_back(archive);
+  }
+  return instance_handle;
+}
+
+TarArchive::TarArchive(std::unique_ptr<FileStream>&& stream)
+    : stream(std::forward<std::unique_ptr<FileStream>>(stream)),
```
mzient, line 52: This is incorrect usage of std::forward - this function is meant for use with templates, when you don't know whether you should move or not.

dali/operators/reader/loader/tar_utils.h
```diff
@@ -0,0 +1,61 @@
...
+#include <vector>
+#include "dali/core/common.h"
+#include "dali/util/file.h"
+
+namespace dali {
+namespace detail {
+class TarArchive {
+ public:
+  TarArchive() = default;
+  explicit TarArchive(std::unique_ptr<FileStream>&& stream);
```
mzient, line 30:
```suggestion
  explicit TarArchive(std::unique_ptr<FileStream> stream);
```
unique_ptr is move-only, so there can't be a copy and the caller needs to pass an r-value anyway. The reasons to use `&&` could be twofold, but I think neither applies:
1. Exception handling - if the function fails to complete (prior to moving out the contents of the pointer), the caller retains the original object and may handle the exception and enjoy the ownership of the object.
2. Conditional move - passing an object by `&&` enables the function to move out the contents from the original object, but does not mandate it - under some circumstances (defined by the particular use case) , the function may decide not to move out the object and the caller will retain ownership. If we're not interested in this scenario - or indeed, we want to guarantee that the object is moved out and either moved out by the callee or destroyed - then we should pass a move-only object by value, because this guarantees that the object will be first moved to the argument.

dali/operators/reader/loader/tar_utils.h
```diff
@@ -0,0 +1,61 @@
...
+  std::unique_ptr<FileStream> stream;
+  bool eof = true;
+  std::string filename = "";
+  uint64 filesize = 0;
+  uint64 readoffset = 0;
+
+  int instance_handle = -1;
+  bool ParseHeader();
+
+  uint64 archiveoffset = 0;
```
mzient, line 55: I know we have these typedefs, but they are rather obsolete and we don't use them, in general.
```suggestion
  uint64_t archiveoffset = 0;
```

dali/operators/reader/loader/tar_utils.h
```diff
@@ -0,0 +1,61 @@
...
+  TarArchive& operator=(TarArchive&&);
+
+  bool NextFile();
+  bool IsAtFile() const;
+
+  std::string GetFileName() const;
+  uint64 GetFileSize() const;
+
+  std::vector<uint8_t> Read();
+  std::vector<uint8_t> Read(uint64 count);
```
mzient, line 42: It would be better to read to a user-supplied buffer. It will save an allocation (and likely an expensive one, if it's the size of an entire file) and the memory we read to might have some other fancy properties (CUDA managed or page-locked or whatever) which std::vector can't provide.

dali/operators/reader/loader/tar_utils_test.cc
```diff
@@ -0,0 +1,215 @@
...
+  vector<SimpleTarTestsData> values;
+
+  SimpleTarTestsData filepaths[] = {
+      {dali::filesystem::join_path(std::getenv("DALI_EXTRA_PATH"),
+                                   "db/webdataset/MNIST/devel-0.tar"),
+       false,
+       false,
+       2000,
+       3000,
+       {".cls", ".jpg"}},
```
mzient, line 134: Nitpick - this formatting is weird.
```suggestion
      { dali::filesystem::join_path(std::getenv("DALI_EXTRA_PATH"),
                                    "db/webdataset/MNIST/devel-0.tar"),
        false,
        false,
        2000,
        3000,
        {".cls", ".jpg"}},
```
^^^ with that at least you have an even number of spaces at the beginning of each line.

dali/operators/reader/loader/tar_utils_test.cc
```diff
@@ -0,0 +1,215 @@
...
+  for (int idx = 0; idx < kMultithreadedSamples; idx++) {
+    threads[idx] = std::async(
+        std::launch::async,
+        [](std::unique_ptr<TarArchive>& archive, int beg, int end) {
+          TestArchiveEntries(*archive, {".cls", ".jpg"}, beg, end, GetParam());
+        },
+        std::ref(archives[idx]), ranges[idx].first, ranges[idx].second);
+  }
+
+  std::for_each(threads, threads + kMultithreadedSamples, [](auto& thread) { thread.wait(); });
```
mzient, line 207:
```suggestion
  for (auto &thread : threads)
    thread.wait();
```
seems more readable.

dali/operators/reader/loader/tar_utils_test.cc
```diff
@@ -0,0 +1,215 @@
...
+        filepath_prefix + "0.tar", filepath_prefix + "1.tar", filepath_prefix + "2.tar"};
+
+    for (int idx = 0; idx < kMultithreadedSamples; idx++) {
+      archives[idx] = std::make_unique<TarArchive>(FileStream::Open(filepaths[idx], false, false));
+    }
+  }
+};
+
+TEST_P(MultiTarTests, Index) {
+  std::future<void> threads[kMultithreadedSamples];
```
mzient, line 196: Why call them threads instead of, say, tasks? I don't think std::async guarantees them to be in separate threads.

dali/operators/reader/loader/tar_utils_test.cc
```diff
@@ -0,0 +1,215 @@
...
+  TarArchive archive(TarArchive(TarArchive(FileStream::Open(filepath, false, false))));
+  std::string name_prefix(128, '#');
+  for (int idx = 0; idx < 1000; idx++) {
+    ASSERT_EQ(archive.GetFileName(), name_prefix + to_string(idx));
+    archive.NextFile();
+  }
+}
+
+void TestArchiveEntries(TarArchive& archive, const std::vector<std::string>& prefixes, int beg,
+                        int end, bool preread) {
```
mzient, line 59: Nitpick - when you don't have each argument in separate line (or all arguments in on line), it's good to keep some logical grouping - like `beg` and `end` in the same line.

dali/operators/reader/loader/tar_utils_test.cc
```diff
@@ -0,0 +1,215 @@
...
+  for (int idx = beg; idx < end; idx++) {
+    for (int prefix_idx = 0; prefix_idx < static_cast<int>(prefixes.size()); prefix_idx++) {
+      if (preread) {
+        archive.Read();
+      }
+      ASSERT_EQ(archive.GetFileName(), to_string(idx) + prefixes[prefix_idx]);
+      bool tmp = archive.NextFile() ^
+                 (idx == end - 1 && prefix_idx == static_cast<int>(prefixes.size()) - 1);
+      ASSERT_TRUE(tmp);
```
mzient, line 68: I'd rather put this condition (however long) in `ASSERT_TRUE`, so you can see it when it fails - otherwise it's not very informative.

dali/operators/reader/loader/tar_utils_test.cc
```diff
@@ -0,0 +1,215 @@
...
+  for (int idx = 0; idx < 1000; idx++) {
+    ASSERT_EQ(archive.GetFileName(), name_prefix + to_string(idx));
+    archive.NextFile();
+  }
+}
+
+void TestArchiveEntries(TarArchive& archive, const std::vector<std::string>& prefixes, int beg,
+                        int end, bool preread) {
+  for (int idx = beg; idx < end; idx++) {
+    for (int prefix_idx = 0; prefix_idx < static_cast<int>(prefixes.size()); prefix_idx++) {
```
mzient, line 61: Is there any reason to not use `size_t` here? I don't like unsigned `size_t`, but you have two code uglifying casts (here and in the codition). Alternatively, you can do:
```suggestion
    for (int prefix_idx = 0, num_prefixes = prefixes.size(); prefix_idx < num_prefixes; prefix_idx++) {
```
and use num_prefixes - but in this instance I think that `size_t` is more practical.

dali/operators/reader/loader/CMakeLists.txt
barci2: Removed a space by mistake, Imma put it back

dali/operators/reader/loader/tar_utils.cc
mzient: No, I mean the link symbols. `namespace {` has special semantics, making everything inside private to the translation units. I raise this concern bacause use global variables and highly collision-prone names like `instances` or `instances_mutex`. If you don't place them in an anonymous namespace (or specify internal linkage with `static` keyword), then these names will be visible in the whole library (dali/operators) and there would be conflicts, should someone use the same names elsewhere.

dali/operators/reader/loader/tar_utils.cc
mzient: If you insist on having this rvalue reference, then use `move`.

dali/operators/reader/loader/tar_utils_test.cc
mzient: I'll repeat it once more: code formatting is for _people_ (or Python ;)) - a compiler would be just as happy with random whitespace.

dali/operators/reader/loader/tar_utils_test.cc
mzient: And what about `size_t`?

dali/operators/reader/loader/tar_utils_test.cc
mzient: OMG. Not everyone uses it - and editing such code is unnecessarily hard.

dali/operators/reader/loader/tar_utils_test.cc
mzient: I think they can use threads from a thread pool and there can be fewer of them than there are tasks.

dali/operators/reader/loader/tar_utils.h
mzient: You wouldn't.

dali/operators/reader/loader/tar_utils_test.cc
barci2: No I mean that if I I have:
`for (int prefix_idx = 0; prefix_idx < prefixes.size(); prefix_idx++) `
then it throws warnings

dali/operators/reader/loader/tar_utils.h
```diff
@@ -0,0 +1,61 @@
...
+  TarArchive& operator=(TarArchive&&);
+
+  bool NextFile();
+  bool IsAtFile() const;
+
+  std::string GetFileName() const;
+  uint64_t GetFileSize() const;
+
+  std::shared_ptr<uint8_t> ReadFile();
+  uint64_t Read(uint8_t* buffer, uint64_t count);
```
mzient, line 42:
```suggestion
  size_t Read(uint8_t* buffer, size_t count);
```
This is consistent with FileStream API.

dali/operators/reader/loader/tar_utils.h
```diff
@@ -0,0 +1,61 @@
...
+  ~TarArchive();
+  TarArchive& operator=(TarArchive&&);
+
+  bool NextFile();
+  bool IsAtFile() const;
+
+  std::string GetFileName() const;
+  uint64_t GetFileSize() const;
+
+  std::shared_ptr<uint8_t> ReadFile();
```
mzient, line 41:
```suggestion
  std::shared_ptr<void> ReadFile();
```
I'd suggest using `<void>`, since that's what `FileStream::Get` returns.

barci2: does it make any difference other than that you have to cast it when you want to use it like that?

dali/operators/reader/loader/tar_utils.cc
```diff
@@ -0,0 +1,189 @@
...
+uint64_t TarArchive::GetFileSize() const {
+  return filesize;
+}
+
+std::shared_ptr<uint8_t> TarArchive::ReadFile() {
+  archiveoffset -= readoffset;
+  readoffset = 0;
+  stream->Seek(archiveoffset);
+  auto buffer = std::shared_ptr<uint8_t>(new uint8_t[filesize], [](uint8_t* ptr) { delete[] ptr; });
+  Read(buffer.get(), filesize);
```
mzient, line 124: That's definitely not what we want.
```suggestion
  buffer = stream->Get(filesize);
  stream->Seek(/* whatever the offset to the next header is */);
  return buffer;
```

dali/operators/reader/loader/tar_utils.h
```diff
@@ -0,0 +1,61 @@
...
+  bool eof = true;
+  std::string filename = "";
+  uint64_t filesize = 0;
+  uint64_t readoffset = 0;
+
+  int instance_handle = -1;
+  bool ParseHeader();
+
+  uint64_t archiveoffset = 0;
+  void Skip(int64 count);
```
mzient, line 56:
```suggestion
  void Skip(int64_t count);
```

dali/operators/reader/loader/tar_utils.h
```diff
@@ -0,0 +1,61 @@
...
+  uint64_t GetFileSize() const;
+
+  std::shared_ptr<uint8_t> ReadFile();
+  uint64_t Read(uint8_t* buffer, uint64_t count);
+  bool Eof() const;
+
+ private:
+  std::unique_ptr<FileStream> stream;
+  bool eof = true;
+  std::string filename = "";
```
mzient, line 48:
```suggestion
  std::string filename;
```
The default constructor will do the trick.

dali/operators/reader/loader/tar_utils.cc
```diff
@@ -0,0 +1,189 @@
...
+                    },
+                    [](int, const void*, size_t) -> ssize_t { return 0; }};
+  tar_open(&handle, "", &type, 0, instance_handle, TAR_GNU);
+
+  int errorcode = th_read(handle);
+  if (errorcode) {
+    filename = "";
+    filesize = 0;
+    if (errorcode == -1) {
+      DALI_FAIL(R"(Corrupted tar file at )" + handle->pathname);
```
mzient, line 175:
```suggestion
      DALI_FAIL("Corrupted tar file at " + handle->pathname);
```
No need for a raw string here.

dali/operators/reader/loader/tar_utils.cc
barci2: but for some readers it's not implemented (like for StdFileStream)

dali/operators/reader/loader/tar_utils.h
barci2: alright

dali/operators/reader/loader/tar_utils.cc
barci2: alright

dali/operators/reader/loader/tar_utils.h
barci2: alright

dali/operators/reader/loader/tar_utils.cc
```diff
@@ -0,0 +1,189 @@
...
+constexpr uint64_t kTarArchiveBufferInitSize = 1;
+
+std::mutex instances_mutex;
+std::list<std::vector<TarArchive*>> instances_registry = {
+    std::vector<TarArchive*>(kTarArchiveBufferInitSize)};
+TarArchive** instances = instances_registry.back().data();
+
+}  // namespace
+
+int Register(TarArchive* archive) {
```
mzient, line 42: Move this to the anonymous namespace or add static / inline.

dali/operators/reader/loader/tar_utils.h
barci2: alright

dali/operators/reader/loader/tar_utils.cc
```diff
@@ -0,0 +1,189 @@
...
+
+  std::lock_guard<std::mutex> instances_lock(instances_mutex);
+  instances[instance_handle] = nullptr;
+  instance_handle = other.instance_handle;
+  return *this;
+}
+
+uint64_t RoundToBlockSize(uint64_t count) {
+  return ((count + kBlockSize - 1) / kBlockSize) * kBlockSize;
+}
```
mzient, line 96: use `align_up(count, kBlockSize)` - it only works for pow2, but it's faster and already there.

dali/operators/reader/loader/tar_utils.cc
```diff
@@ -0,0 +1,189 @@
...
+  auto buffer = std::shared_ptr<uint8_t>(new uint8_t[filesize], [](uint8_t* ptr) { delete[] ptr; });
+  Read(buffer.get(), filesize);
+  return buffer;
+}
+
+uint64_t TarArchive::Read(uint8_t* buffer, uint64_t count) {
+  if (eof) {
+    return 0;
+  }
+  count = std::max(std::min(count, filesize - readoffset), 0_u64);
```
mzient, line 132:
```suggestion
  count = clamp(count, 0_u64, filesize - readoffset, 0_u64);
```

barci2: alright

dali/operators/reader/loader/tar_utils.cc
barci2: alright

dali/operators/reader/loader/tar_utils.cc
barci2: alright

dali/operators/reader/loader/tar_utils.cc
```diff
@@ -0,0 +1,196 @@
...
+  return filesize;
+}
+
+std::shared_ptr<void> TarArchive::ReadFile() {
+  archiveoffset -= readoffset;
+  readoffset = 0;
+  stream->Seek(archiveoffset);
+  auto out = stream->Get(filesize);
+  if (out != nullptr) {
+    archiveoffset += readoffset = filesize;
```
mzient, line 130: Fancy as it is, please split it into two lines for readability.

barci2: done

dali/operators/reader/loader/tar_utils.h
```diff
@@ -0,0 +1,61 @@
...
+  bool eof = true;
+  std::string filename;
+  uint64_t filesize = 0;
+  uint64_t readoffset = 0;
+
+  int instance_handle = -1;
+  bool ParseHeader();
+
+  uint64_t archiveoffset = 0;
+  void Skip(int64_t count);
```
mzient, line 56: Could you logically group fields and functions? For a moment I thought that these are some pairs of logically tied entities.

barci2: done

mzient: While you're at it - you can also group by size to enable more efficient packing.
For example - efficient packing
```
offset size   name
     0    8  stream
     8   32  filename
    40    8  filesize
    48    8  readoffset
    56    8  archiveoffset
    64    4  instance_handle
    68    1  eof
    69    3  (padding to structure alignment, which is 8)
    72 --- total size
```
Inefficient packing:
```
offset size   name
     0    8  stream
     8    1  eof
     9    7  (padding to stream alignment)
    16   32  filename
    48    8  filesize
    56    8  readoffset
    64    8  archiveoffset
    72    4  instance_handle
    76    4  (padding to structure alignment, which is 8)
    80 --- total size
```

dali/operators/reader/loader/tar_utils.cc
```diff
@@ -0,0 +1,197 @@
...
+}
+
+TarArchive::~TarArchive() {
+  Unregister(instance_handle);
+}
+
+TarArchive& TarArchive::operator=(TarArchive&& other) {
+  stream = std::move(other.stream);
+  eof = other.eof;
+  filename = other.filename;
```
mzient, line 88:
```suggestion
  filename = std::move(other.filename);
```

dali/operators/reader/loader/tar_utils.cc
```diff
@@ -0,0 +1,197 @@
...
+  filesize = other.filesize;
+  readoffset = other.readoffset;
+  archiveoffset = other.archiveoffset;
+
+  std::lock_guard<std::mutex> instances_lock(instances_mutex);
+  instances[instance_handle] = nullptr;
```
mzient, line 94:
```suggestion
  instances[instance_handle] = this;
```
?

dali/operators/reader/loader/tar_utils.cc
```diff
@@ -0,0 +1,197 @@
...
+}  // namespace
+
+TarArchive::TarArchive(std::unique_ptr<FileStream> stream)
+    : stream(std::move(stream)), archiveoffset(0), instance_handle(Register(this)), eof(false) {
+  this->stream->Seek(0);
+  ParseHeader();
+}
+
+TarArchive::TarArchive(TarArchive&& other) {
+  *this = static_cast<TarArchive&&>(other);
```
mzient, line 78:
```suggestion
  *this = std::move(other);
```

dali/operators/reader/loader/tar_utils.cc
barci2: cause it gets moved and after getting moved this spot is going to be empty

dali/operators/reader/loader/tar_utils.cc
barci2: done

dali/operators/reader/loader/tar_utils.cc
```diff
@@ -0,0 +1,197 @@
...
+}
+
+TarArchive::~TarArchive() {
+  Unregister(instance_handle);
```
mzient, line 82:
```suggestion
  if (instance_handle >= 0)
    Unregister(instance_handle);
```

dali/operators/reader/loader/tar_utils.cc
barci2: done

dali/operators/reader/loader/tar_utils.cc
```diff
@@ -0,0 +1,197 @@
...
+  stream = std::move(other.stream);
+  eof = other.eof;
+  filename = other.filename;
+  filesize = other.filesize;
+  readoffset = other.readoffset;
+  archiveoffset = other.archiveoffset;
+
+  std::lock_guard<std::mutex> instances_lock(instances_mutex);
+  instances[instance_handle] = nullptr;
+  instance_handle = other.instance_handle;
```
mzient, line 95:
```suggestion
  instance_handle = other.instance_handle;
  other.instance_handle = -1;
```

mzient: It would also be nice to reset the other fields in the other object.

dali/operators/reader/loader/tar_utils.cc
barci2: done

dali/operators/reader/loader/tar_utils.cc
barci2: done

dali/operators/reader/loader/tar_utils.cc
```diff
@@ -0,0 +1,208 @@
...
+                    [](int, const void*, size_t) -> ssize_t { return 0; }};
+  tar_open(&handle, "", &type, 0, instance_handle, TAR_GNU);
+
+  int errorcode = th_read(handle);
+  if (errorcode) {
+    filename = "";
+    filesize = 0;
+    if (errorcode == -1) {
+      DALI_FAIL("Corrupted tar file at " + handle->pathname);
+    }
```
szalpal, line 195: You may use `DALI_ENFORCE` here

dali/operators/reader/loader/tar_utils_test.cc
```diff
@@ -0,0 +1,215 @@
...
+INSTANTIATE_TEST_CASE_P(LibTarUtilsTestParametrized, SimpleTarTests, SimpleTarTestsValues());
+
+
+constexpr int kMultithreadedSamples = 3;
+
+class MultiTarTests : public ::testing::TestWithParam<bool> {
+ protected:
+  std::unique_ptr<TarArchive> archives[kMultithreadedSamples];
+  const std::pair<int, int> ranges[kMultithreadedSamples] = {{2000, 3000}, {0, 1000}, {1000, 2000}};
+  void SetUp() override {
```
szalpal, line 183: That's a nitpick, but this should rather be `final`

dali/operators/reader/loader/tar_utils.h
```diff
@@ -0,0 +1,59 @@
...
+  size_t Read(uint8_t* buffer, size_t count);
+  bool Eof() const;
+
+ private:
+  std::unique_ptr<FileStream> stream;
+  std::string filename;
+  uint64_t filesize = 0;
+  uint64_t readoffset = 0;
+  uint64_t archiveoffset = 0;
+  int instance_handle = -1;
```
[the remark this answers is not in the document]
barci2: done

dali/operators/reader/loader/tar_utils.h
```diff
@@ -0,0 +1,59 @@
+// Copyright (c) 2021, NVIDIA CORPORATION. All rights reserved.
```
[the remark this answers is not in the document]
barci2: done

dali/operators/reader/loader/tar_utils.cc
```diff
@@ -0,0 +1,208 @@
...
+  *this = std::move(other);
+}
+
+TarArchive::~TarArchive() {
+  if (instance_handle >= 0) {
+    Unregister(instance_handle);
+  }
+}
+
+TarArchive& TarArchive::operator=(TarArchive&& other) {
```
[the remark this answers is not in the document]
barci2: thanks for pointing it out, done

dali/operators/reader/loader/CMakeLists.txt
```diff
@@ -52,4 +50,13 @@ if (BUILD_LIBSND)
     "${CMAKE_CURRENT_SOURCE_DIR}/nemo_asr_loader_test.cc")
 endif()
 
+if (BUILD_LIBTAR)
+  set(DALI_OPERATOR_SRCS ${DALI_OPERATOR_SRCS}
+    "${CMAKE_CURRENT_SOURCE_DIR}/tar_utils.cc")
+  set(DALI_OPERATOR_TEST_SRCS ${DALI_OEPERATOR_TEST_SRCS}
+    "${CMAKE_CURRENT_SOURCE_DIR}/tar_utils.cc"
+    "${CMAKE_CURRENT_SOURCE_DIR}/tar_utils_test.cc")
+endif()
```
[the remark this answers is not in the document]
barci2: and put the CMakeFiles.txt there for setting that?

dali/operators/reader/loader/tar_utils.cc
barci2: done

dali/operators/reader/loader/tar_utils_test.cc
barci2: Done

dali/operators/reader/loader/tar_utils.h
```diff
@@ -0,0 +1,59 @@
...
+#include <memory>
+#include <mutex>
+#include <string>
+#include <vector>
+#include "dali/core/common.h"
+#include "dali/util/file.h"
+
+namespace dali {
+namespace detail {
+class TarArchive {
```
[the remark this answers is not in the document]
barci2: done

dali/operators/reader/loader/CMakeLists.txt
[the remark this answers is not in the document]
mzient: Yes, but we have a copy-pasteable one.

dali/operators/reader/loader/tar_utils.h
```diff
@@ -0,0 +1,84 @@
...
+  explicit TarArchive(std::unique_ptr<FileStream> stream_);
+  TarArchive(TarArchive&&);
+  ~TarArchive();
+  TarArchive& operator=(TarArchive&&);
+
+  bool NextFile();
+  bool IsAtFile() const;
+
+  std::string GetFileName() const;
+  uint64_t GetFileSize() const;
```
mzient, line 64: Please provide documentation for these APIs - the name isn't immediately clear as to which file's name or size it returns - is it the current file object or the archive?

dali/operators/reader/loader/tar_utils.h
```diff
@@ -0,0 +1,84 @@
...
+
+  std::shared_ptr<void> ReadFile();
+  size_t Read(uint8_t* buffer, size_t count);
+  bool Eof() const;
```
mzient, line 68: Given that `Read` supports partial reading of file contents (and I think it would stop reading when it reaches the end of file), this name seems a bit misleading. Other instances of the word "file" here refer to the file object inside the tarball, but here it means the archive. Perhaps we should even have to methods:
```suggestion
  /// @brief Indicates that the current file object has been read to the end
  bool EndOfFile() const;
  /// @brief Indicates that there are no more file objects in the archive
  bool EndOfArchive() const;
```

barci2: done

dali/operators/reader/loader/tar_utils.h
barci2: done

dali/operators/reader/loader/webdataset/tar_utils.cc
```diff
@@ -0,0 +1,207 @@
...
+  return filename_;
+}
+
+uint64_t TarArchive::GetFileSize() const {
+  return filesize_;
+}
+
+std::shared_ptr<void> TarArchive::ReadFile() {
+  archiveoffset_ -= readoffset_;
+  int old_readoffset = readoffset_;
```
mzient, line 134:
```suggestion
  int64_t old_readoffset = readoffset_;
```

dali/operators/reader/loader/webdataset/tar_utils.cc
```diff
@@ -0,0 +1,207 @@
...
+
+    if (instance_handle_ >= 0) {
+      Unregister(instance_handle_);
+    }
+    instance_handle_ = other.instance_handle_;
+  }
+  return *this;
+}
+
+uint64_t RoundToBlockSize(uint64_t count) {
```
mzient, line 107:
```suggestion
constexpr uint64_t RoundToBlockSize(uint64_t count) {
```
or inline

dali/operators/reader/loader/webdataset/tar_utils.cc
```diff
@@ -0,0 +1,207 @@
...
+  readoffset_ += num_read_bytes;
+  archiveoffset_ += num_read_bytes;
+  return num_read_bytes;
+}
+
+bool TarArchive::EndOfFile() const {
+  return readoffset_ >= filesize_;
+}
+
+inline void TarArchive::Skip(int64 count) {
```
mzient, line 163:
```suggestion
inline void TarArchive::Skip(int64_t count) {
```

dali/operators/reader/loader/webdataset/tar_utils.h
```diff
@@ -0,0 +1,115 @@
...
+   * @brief Returns whether the file cursor is at the end of file.
+   */
+  bool EndOfFile() const;
+
+ private:
+  std::unique_ptr<FileStream> stream_;
+  std::string filename_;
+  uint64_t filesize_ = 0;
+  uint64_t readoffset_ = 0;
+  uint64_t archiveoffset_ = 0;
```
mzient, line 106: Nitpick: I think it would be safer if these values were signed. Unsigned integers can cause nasty surprises and it's not like we desperately need this extra bit of capacity.

dali/operators/reader/loader/webdataset/tar_utils.h
```diff
@@ -0,0 +1,115 @@
...
+  bool EndOfArchive() const;
+
+  /**
+   * @brief Returns the name of the file in the archive that is currently being viewed.
+   */
+  std::string GetFileName() const;
+  /**
+   * @brief Returns the size (in bytes) of the file in the archive that is currently being viewed.
+   */
+  uint64_t GetFileSize() const;
```
mzient, line 77:
```suggestion
  size_t GetFileSize() const;
```
?

dali/operators/reader/loader/webdataset/tar_utils.cc
```diff
@@ -0,0 +1,207 @@
...
+  tartype_t type = {LibtarOpenTarArchive, [](int) -> int { return 0; },
+                    [](int instance_handle_, void* buf, size_t count) -> ssize_t {
+                      const auto current_archive = instances[instance_handle_];
+                      const ssize_t num_read =
+                          current_archive->stream_->Read(reinterpret_cast<uint8_t*>(buf), count);
+                      current_archive->archiveoffset_ += num_read;
+                      return num_read;
+                    },
+                    [](int, const void*, size_t) -> ssize_t { return 0; }};
+  tar_open(&handle, "", &type, 0, instance_handle_, TAR_GNU);
```
mzient, line 188: This is a bit wasteful - I think that this logic should be done once and here there should only be th_read.

dali/operators/reader/loader/webdataset/tar_utils.h
```diff
@@ -0,0 +1,118 @@
...
+ *     - @ref GetFileName - Returns the name of the file the archive cursor currently points at.
+ *     - @ref GetFileSize - Returns the size of the archive cursor currently points at (in bytes).
+ *     - @ref ReadFile - Reads the contents of the file and returns them. In the case of success
+ *                       places the file cursor at the end of file. In the other case keeps the file
+ *                       cursor intact.
+ *     - @ref Read - Reads the given number of bytes into a given buffer, assuming that said buffer
+ *                   is big enough to perform this operation. Returns the number of bytes actually
+ *                   read.
+ *     - @ref EndOfFile - Returns whether the file cursor is at the end of file.
+ */
```
awolant, line 50: As methods are documented below I think that repeating it here does not add much value and makes the documentation harder to maintain. Maybe it should be removed?

dali/operators/reader/loader/webdataset/tar_utils_test.cc
```diff
@@ -0,0 +1,222 @@
...
+#include "dali/operators/reader/loader/filesystem.h"
+#include "dali/util/std_file.h"
+
+namespace dali {
+namespace detail {
+
+TEST(LibTarUtilsTestSimple, Interface) {
+  std::string filepath(dali::filesystem::join_path(std::getenv("DALI_EXTRA_PATH"),
+                                                   "db/webdataset/MNIST/devel-1.tar"));
+  TarArchive archive(TarArchive(TarArchive(FileStream::Open(filepath, false, false))));
```
awolant, line 34: If the purpose of this is to test move constructor I think it should be tested in separate test for better readability and in case of issues easier debugging.
Also I don't see the added value of multiple nested constructor calls in that context.

dali/operators/reader/loader/webdataset/tar_utils_test.cc
```diff
@@ -0,0 +1,222 @@
...
+  
+  archive.Read(buffer.data() + 10, filesize);
+  ASSERT_TRUE(archive.EndOfFile());
+  ASSERT_EQ(archive.GetFileSize(), filesize);
+}
+
+TEST(LibTarUtilsTestSimple, LongNameIndexing) {
+  std::string filepath(dali::filesystem::join_path(std::getenv("DALI_EXTRA_PATH"),
+                                                   "db/webdataset/sample-tar/gnu.tar"));
+  TarArchive archive(TarArchive(TarArchive(FileStream::Open(filepath, false, false))));
```
awolant, line 58: Same as above.

dali/operators/reader/loader/webdataset/tar_utils.cc
```diff
@@ -0,0 +1,224 @@
...
+TarArchive& TarArchive::operator=(TarArchive&& other) {
+  if (&other != this) {
+    stream_ = std::move(other.stream_);
+
+    tar_close(ToTarHandle(handle_));
+    handle_ = other.handle_;
+    other.handle_ = nullptr;
+    
+    filename_ = std::move(other.filename_);
+    other.filename_ = "";
```
mzient, line 128:
```suggestion
    other.filename_.clear();
```

dali/operators/reader/loader/webdataset/tar_utils.cc
```diff
@@ -0,0 +1,228 @@
...
+  instances[instance_handle_] = nullptr;
+}
+
+inline TAR* ToTarHandle(void* handle) {
+  return reinterpret_cast<TAR*>(handle);
+}
+
+inline TAR** ToTarHandle(void** handle) {
+  return reinterpret_cast<TAR**>(handle);
+}
```
awolant, line 75: It is a matter of taste but it is not obvious to me that these wrappers add any value. In places where they are used they obfuscate a fact that this is just a change of pointer type. Maybe you can consider removing them?

dali/operators/reader/loader/webdataset/tar_utils_test.cc
barci2: Yes, that is what the whole tests is about. The subsequent reads just test if the moving was successful.

dali/operators/reader/loader/webdataset/tar_utils_test.cc
barci2: Removed the moves

dali/operators/reader/loader/webdataset/tar_utils.cc
barci2: They are here to avoid lots of reinterpret_cast

mzient: @awolant It's not perfect, but probably better than having casts all over the place. Ideally we'd forward-declare `TAR`, but it has a C-style declaration which in C++ becomes an anonymous struct which cannot be forward declared:
```
typedef struct {... } TAR;
```

mzient (approved):

awolant (approved):

barci2:
!build

barci2:
!build

barci2:
!build

barci2:
!build

dali/operators/reader/loader/webdataset/CMakeLists.txt
```diff
@@ -0,0 +1,23 @@
...
+
+collect_headers(DALI_INST_HDRS PARENT_SCOPE)
+collect_sources(DALI_WEBDATASET_SRCS)
+collect_test_sources(DALI_OPERATOR_TEST_SRCS PARENT_SCOPE)
+set(DALI_OPERATOR_SRCS ${DALI_OPERATOR_SRCS} ${DALI_WEBDATASET_SRCS} PARENT_SCOPE)
+set(DALI_OPERATOR_TEST_SRCS ${DALI_OPERATOR_TEST_SRCS} ${DALI_WEBDATASET_SRCS}
+    "${DALI_ROOT}/dali/util/file.cc"
+    "${DALI_ROOT}/dali/util/std_file.cc"
+    "${DALI_ROOT}/dali/util/mmaped_file.cc"
+    PARENT_SCOPE)
```
szalpal, line 23: This does not look right. Those files are already compiled in `libdali.so`, there's no need to compile them again. What is the problem you wanted to solve?

barci2:
!build

dali/CMakeLists.txt
```diff
@@ -83,6 +83,7 @@ if (BUILD_CUFILE)
   target_link_libraries(dali PRIVATE dynlink_cufile)
 endif()
 
+###############################################
```
mzient, line 86: ?

dali/operators/reader/loader/CMakeLists.txt
```diff
@@ -40,8 +44,6 @@ if (BUILD_NVDEC)
     "${CMAKE_CURRENT_SOURCE_DIR}/video_loader.cc")
 endif()
 
-set(DALI_OPERATOR_SRCS ${DALI_OPERATOR_SRCS} PARENT_SCOPE)
```
mzient, line 43: ?

barci2:
!build

PR #3200, opened 2021-07-29
Add documentation for indexing.

mzient:

## Description
- [X] **Other** (*e.g. Documentation, Tests, Configuration*)

### What happened in this PR

It adds documentation for DataNode indexing.

#### Additional information

Documentation

### Tests
- [ ] Existing tests apply
- [ ] New tests added
  - [ ] Python tests
  - [ ] GTests
  - [ ] Benchmark
  - [ ] Other
- [X] N/A

### Documentation
- [ ] Documentation updated
  - [ ] Docstring
  - [ ] Doxygen
  - [X] RST
  - [ ] Jupyter
  - [ ] Other
- [ ] N/A

#### Requirements
- [X] N/A

**REQ IDs**: N/A

**JIRA TASK**: DALI-2219


docs/index.rst
```diff
@@ -38,8 +38,7 @@ NVIDIA DALI Documentation
 
    pipeline
    data_types
-   supported_ops
-   math
+   data_manip
```
JanuszL, line 41:
```suggestion
   data_manipulation
```
I don't think we need to shorten it here.

docs/index.rst
```diff
@@ -38,8 +38,7 @@ NVIDIA DALI Documentation
 
    pipeline
    data_types
-   supported_ops
```
JanuszL, line 41: Should we move supported_ops into data manipulation?

docs/indexing.rst
```diff
@@ -0,0 +1,123 @@
...
+The simplest case is scalar indexing with a constant index::
+
+    raw_files = fn.readers.file(...)
+    third_bytes = raw_files[2]
+
+The snippet above extracts the third (index 2) byte from the files.
+
+.. note::
+    The batch dimension is implicit and cannot be indexed. In this example, the index 2
+    is broadcast to the whole batch. The section :ref:`Run-time indices` for per-sample indexing.
```
JanuszL, line 20:
```suggestion
    is broadcast to the whole batch. Check the section :ref:`Run-time indices` for per-sample indexing.
```

docs/indexing.rst
```diff
@@ -0,0 +1,123 @@
...
+
+DALI data nodes can be indexed and sliced using familiar Python syntax for array indexing.
+
+Scalar indexing
+~~~~~~~~~~~~~~~
+
+The simplest case is scalar indexing with a constant index::
+
+    raw_files = fn.readers.file(...)
+    third_bytes = raw_files[2]
```
JanuszL, line 14: Maybe it is better to show the example with the shape in the tensor (and how easy is to extract w and h now) - is it more real life.

docs/indexing.rst
```diff
@@ -0,0 +1,123 @@
...
+
+The index can contain multiple axes, separated by comma. If an index is a scalar, the corresponding
+dimension is removed from the output::
+
+    image = fn.decoders.image(jpegs, device="mixed")  # RGB images in HWC layout
+    red =   image[:,:,0]
+    green = image[:,:,1]
+    blue =  image[:,:,2]
+
+The ``red``, ``green``, ``blue`` are 2D tensors.
```
JanuszL, line 72: So ```
    red =   image[:,:,:1]
    green = image[:,:,1:2]
    blue =  image[:,:,2:3]
```
will keep the dimensions, right?

File: docs/indexing.rst [hunk 6]
```diff
@@ -0,0 +1,123 @@
...
+    video = ... # a 4D tensor in FHWC layout
+    fifth_frame = video[5]         # equivalent to...
+    fifth_frame = video[5,:]       # ...equivalent to...
+    fifth_frame = video[5,:,:]     # ...equivalent to...
+    fifth_frame = video[5,:,:,:]
+
+Strided slices
+~~~~~~~~~~~~~~
+
+Indexing with a custom step is not implemented.
```
JanuszL, line 93:
```suggestion
Indexing with a custom step is not supported.
```

docs/indexing.rst
```diff
@@ -0,0 +1,123 @@
...
+    blue =  image[:,:,2]
+
+The ``red``, ``green``, ``blue`` are 2D tensors.
+
+Using ranges doesn't remove dimensions::
+
+    letterboxed = image[20:-20,:,:]   # the result is still a 3D HWC tensor
+
+.. note::
+    See also :func:`fn.crop` and :func:`fn.slice` for operations tailored for image processing.
```
JanuszL, line 79: Do we have any perf numbers comparing the indexing and slice?

docs/indexing.rst
```diff
@@ -0,0 +1,123 @@
...
+
+DALI tensors can have a layout specifier which affects how the data is interpreted by some
+operators - typically, an image would have ``"HWC"`` layout.
+
+When applying a scalar index to an axis, that axis is removed from the output along with the
+layout name for this axis::
+
+    image = ...             # layout HWC
+    first_row = image[0]    # layout WC
+    last_col = image[:,-1]  # lauout HWC
```
JanuszL, line 115:
```suggestion
    last_col = image[:,-1]  # layout HWC
```

docs/indexing.rst
mzient: Yes.

docs/indexing.rst
mzient: I'd prefer to call it "not implemented", because we intend to support it.

docs/indexing.rst
mzient: Hmm... what happened to my `see`?

docs/index.rst
mzient: That's what I'd like to discuss.

docs/indexing.rst
JanuszL: I think `implemented` is better for the code documentation, supported sounds more like a term from manual.

docs/indexing.rst
JanuszL: 🤷

docs/indexing.rst
klecki: This sounds like a good idea.

docs/indexing.rst
```diff
@@ -0,0 +1,123 @@
...
+
+Run-time indices
+~~~~~~~~~~~~~~~~
+
+Indexing by constant is often insufficient. With DALI, you can use a result of other computations
+to access tensor elements::
+
+    length = fn.shapes(raw_files)[0]
+    index = fn.cast(fn.random.uniform(range=(0, 1)) * length, dtype=dali.types.INT64)
+    random_byte = raw_files[index]
```
klecki, line 40: Won't we get values in range [0, length] due to rounding, thus doing and out of bounds access?

klecki: I also think we should focus on this line more, maybe first constructing it by hand or something? This random is a bit cryptic, and also there is this length already using subscript.

docs/indexing.rst
```diff
@@ -0,0 +1,123 @@
...
+    header = raw_files[:16]
+
+If the start of the slice is omitted, the slice starts at 0. If the end is omitted, the slice
+ends at the end of given axis. Negative indices can be used for both start and end of the slice.
+Either end can be a constant, a run-time value (a DataNode) or can be skipped.
+
+Multidimensional indexing
+~~~~~~~~~~~~~~~~~~~~~~~~~
+
+The index can contain multiple axes, separated by comma. If an index is a scalar, the corresponding
```
klecki, line 64: I'm thinking how easy one can confuse scalars and constants here. Maybe we can find some better wording? Even writing `non-range` might help.

Also, the "dimension removal" can be bit misleading imo, maybe we should add a sentence, that the data is there, we just don't keep the 1-element-sized dummy axis?

docs/indexing.rst
```diff
@@ -0,0 +1,123 @@
...
+dimension is removed from the output::
+
+    image = fn.decoders.image(jpegs, device="mixed")  # RGB images in HWC layout
+    red =   image[:,:,0]
+    green = image[:,:,1]
+    blue =  image[:,:,2]
```
klecki, line 70: General comment, I think we should add spaces after commas, as it's hard to sometimes separate dimensions visually, as the colons and commas stick together, especially if there are few of them side by side:
```suggestion
    red =   image[:, :, 0]
    green = image[:, :, 1]
    blue =  image[:, :, 2]
```

docs/indexing.rst
```diff
@@ -0,0 +1,123 @@
...
+
+The ``red``, ``green``, ``blue`` are 2D tensors.
+
+Using ranges doesn't remove dimensions::
+
+    letterboxed = image[20:-20,:,:]   # the result is still a 3D HWC tensor
```
klecki, line 76: Maybe it's worth to compare it to R, G, B slices but with the dummy dimension so:
```
    red =   image[:, :, 0:1]
    green = image[:, :, 1:2]
    blue =  image[:, :, 2:3]
```

docs/indexing.rst
klecki: not supported yet?

docs/indexing.rst
```diff
@@ -0,0 +1,123 @@
...
+size 1 in the output::
+
+    trailing_channel = grayscale[:,:,dali.newaxis]
+    leading_channel = grayscale[dali.newaxis]
+
+Layout specifiers
+~~~~~~~~~~~~~~~~~
+
+DALI tensors can have a layout specifier which affects how the data is interpreted by some
+operators - typically, an image would have ``"HWC"`` layout.
```
klecki, line 108: Could you maybe add a link to: https://docs.nvidia.com/deeplearning/dali/user-guide/docs/data_types.html#data-layouts

docs/indexing.rst
klecki:
```suggestion
    last_col = image[:,-1]  # layout HC
```

docs/index.rst
klecki: How about something like:
```suggestion
.. toctree::
   :hidden:
   :caption: Data Manipulation
   graph_building # Put some links to Pipeline doc and Getting started here
   indexing
   math
   supported_ops
```

klecki: To keep it top-level

docs/indexing.rst
```diff
@@ -0,0 +1,135 @@
...
+
+Indexing by constant is often insufficient. With DALI, you can use a result of other computations
+to access tensor elements::
+
+    raw_files = fn.readers.file(...)
+    length = fn.shapes(raw_files)[0]
+
+    # calculate a random index from 0 to file_length-1
+    random_01 = fn.random.uniform(range=(0, 1))  # random numbers in range [0..1)
+    index = fn.floor(random_01 * length)  # calculate indices from [0..length)
```
JanuszL, line 48: Or use cast?

docs/indexing.rst
JanuszL: On the other hand that is how python works as well.

JanuszL (approved):

docs/indexing.rst
klecki: It rounds.

docs/indexing.rst
mzient: Fixed.

docs/indexing.rst
mzient: I don't think there's any reason for a difference there - they use the same kernel. Indexing might be ever-so-slightly more efficient when collapsing is used, but I'm not pushing it too much.

docs/indexing.rst
mzient: Done.

docs/indexing.rst
JanuszL: Understood

docs/indexing.rst
mzient: @klecki We don't keep the axis. That''s the point.

mzient: I'll reword it to "indexing" (scalar) and "slicing" (range).

docs/indexing.rst
mzient: I still think that "not implemented" has a slightly different vibe.
For me, "not implemented" means that it's a "simple matter of programming" to add the feature and that we're aware of the incompleteness of the existing solution.
Conversely, "not supported" sounds like we're not just supporting it. "not supported **yet**" is more in the "nice to have" realm.

JanuszL: > For me, "not implemented" means that it's a "simple matter of programming" to add the feature and that we're aware of the incompleteness of the existing solution.

I don't think the user care why it doesn't work now and how difficult is it to add. All it matters is "we're not just supporting it.".

docs/indexing.rst
```diff
@@ -0,0 +1,139 @@
...
+Negative indices can be used to index the tensor starting from the end. The index of -1 denotes
+the last element::
+
+    channels = sizes[-1]   # channels go last
+    widths = sizes[-2]     # widths are the innermost dimension after channels
+
+Indexing with run-time values
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Indexing by constant is often insufficient. With DALI, you can use a result of other computations
```
awolant, line 45: Maybe `by a constant`. But I'm not sure, tbh.

docs/indexing.rst
```diff
@@ -0,0 +1,139 @@
...
+    header = raw_files[:16]  # extract 16-byte headers from files in the batch
+
+If the start of the slice is omitted, the slice starts at 0. If the end is omitted, the slice
+ends at the end of given axis. Negative indices can be used for both start and end of the slice.
+Either end can be a constant, a run-time value (a DataNode) or can be skipped.
+
+Multidimensional selection
+~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+The slection can contain multiple axes, separated by comma. If an index is a scalar, the
```
awolant, line 82: Again, maybe `by a comma` ?

docs/indexing.rst
```diff
@@ -0,0 +1,139 @@
...
+
+When indexing and slicing multidimensional data, the trailing dimenions can be omitted. This is
+equivalent to passing a full-range slice to all trailing dimensions::
+
+    wide = letterboxed[20:-20,:,:]   # slice height, keep width and channels
+    wide = letterboxed[20:-20,:]     # this line is equivalent to the previous one
+    wide = letterboxed[20:-20]       # ...and so is this one
+
+.. note::
+    See also :func:`fn.crop` and :func:`fn.slice` for operations tailored for image processing.
```
awolant, line 104: Maybe it would be useful to add very small code example on how you can replace slice with indexing somewhere in this doc?

awolant (approved):

docs/indexing.rst
mzient: True.

mzient: or rather commas?

JanuszL (approved):

Issue #3243, opened 2021-08-11
readers.video hangs when reading HEVC


castim:
Hi, 

When I run the following code

```python
from tqdm import tqdm

from nvidia import dali
import nvidia.dali.fn as fn

@dali.pipeline_def
def video_pipeline():
    seq = fn.readers.video(device="gpu", filenames="sintel_trailer-1080p-g5-hevc.mp4", sequence_length=3)
    return seq

pipe = video_pipeline(batch_size=16, num_threads=1, device_id=0)

pipe.build()

for i in tqdm(range(99999999)):
    pipe.run()
```
The loop gets stuck after 21 iterations.

I use the [sintel trailer 1080p](https://download.blender.org/durian/trailer/sintel_trailer-1080p.mp4) which I reencoded to get a smaller gop size using the HEVC codec with:
```
ffmpeg -i sintel_trailer-1080p.mp4 -c:v hevc -an sintel_trailer-1080p-g5-hevc.mp4
```

I use DALI 1.4.0 with cuda 10 (also tested with cuda 11).

With H264, it works fine, and it is also much (6-7x) faster.
Is HEVC not supported by DALI?

JanuszL:
Hi @castim,

DALI does support the HEVC codec. I just checked your script and on my NVIDIA TITAN XP, I get ~2-3s per batch with the original encoding while ~1.2s per batch with HEVC.
Regarding the hang, it seems in the case of HEVC preceding frames may appear after the next keyframe so [the condition](https://github.com/NVIDIA/DALI/blob/main/dali/operators/reader/loader/video_loader.cc#L583L602) of not reading the video stream anymore is no longer correct. Let us check if we can do anything with that.

castim:
Regarding the H264 HEVC speed difference, my bad, I was testing between two files with different GOP structures, probably the source of the speed difference.

(My command should have been 
```
ffmpeg -i sintel_trailer-1080p.mp4 -c:v hevc -an -x265-params "keyint=5" sintel_trailer-1080p-g5-hevc.mp4
```
to get a GOP size of 5; ffmpeg -g does not work with hevc)

JanuszL:
I think I have found a fix for the issue you reported - https://github.com/NVIDIA/DALI/pull/3247. Please check the nightly build that follows the day the fix is merged.

castim:
Great, will do!

JanuszL:
The fix has been merged. It should be available in tomorrow's nightly build and 1.6 release.

[...]

PR #3247, opened 2021-08-12
Fix HEVC video handling

JanuszL:
- it seems that preceding frames may appear after the following
  keyframe so the logic that stops sending frames to the decoder
  assuming that each keyframe is IDR doesn't work anymore. Improve
  the logic to stop after seeing the next keyframe as it should be
  very unlikely to have a frame from two keyframes behind
- adds HEVC test

## Description

### What happened in this PR
 - it seems that preceding frames may appear after the following
   keyframe so the logic that stops sending frames to the decoder
   assuming that each keyframe is IDR doesn't work anymore. Improve
   the logic to stop after seeing the next keyframe as it should be
   very unlikely to have a frame from two keyframes behind

#### Additional information
  - video_loader.cc
  - video_reader_op_test.cc

  - NA

### Tests
- [ ] Existing tests apply
- [x] New tests added
  - [ ] Python tests
  - [x] GTests
  - [ ] Benchmark
  - [ ] Other
- [ ] N/A

### Documentation
- [ ] Documentation updated
  - [ ] Docstring
  - [ ] Doxygen
  - [ ] RST
  - [ ] Jupyter
  - [ ] Other
- [x] N/A

#### Requirements
- [x] N/A

**REQ IDs**: N/A

Relates to https://github.com/NVIDIA/DALI/issues/3243
**JIRA TASK**: DALI-2243


JanuszL:
https://github.com/NVIDIA/DALI_extra/pull/65 is a prerequisite.

klecki (commented):
I think some counter how many key frames we've seen and the required number as a constant would feel more natural. 

Also, doesn't this impact performance? Shouldn't we do this based on the format that we are dealing with?

JanuszL:
> I think some counter how many key frames we've seen and the required number as a constant would feel more natural.

I can add it.

> Also, doesn't this impact performance?

It may but I don't see any other way to fix that.

>  Shouldn't we do this based on the format that we are dealing with?

Probably but I don't have sufficient knowledge which one needs that.

JanuszL:
Anyway done.

dali/operators/reader/loader/video_loader.cc
```diff
@@ -533,6 +533,11 @@ void VideoLoader::read_file() {
     bool is_first_frame = true;
     bool key = false;
     bool seek_must_succeed = false;
+    // how many key frames following the last requested frames we need to see so far
```
klecki, line 536:
```suggestion
    // how many key frames following the last requested frames we saw so far
```

dali/operators/reader/loader/video_loader.cc
```diff
@@ -533,6 +533,11 @@ void VideoLoader::read_file() {
...
+    int key_frames_count = 0;
+    // how many key frames following the last requested frames we need to see before we stop
+    // feeding the decoder
+    const int key_frames_treshold = 1;
```
klecki, line 540: IMO we want to see two, so I would probably write here 2, and use greater equal, but it's a nitpick:
```suggestion
    const int key_frames_treshold = 2;
```

dali/operators/reader/loader/video_loader.cc
```diff
@@ -585,10 +590,16 @@ void VideoLoader::read_file() {
           if (!is_first_frame) {
             nonkey_frame_count = 0;
             if (frame > req.frame + req.count) {
```
klecki, line 592: Shouldn't we check for:
```suggestion
            if (frame > req.frame + (1 + key_frames_count) * req.count) {
```
or something?
I mean, if key frame is every `req.count` (just guessing), we need to see the multiple of the distance to detect key frame.

dali/operators/reader/loader/video_loader.cc
```diff
@@ -585,10 +590,16 @@ void VideoLoader::read_file() {
...
               // the next key frame)
-              break;
+              // in case HEVC it seems that preceding frames can appear after a given key frame
+              // but rather not after the next one
+              if (key_frames_count > key_frames_treshold) {
+                key_frames_count = 0;
+                break;
+              }
+              ++key_frames_count;
             }
```
klecki, line 603: If my comment is right, this can be replaced with (note the 2):
```
const int key_frames_treshold = 2;

if (frame > req.frame + key_frames_treshold * req.count) {
   break;
}
```
Which is probably simpler.

klecki (commented):
I'm not sure if I understand this correctly. Is the req.count the distance between keyframes? If so, there is a bug and we can simplify.

JanuszL:
> I'm not sure if I understand this correctly. Is the req.count the distance between keyframes? If so, there is a bug and we can simplify.

It is the number of frames between the first and the last in the sequence including stride.

dali/operators/reader/loader/video_loader.cc
JanuszL: No, keyframe is the video stream property and it doesn't have to be equally placed in the video. `req.count` is how many frames we need to decode to between first and the last frame in the sequence.

dali/operators/reader/loader/video_loader.cc
JanuszL: `key_frames_treshold * req.count` doesn't make any sense. `req.count` has nothing to do with keyframes.

dali/operators/reader/loader/video_loader.cc
```diff
@@ -533,6 +533,11 @@ void VideoLoader::read_file() {
     bool is_first_frame = true;
     bool key = false;
     bool seek_must_succeed = false;
+    // how many key frames following the last requested frames we saw so far
+    int key_frames_count = 0;
+    // how many key frames following the last requested frames we need to see before we stop
+    // feeding the decoder
+    const int key_frames_treshold = 2;
```
mzient, line 540: Can we infer this from the file info somehow? I think that there are relatively few formats that support bidirectional prediction - for others, that's just waste of time.
Even in case of HEVC, I _think_ that you don't need the whole interval between keyframes - just from the keyframe to the requested frame (in forward prediction) and from the requested frame to the keyframe (in reverse prediction) - but determining which keyframe is relevant might be non-trivial.

mzient:
Please rename the PR from "Improve" to "Fix" - after all, before this PR it just doesn't work for some valid HEVC videos.

dali/operators/reader/loader/video_loader.cc
JanuszL: I can add check only for HEVC but I'm not aware of any simple way to determine what kind of prediction is used in a given stream - I afraid it would require parsing headers/stream on our own what is much beyond the scope of this task.

> Even in case of HEVC, I think that you don't need the whole interval between keyframes - just from the keyframe to the requested frame (in forward prediction) and from the requested frame to the keyframe (in reverse prediction) - but determining which keyframe is relevant might be non-trivial.

In some cases we don't get any meaningful timestamp from the frame we read from the stream. So it may not work at all.

mzient (approved):

klecki (approved):

PR #3257, opened 2021-08-17
Fix hw decoder tests disabled on old drivers

klecki:

## Description

### What happened in this PR
Refactored out function to check driver version.
Skip tests that check for use of HW decoder when they are running on unsupported driver.

#### Additional information

### Tests
- [ ] Existing tests apply
- [ ] New tests added
  - [ ] Python tests
  - [x] GTests
  - [ ] Benchmark
  - [ ] Other
- [ ] N/A

### Documentation
- [ ] Documentation updated
  - [ ] Docstring
  - [ ] Doxygen
  - [ ] RST
  - [ ] Jupyter
  - [ ] Other
- [ ] N/A

#### Requirements
- [ ] N/A

**REQ IDs**: N/A

**JIRA TASK**: N/A

### DIFF
--- a/dali/operators/decoder/nvjpeg/nvjpeg_decoder_decoupled_api.h
+++ b/dali/operators/decoder/nvjpeg/nvjpeg_decoder_decoupled_api.h
@@ -101,12 +101,8 @@ class nvJPEGDecoder : public Operator<MixedBackend>, CachedDecoderImpl {
     // disable HW decoder for drivers < 455.x as the memory pool for it is not available
     // and multi GPU performance is far from perfect due to frequent memory allocations
 #if NVML_ENABLED
-      char version[NVML_SYSTEM_DRIVER_VERSION_BUFFER_SIZE];
-      CUDA_CALL(nvmlInitChecked());
-      CUDA_CALL(nvmlSystemGetDriverVersion(version, sizeof version));
-
-      float driverVersion = 0;
-      driverVersion = std::stof(version);
+      nvml::Init();
+      float driverVersion = nvml::GetDriverVersion();
       if (driverVersion < 455) {
         try_init_hw_decoder = false,
         hw_decoder_load_ = 0;
@@ -269,6 +265,10 @@ class nvJPEGDecoder : public Operator<MixedBackend>, CachedDecoderImpl {
   ~nvJPEGDecoder() override {
     try {
       DeviceGuard g(device_id_);
+
+#if NVML_ENABLED
+      nvml::Shutdown();
+#endif
 
       sample_data_.clear();
 
--- a/dali/operators/decoder/nvjpeg/nvjpeg_decoder_decoupled_api_test.cc
+++ b/dali/operators/decoder/nvjpeg/nvjpeg_decoder_decoupled_api_test.cc
@@ -225,7 +225,10 @@ class HwDecoderUtilizationTest : public ::testing::Test {
     auto node = pipeline_.GetOperatorNode(decoder_name_);
     if (!node->op->GetDiagnostic<bool>("using_hw_decoder")) {
       PrintDeviceInfo();
-      if (nvml::isHWDecoderSupported()) {
+      float driver_version = nvml::GetDriverVersion();
+      // HW decoder is disabled for drivers < 455.x, see
+      // dali/operators/decoder/nvjpeg/nvjpeg_decoder_decoupled_api.h for details
+      if (driver_version >= 455 && nvml::isHWDecoderSupported()) {
         FAIL() << "HW Decoder exists in the system and failed to open";
       }
       GTEST_SKIP();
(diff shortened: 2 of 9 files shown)


klecki:
!build

klecki:
!build

dali/util/nvml.h
```diff
@@ -252,6 +252,20 @@ inline bool HasCuda11NvmlFunctions() {
...
 
+inline float GetDriverVersion() {
+  if (!nvmlIsInitialized()) {
+    return 0;
+  }
+
+  float driver_version = 0;
+  char version[NVML_SYSTEM_DRIVER_VERSION_BUFFER_SIZE];
+
+  CUDA_CALL(nvmlSystemGetDriverVersion(version, sizeof version));
```
mzient, line 263: I suggest that we switch to CUDA Driver API for that.
```CUresult cuDriverGetVersion ( int* driverVersion )```
The documentation wasn't clear before CUDA 10, but in 10.2 it says that it
> Returns in *driverVersion the version of CUDA supported by the driver. The version is returned as (1000 major + 10 minor). For example, CUDA 9.2 would be represented by 9020.

Comment (klecki) [reply dali/util/nvml.h:263]: But the cuDriverGetVersion gives me CUDA version, not the driver we are using. Can we safely map those two things?

Comment (mzient) [reply dali/util/nvml.h:263]: I guess (in)compatibility mode could mess it up.

**Comment by klecki** [comment]
!build

File: dali/util/nvml.h (hunk 1 above)
Comment (mzient) [reply dali/util/nvml.h:263]: Now that I think of it, most (all?) of the features we would query the driver for depend on user-mode features. As such, if NVML is not available, we could just conservatively report the oldest driver that supports the CUDA version reported by `cuDriverGetVersion`.

File: dali/util/nvml.h [hunk 2]
```diff
@@ -252,6 +252,20 @@ inline bool HasCuda11NvmlFunctions() {
   return nvmlHasCuda11NvmlFunctions();
 }
 
+inline float GetDriverVersion() {
```
mzient, line 255: I don't think this function should be inline. I know we don't have a translation unit for it - but maybe we should?

mzient: I'd suggest splitting this function in the following way:
```
float GetDriverVersionImpl() {
  // do all the stuff that's here
}

inline float GetDriverVersion() {
  static float version = GetDriverVersionImpl();  // this is called once
  return version;
}
```

dali/operators/decoder/nvjpeg/nvjpeg_decoder_decoupled_api.h
```diff
@@ -270,6 +266,10 @@ class nvJPEGDecoder : public Operator<MixedBackend>, CachedDecoderImpl {
     try {
       DeviceGuard g(device_id_);
 
+#if NVML_ENABLED
+      nvml::Shutdown();
```
mzient, line 270: We have altered the semantics of shutdown WRT plain nvmlShutdown, which makes this kind of behaviour incorrect.
Please either revert or change:
```C++
inline void Shutdown() {
  std::lock_guard<std::mutex> lock(Mutex());
  if (!nvmlIsInitialized()) {
    return;
  }
  CUDA_CALL(nvmlShutdown());
}
```
to
```C++
inline void Shutdown() {
  CUDA_CALL(nvmlShutdown());
}
```

dali/operators/decoder/nvjpeg/nvjpeg_decoder_decoupled_api_test.cc
```diff
@@ -225,7 +225,10 @@ class HwDecoderUtilizationTest : public ::testing::Test {
     auto node = pipeline_.GetOperatorNode(decoder_name_);
     if (!node->op->GetDiagnostic<bool>("using_hw_decoder")) {
       PrintDeviceInfo();
-      if (nvml::isHWDecoderSupported()) {
+      float driver_version = nvml::GetDriverVersion();
+      // HW decoder is disabled for drivers < 455.x, see
+      // dali/operators/decoder/nvjpeg/nvjpeg_decoder_decoupled_api.h for details
+      if (driver_version >= 455 && nvml::isHWDecoderSupported()) {
```
mzient, line 231: Perhaps we should have a function `ShouldUseHWDecoder()` which aggregates these two checks? I see this code repeated 4 times.

klecki: Good point, will do.

awolant (approved):

dali/util/nvml.h
klecki: I was thinking about that, I guess we won't be changing driver version at runtime, will we?

klecki:
!build

klecki:
!build

mzient (approved):

PR #3269, opened 2021-08-20
Support lambdas and local functions as callbacks in parallel ExternalSource

stiepan:
## Description

### What happened in this PR
It adds customized serialization of parallel externalsource callbacks so that lambdas and local functions are supported, usage of
functions defined inside Jupyter notebooks is easier and enables user to specify external package to be used to serialize callback if more specialized serialization is needed,

#### Additional information
pythonic parts of parallel external source implementation

### Tests
- [ ] Existing tests apply
- [ ] New tests added
  - [ ] Python tests
  - [ ] GTests
  - [ ] Benchmark
  - [ ] Other
- [ ] N/A

### Documentation
- [ ] Documentation updated
  - [ ] Docstring
  - [ ] Doxygen
  - [ ] RST
  - [ ] Jupyter
  - [ ] Other
- [ ] N/A

#### Requirements
- [ ] N/A

**REQ IDs**: N/A

**JIRA TASK**: DALI-2214


stiepan:
!build

dali/python/nvidia/dali/_multiproc/pool.py
```diff
@@ -202,10 +204,14 @@ def __init__(
             task_r, task_w = mp.Pipe(duplex=False)
             res_r, res_w = mp.Pipe(duplex=False)
             sock_reader, sock_writer = socket.socketpair()
+            if callback_pickler is None:
+                callbacks_ = callbacks
+            else:
+                callbacks_ = callback_pickler.dumps(callbacks)
```
mzient, line 210: So now `callbacks` contains either pickled or raw callbacks?

dali/python/nvidia/dali/_multiproc/pool.py
```diff
@@ -202,10 +204,14 @@ def __init__(
             task_r, task_w = mp.Pipe(duplex=False)
             res_r, res_w = mp.Pipe(duplex=False)
             sock_reader, sock_writer = socket.socketpair()
+            if callback_pickler is None:
+                callbacks_ = callbacks
```
mzient, line 208: How about this?
```suggestion
                callbacks_arg = callbacks
```
I'm not a fan of trailing underscore and leading underscore has a meaning in Python which I guess we should not abuse.

dali/python/nvidia/dali/pickling.py
```diff
@@ -0,0 +1,144 @@
...
+    f = io.BytesIO()
+    DaliCallbackPickler(f, protocol, **kwargs).dump(obj)
+    return f.getvalue()
+
+
+loads = pickle.loads
+
+
+class CustomPickler:
+
```
mzient, line 103: Nitpick: weird empty line

dali/python/nvidia/dali/pickling.py
```diff
@@ -0,0 +1,144 @@
...
+dummy_lambda = lambda : 0
+
+def set_funcion_state(fun, state):
+    fun.__globals__.update(state['global_refs'])
+    fun.__defaults__ = state['defaults']
+    fun.__kwdefaults__ = state['kwdefaults']
+
+def function_unpickle(name, qualname, code, closure):
+    code = marshal.loads(code)
+    globs = {'__builtins__': __builtins__}
```
mzient, line 32: Why "globs"?
```suggestion
    builtins = {'__builtins__': __builtins__}
```

dali/python/nvidia/dali/pickling.py
```diff
@@ -0,0 +1,144 @@
...
+
+def cell_unpickle():
+    return types.CellType(None)
+
+def cell_reducer(cell):
+    return (cell_unpickle, tuple(), {'cell_contents': cell.cell_contents}, None, None, set_cell_state)
+
+
+class DaliCallbackPickler(pickle.Pickler):
+
```
mzient, line 74: Nitpick: weird empty line

dali/python/nvidia/dali/pickling.py
```diff
@@ -0,0 +1,144 @@
...
+    def reducer_override(self, obj):
+        if inspect.ismodule(obj):
+            return module_reducer(obj)
+        if isinstance(obj, types.CellType):
+            return cell_reducer(obj)
+        if inspect.isfunction(obj):
+            if isinstance(obj, type(dummy_lambda)) and obj.__name__ == dummy_lambda.__name__ or \
+                    getattr(obj, '_dali_pickle_by_value', False):
+                return function_by_value_reducer(obj)
+            if '<locals>' in obj.__qualname__:
```
mzient, line 84: Does it work with regular local functions, such as:
```
def f():
  def g(): return 42;
```
?
`g.__qualname__ == 'f.<locals>.g'`, but it's otherwise quite trivial.
How about switching the logic to:
```Python:
try:
  pickle.dumps(obj)
except (whatever it raises):
  return function_by_value_reducer_(obj)
```
?

dali/test/python/test_external_source_parallel_custom_serialization.py
```diff
@@ -0,0 +1,432 @@
...
+            encoded_img = np.frombuffer(f.read(), dtype=np.uint8)
+        return encoded_img, label
+
+    return py_file_reader
+
+
+def create_simple_pipeline(callback, py_callback_pickler, batch_size, parallel=True, py_num_workers=None):
+
+    extra = {}
+    if py_callback_pickler != UseDefault:
```
mzient, line 155: Why not:
```suggestion
    if py_callback_pickler is nt None:
```
like we do elsewhere?

dali/python/nvidia/dali/pipeline.py
```diff
@@ -135,13 +136,34 @@ class Pipeline(object):
     you will need to call :meth:`start_py_workers` before calling :meth:`build` of any
     of the pipelines. You can find more details and caveats of both methods in Python's
     ``multiprocessing`` module documentation.
+`py_callback_pickler` : module or tuple, default = nvidia.dali.pickling
+    If `py_start_method` is set to *spawn* callback passed to parallel ExternalSource must be picklable.
+    If run in Python3.8 or newer, DALI uses customized pickle (`nvidia.dali.pickling`) when
```
mzient, line 141: Do we want to advertise this module to the users? I think it could be internal.

dali/python/nvidia/dali/pipeline.py
```diff
@@ -135,13 +136,34 @@ class Pipeline(object):
...
+    serializing callbacks to support serialization of local functions and lambdas.
+
+    However, if you need to serialize more complex objects like local classes or you are running
+    older version of Python you can provide external serialization package such as dill or cloudpickle
```
mzient, line 145: For me it wasn't immediately clear that I can pass `dill` _module_ as an argument. How about adding an example?

dali/python/nvidia/dali/pipeline.py
```diff
@@ -135,13 +136,34 @@ class Pipeline(object):
...
+    that implements two methods: `dumps` and `loads` to make DALI use them to serialize
+    external source callbacks.
+
```
mzient, line 148:
```suggestion
    external source callbacks. You can even pass the module directly as ``py_callback_pickler``::

        import dill
        src = fn.external_source(lambda sample_info: 42, batch=False, parallel=True, py_callback_pickler=dill)

```

dali/test/python/import_module_test_helper.py
```diff
@@ -0,0 +1,24 @@
...
+
+# Used in test_external_source_parallel_custom_serialization to check if modules
+# are properly imported during callback deserialization. Such test only makes sense
+# if this module is not automatically imported when worker process starts, so don't 
+# import this file globally
+
+import numpy as np
+
+def cb(x):
+    return np.full((10, 100), x.idx_in_epoch)
```
mzient, line 24: Do you really need this in a separate file?

dali/test/python/test_utils.py
```diff
@@ -543,3 +543,19 @@ def get_files(path, ext):
       if re.match(f".*\.{ext}", f) is not None
   ]
   return audio_files
+
+
+def _test_omitted():
+    print("Omitting tests suite for Python3.8+ serialization")
```
mzient, line 549: Please move this to the relevant test module or parameterize it. The name suggests a generic utility, but it produces a very specific message.

dali/test/python/test_utils.py
```diff
@@ -543,3 +543,19 @@ def get_files(path, ext):
...
+
+
+def restrict_python_version(major, minor=None):
+
+    def decorator(test_case):
+        version_info = sys.version_info
+        if version_info.major > major or \
+                (version_info.major == major and (minor is None or version_info.minor >= minor)):
+            return test_case
+        return _test_omitted
```
mzient, line 559:
```suggestion
        return lambda: test_skipped(f"Insufficient Python version {version_info.major}.{version_info.minor} - required {major}.{minor}")
```

dali/python/nvidia/dali/pickling.py
```diff
@@ -0,0 +1,144 @@
...
+        if isinstance(py_callback_pickler, (tuple, list)):
+            params = [None] * 3
+            for i, item in enumerate(py_callback_pickler):
+                params[i] = item
+            reducer, kwargs_dumps, kwargs_loads = params
+            return cls.of_reducer(reducer, kwargs_dumps, kwargs_loads)
+        raise ValueError("Unsupported py_callback_pickler value provided.")
+
+    @classmethod
+    def of_reducer(cls, reducer, dumps_kwargs=None, loads_kwargs=None):
```
mzient, line 119: This is a weird name. Is it required by some API? What does this `of` stand for?

jantonguirao (approved):

dali/python/nvidia/dali/_multiproc/pool.py
stiepan: Yes, so in the first case it creates kind of additional layer of serialization where multiprocessing pickler just sees already serialized data. At first I simply set Dali customized pickler to be used directly by multiprocessing but it is unfortunately a global change for the whole process and: 
1. does not work if multiprocessing already started some actual work before DALI was able to modify the context
2. there's a question if that's not going to interfere with other packages using multiprocessing, like pytorch (it seemed to work okay, but I think that if we can avoid worrying about that let's avoid it.)
3. for user provided packages for serialization I still would go for this type of double serialization

dali/python/nvidia/dali/pickling.py
stiepan: It is a dictionary containing "global scope" of the function, the one available later on as `fun.__globals__` immutable attribute (you cannot replace the dict but can update its contents). If one called `globals()` inside the instantiated function this is the object that would be returned. So it is rather a coincidence that this dictionary contains only builtins at this point - all other global references (if any) are added at the set state stage to handle possible cyclic references.
I didn't call it globals on the other hand to avoid shadowing `globals()` inside `fun.__globals__`

dali/python/nvidia/dali/pipeline.py
stiepan: I decided to advertise it because apart from serving as default callback pickler it also contains @pickle_by_value decorator whose whole purpose is to be available for manual usage by users.

dali/python/nvidia/dali/pipeline.py
stiepan: I've added the example in the notebook with parallel External Source spawn use cases. Do you think abut explicitly putting some code into the docstring too?

dali/python/nvidia/dali/pipeline.py
stiepan: Never mind, yo do. :D

dali/python/nvidia/dali/pickling.py
stiepan: Nah, it's not required by any API, I meant create_instance_of_this_class_from_reducer and for some reason used of rather than from.

dali/python/nvidia/dali/pickling.py
stiepan: The one line space between first method and the `class ... :`?
I can remove it if you think it somehow clashes with the code formatting in DALI though I tried to use it everywhere both in this and other PRs. D:
Most of the classes I can find have docstrings but those are followed with a single empty line too.

dali/python/nvidia/dali/pickling.py
stiepan: Yep, g will be serialized by value and plain pickling fails as pickle recursively steps down the attributes and raises AttributeError when encountering <locals>.

I'd rather limit this `pickle.dumps` test to the situations when I am nearly sure that it is going to fail (or maybe should get rid of the test whatsoever?) for efficiency reasons. If the test succeeded it serialized something only for us to discard that result and proceed with a single step of serialization.

It would be nice if we could catch the exception by overriding dump method of the pickler, catch AttributeError and only then fall back to obj.__dali_pickle_by_value = True. Unfortunately it doesn't seem to be (easily) doable as the pickle.Pickler class is in fact a wrapper around C implementation of pickle and overridden dump method would be called only once for a top level object; recursive steps for dependencies/contents will be serialized by C code unaware of override.

stiepan:
!build

dali/python/nvidia/dali/_multiproc/pool.py
stiepan: done

dali/test/python/test_external_source_parallel_custom_serialization.py
stiepan: done

dali/python/nvidia/dali/pipeline.py
stiepan: done

dali/python/nvidia/dali/pipeline.py
stiepan: done

dali/test/python/test_utils.py
stiepan: done

dali/test/python/test_utils.py
stiepan: done

dali/python/nvidia/dali/pipeline.py
```diff
@@ -135,13 +136,34 @@ class Pipeline(object):
...
+    Valid value for `py_callback_pickler` is either a module/object implementing
```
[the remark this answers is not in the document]
stiepan: done

dali/python/nvidia/dali/pipeline.py
```diff
@@ -135,13 +136,34 @@ class Pipeline(object):
...
+    dumps and loads methods or a tuple where first item is the module/object and the next
```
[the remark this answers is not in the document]
stiepan: done

dali/python/nvidia/dali/pipeline.py
```diff
@@ -135,13 +136,34 @@ class Pipeline(object):
...
+    two optional parameters are extra kwargs to be passed when calling dumps and loads respectively.
+    Provided methods and kwargs must themselves be picklable.
+
+    If you run Python3.8 or newer and use default `nvidia.dali.pickling` you can hint DALI to serialize
```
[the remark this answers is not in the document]
stiepan: done

dali/test/python/import_module_test_helper.py
```diff
@@ -0,0 +1,24 @@
+# Copyright (c) 2020-2021, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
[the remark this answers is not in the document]
stiepan: done

dali/test/python/test_external_source_parallel_custom_serialization.py
```diff
@@ -0,0 +1,432 @@
+# Copyright (c) 2020-2021, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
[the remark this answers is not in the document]
stiepan: done

dali/test/python/test_external_source_parallel_custom_serialization.py
```diff
@@ -0,0 +1,432 @@
...
+    return np.full((10, 20), i, dtype=np.uint8)
+
+
+class NoDumpsParam(ValueError):
+    pass
+
+
+def dumps(obj, **kwargs):
+    if kwargs.get('special_dumps_param') != 42:
+        raise NoDumpsParam("Expected special_dumps_pram among kwargs, got {}".format(kwargs))
```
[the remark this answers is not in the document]
stiepan: done

dali/test/python/test_external_source_parallel_custom_serialization.py
```diff
@@ -0,0 +1,432 @@
...
+
+def create_simple_pipeline(callback, py_callback_pickler, batch_size, parallel=True, py_num_workers=None):
+
+    extra = {}
+    if py_callback_pickler != UseDefault:
+        extra['py_callback_pickler'] = py_callback_pickler
+
+    @pipeline_def(batch_size=batch_size, num_threads=2, device_id=0, py_num_workers=py_num_workers,
+        py_start_method="spawn", **extra)
+    def crate_pipline():
```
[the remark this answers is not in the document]
stiepan: done

dali/python/nvidia/dali/pipeline.py
stiepan: done

PR #3297, opened 2021-08-30
Use libcu++ interfaces.

mzient:

## Description

### What happened in this PR
- It switches from using RMM to libcu++.
- It implements the simple resources from RMM which we had previously used.
- It **changes the behavior of the linter** so that headers without an extension are interpreted as C++

* Memory management
* Linter

### Tests
- [X] Existing tests apply
- [ ] New tests added
  - [ ] Python tests
  - [ ] GTests
  - [ ] Benchmark
  - [ ] Other
- [ ] N/A

### Documentation
- [ ] Documentation updated
  - [ ] Docstring
  - [ ] Doxygen
  - [ ] RST
  - [ ] Jupyter
  - [X] Other
- [ ] N/A
Acknowledgements.

#### Requirements
- [X] N/A

**REQ IDs**: N/A

**JIRA TASK**: DALI-2268

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -26,15 +26,9 @@
 [submodule "third_party/cocoapi"]
 	path = third_party/cocoapi
 	url = https://github.com/cocodataset/cocoapi
-[submodule "third_party/rmm"]
-	path = third_party/rmm
-	url = https://github.com/mzient/rmm.git
-[submodule "third_party/spdlog"]
-	path = third_party/spdlog
-	url = https://github.com/gabime/spdlog.git
-[submodule "third_party/fmt"]
-	path = third_party/fmt
-	url = https://github.com/fmtlib/fmt
 [submodule "third_party/thrust"]
 	path = third_party/thrust
 	url = https://github.com/NVIDIA/thrust
+[submodule "third_party/libcudacxx"]
+	path = third_party/libcudacxx
+	url = https://github.com/mzient/libcudacxx.git
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,4 @@
-# Copyright (c) 2017-2018, NVIDIA CORPORATION. All rights reserved.
+# Copyright (c) 2017-2021, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
@@ -254,6 +254,8 @@ if ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "Clang")
     set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wno-unused-command-line-argument")
     # Ignore warnings coming from cutlass
     set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} --system-header-prefix=cutlass/")
+    # Enable SFINAE with ellipsis in device code
+    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Xclang -fcuda-allow-variadic-functions")
   endif()
 
   # CUDA does not support current Clang as host compiler, we need use gcc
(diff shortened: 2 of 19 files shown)


dali/core/mm/async_pool_test.cu
```diff
@@ -163,12 +162,14 @@ void AsyncPoolTest(Pool &pool, vector<block> &blocks, Mutex &mtx, CUDAStream &st
       CUDA_CALL(cudaMemsetAsync(ptr, fill, size, stream));
       {
         std::lock_guard<Mutex> guard(mtx);
+        (void)guard;  // for dummy mutexes
```
szalpal, line 165: Could you explain this? I'm not sure I understand this line correctly

.gitmodules
```diff
@@ -26,15 +26,9 @@
...
-	url = https://github.com/gabime/spdlog.git
-[submodule "third_party/fmt"]
-	path = third_party/fmt
-	url = https://github.com/fmtlib/fmt
 [submodule "third_party/thrust"]
 	path = third_party/thrust
 	url = https://github.com/NVIDIA/thrust
+[submodule "third_party/libcudacxx"]
+	path = third_party/libcudacxx
+	url = https://github.com/mzient/libcudacxx.git
```
szalpal, line 34: Is linking from your private fork on purpose?

banasraf (approved):

third_party/cpplint.py
```diff
@@ -4480,7 +4480,8 @@ def _ClassifyInclude(fileinfo, include, is_system):
             or include.endswith(".hpp") \
             or include.endswith(".hxx") \
             or include.endswith(".H") \
-            or include.endswith(".hh")
+            or include.endswith(".hh") \
+            or '.' not in include[include.rfind('/')]  # no extension
```
szalpal, line 4484: Are we sure we want this? I recall you saying, that that's in order to have `#include <cuda/memory_resource>` recognized as C++. However, we are not really running linter for third party headers, so I don't feel it's necessary to run it for `<cuda/memory_resource>`. On the other hand, this may lead to strange linter behaviour in future.

@banasraf , what's your opinion about it?

CMakeLists.txt
```diff
@@ -254,6 +254,8 @@ if ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "Clang")
     set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wno-unused-command-line-argument")
     # Ignore warnings coming from cutlass
     set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} --system-header-prefix=cutlass/")
+    # Enable SFINAE with ellipsis in device code
+    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Xclang -fcuda-allow-variadic-functions")
```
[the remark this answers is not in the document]
klecki: I agree, no issue with turning this on, probably split it into warning section and useful feature section.

third_party/cpplint.py
banasraf: If I understand correctly, it's not about running linter on a third party header, but about ordering of includes. Linter requires some certain order of includes, and this change is to correctly qualify the cuda/memory_resource as C++ include.

mzient: ^^^ exactly that.

.gitmodules
mzient: It's not been upstreamed yet, so yes, it's necessary for the time being.

CMakeLists.txt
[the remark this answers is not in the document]
mzient: I can move the comment and run build stage (only) again if you _REALLY_ insist.

dali/core/mm/async_pool_test.cu
mzient: Sometimes (in the tests) the mutex is a dummy object where the lock/unlock functions are empty and the compiler is smart enough to see that the construction/destruction has no effect and emits a warning.

CMakeLists.txt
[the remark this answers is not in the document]
mzient: Done - I had to fix basename extraction in linter, so another build is in order.

szalpal (approved):

PR #3302, opened 2021-08-31
Add assert_raises_pattern and raises_pattern utilities

stiepan:

## Description

### What happened in this PR
Added `assert_raises_pattern` function and `raises_pattern` decorator that work as nose `assert_raises` and `raises` but
with additional pattern check on exception message.

#### Additional information

### Tests
- [ ] Existing tests apply
- [ ] New tests added
  - [ ] Python tests
  - [ ] GTests
  - [ ] Benchmark
  - [ ] Other
- [x] N/A

### Documentation
- [ ] Documentation updated
  - [ ] Docstring
  - [ ] Doxygen
  - [ ] RST
  - [ ] Jupyter
  - [ ] Other
- [x] N/A

#### Requirements
- [x] N/A

**REQ IDs**: N/A

**JIRA TASK**: DALI-2233

### DIFF
--- a/dali/test/python/nose_utils.py
+++ b/dali/test/python/nose_utils.py
@@ -1,0 +1,64 @@
+# Copyright (c) 2021, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import nose.tools as tools
+import re
+import fnmatch
+
+
+def assert_raises_pattern(exception, pattern, *args, **kwargs):
+
+    """
+    There already is an assert_raises_regex function available in nose.tools
+    (which is actually a method assertRaisesRegex copied from python's unittest.TestCase dummy instance).
+    This is a simple wrapper around the function that allows to:
+    * specify if the pattern searching should be case-sensitive (by default it is not, use `match_case`=True to change that)
+    * use simple glob pattern instead of regex (specify `use_glob`=False if you want to use regex)
+
+    `exception` can be either a class or list of classes
+    """
+
+    match_case = kwargs.pop("match_case", None)
+    use_glob = kwargs.pop("use_glob", None)
+
+    if not isinstance(pattern, str): # it could be an already compiled regex
+        if use_glob is not None or match_case is not None:
+            raise ValueError("Pattern must be a string if `match_case` or `use_glob` is specified when calling assert_raises_pattern")
+    else:
+        if use_glob is None:
+            use_glob = True
+        if match_case is None:
+            match_case = False
+
+        if use_glob:
+            pattern = fnmatch.translate(pattern)
+        flags = 0
+        if not match_case:
+            flags |= re.IGNORECASE
+        pattern = re.compile(pattern, flags)
+
+    return tools.assert_raises_regex(exception, pattern, *args, **kwargs)
+
+
+def raises_pattern(exception, pattern, match_case=None, use_glob=None):
+
+    def decorator(func):
+
+        def new_func(*args, **kwargs):
+            with assert_raises_pattern(exception, pattern, match_case=match_case, use_glob=use_glob):
+                return func(*args, **kwargs)
+
+        return tools.make_decorator(func)(new_func)
+
+    return decorator
--- a/dali/test/python/nose_utils.py
+++ b/dali/test/python/nose_utils.py
@@ -17,46 +17,71 @@ import re
 import fnmatch
 
 
-def assert_raises_pattern(exception, pattern, *args, **kwargs):
+def assert_raises(exception, *args, **kwargs):
 
     """
-    There already is an assert_raises_regex function available in nose.tools
-    (which is actually a method assertRaisesRegex copied from python's unittest.TestCase dummy instance).
-    This is a simple wrapper around the function that allows to:
-    * specify if the pattern searching should be case-sensitive (by default it is not, use `match_case`=True to change that)
-    * use simple glob pattern instead of regex (specify `use_glob`=False if you want to use regex)
-
-    `exception` can be either a class or list of classes
+    Wrapper combining `nose.tools.assert_raises` and `nose.tools.assert_raises_regex`.
+    Specify ``regex=pattern`` or ``glob=pattern`` to check error message of expected exception against the pattern.
+    Value for `glob` must be a string, `regex` can be either a literal or compiled regex pattern.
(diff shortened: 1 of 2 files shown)


dali/test/python/test_external_source_parallel_custom_serialization.py
```diff
@@ -305,7 +301,7 @@ def _test_pickle_by_value_decorator_on_global_function(name, py_callback_pickler
 
 
 @register_case(tests_dali_pickling)
-@raises(NoDumpsParam)
+@raises_pattern(ValueError, "Expected special_dumps_param among kwargs, got *")
```
klecki, line 304: General comment, do we want to make the name shorter? 
Would we be fine with just `raises`?
Can we make the API something like:
```
raises(exception_type, glob=None, regex=None, case_sensitive=False)
```
That way, if you want the glob, you do
```
@raises(ValueError, "My * error")
```
for regex, you do:
```
@raises(ValueError, regex="My .*ror")
```

How easy it is to pass a compiled regex? Shouldn't we just allow the regex string, as it is usually defined above the tests in the global scope, so we don't want to deal with adding re.compile("pattern") boilerplate to the match.

stiepan: It allows using regex as string, compiled patterns are just passed through (because why not, maybe there will be some use case for special hand-crafted pattern compiled with extra flags).

For now this is how using regex looks like (this builds with case-insensitive flag):

```
@raises_pattern(ValueError, "My [a-z]*\\s[0-9]{2} message", use_glob=False)
def test_regex_insensitive():
    raise ValueError("my ahgsfds 42 Message")
```

And manually built pattern:

```
import re
pattern = re.compile("[a-z]{3}[0-9]+")

@raises_pattern(ValueError, pattern)
def test_precompiled_regex():
    raise ValueError("abc42")
```

klecki: Ah, ok, I misread. Still, I think, `glob` and `regex` patterns would be probably easier than the boolean to switch between them?

dali/test/python/nose_utils.py
```diff
@@ -0,0 +1,64 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import nose.tools as tools
+import re
+import fnmatch
+
+
+def assert_raises_pattern(exception, pattern, *args, **kwargs):
```
mzient, line 20: I think that the most useful thing would be something like:
```assert_raises_msg(exception, message, *args, **kwargs)```
which would catch all exceptions where the error message contains `message`.

File: dali/test/python/nose_utils.py [hunk 3]
```diff
@@ -0,0 +1,64 @@
...
+            pattern = fnmatch.translate(pattern)
+        flags = 0
+        if not match_case:
+            flags |= re.IGNORECASE
+        pattern = re.compile(pattern, flags)
+
+    return tools.assert_raises_regex(exception, pattern, *args, **kwargs)
+
+
+def raises_pattern(exception, pattern, match_case=None, use_glob=None):
```
mzient, line 54: Likewise - `raises_msg` - in most cases we're interested in the string appearing inside the message.

dali/test/python/test_external_source_parallel_custom_serialization.py
mzient: @stiepan  Why do we need regex strings? We could just do:
```
@raises(ValueError, r"lorem ipsum")
```
which would match all strings that contain `lorem ipsum`.

dali/test/python/nose_utils.py
```diff
@@ -0,0 +1,64 @@
...
+    (which is actually a method assertRaisesRegex copied from python's unittest.TestCase dummy instance).
+    This is a simple wrapper around the function that allows to:
+    * specify if the pattern searching should be case-sensitive (by default it is not, use `match_case`=True to change that)
+    * use simple glob pattern instead of regex (specify `use_glob`=False if you want to use regex)
+
+    `exception` can be either a class or list of classes
+    """
+
+    match_case = kwargs.pop("match_case", None)
+    use_glob = kwargs.pop("use_glob", None)
```
mzient, line 33: Why not just have regular named arguments?

dali/test/python/test_external_source_parallel_custom_serialization.py
stiepan: @mzient yep, you can use r"" literals too.

mzient: We should consider what's the most common use case and make that shortest and easiest to use. For example, if the most common use case is to check for a presence of a string inside the message, then we should have sth like:
```@raises(ErrorType, "part of the message")```
and to supplement that:
```Python
@raises(ErrorType, regex="^Could not allocate.*bytes")```
```Python
@raises(ErrorType, glob=Could not allocate*bytes*")```

Comment (stiepan) [reply dali/test/python/test_external_source_parallel_custom_serialization.py]: @mzient 

```
@raises(ErrorType, "part of the message")
```
is now by default
@raises(ErrorType, glob="part of the message") that will search for any occurrence inside the message.

To enforce matching the whole message, one could use regex with beginning/end of string explicitly stipulated.

File: dali/test/python/nose_utils.py (hunk 4 above)
Comment (stiepan) [reply dali/test/python/nose_utils.py]: done

File: dali/test/python/nose_utils.py [hunk 5]
```diff
@@ -0,0 +1,94 @@
...
+    Wrapper combining `nose.tools.assert_raises` and `nose.tools.assert_raises_regex`.
+    Specify ``regex=pattern`` or ``glob=pattern`` to check error message of expected exception against the pattern.
+    Value for `glob` must be a string, `regex` can be either a literal or compiled regex pattern.
+    By default, the check will ignore case, if called with `glob` or a literal for `regex`.
+    To enforce case sensitive check pass ``match_case=True``.
+    Don't specify `match_case` if passing already compiled regex pattern.
+    """
+
+    glob = kwargs.pop("glob", None)
+    regex = kwargs.pop("regex", None)
```
mzient, line 32: Again - why not named arguments in the function signature?
```Python
def assert_raises(exception, *args, regex = None, glob = None, **kwargs)```

Comment (stiepan) [reply dali/test/python/nose_utils.py]: Sorry, I misunderstood what you meant.

**Review by klecki — APPROVED** [review]

File: dali/test/python/nose_utils.py [hunk 6]
```diff
@@ -0,0 +1,89 @@
...
+            raise ValueError("Regex must be a string if `match_case` is specified when calling assert_raises_pattern")
+        pattern = regex
+
+    if isinstance(pattern, str) and not match_case: # ignore case by default
+        pattern = re.compile(pattern, re.IGNORECASE)
+
+    return tools.assert_raises_regex(exception, pattern, *args, **kwargs)
+
+
+def raises(exception, glob=None, regex=None, match_case=None):
```
mzient, line 57: I'm not sure about the order (glob, regex or the reverse). @klecki ?

mzient (approved):

stiepan:
!build

dali/test/python/nose_utils.py
klecki: I think glob is easier to use, it's not like if we see a loot of regex we can't change

PR #3311, opened 2021-09-03
Rework diplacement filter to sample-based approach

klecki:
## Description

### What happened in this PR
Refactoring:
* Get rid of unnecessary DataDependantSetup
* Introduce SetupImpl
* Change pass-by-pointer to pass-by-reference
* Rework GPU Op to process flattened blocks instead
  of whole images accessed via offset to one global
  batch pointer.
* Instead of accessing underlying contiguous
  TL buffer we access individual samples in GPU Op.
* Rework CPU Op to use HostWorkspace
* Keep the optimized implementation for aligned 
  samples.

Fixes:
* Fix masking in CPU Op - access it as int instead
  of bool to be consistent with the default and GPU Op.
* Fix masking in CPU Op - copy does not try to access
  stream in CPU workspace which causes error.

Tests:
* Add paths in water test to cover more code paths
   * prime-sized image to fall into the non-optimized
     kernel
   * mask support
   * both input types: uint8 and float32.

#### Additional information
Old displacement filter implementation, basis for Water, Sphere and Jitter.

- 
### Tests
- [x] Existing tests apply
- [ ] New tests added
  - [x] Python tests
  - [ ] GTests
  - [ ] Benchmark
  - [ ] Other
- [ ] N/A

### Documentation
- [ ] Documentation updated
  - [ ] Docstring
  - [ ] Doxygen
  - [ ] RST
  - [ ] Jupyter
  - [ ] Other
- [ ] N/A

#### Requirements
- [x] N/A

**REQ IDs**: N/A

**JIRA TASK**: N/A

### DIFF
--- a/dali/operators/image/remap/displacement_filter_impl_gpu.cuh
+++ b/dali/operators/image/remap/displacement_filter_impl_gpu.cuh
@@ -65,7 +65,7 @@ struct PrepareParam {
 template <class Displacement>
 struct PrepareParam<Displacement, true> {
   __device__ __host__ inline void operator()(Displacement &displace, const void *raw_params) {
-    const auto *const params = reinterpret_cast<const typename Displacement::Param *>(raw_params);
+    const auto *const params = static_cast<const typename Displacement::Param *>(raw_params);
     displace.param = *params;
   }
 };
@@ -77,8 +77,8 @@ __global__ void DisplacementKernel(const DisplacementSampleDesc *samples,
   const auto &block = blocks[blockIdx.x];
   const auto &sample = samples[block.sample_idx];
 
-  auto *image_out = reinterpret_cast<T *>(sample.output);
-  const auto *image_in = reinterpret_cast<const T *>(sample.input);
+  auto *image_out = static_cast<T *>(sample.output);
+  const auto *image_in = static_cast<const T *>(sample.input);
 
   const int H = sample.shape[0];
   const int W = sample.shape[1];
@@ -92,6 +92,12 @@ __global__ void DisplacementKernel(const DisplacementSampleDesc *samples,
   auto end = block.end.x;
   for (int64_t out_idx = threadIdx.x + start; out_idx < end; out_idx += blockDim.x) {
     if (m) {
+      // int64_t idx = out_idx;
+      // const int c = idx % C;
+      // idx /= C;
+      // const int w = idx % W;
+      // idx /= W;
+      // const int h = idx;
       const int c = out_idx % C;
       const int w = (out_idx / C) % W;
       const int h = (out_idx / W / C);
@@ -222,7 +228,7 @@ class DisplacementFilter<GPUBackend, Displacement,
       interp_type_(spec.GetArgument<DALIInterpType>("interp_type")),
       flat_block_setup_(32),
       channel_block_setup_(32) {
-    channel_block_setup_.SetDefaultBlockSize({kAlignedBlockDim});
+    channel_block_setup_.SetBlockDim(ivec3{kAlignedBlockDim, 1, 1});
     has_mask_ = spec.HasTensorArgument("mask");
     DALI_ENFORCE(interp_type_ == DALI_INTERP_NN || interp_type_ == DALI_INTERP_LINEAR,
         "Unsupported interpolation type, only NN and LINEAR are supported for this operation");
@@ -317,7 +323,6 @@ class DisplacementFilter<GPUBackend, Displacement,
     auto stream = ws.stream();
 
     const auto num_samples = shape.num_samples();
-    // const int pitch = nDims + 1;  // shape and offset
     samples_.resize(num_samples);
     for (int sample_idx = 0; sample_idx < num_samples; sample_idx++) {
       auto &sample = samples_[sample_idx];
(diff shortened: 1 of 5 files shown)


klecki:
!build

klecki:
!build

dali/operators/image/remap/displacement_filter_impl_gpu.cuh
```diff
@@ -47,155 +59,154 @@ __device__ inline void GetPixelValueMultiC(int h, int w,
...
 template <class Displacement>
 struct PrepareParam<Displacement, true> {
- public:
-  __device__ __host__ inline void operator() (Displacement &displace,  // NOLINT(*)
-                                              const void * raw_params, const int n) {
-    const typename Displacement::Param * const params =
-      reinterpret_cast<const typename Displacement::Param *>(raw_params);
-    displace.param = params[n];
+  __device__ __host__ inline void operator()(Displacement &displace, const void *raw_params) {
+    const auto *const params = reinterpret_cast<const typename Displacement::Param *>(raw_params);
```
mzient, line 68: Cast from `void*` is a static_cast.
```suggestion
    const auto *const params = static_cast<const typename Displacement::Param *>(raw_params);
```

dali/operators/image/remap/displacement_filter_impl_gpu.cuh
```diff
@@ -47,155 +59,154 @@ __device__ inline void GetPixelValueMultiC(int h, int w,
...
+                                   Displacement displace) {
+  const auto &block = blocks[blockIdx.x];
+  const auto &sample = samples[block.sample_idx];
+
+  auto *image_out = reinterpret_cast<T *>(sample.output);
+  const auto *image_in = reinterpret_cast<const T *>(sample.input);
+
+  const int H = sample.shape[0];
+  const int W = sample.shape[1];
+  const int C = sample.shape[2];
```
mzient, line 85: These could be `fast_div` in `sample.shape`....

dali/operators/image/remap/displacement_filter_impl_gpu.cuh
```diff
@@ -47,155 +59,154 @@ __device__ inline void GetPixelValueMultiC(int h, int w,
...
+  PrepareParam<Displacement> pp;
+  pp(displace, sample.raw_params);
+
+  auto start = block.start.x;
+  auto end = block.end.x;
+  for (int64_t out_idx = threadIdx.x + start; out_idx < end; out_idx += blockDim.x) {
+    if (m) {
+      const int c = out_idx % C;
+      const int w = (out_idx / C) % W;
+      const int h = (out_idx / W / C);
```
mzient, line 97:
```suggestion
      int64_t idx = out_idx;
      const int c = idx % C;
      idx /= C;
      const int w = idx % W;
      idx /= W;
      const int h = idx;
```
or at least
```suggestion
      const int c = out_idx % C;
      const int w = (out_idx / C) % W;
      const int h = (out_idx / C / W);
```
The way it was written, it prevented optimization of the last two divisions.

mzient (approved):
Nitpicks, mostly.

dali/operators/image/remap/displacement_filter_impl_gpu.cuh
```diff
@@ -208,7 +219,10 @@ class DisplacementFilter<GPUBackend, Displacement,
   explicit DisplacementFilter(const OpSpec &spec) :
       Operator(spec),
       displace_(spec),
-      interp_type_(spec.GetArgument<DALIInterpType>("interp_type")) {
+      interp_type_(spec.GetArgument<DALIInterpType>("interp_type")),
+      flat_block_setup_(32),
+      channel_block_setup_(32) {
```
mzient, line 224: Why not just put it in the member definition?

dali/operators/image/remap/displacement_filter_impl_gpu.cuh
```diff
@@ -208,7 +219,10 @@ class DisplacementFilter<GPUBackend, Displacement,
   explicit DisplacementFilter(const OpSpec &spec) :
       Operator(spec),
       displace_(spec),
-      interp_type_(spec.GetArgument<DALIInterpType>("interp_type")) {
+      interp_type_(spec.GetArgument<DALIInterpType>("interp_type")),
```
mzient, line 222: Order of initialization is wrong - and this could be an assignment inside the constructor body - the type of the variable is trivial whereas the initialization expression is not.

stiepan (approved):

dali/operators/image/remap/displacement_filter_impl_gpu.cuh
klecki: done

dali/operators/image/remap/displacement_filter_impl_gpu.cuh
klecki: Fixed the order.

dali/operators/image/remap/displacement_filter_impl_gpu.cuh
klecki: No particular reason, but due to the intricacies of C++, I tried it and could not use `flat_block_setup_ = {32};` only, `FlatBlockSetup flat_block_setup_ = FlatBlockSetup(32);`. Not sure which one is nicer.

dali/operators/image/remap/displacement_filter_impl_gpu.cuh
klecki: done

dali/operators/image/remap/displacement_filter_impl_gpu.cuh
klecki: Hmm, I can `flat_block_stup_{32}`.

dali/operators/image/remap/displacement_filter_impl_gpu.cuh
klecki: I won't be trying to benchmark that.

klecki:
!build

PR #3332, opened 2021-09-09
Improve tests for expected errors + fix PythonFunction

mzient:

## Description
- [X] **Bug fix** (*non-breaking change which fixes an issue*)

### What happened in this PR
* Add messages to raises and assert_raises in selected tests.
* Add more internal error checking in PythonFunction.
* Fix image decoder tests.

ops.py - PythonFunction operator family
Tests:
* python/test_pipeline_decorator.py
* python/test_pipeline.py
* python/test_operator_python_function.py
* python/test_dali_tf_dataset_mnist_eager.py
* python/test_operator_decoders_image.py
* python/test_operator_subscript.py
* python/test_backend_impl.py

N/A

### Tests
- [X] Existing tests apply
- [X] New tests added
  - [X] Python tests
  - [ ] GTests
  - [ ] Benchmark
  - [ ] Other
- [ ] N/A

New tests:
- invalid number of outputs
- multi-output return value not a tuple

### Documentation
- [ ] Documentation updated
  - [ ] Docstring
  - [ ] Doxygen
  - [ ] RST
  - [ ] Jupyter
  - [ ] Other
- [ ] N/A

#### Requirements
- [ ] N/A

**REQ IDs**: N/A

**JIRA TASK**: DALI-2234


dali/test/python/test_operator_decoders_image.py
```diff
@@ -356,8 +357,9 @@ def pipe(device):
         decoded = fn.decoders.image_slice(encoded, device=device, end=[10000], axes=[1])
         return decoded
     p = pipe(device)
-    assert_raises(RuntimeError, p.run)
+    p.build()
+    assert_raises(RuntimeError, p.run, glob='*IsInRange*failed*')
```
JanuszL, line 361: `IsInRange`? Is this a part of the assert or error message itself?

dali/test/python/test_operator_python_function.py
```diff
@@ -204,6 +204,12 @@ def flip(image):
     return numpy.fliplr(image)
 
 
+def dlflip(image):
```
JanuszL, line 207: Do you use it anywhere?

dali/test/python/test_operator_python_function.py
```diff
@@ -289,15 +295,15 @@ def invalid_function(image):
     return img
 
 
-@raises(Exception)
+@raises(RuntimeError, regex="img.*not defined")
```
JanuszL, line 298: `img`? Shouldn't we print `image`?

dali/test/python/test_pipeline.py
```diff
@@ -1581,10 +1581,9 @@ def test_pipeline_wrong_device_id():
     pipe = dali.Pipeline(batch_size=1, num_threads=1, device_id=-123)
     with pipe:
         pipe.set_outputs(np.int32([1,2,3]))
-    with assert_raises(RuntimeError) as x:
+    with assert_raises(RuntimeError, regex="device_id"):
```
JanuszL, line 1584: Do we really expect to have only `device_id` in the error message? I would extend the regexp here,

dali/test/python/test_dali_tf_dataset_mnist_eager.py
```diff
@@ -29,7 +29,7 @@ def test_keras_single_cpu():
     run_keras_single_device('cpu', 0)
 
 
-@raises(Exception)
+@raises(Exception, regex="TF device and DALI device mismatch")
```
JanuszL, line 32: Why not use glob?

dali/test/python/test_dali_tf_dataset_mnist_eager.py
```diff
@@ -41,7 +41,7 @@ def test_keras_wrong_placement_gpu():
         steps_per_epoch=ITERATIONS)
 
 
-@raises(Exception)
+@raises(Exception, regex="TF device and DALI device mismatch")
```
JanuszL, line 44: Why not use glob?

dali/test/python/test_backend_impl.py
```diff
@@ -55,9 +55,9 @@ def test_tensorlist_getitem_cpu():
     assert(type(tensorlist[0]) == TensorCPU)
     assert(type(tensorlist[-3]) == TensorCPU)
     assert(len(list_of_tensors) == len(tensorlist))
-    with assert_raises(IndexError):
+    with assert_raises(IndexError, regex="out of range"):
```
JanuszL, line 58: Why not use glob?

dali/test/python/test_backend_impl.py
```diff
@@ -55,9 +55,9 @@ def test_tensorlist_getitem_cpu():
...
         tensorlist[len(tensorlist)]
-    with assert_raises(IndexError):
+    with assert_raises(IndexError, regex="out of range"):
```
JanuszL, line 60: Why not use glob?

dali/test/python/test_operator_python_function.py
mzient: `img` - look at the function above - it complains about `img` being undefined.

dali/test/python/test_dali_tf_dataset_mnist_eager.py
mzient: Because with regex it automatically matches it anywhere - I don't have to put `*`.

JanuszL: In case of glob you can omit leading and trailing as well. In that case it should be equivalent.

dali/test/python/test_operator_decoders_image.py
mzient: There's no message there.
```C++
DALI_ENFORCE(crop.IsInRange(shape));
```
```C++
DALI_ENFORCE(data.roi.IsInRange(dims));
```

dali/test/python/test_operator_python_function.py
mzient: Hmmm... I had it, but apparently it got lost somehow... strange.

dali/test/python/test_operator_python_function.py
JanuszL: Got it.

dali/test/python/test_operator_decoders_image.py
JanuszL: I would add it then. I though we want to get rid of printing assert itself and leave only error messages.

dali/test/python/test_pipeline.py
mzient: I can take a look what the message looks like. I don't remember what it was - but judging by the fact that I've only left `device_id`, it wasn't very clear.

JanuszL: So I would fix the message at the first place.

dali/test/python/test_dali_tf_dataset_mnist_eager.py
klecki: Yes, globs behave the same way.

dali/test/python/test_operator_decoders_image.py
klecki: Agreed, we decided to not rely on the DALI_ENFORCE condition, in this case I would introduce a proper message.

dali/test/python/test_dali_tf_dataset_mnist_eager.py
mzient: So, they are not really globs, are they?

dali/test/python/test_operator_decoders_image.py
mzient: Agreed... that we shouldn't. What's not necessarily agreed whether this is a part of this task or not.

dali/test/python/test_dali_tf_dataset_mnist_eager.py
JanuszL: https://docs.python.org/3/library/fnmatch.html#fnmatch.translate is used to translate glob into regexp. And then it is used to find any match, not necessarily whole string.

dali/test/python/test_backend_impl.py
mzient: Here I have to specify one or the other anyway (unlike assert, it requires a keyword argument), so it's a matter of taste.

JanuszL (approved):

dali/python/nvidia/dali/ops.py
```diff
@@ -949,35 +949,46 @@ def current_stream():
...
+            if len(outputs) != num_outputs:
+                raise ValueError(f"Unexpected number of outputs from Python function operator - got {len(outputs)}, expected {num_outputs}")
+
+    @staticmethod
+    def function_wrapper_per_sample(function, num_outputs, from_dlpack, to_dlpack, *dlpack_inputs):
         arrays = [from_dlpack(dlpack) for dlpack in dlpack_inputs]
         arr_out = function(*arrays)
         if arr_out is None:
             return
+        PythonFunction.check_outputs(arr_out, num_outputs)
```
klecki, line 965: Judging by the check below, that allowed `list`, aren't we restricting this operator only to return tuples and thus making a breaking change? 
I know that it is not like to happen in production code.

klecki (approved):
We might be introducing a breaking change, not sure what to do here, otherwise looks ok.

dali/util/crop_window.h
```diff
@@ -68,6 +73,26 @@ struct CropWindow {
   }
 };
 
+inline std::ostream &operator<<(std::ostream &os, const CropWindow &wnd) {
+  int ndim = wnd.anchor.size();
+  assert(wnd.shape.size() == ndim);
+  os << "[";
+  for (int i = 0; i < ndim; i++) {
+    if (i)
```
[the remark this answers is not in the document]
mzient: No? We specifically don't want one?

dali/python/nvidia/dali/ops.py
mzient: If it was ever possible, then it was a bug and likely against the documentation. If the function has multiple outputs, then they should be returned as a tuple, like `return a, b, c` would do.

dali/util/crop_window.h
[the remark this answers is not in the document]
klecki: Oh, I'm blind, we add only comma before.

dali/python/nvidia/dali/ops.py
klecki: Bugfix it is then.

dali/test/python/test_operator_python_function.py
mzient: Tests added.

PR #3338, opened 2021-09-14
Use DALIDataType where possible.

mzient:

## Description

### What happened in this PR
- modified type queries and setting the type to use DALIDataType instead of TypeInfo
- add templated set_type to buffer
- fix numerous inefficient usages of `TypeInfo::Create`

#### Additional information
- Everything

### Tests
- [X] Existing tests apply
- [ ] New tests added
  - [ ] Python tests
  - [ ] GTests
  - [ ] Benchmark
  - [ ] Other
- [ ] N/A

### Documentation
- [ ] Documentation updated
  - [ ] Docstring
  - [ ] Doxygen
  - [ ] RST
  - [ ] Jupyter
  - [ ] Other
- [ ] N/A

#### Requirements
- [x] N/A

**REQ IDs**: N/A

**JIRA TASK**: DALI-2283


dali/operators/generic/one_hot.h
```diff
@@ -102,12 +102,10 @@ class OneHot : public Operator<Backend> {
     for (int i = 0; i < num_samples; i++) {
       output_desc[0].shape.set_tensor_shape(i, determine_shape(shape[i], output_sample_dim));
     }
-    TYPE_SWITCH(output_type_, type2id, DType, ONE_HOT_TYPES, ({
-                  TypeInfo type;
-                  type.SetType<DType>(output_type_);
-                  output_desc[0].type = type;
-                }),
+    TYPE_SWITCH(output_type_, type2id, DType, ONE_HOT_TYPES, (
```
JanuszL, line 105: So now this switch is only for supported type validation, right?

mzient: Yes. I guess it could be done more elegantly, but I don't know if this is strictly necessary in this PR.

dali/pipeline/data/tensor_list.h
```diff
@@ -132,8 +132,8 @@ class DLL_PUBLIC TensorList : public Buffer<Backend> {
 
     use_copy_kernel &= (std::is_same<SrcBackend, GPUBackend>::value || other.is_pinned()) &&
                        (std::is_same<Backend, GPUBackend>::value || pinned_);
-    type.template Copy<SrcBackend, Backend>(dsts.data(), srcs.data(), sizes.data(),
-                                            nsamples, stream, use_copy_kernel);
+    type_.template Copy<SrcBackend, Backend>(dsts.data(), srcs.data(), sizes.data(),
```
JanuszL, line 135: What difference does it make?

dali/pipeline/data/tensor_vector.cc
```diff
@@ -149,8 +149,10 @@ void TensorVector<Backend>::SetSize(int new_size) {
 
 
 template <typename Backend>
-void TensorVector<Backend>::set_type(const TypeInfo &new_type) {
-  type_ = new_type;
+void TensorVector<Backend>::set_type(DALIDataType new_type) {
+  if (type_.id() == new_type)
+    return;
+  type_ = TypeTable::GetTypeInfo(new_type);
   tl_->set_type(new_type);
```
JanuszL, line 156: Maybe we should keep `inline void set_type(const TypeInfo& new_type) {` and use it here, as you call `GetTypeInfo` anyway you can use it directly instead calling it again inside `set_type`.

docs/examples/custom_operations/custom_operator/create_a_custom_operator.ipynb
```diff
@@ -186,13 +186,16 @@
...
+      "  const auto &shape = input.shape();\r\n",
       "  auto &output = ws.Output<::dali::GPUBackend>(0);\r\n",
-      "  CUDA_CALL(cudaMemcpyAsync(\r\n",
-      "          output.raw_mutable_data(),\r\n",
-      "          input.raw_data(),\r\n",
-      "          input.nbytes(),\r\n",
-      "          cudaMemcpyDeviceToDevice,\r\n",
-      "          ws.stream()));\r\n",
+      "  for (int sample_idx = 0; sample_idx < shape.num_samples(); sample_idx++) {\r\n",
+      "    CUDA_CALL(cudaMemcpyAsync(\r\n",
```
JanuszL, line 192: Is this due to rework towards non-contiguous TL/TV memory?

JanuszL (approved):

dali/pipeline/data/tensor_list.h
mzient: Unlike `type_`, `type` is just an identifier. This line, however, takes place after set_type (line 114), so we can safely use the member `type_` to access its extended functionality.

dali/pipeline/data/tensor_vector.cc
mzient: I specifically want to discourage that.

docs/examples/custom_operations/custom_operator/create_a_custom_operator.ipynb
mzient: Looks like it wasn't regenerated after that change.

dali/pipeline/data/tensor_vector.cc
```diff
@@ -149,8 +149,10 @@ void TensorVector<Backend>::SetSize(int new_size) {
 
 
 template <typename Backend>
-void TensorVector<Backend>::set_type(const TypeInfo &new_type) {
-  type_ = new_type;
+void TensorVector<Backend>::set_type(DALIDataType new_type) {
+  if (type_.id() == new_type)
+    return;
+  type_ = TypeTable::GetTypeInfo(new_type);
```
klecki, line 155: Just a nitpick, but the tl_->set_type checks for valid type, so maybe, we should:
```suggestion
  tl_->set_type(new_type);
  for (auto t : tensors_) {
    t->set_type(new_type);
  }
  type_ = tl_->type_info();
```


This also works slightly differently if we have NoType, and try to set it - it will be a no op - but it was already broken a bit, so we probably don't need to mix in those changes here.

dali/pipeline/data/tensor_list.h
klecki: type is now an enum.

dali/pipeline/data/tensor_vector.cc
mzient: If anything, it is _less_ broken now :)

dali/operators/reader/loader/numpy_loader.h
```diff
@@ -36,21 +36,25 @@
...
 namespace dali {
 
-TypeInfo TypeFromNumpyStr(const std::string &format);
+const TypeInfo &TypeFromNumpyStr(const std::string &format);
 
 class NumpyParseTarget{
  public:
   std::vector<int64_t> shape;
-  TypeInfo type_info;
+  const TypeInfo *type_info;
```
klecki, line 44: Shouldn't we initialize it with nullptr?

klecki (commented):
There are few places where we can get rid of type to enum trait, dunno if it's worth the effort.

One suspected uninitialized ptr. 
Not a fan of taking pointer to `const TypeInfo&` returned by the TypeTable - I know it's static, but still if I found that line without that context it would be a bit weird.

dali/operators/audio/mfcc/mfcc.cc
```diff
@@ -128,12 +128,12 @@ bool MFCC<CPUBackend>::SetupImpl(std::vector<OutputDesc> &output_desc,
...
                make_string("Axis ", axis_, " is out of bounds [0,", ndim, ")"));
 
-  TYPE_SWITCH(input.type().id(), type2id, T, MFCC_SUPPORTED_TYPES, (
+  TYPE_SWITCH(input.type(), type2id, T, MFCC_SUPPORTED_TYPES, (
     VALUE_SWITCH(in_shape.sample_dim(), Dims, MFCC_SUPPORTED_NDIMS, (
       using DctKernel = kernels::signal::dct::Dct1DCpu<T, T, Dims>;
       kmgr_.Initialize<DctKernel>();
       kmgr_.Resize<DctKernel>(nthreads, nsamples);
-      output_desc[0].type = TypeInfo::Create<T>();
+      output_desc[0].type = type2id<T>::value;
```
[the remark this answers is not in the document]
mzient: I had mixed feelings here, too - using T is consistent with the rest of the code in TYPE_SWITCH body.

dali/operators/reader/loader/numpy_loader.h
mzient: I guess we should - that, and other members.

dali/operators/audio/mfcc/mfcc.cc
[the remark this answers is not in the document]
mzient: But I guess I mixed it up already.

dali/operators/reader/loader/numpy_loader.h
mzient: Done.

mzient:
> There are few places where we can get rid of type to enum trait, dunno if it's worth the effort.

I think it will help avoid bugs when copy-pasting in cases where the output type is not equal to the input type.

> One suspected uninitialized ptr.

Fixed

> Not a fan of taking pointer to `const TypeInfo&` returned by the TypeTable - I know it's static, but still if I found that line without that context it would be a bit weird.

I know it's not pretty., but it's better than keeping copies of TypeInfo.
I had an idea of making `TypeInfo` a wrapper for a pointer to the actual (and unique) type descirption object. The actual type description would have private constructors, available only to `TypeTable`. I just don't think it's worth the effort. Perhaps if we need more elaborate type system (with some kind of reflection, like generic ifrastructure for obtaining element types and the like), this effort would be justified - now it seems excessive.

dali/operators/image/crop/crop_mirror_normalize.cc
```diff
@@ -86,7 +86,7 @@ bool CropMirrorNormalize<CPUBackend>::SetupImpl(std::vector<OutputDesc> &output_
       VALUE_SWITCH(ndim, Dims, CMN_NDIMS, (
         using Kernel = kernels::SliceFlipNormalizePermutePadCpu<OutputType, InputType, Dims>;
         using Args = kernels::SliceFlipNormalizePermutePadArgs<Dims>;
-        output_desc[0].type = TypeInfo::Create<OutputType>();
+        output_desc[0].type = type2id<OutputType>::value;
```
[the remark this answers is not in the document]
mzient: Fixed.

dali/pipeline/data/tensor_vector.cc
mzient: Fixed.

dali/operators/audio/mfcc/mfcc.cc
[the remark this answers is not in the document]
mzient: Done.

dali/operators/reader/loader/webdataset_loader.cc
```diff
@@ -34,8 +34,8 @@ namespace detail {
 namespace wds {
 
 inline MissingExtBehavior ParseMissingExtBehavior(std::string missing_component_behavior) {
-  std::transform(missing_component_behavior.begin(), missing_component_behavior.end(),
-                 missing_component_behavior.begin(), static_cast<int (*)(int)>(std::tolower));
+  for (auto &c : missing_component_behavior)
+      c = std::tolower(static_cast<unsigned char >(ctr));
```
mzient, line 38: According to C standard, it's UB to call `std::tolower` with `char` when out-of-ascii values are expected. Since this values come from the user, _any_ values are expected, including maliciously crafted ones.

JanuszL (approved):

dali/operators/reader/loader/webdataset_loader.cc
```diff
@@ -34,8 +34,8 @@ namespace detail {
 namespace wds {
 
 inline MissingExtBehavior ParseMissingExtBehavior(std::string missing_component_behavior) {
-  std::transform(missing_component_behavior.begin(), missing_component_behavior.end(),
-                 missing_component_behavior.begin(), static_cast<int (*)(int)>(std::tolower));
+  for (auto &c : missing_component_behavior)
+      c = std::tolower(static_cast<unsigned char >(c));
```
barci2, line 38:
```suggestion
      c = std::tolower(static_cast<unsigned char>(c));
```

dali/operators/reader/loader/webdataset_loader.cc
```diff
@@ -157,15 +157,12 @@ WebdatasetLoader::WebdatasetLoader(const OpSpec& spec)
...
-  dtypes_.reserve(dtypes_ids.size());
-  std::transform(dtypes_ids.begin(), dtypes_ids.end(), std::back_inserter(dtypes_),
-                 TypeTable::GetTypeInfo);
+  dtypes_ = spec.HasArgument("dtypes") ? spec.GetRepeatedArgument<DALIDataType>("dtypes")
+                                       : std::vector<DALIDataType>(ext_.size(), DALI_UINT8);
 
-  for (auto& dtype : dtypes_) {
-    DALI_ENFORCE(detail::wds::kSupportedTypes.count(dtype.id()),
-                 make_string("Unsupported output dtype ", dtype.name(),
+  for (DALIDataType dtype : dtypes_) {
```
barci2, line 163: why remove a reference here?

dali/operators/reader/loader/webdataset_loader.cc
```diff
@@ -157,15 +157,12 @@ WebdatasetLoader::WebdatasetLoader(const OpSpec& spec)
...
+    DALI_ENFORCE(detail::wds::kSupportedTypes.count(dtype),
+                 make_string("Unsupported output dtype ", dtype,
```
barci2, line 165: won't that get converted into an integer? Can you confirm that it produces a meaningful error?

dali/operators/reader/loader/webdataset_loader.cc
```diff
@@ -300,6 +297,11 @@ void WebdatasetLoader::PrepareMetadataImpl() {
   bitmask was_output_set;
   was_output_set.resize(ext_.size(), false);
   output_indicies_.reserve(ext_.size());
+
+  std::vector<size_t> dtype_sizes_(dtypes_.size());
+  for (size_t i = 0; i < dtypes_.size(); i++)
+    dtype_sizes_[i] = TypeTable::GetTypeInfo(dtypes_[i]).size();
+
```
barci2, line 304: See the coment below

dali/operators/reader/loader/webdataset_loader.cc
```diff
@@ -320,7 +322,7 @@ void WebdatasetLoader::PrepareMetadataImpl() {
         for (auto& output : ext_map[component.ext]) {
           if (!was_output_set[output]) {
             DALI_ENFORCE(
-                component.size % dtypes_[output].size() == 0,
+                component.size % dtype_sizes_[output] == 0,
```
barci2, line 325: why not just call TypeTable here? creating dtype_sizes_ produces unnecessary code, and I believe that it's not that much of a performance penalty since for example reallocating vectors or reading component data is much more expensive

dali/pipeline/data/types.h
```diff
@@ -436,8 +439,9 @@ class DLL_PUBLIC TypeTable {
   }
 
   template <typename T>
-  DLL_PUBLIC static const TypeInfo& GetTypeInfoFromStatic() {
-    return GetTypeInfo(GetTypeID<T>());
+  DLL_PUBLIC static const TypeInfo& GetTypeInfo() {
+    static const TypeInfo &type_info = GetTypeInfo(GetTypeID<T>());
+    return type_info;
```
barci2, line 444: Why can't we just return the reference directly?

barci2 (commented):
Some issues with the changes to WDS

dali/operators/reader/loader/webdataset_loader.cc
barci2: what is ctr?

dali/pipeline/data/types.h
mzient: Because we want to get it only once in the lifetime of the process.

dali/operators/reader/loader/webdataset_loader.cc
mzient: Because it's effectively a number - dereferencing it is counterproductive.

dali/operators/reader/loader/webdataset_loader.cc
mzient: We do this everywhere, all the time.

dali/operators/reader/loader/webdataset_loader.cc
mzient: Creating dtype_sizes_ once per dataset is much cheaper than querying the type table by dynamic id over and over again for each sample.

klecki (approved):

dali/operators/reader/loader/webdataset_loader.cc
klecki:
```suggestion
      c = std::tolower(static_cast<unsigned char>(c));
```
I wonder why the linter didn't scream. But I guess we can fix in a followup.

PR #3347, opened 2021-09-17
Handle SIGV situation when trying to load prebuild DALI TF Plugin

JanuszL:
- during DALI TF plugin installation first, prebuild plugins are tested
  for compatibility. If the attempt to load it causes SIGV
  the installation process is interrupted and no attempt to build
  a plugin from the source is made. This PR moves the loading to
  a subprocess to avoid the main process crash.

## Description

### What happened in this PR
- during DALI TF plugin installation first, prebuild plugins are tested
  for compatibility. If the attempt to load it causes SIGV
  the installation process is interrupted and no attempt to build
  a plugin from the source is made. This PR moves the loading to
  a subprocess to avoid the main process crash.

#### Additional information
  - dali_tf_plugin_install_tool.py

  - NA

### Tests
- [x] Existing tests apply
- [ ] New tests added
  - [ ] Python tests
  - [ ] GTests
  - [ ] Benchmark
  - [ ] Other
- [ ] N/A

### Documentation
- [ ] Documentation updated
  - [ ] Docstring
  - [ ] Doxygen
  - [ ] RST
  - [ ] Jupyter
  - [ ] Other
- [ ] N/A

#### Requirements
- [x] N/A

**REQ IDs**: N/A

**JIRA TASK**: N/A

### DIFF
--- a/dali_tf_plugin/dali_tf_plugin_install_tool.py
+++ b/dali_tf_plugin/dali_tf_plugin_install_tool.py
@@ -20,6 +20,7 @@ from distutils.version import StrictVersion
 from pathlib import Path
 import tempfile
 from stubgen import stubgen
+from multiprocessing import Process
 
 class InstallerHelper:
     def __init__(self, plugin_dest_dir = None):
@@ -122,8 +123,15 @@ class InstallerHelper:
 
             try:
                 print("Loading DALI TF library: ", lib_path_tmpdir)
-                tf.load_op_library(lib_path_tmpdir)
-                return True
+                # try in a separate process just in case it recives SIGV
+                p = Process(target=tf.load_op_library, args=(lib_path_tmpdir, ))
+                p.start()
+                p.join(5)
+                ret = p.exitcode
+                if ret is None:
+                    p.terminate()
+                    p.join()
+                return ret == 0
             except Exception as e:
                 print("Failed to import TF library: ", str(e))
                 return False


stiepan (approved):

awolant (approved):

dali_tf_plugin/dali_tf_plugin_install_tool.py
```diff
@@ -122,8 +123,15 @@ def check_import(self, lib_path, dali_stub):
 
             try:
                 print("Loading DALI TF library: ", lib_path_tmpdir)
-                tf.load_op_library(lib_path_tmpdir)
-                return True
+                # try in a separate process just in case it recives SIGV
+                p = Process(target=tf.load_op_library, args=(lib_path_tmpdir, ))
```
klecki, line 127: Just a question, do the exceptions propagate somehow through the Process, or we just basically look at the exit code now?

stiepan: They don't out of the box. I think in this case exception from the worker process only results in non zero exit code.

JanuszL: I checked:
```
from multiprocessing import Process

def my_f():
  1/0

p = Process(target=my_f)
p.start()
p.join()
p.exitcode
```
It shows 1. Other exceptions, like `StopIteration` also shows 1. I think the change covers this case.

PR #3354, opened 2021-09-20
Fix variable batch size handling in pad operator

banasraf:
Signed-off-by: Rafal <Banas.Rafal97@gmail.com>

## Description

### What happened in this PR
Currently, when shape is not provided, it is set to default value, but only once, due to the check in line 95 (`if shape.empty()`).
This caused problems when the batch size was changed because the shape was not adjusted to that. 
My fix checks if the shape is provided in the current iteration instead of checking if it's empty, so whenever the shape arg is not provided, the shape will be filled with default value of a correct size. 

#### Additional information
Pad setup and var BS tests

Pad

### Tests
- [ ] Existing tests apply
- [ ] New tests added
  - [x] Python tests
  - [ ] GTests
  - [ ] Benchmark
  - [ ] Other
- [ ] N/A

### Documentation
- [ ] Documentation updated
  - [ ] Docstring
  - [ ] Doxygen
  - [ ] RST
  - [ ] Jupyter
  - [ ] Other
- [x] N/A

#### Requirements
- [x] N/A

**REQ IDs**: N/A

**JIRA TASK**: DALI-2290

### DIFF
--- a/dali/operators/generic/pad.h
+++ b/dali/operators/generic/pad.h
@@ -67,6 +67,7 @@ class Pad : public Operator<Backend> {
     auto in_layout = input.GetLayout();
     int ndim = in_shape.sample_dim();
     int nsamples = in_shape.num_samples();
+    bool shape_arg_provided = false;
 
     if (!axis_names_.empty()) {
       axes_ = GetDimIndices(in_layout, axis_names_).to_vector();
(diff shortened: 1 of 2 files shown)


dali/operators/generic/pad.h
```diff
@@ -85,12 +86,13 @@ class Pad : public Operator<Backend> {
...
     if (spec.ArgumentDefined("shape")) {
       GetShapeArgument(shape_, spec, "shape", ws, curr_batch_size);
+      shape_arg_provided = true;
     }
     if (spec.ArgumentDefined("align")) {
       GetShapeArgument(align_, spec, "align", ws, curr_batch_size);
     }
 
-    if (shape_.empty())
+    if (!shape_arg_provided)
```
JanuszL, line 95: Can you write more in the PR description on how this fixes the bug?
Can the shape be provided and still be empty?

mzient: I think it would be taken care of in line 105 - if `axes` are empty, they will be implicitly extended to cover all axes in the input (which, by the way, is not ideal) and the comparison will fail.

mzient (approved):

dali/operators/generic/pad.h
banasraf: I updated the description (__What happened in this PR__ section)

banasraf:
!build

JanuszL (approved):

PR #3355, opened 2021-09-20
Fix Video reader race condition

JanuszL:
- there is a race condition between the video reader
  destructor and the reader Run method. After the reader
  run method, all current batch samples are zeroed
  to recycle the underlying storage. However, the video
  reader destructor calls the` wait` method on all samples
  in the prefetch buffer to make sure that all underlying
  kernels that operate on them ended. There is a short period
  between zeroing samples and creating a fresh batch where
  the destructor may attempt to access nullptr. This PR fixes
  the issue but not only clearing all samples but reducing
  vector length that holds them to 0 so the destructor won't
  access them as it iterates only over available samples in the vector

## Description

### What happened in this PR
- there is a race condition between the video reader
  destructor and the reader Run method. After the reader
  run method, all current batch samples are zeroed
  to recycle the underlying storage. However, the video
  reader destructor calls the` wait` method on all samples
  in the prefetch buffer to make sure that all underlying
  kernels that operate on them ended. There is a short period
  between zeroing samples and creating a fresh batch where
  the destructor may attempt to access nullptr. This PR fixes
  the issue but not only clearing all samples but reducing
  vector length that holds them to 0 so the destructor won't
  access them as it iterates only over available samples in the vector

#### Additional information
  - reader_op.

  - NA 

### Tests
- [x] Existing tests apply
- [ ] New tests added
  - [ ] Python tests
  - [ ] GTests
  - [ ] Benchmark
  - [ ] Other
- [ ] N/A

### Documentation
- [ ] Documentation updated
  - [ ] Docstring
  - [ ] Doxygen
  - [ ] RST
  - [ ] Jupyter
  - [ ] Other
- [ ] N/A

#### Requirements
- [x] N/A

**REQ IDs**: N/A

**JIRA TASK**: DALI-2292

### DIFF
--- a/dali/operators/reader/reader_op.h
+++ b/dali/operators/reader/reader_op.h
@@ -179,9 +179,8 @@ class DataReader : public Operator<Backend> {
     // Consume batch
     Operator<Backend>::Run(ws);
     CUDA_CALL(cudaStreamSynchronize(ws.stream()));
-    for (int sample_idx = 0; sample_idx < max_batch_size_; sample_idx++) {
-      auto sample = MoveSample(sample_idx);
-    }
+    // clear samples from the current batch so they can be recycled
+    ClearCurentBatch();
 
     // Notify we have consumed a batch
     ConsumerAdvanceQueue();
@@ -214,11 +213,8 @@ class DataReader : public Operator<Backend> {
     return *GetCurrBatch()[sample_idx];
   }
 
-  LoadTargetPtr MoveSample(int sample_idx) {
-    auto &sample = prefetched_batch_queue_[curr_batch_consumer_][sample_idx];
-    auto sample_ptr = std::move(sample);
-    sample = {};
-    return sample_ptr;
+  void ClearCurentBatch() {
+    prefetched_batch_queue_[curr_batch_consumer_].clear();
   }
 
  protected:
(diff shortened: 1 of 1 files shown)


JanuszL:
!build

mzient (approved):

dali/operators/reader/reader_op.h
```diff
@@ -179,9 +179,8 @@ class DataReader : public Operator<Backend> {
     // Consume batch
     Operator<Backend>::Run(ws);
     CUDA_CALL(cudaStreamSynchronize(ws.stream()));
-    for (int sample_idx = 0; sample_idx < max_batch_size_; sample_idx++) {
-      auto sample = MoveSample(sample_idx);
-    }
+    // clear samples from the current batch so they can be recycled
+    ClearCurentBatch();
```
mzient, line 183: So, after `Run` this queue entry is empty, right? It's now the responsibility of `Prefetch` to resize it to the right size.
While you're at it, could you swap `reserve` and `clear` in `Prefetch`? Now `reserve` goes before `clear` which is rather counterproductive.

mzient (approved):

dali/operators/reader/reader_op.h
JanuszL: > So, after Run this queue entry is empty, right?

Yes

> It's now the responsibility of Prefetch to resize it to the right size.

It always was but was used only during the first iterations. Now it happens always.

> While you're at it, could you swap reserve and clear in Prefetch? Now reserve goes before clear which is rather counterproductive.

Done.

JanuszL:
!build

stiepan (approved):

PR #3373, opened 2021-09-24
Remove possibility of access to contiguous TL buffer

klecki:
## Description

### What happened in this PR
Keep escape-hatch functions for the purpose of Pipeline
output.

This is intended as intermediate step, the main
purpose is to not introduce the access to the underlying
buffer again into the code-base.

Get rid of those functions from TL tests.

#### Additional information
Buffer, Tensor, Tensor List, Pipeline outputs

### Tests
- [x] Existing tests apply
- [ ] New tests added
  - [ ] Python tests
  - [x] GTests
  - [ ] Benchmark
  - [ ] Other
- [ ] N/A

### Documentation
- [x] Documentation updated
  - [x] Docstring
  - [ ] Doxygen
  - [ ] RST
  - [ ] Jupyter
  - [ ] Other
- [ ] N/A

#### Requirements
- [x] N/A

**REQ IDs**: N/A

**JIRA TASK**: DALI-2255

### DIFF
--- a/dali/pipeline/data/buffer.h
+++ b/dali/pipeline/data/buffer.h
@@ -87,6 +87,71 @@ class DLL_PUBLIC Buffer {
   virtual ~Buffer() = default;
 
   /**
+   * @brief Returns a typed pointer to the underlying storage. If the
+   * buffer has not been allocated because it does not yet have a type,
+   * the calling type is taken to be the type of the data and the memory
+   * is allocated.
+   *
+   * If the buffer already has a valid type, and the calling type does
+   * not match, the type of the buffer is reset and the underlying
+   * storage is re-allocated if the buffer does not currently own
+   * enough memory to store the current number of elements with the
+   * new data type.
+   */
+  template <typename T>
+  inline T* mutable_data() {
+    // Note: Call to 'set_type' will immediately return if the calling
+    // type matches the current type of the buffer.
+    set_type<T>();
+    return static_cast<T*>(data_.get());
+  }
+
+  /**
+   * @brief Returns a const, typed pointer to the underlying storage.
+   * The calling type must match the underlying type of the buffer.
+   */
+  template <typename T>
+  inline const T* data() const {
+    // clang-format off
+    DALI_ENFORCE(IsValidType(type_),
+                 "Buffer has no type, 'mutable_data<T>()' must be called "
+                 "on non-const buffer to set valid type for " + type_.name());
+    DALI_ENFORCE(type_.id() == TypeTable::GetTypeID<T>(),
+                 "Calling type does not match buffer data type: " +
+                 TypeTable::GetTypeName<T>() + " vs " + type_.name());
+    // clang-format on
+    return static_cast<T*>(data_.get());
+  }
+
+  /**
+   * @brief Return an un-typed pointer to the underlying storage.
+   * A valid type must be set prior to calling this method by calling
+   * the non-const version of the method, or calling 'set_type'.
+   */
+  inline void* raw_mutable_data() {
+    // Empty tensor
+    if (data_ == nullptr) return nullptr;
+    DALI_ENFORCE(IsValidType(type_),
+                 "Buffer has no type, 'mutable_data<T>()' or 'set_type' must "
+                 "be called on non-const buffer to set valid type");
+    return static_cast<void*>(data_.get());
+  }
+
+  /**
+   * @brief Return an const, un-typed pointer to the underlying storage.
+   * A valid type must be set prior to calling this method by calling
+   * the non-const version of the method, or calling 'set_type'.
+   */
+  inline const void* raw_data() const {
+    // Empty tensor
+    if (data_ == nullptr) return nullptr;
+    DALI_ENFORCE(IsValidType(type_),
+                 "Buffer has no type, 'mutable_data<T>()' or 'set_type' must "
+                 "be called on non-const buffer to set valid type");
+    return static_cast<void*>(data_.get());
+  }
+
+  /**
    * @brief Returns the size in elements of the underlying data
    */
   inline Index size() const {
@@ -277,71 +342,6 @@ class DLL_PUBLIC Buffer {
   static constexpr double kMaxGrowthFactor = 4;
 
  protected:
-  /**
-   * @brief Returns a typed pointer to the underlying storage. If the
-   * buffer has not been allocated because it does not yet have a type,
-   * the calling type is taken to be the type of the data and the memory
-   * is allocated.
-   *
-   * If the buffer already has a valid type, and the calling type does
-   * not match, the type of the buffer is reset and the underlying
-   * storage is re-allocated if the buffer does not currently own
-   * enough memory to store the current number of elements with the
-   * new data type.
-   */
-  template <typename T>
-  inline T* mutable_data() {
-    // Note: Call to 'set_type' will immediately return if the calling
-    // type matches the current type of the buffer.
-    set_type<T>();
-    return static_cast<T*>(data_.get());
-  }
-
-  /**
-   * @brief Returns a const, typed pointer to the underlying storage.
-   * The calling type must match the underlying type of the buffer.
-   */
-  template <typename T>
-  inline const T* data() const {
-    // clang-format off
-    DALI_ENFORCE(IsValidType(type_),
-                 "Buffer has no type, 'mutable_data<T>()' must be called "
-                 "on non-const buffer to set valid type for " + type_.name());
-    DALI_ENFORCE(type_.id() == TypeTable::GetTypeID<T>(),
-                 "Calling type does not match buffer data type: " +
-                 TypeTable::GetTypeName<T>() + " vs " + type_.name());
-    // clang-format on
-    return static_cast<T*>(data_.get());
-  }
-
-  /**
-   * @brief Return an un-typed pointer to the underlying storage.
-   * A valid type must be set prior to calling this method by calling
-   * the non-const version of the method, or calling 'set_type'.
-   */
-  inline void* raw_mutable_data() {
-    // Empty tensor
-    if (data_ == nullptr) return nullptr;
-    DALI_ENFORCE(IsValidType(type_),
-                 "Buffer has no type, 'mutable_data<T>()' or 'set_type' must "
-                 "be called on non-const buffer to set valid type");
-    return static_cast<void*>(data_.get());
-  }
-
-  /**
-   * @brief Return an const, un-typed pointer to the underlying storage.
-   * A valid type must be set prior to calling this method by calling
-   * the non-const version of the method, or calling 'set_type'.
-   */
-  inline const void* raw_data() const {
-    // Empty tensor
-    if (data_ == nullptr) return nullptr;
-    DALI_ENFORCE(IsValidType(type_),
-                 "Buffer has no type, 'mutable_data<T>()' or 'set_type' must "
-                 "be called on non-const buffer to set valid type");
-    return static_cast<void*>(data_.get());
-  }
-
   // Helper to resize the underlying allocation
   inline void ResizeHelper(Index new_size) {
     ResizeHelper(new_size, type_);
--- a/dali/pipeline/data/buffer.h
+++ b/dali/pipeline/data/buffer.h
@@ -87,71 +87,6 @@ class DLL_PUBLIC Buffer {
   virtual ~Buffer() = default;
 
   /**
-   * @brief Returns a typed pointer to the underlying storage. If the
-   * buffer has not been allocated because it does not yet have a type,
-   * the calling type is taken to be the type of the data and the memory
-   * is allocated.
-   *
-   * If the buffer already has a valid type, and the calling type does
-   * not match, the type of the buffer is reset and the underlying
-   * storage is re-allocated if the buffer does not currently own
-   * enough memory to store the current number of elements with the
-   * new data type.
-   */
-  template <typename T>
-  inline T* mutable_data() {
-    // Note: Call to 'set_type' will immediately return if the calling
-    // type matches the current type of the buffer.
-    set_type<T>();
-    return static_cast<T*>(data_.get());
-  }
-
-  /**
-   * @brief Returns a const, typed pointer to the underlying storage.
-   * The calling type must match the underlying type of the buffer.
-   */
-  template <typename T>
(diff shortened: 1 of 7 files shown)


dali/pipeline/data/buffer.h
```diff
@@ -86,71 +86,6 @@ class DLL_PUBLIC Buffer {
   inline Buffer() = default;
   virtual ~Buffer() = default;
 
-  /**
```
klecki, line 89: This was just moved from `public:` to `protected:`

dali/pipeline/data/tensor.h
```diff
@@ -46,6 +46,57 @@ class Tensor : public Buffer<Backend> {
   inline ~Tensor() override = default;
 
 
+  /** @defgroup ContiguousAccessorMembers Contiguous accessor member functions
```
klecki, line 49: Reexposing protected functions.

dali/pipeline/data/tensor.h
```diff
@@ -46,6 +46,57 @@ class Tensor : public Buffer<Backend> {
...
+   * Public buffer access to Tensor underlying memory - Tensors are always contiguous.
+   * @{
+   */
+
+  /**
+   * @brief Returns a typed pointer to the underlying storage. If the
+   * buffer has not been allocated because it does not yet have a type,
```
JanuszL, line 56:
```suggestion
   * tensor has not been allocated because it does not yet have a type,
```

dali/pipeline/data/tensor.h
```diff
@@ -46,6 +46,57 @@ class Tensor : public Buffer<Backend> {
...
+   * the calling type is taken to be the type of the data and the memory
+   * is allocated.
+   *
+   * If the buffer already has a valid type, and the calling type does
```
JanuszL, line 60:
```suggestion
   * If the tensor already has a valid type, and the calling type does
```

dali/pipeline/data/tensor.h
```diff
@@ -46,6 +46,57 @@ class Tensor : public Buffer<Backend> {
...
+   * new data type.
+   */
+  template <typename T>
+  inline T* mutable_data() {
+    return Buffer<Backend>::template mutable_data<T>();
+  }
+
+  /**
+   * @brief Returns a const, typed pointer to the underlying storage.
+   * The calling type must match the underlying type of the buffer.
```
JanuszL, line 73:
```suggestion
   * The calling type must match the underlying type of the tensor.
```

dali/pipeline/data/tensor.h
JanuszL: As this is the Tensor API I would stick to using `tensor` in the docs.

dali/pipeline/data/tensor_list.h
```diff
@@ -603,6 +609,62 @@ class DLL_PUBLIC TensorList : public Buffer<Backend> {
...
+   * and can be used only for outputs of the pipeline that were made sure to be contiguous.
+   * Currently TensorList is contiguous by design, but it is up to change.
+   * @{
+   */
+
+  /**
+   * @brief Return an un-typed pointer to the underlying storage.
+   * Buffer must be either empty or have a valid type and be contiguous.
+   */
+  friend void *unsafe_raw_mutable_data(TensorList<Backend> &tl) {
```
JanuszL, line 653: Why we need that function? To highlight that this approach is going to be removed?

dali/pipeline/data/tensor_list.h
```diff
@@ -603,6 +609,62 @@ class DLL_PUBLIC TensorList : public Buffer<Backend> {
...
+  /** @defgroup ContiguousAccessorFunctions Fallback contiguous accessors
+   * Fallback access to contiguous data to TensorList. It should not be used for processing,
+   * and can be used only for outputs of the pipeline that were made sure to be contiguous.
+   * Currently TensorList is contiguous by design, but it is up to change.
+   * @{
+   */
+
+  /**
+   * @brief Return an un-typed pointer to the underlying storage.
+   * Buffer must be either empty or have a valid type and be contiguous.
```
JanuszL, line 651:
```suggestion
   * The memory must be either empty or have a valid type and be contiguous.
```

dali/pipeline/data/tensor_list.h
```diff
@@ -603,6 +609,62 @@ class DLL_PUBLIC TensorList : public Buffer<Backend> {
...
+   */
+  friend void *unsafe_raw_mutable_data(TensorList<Backend> &tl) {
+    DALI_ENFORCE(tl.IsContiguous(), "Data pointer can be obtain only for contiguous TensorList.");
+    return tl.raw_mutable_data();
+  }
+
+  /**
+   * @brief Return an un-typed const pointer to the underlying storage.
+   * Buffer must be either empty or have a valid type and be contiguous.
```
JanuszL, line 660:
```suggestion
   * The memory must be either empty or have a valid type and be contiguous.
```

dali/pipeline/data/tensor.h
klecki: I just copied the old Doxygen verbatim from tensor, I can adjust if you want.

dali/pipeline/data/tensor_list.h
klecki: So, I think we are going to keep some sort of this behaviour for pipeline output. But I don't want anyone using TensorList.raw_data() & co inside of regular code base (that is operator implementation etc). So this is the escape hatch that I mention in the PR description, but it's also documented and named with the `unsafe_` prefix to discourage the usage. The "I know what I'm doing" kind of code.

dali/pipeline/data/tensor_list_test.cc
```diff
@@ -73,12 +73,13 @@ class TensorListTest : public DALITest {
     }
 
     // Resize the buffer
-    tensor_list->Resize(shape);
+    tensor_list->Resize(shape, DALI_FLOAT);
 
     // Check the internals
-    ASSERT_NE(tensor_list->template mutable_data<float>(), nullptr);
+    ASSERT_TRUE(tensor_list->has_data());
```
JanuszL, line 79: Shouldn't you specialize `has_data` to the TensorList now (as it is implemented in the buffer)?

klecki: I maybe can do something, but I think it needs to wait for the proper changes.

dali/pipeline/data/tensor_list_test.cc
```diff
@@ -418,23 +457,19 @@ TYPED_TEST(TensorListTest, TestTypeChangeSmaller) {
...
   size_t nbytes = tensor_list.nbytes();
 
   // Change the data type to something smaller
-  tensor_list.template mutable_data<uint8>();
+  tensor_list.template set_type<uint8>();
 
   // Check the internals
   ASSERT_EQ(tensor_list.ntensor(), shape.size());
   for (size_t i = 0; i < tensor_list.ntensor(); ++i) {
+    // ASSERT_EQ(ptrs[i], tensor_list.raw_tensor(i));
```
JanuszL, line 468: ?

dali/pipeline/data/tensor_list_test.cc
```diff
@@ -418,23 +457,19 @@ TYPED_TEST(TensorListTest, TestTypeChangeSmaller) {
...
     ASSERT_EQ(tensor_list.tensor_shape(i), shape[i]);
     ASSERT_EQ(tensor_list.tensor_offset(i), offsets[i]);
   }
 
-  // No memory allocation should have occured
-  ASSERT_EQ(ptr, tensor_list.raw_data());
```
JanuszL, line 436: Maybe we can save ptrs to each tensor and then check if no reallocation has happened.

dali/pipeline/data/copy_to_external.h
```diff
@@ -44,39 +44,76 @@ struct kind2backend<mm::memory_kind::managed> {
...
 }
 
+template <typename DstBackend, typename SrcBackend>
+inline void CopyToExternalImpl(void* dst,
+                               const TensorList<SrcBackend> &src,
+                               cudaStream_t stream, bool use_copy_kernel) {
+  DeviceGuard d(src.device_id());
+  const auto &type_info = src.type_info();
+
+  // TODO(klecki): Add a proper test for non-contiguous access when we can have non-contiguous
```
JanuszL, line 62: Do think we will have such case?
I imagine that DALI still should return raw outputs as continuous tensors so we can wrap them into CuPy/NumPy or DLPack directly.

dali/pipeline/data/copy_to_external.h
```diff
@@ -44,39 +44,76 @@ struct kind2backend<mm::memory_kind::managed> {
...
+    for (int i = 0; i < num_samples; i++) {
+      sizes.push_back(src_shape.tensor_size(i));
+      to.push_back(dst_buf);
+      dst_buf += sizes[i] * type_info.size();
+      from.push_back(src.raw_tensor(i));
+    }
+
+    type_info.template Copy<DstBackend, SrcBackend>(to.data(), from.data(), sizes.data(),
+                                                    num_samples, stream, use_copy_kernel);
+  }
```
JanuszL, line 85:
```suggestion
    const auto &src_shape = src.shape();
    SmallVector<const void *, 256> from;
    SmallVector<int64_t, 256> sizes;
    int num_samples = src_shape.num_samples();
    sizes.reserve(num_samples);
    to.reserve(num_samples);
    from.reserve(num_samples);
    for (int i = 0; i < num_samples; i++) {
      sizes.push_back(src_shape.tensor_size(i));
      from.push_back(src.raw_tensor(i));
    }

    type_info.template Copy<DstBackend, SrcBackend>(dst, from.data(), sizes.data(),
                                                    num_samples, stream, use_copy_kernel);
  }
```
As we have:
```
template <typename DstBackend, typename SrcBackend>
void TypeInfo::Copy(void *dst, const void** srcs, const Index* sizes, int n,
                    cudaStream_t stream, bool use_copy_kernel) const {```

File: dali/pipeline/data/tensor.h (hunk 3 above)
Comment (JanuszL) [reply dali/pipeline/data/tensor.h:56]: I would appreciate that.

File: dali/pipeline/data/tensor_list.h (hunk 6 above)
Comment (JanuszL) [reply dali/pipeline/data/tensor_list.h:653]: Usually it is rather `I do whatever possible to make that working`.

File: dali/pipeline/data/buffer.h (hunk 1 above)
Comment (mzient) [reply dali/pipeline/data/buffer.h:89]: I don't like it. The `Buffer` class is really quite useless after this change. How about making the inheritance of TensorList from Buffer protected (or even private) instead?

Comment (JanuszL) [reply dali/pipeline/data/buffer.h:89]: useless -> redundant ?

Comment (klecki) [reply dali/pipeline/data/buffer.h]: Switched to private inheritance.

File: dali/pipeline/data/tensor_list.h (hunk 7 above)
Comment (klecki) [reply dali/pipeline/data/tensor_list.h:647]: done

File: dali/pipeline/data/tensor_list.h (hunk 8 above)
Comment (klecki) [reply dali/pipeline/data/tensor_list.h:656]: done

File: dali/pipeline/data/tensor_list_test.cc (hunk 11 above)
Comment (klecki) [reply dali/pipeline/data/tensor_list_test.cc:436]: I cannot check if no reallocation happen this way, as it will match only for the first pointer. I can add the unsafe call.

File: dali/pipeline/data/tensor_list_test.cc (hunk 10 above)
Comment (klecki) [reply dali/pipeline/data/tensor_list_test.cc:468]: done

File: dali/pipeline/data/copy_to_external.h (hunk 13 above)
Comment (klecki) [reply dali/pipeline/data/copy_to_external.h:85]: Done

File: dali/pipeline/data/copy_to_external.h (hunk 12 above)
Comment (klecki) [reply dali/pipeline/data/copy_to_external.h:62]: I just wrote it for the sake of TODO for the next PR mostly, I guess we won't be returning non-contiguous stuff soon, but either way I will add an error here or test this code-path and error out somewhere else.

File: dali/pipeline/data/tensor.h (hunk 3 above)
Comment (klecki) [reply dali/pipeline/data/tensor.h]: All removed, will resolve stale comments.

**Comment by klecki** [comment]
!build

**Review by awolant — APPROVED** [review]

File: dali/util/user_stream.h [hunk 14]
```diff
@@ -36,48 +39,53 @@ class DLL_PUBLIC UserStream {
...
-   * @brief Synchronizes on the the stream where buffer b was created
+   * @brief Synchronizes on the device where given Tensor t exists
    */
-  DLL_PUBLIC void Wait(const dali::Buffer<GPUBackend> &b) {
-    size_t dev = GetDeviceForBuffer(b);
-    DALI_ENFORCE(streams_.find(dev) != streams_.end(),
-        "Can only wait on user streams");
-    DeviceGuard g(dev);
-    CUDA_CALL(cudaStreamSynchronize(streams_[dev]));
+  DLL_PUBLIC void WaitForDevice(const dali::Tensor<GPUBackend> &t) {
```
JanuszL, line 66:
```suggestion
  template<template<GPUBackend>class BatchContainer>
  DLL_PUBLIC void WaitForDevice(const BatchContainer<GPUBackend> &t) {
```
?

JanuszL (approved):

dali/util/user_stream.h
klecki: That's the other alternative, I just went with two overloads, I guess a coin flip to decide.

JanuszL: Up to you

klecki:
!build

dali/pipeline/data/tensor_list.h
```diff
@@ -63,6 +63,31 @@ class DLL_PUBLIC TensorList : public Buffer<Backend> {
...
+  using Buffer<Backend>::device_id;
+  using Buffer<Backend>::set_device_id;
+  using Buffer<Backend>::set_type;
+  using Buffer<Backend>::reserve;
+  // using Buffer<Backend>::reset;  // Available via USE_BUFFER_MEMBERS
+  using Buffer<Backend>::shares_data;
+  using Buffer<Backend>::SetGrowthFactor;
+  using Buffer<Backend>::SetShrinkThreshold;
+  using Buffer<Backend>::GetGrowthFactor;
+  using Buffer<Backend>::GetShrinkThreshold;
```
mzient, line 89: Do we need those 4?

mzient (approved):

PR #3397, opened 2021-10-04
Parallel external source batch mode

stiepan:
## Description

* Add support for sources that return batches in parallel mode
Iterators, generator functions and callbacks that return batches are now supported.
For a source that is a generator function or an iterator a single worker from the poll is selected and a single instance of the source will always be run there to preserve state. Callbacks are assumed to be stateless, they should produce batch based on the batch index within the epoch - they can benefit from parallelization if the prefetch_queue_depth is set set accordingly for the given external source.

Implementation changes:
* Shared memory chunks are now shared among all workers that can compute given callback
* Shared memory fds now can be and are closed in the main process as soon as the pool is setup
* Tasks for stateless callbacks are pushed into a queue shared by all workers and taken whenever worker is idle
* Pipes are mostly removed and replaced with simple shared memory queue
#### Additional information

### Tests
- [ ] Existing tests apply
- [ ] New tests added
  - [ ] Python tests
  - [ ] GTests
  - [ ] Benchmark
  - [ ] Other
- [ ] N/A

### Documentation
- [ ] Documentation updated
  - [ ] Docstring
  - [ ] Doxygen
  - [ ] RST
  - [ ] Jupyter
  - [ ] Other
- [ ] N/A

#### Requirements
- [ ] N/A

**REQ IDs**: N/A

**JIRA TASK**: N/A


stiepan:
!build

include/dali/core/os/shared_mem.h
```diff
@@ -153,6 +153,8 @@ class DLL_PUBLIC SharedMem {
    */
   DLL_PUBLIC void resize(uint64_t size, bool trunc = false);
 
+  DLL_PUBLIC void seal();
```
JanuszL, line 156: Could you add docs for this function?

stiepan: Added brief description

stiepan:
!build

dali/python/nvidia/dali/_multiproc/messages.py
```diff
@@ -13,29 +13,152 @@
...
+        values = self._struct.unpack_from(buf, offset)
+        self.set_values(*values)
+        return self
+
+    def get_size(self):
+        return self._struct.size
+
+
+class ShmMessage(Structure):
+    _fields = ("worker_id", "i"), ("shm_chunk_id", "i"), ("shm_capacity", "i"), ("offset", "i"), ("num_bytes", "i")
```
JanuszL, line 65: Do you use the second element of each tuple anywhere? I see it is always referred as `_` and discarded.

dali/python/nvidia/dali/_multiproc/messages.py
```diff
@@ -13,29 +13,152 @@
...
+
+
+class BufShmChunkMeta:
+
```
JanuszL, line 69: Can you add a docs for it as well?

dali/python/nvidia/dali/_multiproc/shared_batch.py
```diff
@@ -178,20 +151,24 @@ def _apply_to_sample(func, sample, *args, nest_with_sample=0):
...
+    def __init__(self, shm_chunk: shared_mem.SharedMem, batch, min_trailing_offset=1024*1024):
         import_numpy()
-        self.mem_batch = mem_batch
+        self.shm_chunk = shm_chunk
         self.data_size = 0
         self.meta_data_size = 0
+        self.total_size = 0
+        # hint how much space should be left in case of the resize at the end of the shm chunk
+        # after batch data to accommodate meta data of the task
+        self.min_trailing_offset = min_trailing_offset
```
JanuszL, line 168: I think it should be aligned.

dali/python/nvidia/dali/_multiproc/shared_batch.py
```diff
@@ -217,42 +192,60 @@ def _add_array_to_batch(self, np_array, meta, memview):
...
-        if self.mem_batch.capacity < needed_capacity:
-            new_capacity = max(needed_capacity, 2 * self.mem_batch.capacity)
-            new_capacity = _align_up(new_capacity, self.BUFFER_ALIGNMENT)
-            self.mem_batch.resize(new_capacity)
-        memview = self.mem_batch.shm_chunk.buf
-        for (idx, sample), (meta_idx, sample_meta) in zip(batch, meta):
-            assert idx == meta_idx
+        self.total_size = _align_up(self.data_size + self.meta_data_size, self.SAMPLE_ALIGNMENT)
+        if self.shm_chunk.capacity < self.total_size:
+            self.resize_shm_chunk(self.shm_chunk, self.total_size + self.min_trailing_offset)
```
JanuszL, line 203: Similar to L168 - shouldn't it be aligned?

stiepan: I can make it explicitly aligned here, but resize_shm_chunk aligns that value up to the BUFFER_ALIGNMENT.

dali/python/nvidia/dali/_multiproc/shared_batch.py
```diff
@@ -217,42 +192,60 @@ def _add_array_to_batch(self, np_array, meta, memview):
...
+    serialized_message = pickle.dumps(message)
+    num_bytes = len(serialized_message)
+    if num_bytes > shm.shm_chunk.capacity - offset:
+        if resize:
+            SharedBatchWriter.resize_shm_chunk(shm.shm_chunk, offset + num_bytes)
+        else:
+            raise RuntimeError(
+                "Could not put message into shared memory region, not enough space in the buffer. "
+                "Consider specifying `bytes_per_minibatch_hint` of parallel external source to at "
+                "least {}".format(offset + num_bytes))
```
JanuszL, line 248: I don't see much value in that message.
I think it should always be resized. Because the fact the user adjusts the value to `offset + num_bytes` doesn't give any guarantee it will work for the next chunk/batch/message.

dali/test/python/test_external_source_parallel_shared_batch.py
```diff
@@ -58,5 +54,103 @@ def test_serialize_deserialize_random():
...
+    proc.start()
+    if start_method == "spawn":
+        pid = os.getppid()
+        multiprocessing.reduction.send_handle(socket_w, task_queue.shm.handle, pid)
+        multiprocessing.reduction.send_handle(socket_w, res_queue.shm.handle, pid)
+    return proc, task_queue, res_queue
+
+
+def _put_msgs(queue, msgs, one_by_one):
+    if not one_by_one:
```
JanuszL, line 88: What is the difference between these two?

dali/python/nvidia/dali/_multiproc/shared_batch.py
JanuszL: No need in such case.

dali/python/nvidia/dali/_multiproc/shared_batch.py
stiepan: This would happen only if parent process tried to put task description that doesn't fit into the given chunk. They are initialized to 1MB and resized by the worker if actual batch needs more space. I thought it is rather unusual situation to have such a big task description (it would have to be sample mode with batch size so big that mere SampleInfo list exceeds that size). Additionally I assumed that when workers resize the chunk for data then it later on it is even more unlikely for the task description to take more space than actual data.

If, however, that would be the case, user could specify bytes_per_minibatch_hint to be big enough.

Parent process could resize that if needed but then I need to drop the idea of closing file descriptors for shm in parent process.

dali/test/python/test_RN50_external_source_parallel_utils.py
```diff
@@ -102,24 +97,60 @@ def create_data_set(self):
...
 
-class CV2SampleLoaderPerm(CV2SampleLoader):
+class CV2BatchLoader(BatchLoader):
+    def read_file(self, file_path):
+        img = cv2.imread(file_path)
+        return img
 
-    DATA_SET = ShuffledFilesDataSet
+
+def create_dataset_generator(data_path, batch_size, mixed_decode):
```
JanuszL, line 130:
```suggestion
def create_dataset_generator(data_path, batch_size, read_raw):
```

dali/test/python/test_RN50_external_source_parallel_utils.py
```diff
@@ -136,56 +167,84 @@ def common_pipeline(images):
...
-    def __init__(self, data_path, **kwargs):
+    def __init__(self, data_path, mixed_decode, source_mode, **kwargs):
         super().__init__(**kwargs)
-        self.loader = CV2SampleLoaderPerm(data_path)
+        if source_mode == "generator":
+            self.loader, self.data_set_len = create_dataset_generator(
+                data_path, batch_size=kwargs['batch_size'], mixed_decode=mixed_decode)
+        else:
+            self.data_set_len = None
+            if mixed_decode:
```
JanuszL, line 200:
```suggestion
            if needs_decode:
```

dali/test/python/test_RN50_external_source_parallel_utils.py
```diff
@@ -225,6 +284,13 @@ def parse_test_arguments(supports_distributed):
     parser.add_argument(
         "--test_pipes", nargs="+", default=["parallel", "file_reader", "scalar"],
         help="Pipelines to be tested, allowed values: 'parallel', 'file_reader', 'scalar'")
+    parser.add_argument('--source_mode', default="sample", choices=['sample', 'batch', 'generator'], type=str,
+                        help='Available modes: sample, batch, generator. First two run stateless '
+                        'callbacks that return sample or batch given the index, the '
+                        'generator mode iterates over a generator. '
+                        'Parameter value has no effect on file reader pipeline.')
+    parser.add_argument('--mixed_decode', default=False, type=bool,
```
JanuszL, line 292:
```suggestion
    parser.add_argument('--dali_decode', default=False, type=bool,
```

dali/test/python/test_RN50_external_source_parallel_utils.py
```diff
@@ -225,6 +284,13 @@ def parse_test_arguments(supports_distributed):
...
+                        help='If True decodes with mixed decoder, otherwise decodes on cpu and moves with tensor.gpu()')
```
JanuszL, line 293:
```suggestion
                        help='If True decodes images with DALI's mixed decoder, otherwise decodes on cpu inside external source callback and moves with tensor.gpu()')
```

dali/python/nvidia/dali/_multiproc/shared_batch.py
JanuszL: Ok. Up to you then.

dali/python/nvidia/dali/_multiproc/shared_queue.py
```diff
@@ -0,0 +1,233 @@
...
+            return
+        with self.lock:
+            self.read_meta()
+            self.is_closed = True
+            if not self.meta.is_closed:
+                self.meta.is_closed = 1
+                self.write_meta()
+                # Notify only one waiting worker about closing of the queue, the woken up worker
+                # will notify the next one. Avoid notify_all at this point, due to possible deadlock if
+                # one of the notified workers exited abruptly when waiting on cv_empty without proper releasing
```
JanuszL, line 99: So if we notifying on one worker, how we avoid the deadlock? Still there could be a one process that crashed and hasn't released the underlying semaphore.

dali/python/nvidia/dali/_multiproc/shared_queue.py
```diff
@@ -0,0 +1,233 @@
...
+        with self.lock:
+            self.read_meta()
+            assert self.meta.size + len(msgs) <= self.meta.capacity, "The queue is full"
+            if self.meta.is_closed:
+                self.is_closed = True
+                return
+            msgs_len = len(msgs)
+            next_slot = (self.meta.begining + self.meta.size) % self.meta.capacity
+            for msg in msgs:
+                self.write_msg(next_slot, msg)
```
JanuszL, line 115: Can this handle wrap up?

dali/python/nvidia/dali/_multiproc/shared_queue.py
```diff
@@ -0,0 +1,233 @@
...
+            self.write_meta()
+            self.cv_empty.notify()
+        return msgs_len
+
+    def _recv_samples(self, num_samples):
+        num_take = self.meta.size
+        if num_samples is not None and num_samples < num_take:
+            num_take = num_samples
+        recv = [self.read_msg((self.meta.begining + i) % self.meta.capacity) for i in range(num_take)]
+        self.meta.size -= num_take
```
JanuszL, line 127:
```suggestion
        self.meta.size = 0
```
Or it is possible we won't read everything?

dali/python/nvidia/dali/_multiproc/shared_queue.py
```diff
@@ -0,0 +1,233 @@
...
+            if self.meta.is_closed:
+                self.is_closed = True
+                return
+            msgs_len = len(msgs)
+            next_slot = (self.meta.begining + self.meta.size) % self.meta.capacity
+            for msg in msgs:
+                self.write_msg(next_slot, msg)
+                next_slot = (next_slot + 1) % self.meta.capacity
+            self.meta.size += msgs_len
+            self.write_meta()
```
JanuszL, line 118: Can `_recv_samples` call `write_meta` between L107 and L118 in parallel?

dali/python/nvidia/dali/_multiproc/shared_queue.py
```diff
@@ -0,0 +1,233 @@
...
+        self.read_meta()
+        while not self.meta.size > 0 and not self.meta.is_closed:
+            self.cv_empty.wait()
+            waited = True
+            self.read_meta()
+        return waited
+
+    def get(self, num_samples=1, get_if_waited=None) -> Optional[List[MSG_CLASS]]:
+        if self.is_closed:
+            return
```
JanuszL, line 143: Or it should assert?

dali/python/nvidia/dali/_multiproc/shared_queue.py
```diff
@@ -0,0 +1,233 @@
...
+            self.pending_cv.notify()
+
+    def append(self, msg):
+        with self.pending_cv:
+            self.pending.append(msg)
+            self.pending_cv.notify()
+
+
+class DispatcherWorker:
+
```
JanuszL, line 200: You could add a docs here as well.

dali/python/nvidia/dali/external_source.py
```diff
@@ -340,31 +344,42 @@ class ExternalSource():
...
-    allows to add an initial state to the object instance.
+    Acceptable sources depend on the value specified for ``batch`` parameter.
 
-    Keep in mind, that **copies** of the ``source`` will be distributed between Python workers,
-    and no global state can be shared between them.
+    If batch is set to True, source must be a callable (a function or an object with __call__ method)
+    that accepts exactly one argument (:meth:`~nvidia.dali.types.SampleInfo` objects that
+    represent the index of the requested sample).
+    If ``batch=True``, then ``source`` can be either a callable, an iterable or a generator function.
+    Callable in batch mode must accept exactly one argument - an integer that represents the index of the
```
JanuszL, line 356: Does it accept `~nvidia.dali.types.SampleInfo` or just an integer?

dali/python/nvidia/dali/external_source.py
```diff
@@ -340,31 +344,42 @@ class ExternalSource():
...
+
+    The ``source`` callback must raise StopIteration when the end of data is reached.
 
 `prefetch_queue_depth` : int, option, default = 1
     When run in ``parallel=True`` mode, specifies the number of batches to be computed in advance and stored
     in the internal buffer, otherwise parameter is ignored.
+
+`bytes_per_minibatch_hint` : int, default = None
+    Hint for the initial size of each buffer accommodating (part of) the batch returned from worker
+    in parallel mode. You may want to specify custom value to lower the number of resizes necessary when
```
JanuszL, line 375: Is there any reasonable way to learn what is the right value? Or just some low level debugger is needed?

dali/test/python/test_pool.py
```diff
@@ -30,12 +31,44 @@ def simple_callback(info):
...
+class IteratorCb:
+    def __init__(self):
+        self.count = 0
+
+    def __iter__(self):
+        self.count = 0
+        return self
+
+    def __next__(self):
+        pid = os.getpid()
```
JanuszL, line 46: Maybe it can be moved to `__iter__` as we don't expect pid to change.

dali/test/python/test_external_source_parallel.py
```diff
@@ -59,21 +102,23 @@ def check_source_build(source):
...
-    common_msg = "External Source in parallel mode (when `parallel=True`) accepts as `source` only *. Got {} instead"
+    callable_msg = ("Callable passed to External Source in parallel mode (when `parallel=True`) "
+        "must accept exactly one argument*. Got {} instead.")
+    batch_required_msg = "Parallel external source with {} must be run in a batch mode"
     expected_error_msgs = [
-        common_msg.format("a callable that does not accept arguments"),
-        "External source callback must be a callable with 0 or 1 argument",
-        common_msg.format("an iterable"),
-        common_msg.format("a generator function")]
+        (TypeError, callable_msg.format("a callable that does not accept arguments")),
```
JanuszL, line 109: Maybe it is better to move this error messages next to `disallowed_sources` to make sure they match and are up to date.

dali/test/python/test_external_source_parallel.py
```diff
@@ -184,39 +240,259 @@ def __init__(self, name, shape):
...
+    while True:
+        yield [np.full([10], count + i) for i in range(50)]
 
+def generator_shape_100x3():
+    count = 0
+    while True:
+        yield [np.full([10, 10, 10], count + i) for i in range(50)]
+
+@with_setup(setup_function, teardown_function)
+def _test_generator_vs_non_parallel(cb):
```
JanuszL, line 283: How it differs compared to `_test_vs_non_parallel`?

dali/test/python/test_external_source_parallel.py
```diff
@@ -184,39 +240,259 @@ def __init__(self, name, shape):
...
+                assert len(batch) == len(expected_batch), "Batch length mismatch: expected {}, got {}".format(len(expected_batch), len(batch))
+                for sample, expected_sample in zip(batch, expected_batch):
+                    np.testing.assert_equal(sample, expected_sample)
+                i += 1
+            except StopIteration:
+                pipe.reset()
+                if is_gen_fun:
+                    refer_iter = cb()
+                else:
+                    refer_iter = iter(cb)
```
JanuszL, line 331: What is that for?

dali/python/nvidia/dali/_multiproc/shared_queue.py
stiepan: The main way to avoid this deadlock is notifying this way only under "normal" exit circumstances, i.e. when pool close method is called and if at the point of cleanup all workers are alive. Otherwise, tracker thread won't bother to call queue.close. This seems to be reliable, though one could imagine that workers drop dead just after the aliveness check and before main process calls the queue.close.
So notifying one worker (and properly releasing resources in worker processes if they are not immediately killed by the OS) serve as kind of additional safety measures.

JanuszL: Understood.

dali/python/nvidia/dali/_multiproc/shared_queue.py
```diff
@@ -0,0 +1,233 @@
...
+                # Notify only one waiting worker about closing of the queue, the woken up worker
+                # will notify the next one. Avoid notify_all at this point, due to possible deadlock if
+                # one of the notified workers exited abruptly when waiting on cv_empty without proper releasing
+                # of the underlying semaphore.
+                self.cv_empty.notify()
+
+    def put(self, msgs : List[MSG_CLASS]) -> Optional[int]:
+        if self.is_closed or not msgs:
+            return
+        with self.lock:
```
JanuszL, line 106: Why we use this lock only for put? Shouldn't we use it also for get?

stiepan:
!build

PR #3400, opened 2021-10-04
Fix pool process tracking in parallel ES tests, cleanup batches properly

stiepan:

## Description

Fixing two bugs for parallel external source: 
   * When batch received from workers in coalesced, its parts were not properly deleted in the pool context which results in keeping corrupted arrays (as the underlying buffer is overwritten) throughout the epoch - they were not exposed anywhere but still took some space and accidental access to them could kill process.
   * Tests for ES should check if pool processes are closed at the end of the test, but by mistake pids were compared against process instance, so this oranges to apples check was falsely passing all the time.

#### Additional information
Python part of External Source.

It is separated from bigger PR for batch mode in external source.

### Tests
- [ ] Existing tests apply
- [ ] New tests added
  - [ ] Python tests
  - [ ] GTests
  - [ ] Benchmark
  - [ ] Other
- [ ] N/A

### Documentation
- [ ] Documentation updated
  - [ ] Docstring
  - [ ] Doxygen
  - [ ] RST
  - [ ] Jupyter
  - [ ] Other
- [ ] N/A

#### Requirements
- [ ] N/A

**REQ IDs**: N/A

**JIRA TASK**: N/A

### DIFF
--- a/dali/python/nvidia/dali/_multiproc/pool.py
+++ b/dali/python/nvidia/dali/_multiproc/pool.py
@@ -126,8 +126,7 @@ received so far.
         """Check if given batch notified error and raise it"""
         if batch_i in self.iter_failed:
             exception, traceback_str = self.iter_failed[batch_i]
-            del self.iter_failed[batch_i]
-            del self.partially_received[batch_i]
+            self.clear_scheduled(batch_i)
             if isinstance(exception, StopIteration):
                 raise exception
             else:
@@ -149,11 +148,16 @@ received so far.
         """Check if we didn't receive all results for given tasks and batch_i"""
         return len(self.partially_received[batch_i]) < len(tasks)
 
+    def clear_scheduled(self, batch_i):
+        del self.partially_received[batch_i]
+        if batch_i in self.iter_failed:
+            del self.iter_failed[batch_i]
+
     def get_batch(self, batch_i, tasks):
         """Return the full batch, mark it as cleared and consumed"""
         full_batch = self.partially_received[batch_i]
         res = [full_batch[i] for i, _ in tasks]
-        del full_batch
+        self.clear_scheduled(batch_i)
         return res
 
     def receive_chunk(self, batch_i, sock, serialized_batch):
--- a/dali/test/python/test_external_source_parallel.py
+++ b/dali/test/python/test_external_source_parallel.py
@@ -73,7 +73,6 @@ def test_wrong_source():
 # Test that we can launch several CPU-only pipelines by fork as we don't touch CUDA context.
 @with_setup(setup_function, teardown_function)
 def test_parallel_fork_cpu_only():
-    global pipe_processes
     pipeline_pairs = 4
     batch_size = 10
     iters = 40
@@ -90,7 +89,6 @@ def test_parallel_fork_cpu_only():
         capture_processes(pipe1._py_pool)
         compare_pipelines(pipe0, pipe1, batch_size, iters)
 
-@with_setup(setup_function, teardown_function)
 def test_parallel_no_workers():
     batch_size = 10
     iters = 4
@@ -98,7 +96,6 @@ def test_parallel_no_workers():
     parallel_pipe = create_pipe(callback, 'cpu', batch_size, py_num_workers=0,
                                 py_start_method='spawn', parallel=True, device_id=None)
     parallel_pipe.build()
-    capture_processes(parallel_pipe._py_pool)
     assert parallel_pipe._py_pool is None
     assert parallel_pipe._py_pool_started == False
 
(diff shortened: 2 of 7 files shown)


stiepan:
!build

dali/test/python/test_RN50_external_source_parallel_data.py
```diff
@@ -95,6 +90,9 @@ def iteration_test(args):
                 pipe.reset()
 
         print("OK {}".format(pipe_factory.__name__))
+        # pipline is not immediately released because traceback of last exception kept by
+        # python (i.e. StopIteration) stores reference to the pipeline
+        gc.collect()
```
JanuszL, line 95: So what this gc does here?  Do we have any guarantee it does something?

dali/test/python/test_external_source_parallel.py
```diff
@@ -119,63 +116,73 @@ def test_parallel_fork():
...
         yield check_callback, parallel_pipe, pipe, epoch_size, batch_size, dtype
+        # explicitely call py_pool close as nose might still reference parallel_pipe from the yield above
+        parallel_pipe._py_pool.close()
     # test that another pipline with forking initialization fails as there is CUDA contexts already initialized
     parallel_pipe = create_pipe(callback, 'cpu', 16, py_num_workers=4,
                                 py_start_method='fork', parallel=True)
     yield raises(RuntimeError, "Cannot fork a process when there is a CUDA context already bound to the process.")(
         build_and_run_pipeline), parallel_pipe, 1
 
-@with_setup(setup_function, teardown_function)
```
JanuszL, line 128: Why removing this?

dali/test/python/test_external_source_parallel.py
```diff
@@ -119,63 +116,73 @@ def test_parallel_fork():
...
 def test_num_outputs():
     yield from check_spawn_with_callback(ExtCallbackMultipleOutputs, ExtCallbackMultipleOutputs, num_outputs=2, dtypes=[np.uint8, np.float])
 
-@with_setup(setup_function, teardown_function)
 def test_tensor_cpu():
     yield from check_spawn_with_callback(ExtCallbackTensorCPU)
 
+@with_setup(setup_function, teardown_function)
```
JanuszL, line 142: How much are these changes about refactoring and how much about adding new tests for the issue this PR fixes?

dali/test/python/test_external_source_parallel_mxnet.py
```diff
@@ -30,7 +29,6 @@ def __call__(self, sample_info):
         return mx.nd.array(a, dtype=a.dtype)
 
 
-@with_setup(setup_function, teardown_function)
```
JanuszL, line 33: Why remove that?

dali/test/python/test_external_source_parallel_pytorch.py
```diff
@@ -33,7 +32,6 @@ def __call__(self, sample_info):
 @raises(RuntimeError, "Error*starting Python worker threads for*parallel External Source*"
                       "Cannot fork*CUDA context already bound*"
                       "*start_py_workers*fork*spawn*")
-@with_setup(setup_function, teardown_function)
```
JanuszL, line 36: Why remove that?

dali/test/python/test_external_source_parallel_utils.py
```diff
@@ -97,6 +100,17 @@ def check_callback(parallel_pipe, pipe, epoch_size, batch_size, dtype=None):
     parallel_pipe._py_pool.close()
 
 
+@with_setup(setup_function, teardown_function)
```
JanuszL, line 103: Fix or just refactoring?

dali/python/nvidia/dali/_multiproc/pool.py
```diff
@@ -149,11 +148,16 @@ def is_not_received(self, batch_i, tasks):
         """Check if we didn't receive all results for given tasks and batch_i"""
         return len(self.partially_received[batch_i]) < len(tasks)
 
+    def clear_scheduled(self, batch_i):
```
JanuszL, line 151: This is the fix?

dali/test/python/test_pool.py
```diff
@@ -73,93 +78,65 @@ def wrapper():
...
-def test_pool_no_overwrite_batch(start_method):
+def test_pool_scheduled_tasks_cleanup(start_method):
     callbacks = [simple_callback]
-    for depth in [1, 2, 4, 8]:
-        pool = create_pool(callbacks, queue_depth=depth, num_workers=1, start_method=start_method)
+    with create_pool(callbacks, queue_depth=3, num_workers=1, start_method=start_method) as pool:
         pids = get_pids(pool)
         pid = pids[0]
-        tasks_list = [(i, [(SampleInfo(i, 0, i),)]) for i in range(depth)]
+        tasks_list = [(i, [(SampleInfo(i, 0, i),)]) for i in range(3)]
```
JanuszL, line 110: Why `range(3)`? How that relates to `queue_depth=3`?

dali/test/python/test_pool.py
```diff
@@ -73,93 +78,65 @@ def wrapper():
...
+            assert len(pool.contexts[0].partially_received) == 2 - i
         tasks_batches = zip(tasks_list, batches)
         for (i, tasks), batch in tasks_batches:
             for task, sample in zip(tasks, batch):
                 np.testing.assert_array_equal(answer(pid, *task), sample)
-        pool.close()
+
+
+# Test that we can hold as many results as the queue depth
+@check_pool
```
JanuszL, line 125: Is this a test for the fix?

dali/python/nvidia/dali/_multiproc/pool.py
JanuszL: I'm afraid that there is a lot of unrelated to this fix changes in this PR. How about extracting only the fix and test for it to a small PR and move the rest of the refactor to a separate one.

dali/test/python/test_pool.py
stiepan: Yes, it's the number of batches that can fit into shm circular buffer for data from workers.

dali/test/python/test_external_source_parallel.py
stiepan: Nose test cases generated with yield are not always immediately disposed which once again causes a problem that if yield references the pipeline, the pipeline is not garbage collected immediately and check for pool begin finalized fails. Faulty check in teardown_function didn't pick that up previously.

dali/test/python/test_external_source_parallel_pytorch.py
stiepan: This test doesn't create the pool at all as it fails on check for CUDA context, so it was unnecessary. On the other hand I wanted to make sure that tests with this decorator actually call capture_processes inside, so the teardown_function can make the check. In this case that assertion would error out.

dali/test/python/test_RN50_external_source_parallel_data.py
stiepan: Okay, turns out that the traceback outliving the scope can be avoided and was due to circular reference to traceback created when raising the exception from the worker. I fixed that too and this is no longer needed.

dali/python/nvidia/dali/_multiproc/pool.py
stiepan: I removed the unrelated test case from the test_pool. There's one more unrelated change - I dropped two test cases that were checking that the buffer is overwritten if we schedule too much, because I believe it doesn't test anything in fact. Should I bring those two back?

dali/test/python/test_external_source_parallel_utils.py
JanuszL: Understood.

dali/test/python/test_external_source_parallel_pytorch.py
JanuszL: Understood.

dali/test/python/test_pool.py
JanuszL: Maybe you can bring back a variable that sets this? Now this 3 looks quite magic and I'm not sure where the `3` value is on purpose and where just like that.

stiepan: That's good point, I added it back (and actually testing on a few different depths).

stiepan:
!build

dali/python/nvidia/dali/_multiproc/pool.py
```diff
@@ -123,18 +123,26 @@ def pop_scheduled(self):
...
+                    # message, originating from original exception
+                    raise Exception(
+                        "\n\nException traceback received from worker thread:\n\n" + traceback_str) from exception
+        finally:
+            # Fix circular reference problem on StopIteration - the exception contains reference to the
+            # traceback that referrs a frame that contains local variables and among them the exception.
+            # This traceback is then chained into excpetions reraised along the way
+            # (eventually at the pipeline level) which in effect introduces a reference to the pipline
+            # that would be only removed after garbage collection round, delaying finalization of the pool
+            exception = None
```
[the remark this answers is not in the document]
JanuszL: Let us ignore it.

stiepan:
!build

PR #3471, opened 2021-11-04
Update tensorflow version in conda build

szalpal:

## Description

### What happened in this PR

Due to recent dependency update, conda build started to fail. The reason was that the `protobuf` version that exists after the update is incompatible with TF version, that was used in the conda build. The solution is to update the TF version in conda. I didn't update the TF build to the newest (which is `2.6.0` in conda) to follow as-old-as-possible rule. `tensorboard` package has been updated to `2.5.0`, since `2.4.1` is not available in conda.

#### Additional information

### Tests
- [x] Existing tests apply
- [ ] New tests added
  - [ ] Python tests
  - [ ] GTests
  - [ ] Benchmark
  - [ ] Other
- [ ] N/A

### Documentation
- [ ] Documentation updated
  - [ ] Docstring
  - [ ] Doxygen
  - [ ] RST
  - [ ] Jupyter
  - [ ] Other
- [x] N/A

#### Requirements
- [x] N/A

**REQ IDs**: N/A

**JIRA TASK**: N/A

### DIFF
--- a/conda/recipe/meta.yaml
+++ b/conda/recipe/meta.yaml
@@ -70,9 +70,9 @@ requirements:
     - protobuf >=3.19.0
     - libprotobuf-static >=3.19.0
     - libjpeg-turbo >=2.1.0
-    - tensorflow-gpu =2.2.0
-    - tensorflow-estimator =2.2.0
-    - tensorboard =2.2.2
+    - tensorflow-gpu =2.4.1
+    - tensorflow-estimator =2.4.1
+    - tensorboard =2.5.0
     - dali-opencv
     - dali-ffmpeg
     - boost >=1.76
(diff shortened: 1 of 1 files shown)


JanuszL (approved):

JanuszL (changes requested):
It fails for python 3.6 in conda.

JanuszL:
Even leaving the TF version blank fails the job if the protobuf is suppose to be at least 3.19. Maybe we should revert protobuf changes (conda will pick up the latest compatible anyway).
I would leave a comment why we are doing so.

mzient (approved):

JanuszL (approved):

PR #3487, opened 2021-11-11
Fix DALI paddle plugin stream synchronization error

sneaxiy:
## Description

### What happened in this PR

As far as I know:

- For DALI: if the given stream is `None` when calling the `dali_tensor.copy_to_external()` method, DALI would use an internal stream (non-default stream) to copy the data. Let's call this DALI internal stream to be `dali_stream`.
- For PaddlePaddle: PaddlePaddle uses its own internal stream to launch the CUDA kernel. Let's call this PaddlePaddle internal stream to be `paddle_stream`.

The code of `dali_tensor.copy_to_external(stream=None, non_blocking=False)` is something like this:
```
if stream is None:
    stream = dali_stream

cudaMemcpyAsync(external_tensor_ptr, dali_tensor_ptr, dali_stream)
if not non_blocking:
    cudaStreamSynchronize(dali_stream)
```

When calling `cudaMemcpyAsync`, the `external_tensor_ptr` may be used in the launched CUDA kernel inside the PaddlePaddle framework. Therefore, the `external_tensor_ptr` may be read using `paddle_stream` and be written using `dali_stream` in the same time. This would cause incorrect results. We found this bug when running ResNet50 model of PaddlePaddle.

This PR makes the `cudaMemcpyAsync` use the `paddle_stream` instead of `dali_stream` to make sure the proper synchronization.

#### Additional information
- Affected modules and functionalities: DALI PaddlePaddle plugin

- Key points relevant for the review: CUDA stream synchronization between the DALI stream and PaddlePaddle stream

### Tests
- [x] Existing tests apply
- [ ] New tests added
  - [ ] Python tests
  - [ ] GTests
  - [ ] Benchmark
  - [ ] Other
- [ ] N/A

### Documentation
- [ ] Documentation updated
  - [ ] Docstring
  - [ ] Doxygen
  - [ ] RST
  - [ ] Jupyter
  - [ ] Other
- [ ] N/A

#### Requirements
- [ ] N/A

**REQ IDs**: N/A

**JIRA TASK**: N/A

### DIFF
--- a/dali/python/nvidia/dali/plugin/paddle.py
+++ b/dali/python/nvidia/dali/plugin/paddle.py
@@ -318,12 +318,6 @@ class DALIGenericIterator(_DaliBaseIterator):
                                          category_pd_type[cat])
             data_batches[i] = pd_tensors
 
-            # due to https://github.com/PaddlePaddle/Paddle/issues/35555 the tensors are allocated
-            # first by setting shape and calling _mutable_data, then the device is synchronized
-            # and after that, another call to _mutable_data is made to obtain the pointer and
-            # copy data from the pipeline to the tensor
-            # when the issue is solved the copy can be moved to the preceding loop
-            fluid.core._cuda_synchronize(pd_gpu_place)
             for cat, tensor in category_tensors.items():
                 ptr = pd_tensors[cat]._mutable_data(category_place[cat],
                                                category_pd_type[cat])
--- a/dali/python/nvidia/dali/plugin/paddle.py
+++ b/dali/python/nvidia/dali/plugin/paddle.py
@@ -17,6 +17,7 @@ import ctypes
 import math
 
 import numpy as np
+import paddle
 
 from nvidia.dali import types
 from nvidia.dali.backend import TensorListCPU, TensorGPU, TensorListGPU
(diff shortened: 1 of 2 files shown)


dali/python/nvidia/dali/plugin/paddle.py
```diff
@@ -72,6 +73,8 @@ def feed_ndarray(dali_tensor, ptr, cuda_stream=None):
                     (if not provided, an internal user stream will be selected)
     """
 
+    if cuda_stream is None:
```
JanuszL, line 76: How about moving this to L326 (like in the case of pytorch)?
Please also remove `fluid.core._cuda_synchronize(pd_gpu_place)` which was supposed to be a workaround for the mentioned problem.
Side question - which paddle paddle version is needed to support this maybe there should an if for the version?
And you can update qa/setup_packages.py to make DALI tests use the more recent paddlepaddle-gpu version.

JanuszL:
Hi @sneaxiy,

Thank you for the fix. Would you mind checking my comment?

dali/python/nvidia/dali/plugin/paddle.py
sneaxiy: Done. Thanks for your suggestions!

JanuszL:
!build

JanuszL:
Hi @sneaxiy,

Thank you for the update. Let me see if it passes our CI.

JanuszL (approved):

jantonguirao (approved):

JanuszL:
@sneaxiy - thank you again for your contribution.

PR #3497, opened 2021-11-16
Remove synchronization from deallocate.

mzient:

## Description

### What happened in this PR
Adjusts semantics of deallocation so that `deallocate` doesn't wait for the device to finish scheduled work.

We can do it now, because most of the time, the deallocation of GPU memory is done in stream order. Pinned memory is another matter, but appropriate synchronization is already in place to guarantee safety against overwriting the memory before reuse, so resizing was already covered - regarding deallocation at the object's end of life, it is now done in stream order. The remaining few cases were fixed in this PR.

* Every deallocation of device-accessible memory

##### Key points relevant for the review:
* More like a thought experiment: where else can something break?

### Tests
- [X] Existing tests apply
- [ ] New tests added
  - [ ] Python tests
  - [ ] GTests
  - [ ] Benchmark
  - [ ] Other
- [ ] N/A

### Documentation
- [ ] Documentation updated
  - [ ] Docstring
  - [ ] Doxygen
  - [ ] RST
  - [ ] Jupyter
  - [ ] Other
- [X] N/A

#### Requirements
- [X] N/A

**REQ IDs**: N/A

**JIRA TASK**: DALI-2459


jantonguirao (approved):

dali/pipeline/data/tensor.h
```diff
@@ -45,6 +45,8 @@ class Tensor : public Buffer<Backend> {
   inline Tensor() {}
   inline ~Tensor() override = default;
 
+  using Buffer<Backend>::is_pinned;
+  using Buffer<Backend>::has_data;
```
JanuszL, line 49: Do we need that here?

dali/pipeline/operator/builtin/external_source_test.cc
```diff
@@ -147,7 +148,9 @@ class ExternalSourceTest : public::testing::WithParamInterface<int>,
       vt_gpu_[j].Copy(tensor);
     }
     CUDA_CALL(cudaStreamSynchronize(0));
-    src_op->SetDataSource(vt_gpu_);
+    ExtSrcSettingMode settings = {};
+    settings.sync = true;
+    src_op->SetDataSource(vt_gpu_, {}, settings);
```
JanuszL, line 153: Does it mean that `SetDataSource` behaves differently now?

dali/operators/math/expressions/constant_storage.h
```diff
@@ -81,18 +86,20 @@ class ConstantStorage {
...
+  Tensor<CPUBackend> result_cpu_;
 
   template <typename T>
   void Rewrite(Tensor<GPUBackend> &result, const std::vector<T> &constants,
-               const std::vector<ExprConstant *> &constant_nodes, cudaStream_t stream) {
-    Tensor<CPUBackend> result_cpu;
-    Rewrite(result_cpu, constants, constant_nodes);
-    result.Copy(result_cpu, stream);
+               const std::vector<ExprConstant *> &constant_nodes, AccessOrder order) {
+    Rewrite(result_cpu_, constants, constant_nodes, order);
```
JanuszL, line 94: I think you should set it pinned as we someday change the default behavior.

dali/operators/image/resize/resize_op_impl_gpu.h
```diff
@@ -79,6 +79,7 @@ class ResizeOpImplGPU : public ResizeBase<GPUBackend>::Impl {
       kernels::KernelRequirements &req = kmgr_.Setup<Kernel>(mb_idx, ctx, mb.input, param_slice);
       mb.out_shape = req.output_shapes[0].to_static<frame_ndim>();
     }
+    kmgr_.ReserveMaxScratchpad(0);
```
JanuszL, line 82: Why it is needed here?

dali/core/mm/async_pool_test.cu
```diff
@@ -190,6 +190,7 @@ void AsyncPoolTest(Pool &pool, vector<block> &blocks, Mutex &mtx, CUDAStream &st
       if (stream) {
         pool.deallocate_async(blk.ptr, blk.size, sv);
       } else {
+        CUDA_CALL(cudaStreamSynchronize(stream));
```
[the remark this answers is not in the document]
mzient: I don't understand the question.

dali/operators/image/resize/resize_op_impl_gpu.h
mzient: To avoid reallocation of the scratchpad between minibatches. Since now the reallocation is not syncrhonous (and scratchpad is not stream ordered yet), it caused hard errors. This guarantees that the scratchpad admits the largest minibatch. Actually, this was meant to work like this from the start - I think this line was even there at some point, just dropped during some refactoring.

dali/core/mm/async_pool_test.cu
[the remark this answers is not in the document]
JanuszL: So if `stream==0` means `no stream` then which stream is sued for this sync?

dali/operators/math/expressions/constant_storage.h
mzient: Perhaps, but I think it's a _tiny_ bit out of scope here.

dali/pipeline/data/tensor.h
mzient: I guess not.

dali/pipeline/operator/builtin/external_source_test.cc
mzient: Good question. Need to recheck.

dali/operators/math/expressions/constant_storage.h
JanuszL: True. Up to you.

dali/operators/image/resize/resize_op_impl_gpu.h
JanuszL: Understoood.

dali/pipeline/data/tensor.h
JanuszL: Up to you.

dali/pipeline/operator/builtin/external_source_test.cc
mzient: OK - if I set the access order as stream 0 for the TL, it works. Note that GPU TLs produced by DALI will have a stream set. I don't know how we create TLs for the GPU external source, but nothing seemed to be broken...

dali/core/mm/async_pool_test.cu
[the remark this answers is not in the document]
mzient: Probably 0... which is weird.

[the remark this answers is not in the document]
mzient: OK, got it - when stream==0, the allocations are synchronous, but the memset happens on stream 0 - that's why we synchronize with it. It doesn't read well, but it's correct.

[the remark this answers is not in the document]
JanuszL: It doesn't make much sense to me...

dali/pipeline/operator/builtin/external_source_test.cc
JanuszL: We can use `__cuda_array_interface__` or dlpack for that https://github.com/NVIDIA/DALI/blob/main/dali/python/backend_impl.cc#L912
Btw - I see that v3 adds stream. Maybe we should add it support as well https://numba.readthedocs.io/en/stable/cuda/cuda_array_interface.html?

dali/core/mm/async_pool_test.cu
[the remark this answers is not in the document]
JanuszL: So I would:
```suggestion
        // if stream evaluates to false, allocationsare synchronous, but the memset happens on stream 0, sync with it
        CUDA_CALL(cudaStreamSynchronize(0));
```

dali/pipeline/operator/builtin/external_source_test.cc
mzient: Not this PR, maybe? It seems to work the way it is.

JanuszL: Ok

JanuszL (approved):

PR #3536, opened 2021-11-25
Make data objects stream-aware

mzient:
## Description

### What happened in this PR
* Added a new type, AccessOrder which denotes a CUDA stream or host order
* Added order info to Buffer
  * Buffer is allocated/deallocated according to the specified order
  * Some operations on a buffer synchronize between different orders
  * Deallocation happens on a stream
  * If the shared pointer has a stream-aware deleter, then appropriate synchronization is added between the deletion stream and the stream associated with the buffer object

Buffer
Tensor
TensorList
TensorVector
...and some of the places where they're used

#### Key points relevant for the review:
??

### Tests
- [ ] Existing tests apply
- [ ] New tests added
  - [ ] Python tests
  - [ ] GTests
  - [ ] Benchmark
  - [ ] Other
- [ ] N/A
TODO

### Documentation
- [ ] Documentation updated
  - [X] Docstring
  - [ ] Doxygen
  - [ ] RST
  - [ ] Jupyter
  - [ ] Other
- [ ] N/A

#### Requirements
- [X] N/A

**REQ IDs**: N/A

**JIRA TASK**: DALI-2097


mzient:
!build

dali/c_api/c_api_test.cc
```diff
@@ -251,6 +250,7 @@ TYPED_TEST(CApiTest, FileReaderPipe) {
   pipe_ptr->RunGPU();
 
   ComparePipelinesOutputs<TypeParam>(handle, *pipe_ptr);
+  ws.Clear();
```
JanuszL, line 253: Why we need this call here?

dali/c_api/c_api.cc
```diff
@@ -546,7 +546,8 @@ void daliDeletePipeline(daliPipelineHandle* pipe_handle) {
   auto *bs_map = reinterpret_cast<batch_size_map_t *>(pipe_handle->batch_size_map);
   DALI_ENFORCE(pipeline != nullptr && ws != nullptr, "Pipeline already deleted");
   if (pipe_handle->copy_stream) {
-    CUDA_CALL(cudaStreamDestroy(pipe_handle->copy_stream));
+    CUDA_CALL(cudaStreamSynchronize(pipe_handle->copy_stream));
```
JanuszL, line 549: Why you need to synchronize before putting it back to the pool?

dali/core/mm/default_resources.cc
```diff
@@ -262,10 +263,10 @@ const std::shared_ptr<pinned_async_resource> &ShareDefaultResourceImpl<memory_ki
     std::lock_guard<std::mutex> lock(g_resources.mtx);
     if (!g_resources.pinned_async) {
       static CUDARTLoader init_cuda;  // force initialization of CUDA before creating the resource
+      g_resources.pinned_async = CreateDefaultPinnedResource();
```
JanuszL, line 266: What difference does this change make?

dali/pipeline/data/buffer.h
```diff
@@ -289,34 +339,65 @@ class DLL_PUBLIC Buffer {
...
+  inline void reserve(size_t new_num_bytes, AccessOrder order = {}) {
+    if (new_num_bytes <= num_bytes_)
+      return;
 
     DALI_ENFORCE(!shares_data_,
                  "Cannot reallocate Buffer if it is sharing data. "
                  "Clear the status by `Reset()` first.");
-    data_.reset();
+
+    if (data_ && device_ >= 0 && order.is_device() && order.device_id() != device_)
```
JanuszL, line 350: Is it possible to set device order for device < 0?

dali/pipeline/data/buffer.h
```diff
@@ -289,34 +339,65 @@ class DLL_PUBLIC Buffer {
...
+      DALI_FAIL("Cannot reallocate a buffer on a different device!");
+
+    set_order(order);
+    free_storage();
+
+    if (device_ < 0) {
+      if (order.is_device() && order.device_id() >= 0) {
+        device_ = order.device_id();
```
JanuszL, line 358: Doesn't this make the CPU buffer a GPU one?

dali/pipeline/data/buffer.h
```diff
@@ -289,34 +339,65 @@ class DLL_PUBLIC Buffer {
...
+                      : AllocBuffer<Backend>(new_num_bytes, pinned_, device_, order_);
 
     num_bytes_ = new_num_bytes;
   }
 
-  void reset() {
+  /**
+   * @brief Deallocates the data and clears the type.
+   *
+   * The data, if any, is deallocated. The function will
```
JanuszL, line 380: will...?

dali/pipeline/data/buffer.h
```diff
@@ -433,26 +510,31 @@ class DLL_PUBLIC Buffer {
...
-
+    swap(buffer);
     buffer.reset();
   }
 
+  void free_storage(AccessOrder order = {}) {
+    if (data_) {
+      if (!order)
+        order = order_;
+      if (!set_deletion_order(data_, order))  // TODO(michalz): Check if it helps or hurts.
```
JanuszL, line 528: When and how it is going to be determined?

dali/pipeline/data/buffer.h
```diff
@@ -464,8 +546,9 @@ class DLL_PUBLIC Buffer {
   Index size_ = 0;                   // The number of elements in the buffer
   size_t num_bytes_ = 0;             // To keep track of the true size of the underlying allocation
   int device_ = CPU_ONLY_DEVICE_ID;  // device the buffer was allocated on
+  AccessOrder order_;                // The order of memory access (host or device)
   bool shares_data_ = false;         // Whether we aren't using our own allocation
-  bool pinned_ = !RestrictPinnedMemUsage();  // Whether the allocation uses pinned memory
+  bool pinned_ = false;              // Whether the allocation uses pinned memory
```
JanuszL, line 551: Maybe we should do this in a separate PR?

dali/pipeline/data/buffer.cc
```diff
@@ -18,22 +18,48 @@
...
+    return AccessOrder(del->release_on_stream);
+  else
+    return {};
+}
+
+DLL_PUBLIC bool set_deletion_order(const std::shared_ptr<void> &ptr, AccessOrder order) {
+  if (ptr.use_count() == 1 && order.has_value()) {
+    if (auto *del = std::get_deleter<mm::AsyncDeleter>(ptr)) {
+      del->release_on_stream = order.get();
+      if (ptr.use_count() != 1)
```
JanuszL, line 32: What if it changed to 2 and then returned to 1 between L31 and L32?

dali/pipeline/data/buffer.h
```diff
@@ -289,34 +339,65 @@ class DLL_PUBLIC Buffer {
...
     num_bytes_ = new_num_bytes;
   }
 
-  void reset() {
+  /**
+   * @brief Deallocates the data and clears the type.
+   *
+   * The data, if any, is deallocated. The function will
+   */
+  void reset(AccessOrder order = {}) {
```
JanuszL, line 382: Shouldn't it use `order_`?

dali/pipeline/data/tensor.h
```diff
@@ -284,12 +297,13 @@ class Tensor : public Buffer<Backend> {
...
-    ShareData(ptr, bytes, pinned, { 0 }, type);
+                        DALIDataType type = DALI_NO_TYPE,
+                        AccessOrder order = {}) {
+    ShareData(ptr, bytes, pinned, { 0 }, type, order);
   }
 
-  inline void Reset() {
-    reset();  // free the underlying buffer
+  inline void Reset(AccessOrder stream = {}) {
+    reset(stream);  // free the underlying buffer
```
JanuszL, line 306: Why it doesn't use `order_`?

dali/pipeline/data/tensor.h
```diff
@@ -181,61 +199,57 @@ class Tensor : public Buffer<Backend> {
...
    *
-   * After wrapping the allocation, the Tensors size is set to dot product
-   * of shape vector, and its type is reset to NoType.
    * After calling this function any following call to `set_type` and `Resize`
-   * must match the total size of underlying allocation (`num_bytes_`) of
+   * must note exceed the total size of underlying allocation (`num_bytes_`) of
    * shared data or the call will fail.
    * Size can be set to 0 and type to NoType as intermediate step.
-   *
-   * The Tensor object assumes no ownership of the input allocation, and will
```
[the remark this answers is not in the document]
JanuszL: Ok, it accepts `shared_ptr`, not the raw one like in L272.

dali/core/cuda_stream_pool.cc
```diff
@@ -22,12 +22,13 @@ namespace dali {
...
 CUDAStreamPool::~CUDAStreamPool() {
   Purge();
-  assert(lease_count_ == 0);
 }
 
 CUDAStreamPool::CUDAStreamPool() {
   int num_devices = 0;
-  CUDA_CALL(cudaGetDeviceCount(&num_devices));
+  auto e = cudaGetDeviceCount(&num_devices);
+  if (e != cudaSuccess && e != cudaErrorNoDevice && e != cudaErrorInsufficientDriver)
```
JanuszL, line 30: What we get for CPU only? `cudaErrorNoDevice`?

dali/core/access_order.cc
```diff
@@ -0,0 +1,53 @@
...
+  if (*this == other)
+    return;
+  // If this order is null, do nothing.
+  // If the order to synchronize after is null or host, do nothing - host order is
+  // always considered up-to-date.
+  if (!has_value() || !other.is_device())
+    return;
+  if (is_device()) {
+    auto &pool = CUDAEventPool::instance();
+    int other_dev = other.device_id();
```
JanuszL, line 30: Shouldn't we check if device ids match?

include/dali/core/access_order.h
```diff
@@ -0,0 +1,115 @@
...
+      device_id_ = DeviceFromStream(stream);
+  }
+
+  static constexpr AccessOrder host() {
+    return AccessOrder(host_sync_stream());
+  }
+
+  cudaStream_t get() const noexcept { return has_value() ? stream_ : 0; }
+
+  cudaStream_t stream() const noexcept { return is_device() ? stream_ : 0; }
```
JanuszL, line 52: I'm bit confused by two of this having similar but quite different behavior.

include/dali/core/access_order.h
```diff
@@ -0,0 +1,115 @@
...
+    return stream_ != host_sync_stream() && stream_ != null_stream();
+  }
+
+  bool has_value() const noexcept { return stream_ != null_stream(); }
+
+  explicit operator bool() const noexcept { return has_value(); }
+
+  static constexpr cudaStream_t host_sync_stream() noexcept {
+    // TODO(michalz): Unify with dali::mm::host_sync
+    const int i = 4321;
```
JanuszL, line 68: Where this magic value comes from (can you just use mm:: relevant constant here as well)?

include/dali/core/access_order.h
```diff
@@ -0,0 +1,115 @@
...
+    return static_cast<cudaStream_t>(static_cast<void *>(static_cast<char*>(nullptr) - i));
+#endif
+  }
+
+  /**
+   * @brief Waits in `this` ordering context for the work scheduled in the `other` order.
+   */
+  void join(const AccessOrder &other) const;
+
+  void wait(cudaEvent_t event) const;
```
JanuszL, line 98: Maybe you both should be called wait or join? What is the difference?

dali/pipeline/operator/builtin/make_contiguous.cc
```diff
@@ -27,8 +27,7 @@ void MakeContiguousCPU::RunImpl(HostWorkspace &ws) {
   for (int sample_id = 0; sample_id < batch_size; ++sample_id) {
     thread_pool.AddWork([sample_id, &input, &output] (int tid) {
       // HostWorkspace doesn't have any stream
```
JanuszL, line 29: I guess you can remove th above comment.

dali/python/backend_impl.cc
```diff
@@ -1551,7 +1551,7 @@ PYBIND11_MODULE(backend_impl, m) {
           py::detail::make_caster<Tensor<CPUBackend>&> conv;
           bool is_cpu_data = conv.load(static_cast<py::object>(list[0]), true);
           if (is_cpu_data) {
-            FeedPipeline<CPUBackend>(p, name, list, 0, true);
+            FeedPipeline<CPUBackend>(p, name, list, AccessOrder::host(), true);
```
JanuszL, line 1554: Don't you have to adjust `.def("SetExternalTLInput",` as well? We get the the exact stream there as an agument of `SetExternalInput`.

dali/test/python/test_video_reader_resize.py
```diff
@@ -112,8 +113,8 @@ def ground_truth_pipeline(batch_size, video_reader_params, resize_params):
     pipeline = video_reader_pipeline(batch_size, video_reader_params)
 
     def get_next_frame():
-        pipe_out = pipeline.run()
-        sequences_out = pipe_out[0].as_cpu().as_array()
+        pipe_out, = pipeline.run()
```
JanuszL, line 116: ?
```suggestion
        pipe_out = pipeline.run()
```

dali/test/python/test_video_reader_resize.py
```diff
@@ -160,6 +161,8 @@ def run_for_params(batch_size, video_reader_params, resize_params):
     compare_video_resize_pipelines(
         pipeline, gt_pipeline, batch_size, video_reader_params['sequence_length'])
 
+    gc.collect()
```
JanuszL, line 164: Can you add a comment on what it is for here?

dali/pipeline/executor/async_separated_pipelined_executor.h
```diff
@@ -44,11 +44,16 @@ class DLL_PUBLIC AsyncSeparatedPipelinedExecutor : public SeparatedPipelinedExec
...
+    Shutdown();
+  }
+
+  DLL_PUBLIC void Shutdown() override {
     ShutdownQueue();
 
     cpu_thread_.ForceStop();
     mixed_thread_.ForceStop();
     gpu_thread_.ForceStop();
+    SyncDevice();
```
JanuszL, line 56:
```suggestion
    SeparatedPipelinedExecutor::Shutdown();
```

dali/pipeline/executor/async_pipelined_executor.h
```diff
@@ -44,10 +44,15 @@ class DLL_PUBLIC AsyncPipelinedExecutor : public PipelinedExecutor {
...
   DLL_PUBLIC ~AsyncPipelinedExecutor() override {
+    Shutdown();
+  }
+
+  DLL_PUBLIC void Shutdown() override {
     ShutdownQueue();
     cpu_thread_.ForceStop();
     mixed_thread_.ForceStop();
     gpu_thread_.ForceStop();
+    SyncDevice();
```
JanuszL, line 55:
```suggestion
    PipelinedExecutor::Shutdown();
```

dali/pipeline/executor/async_separated_pipelined_executor.h
JanuszL: Mhhh, won't `~AsyncSeparatedPipelinedExecutor()` call `Executor<WorkspacePolicy, QueuePolicy>::~Executor()` and sync anyway?

dali/pipeline/executor/executor.cc
```diff
@@ -299,6 +306,24 @@ void Executor<WorkspacePolicy, QueuePolicy>::RunHelper(OpNode &op_node, Workspac
...
   SmallVector<int, 16> empty_layout_in_idxs;
 
+  cudaStream_t prev_stage_stream = ws.has_stream() && ws.stream() == gpu_op_stream_
+    ? mixed_op_stream_ : gpu_op_stream_;
+
+  for (int i = 0; i < ws.NumOutput(); i++) {
+    auto set_order = [&](auto &output) {
+      // NOTE: the stage streams are synchronized by the executor
+      bool need_sync = output.order().stream() != prev_stage_stream;
+      auto order = ws.has_stream() ? AccessOrder(ws.stream()) : AccessOrder::host();
```
JanuszL, line 316: I guess you can move this line out of the lambda and do not recompute it over and over.

dali/pipeline/executor/executor.cc
```diff
@@ -299,6 +306,24 @@ void Executor<WorkspacePolicy, QueuePolicy>::RunHelper(OpNode &op_node, Workspac
   const auto &schema = spec.GetSchema();
   SmallVector<int, 16> empty_layout_in_idxs;
 
+  cudaStream_t prev_stage_stream = ws.has_stream() && ws.stream() == gpu_op_stream_
```
JanuszL, line 309: What is the previous stream for the CPU stage?

dali/pipeline/executor/workspace_policy.h
```diff
@@ -376,6 +391,18 @@ template <>
 struct AOT_WS_Policy<SeparateQueuePolicy> {
   AOT_WS_Policy() : depths_(0) {}
 
+  ~AOT_WS_Policy() {
+    for (auto &q : cpu_workspaces_)
+      for (auto &ws : q)
+        SetOrder(ws, AccessOrder::host());
```
JanuszL, line 397: What is it for? To sync?

dali/pipeline/executor/workspace_policy.h
```diff
@@ -503,6 +530,17 @@ struct AOT_WS_Policy<UniformQueuePolicy> {
   template <OpType op_type>
   using ws_t = op_type_to_workspace_t<op_type> &;
 
+  ~AOT_WS_Policy() {
+    for (auto &wss : wss_) {
+      for (auto ws : std::get<static_cast<int>(OpType::CPU)>(wss.op_data))
+        SetOrder(ws, AccessOrder::host());
```
JanuszL, line 536: As above.

dali/c_api/c_api.cc
mzient: I don't think I really need to. I think it's a historical artifact - initially, I destroyed the stream, but then the stream-ordered deletion failed with hard errors (the stream handle was still used) - that's why I added the synchronization. Subsequently, I realized that it's not possible to use stream-aware deleters _and_ destroy streams. That's why the stream pool introduced. I changed the stream destruction to putting it back in the pool, but I haven't removed the synchronization added before.
Having said that, I don't think it's good to put a stream with pending work back in the pool - especially if the stream pool uses LIFO order.

dali/c_api/c_api_test.cc
mzient: Like above, it originated when the streams were created and destroyed, not taken from the pool.

dali/core/access_order.cc
mzient: It's possible to synchronize with streams running on different devices.

dali/core/cuda_stream_pool.cc
mzient: That depends on how it's realized. If it's a machine without a driver, we get `cudaErrorInsufficientDriver`, if a driver is there, but there are no devices (e.g. by setting `export CUDA_VISIBLE_DEVICES=""`), it's `cudaErrorNoDevice`.

dali/core/mm/default_resources.cc
mzient: It fixes static object lifetime issues. If there are any `static` objects created in `CreateDefaultPinnedResource`, they will be alive when `cleanup` object is being destroyed.
For example, if `CreateDefaultPinnedResource` calls `CUDAStreamPool::instance()`, it guarantees that the stream pool singleton will be there when the resource is destroyed. When these lines were swapped, the static variables instantiated as side effects of `CreateDefaultPinnedResource` were instantiated after (and therefore released before), `cleanup`.

dali/pipeline/data/buffer.cc
mzient: Then we're good. The program would be still "correct" (i.e. it wouldn't crash) if we set deletion order every time, but the performance would be affected. Buffer objects are not thread-safe anyway, so if we detected such a problem, then it would indicate a more serious issue.

dali/pipeline/data/buffer.h
mzient: It's for pinned buffers - they may be accessed by GPUs and are subject to stream semantics.

dali/pipeline/data/buffer.h
mzient: Hmmm.... I guess merging didn't go as planned...

dali/pipeline/data/buffer.h
mzient: If `order` is empty, it will: `free_storage` will use `order_` and `set_order` will be a no-op.

dali/pipeline/executor/workspace_policy.h
mzient: Yes; probably no longer necessary, since we use CUDAStreamPool

dali/python/backend_impl.cc
mzient: Will check.

dali/test/python/test_video_reader_resize.py
mzient: Sure.

include/dali/core/access_order.h
mzient: It's just a magic value... It is defined twice. If anything, I'd bring access_order.h to `mm/memory.h` and use AccessOrder::host_sync_stream there.

include/dali/core/access_order.h
mzient: Definitely `wait`. `join` is used for threads - that's why I used it for streams, as per analogy, but if you insist, I can rename it to `wait`.

dali/c_api/c_api.cc
JanuszL: Does it make sense to add synchronization in the `Put` function?

dali/c_api/c_api_test.cc
JanuszL: So the code that uses DALI C API needs to call `Clear` on created ws explicitly?

dali/core/access_order.cc
JanuszL: Understood.

dali/core/cuda_stream_pool.cc
JanuszL: Understood.

dali/core/mm/default_resources.cc
JanuszL: Understood.

dali/pipeline/data/buffer.cc
JanuszL: So why do we throw here if this is just a performance optimization?

dali/pipeline/data/buffer.h
JanuszL: Understood.

dali/pipeline/data/buffer.h
JanuszL: Please add this in the docstring.

dali/pipeline/executor/workspace_policy.h
JanuszL: Can you remove to make code cleaner?

include/dali/core/access_order.h
JanuszL: I was thinking about this. `mm/memory.h` sounds good.

include/dali/core/access_order.h
JanuszL: My point is that we have `wait` and `join` and I as a user of the API I don't know which one I should use and what is the difference, so maybe having just one with a name with different argument types would be better.

dali/c_api/c_api.cc
mzient: Not really. `cudaStreamDestroy` doesn't synchronize - the stream will be kept alive until it drains. I think we should switch to FIFO or do some magic with callbacks/host functions.

JanuszL: I guess this is ToDo for a next PR, right?

include/dali/core/access_order.h
mzient: The latter is mostly for interoperability with legacy functions like TypeInfo::Copy, which use a value of 0 in cases where the stream is not relevant and which cannot handle the `host_sync_stream()` value in a meaningful way.

dali/pipeline/data/buffer.h
mzient: When: when we implement some recycling within the pipeline. For now, it seems that it works fine.

dali/pipeline/data/buffer.h
mzient: I'll revert and see how it goes...

dali/pipeline/data/tensor.h
mzient: I don't think it needs it at all at this point.

include/dali/core/access_order.h
JanuszL: Can you add a comment and write that `get` should be used when possible?

dali/pipeline/data/buffer.h
JanuszL: I'm not saying it is wrong. But it is a big change and maybe it is better to add it separately.

dali/pipeline/executor/async_separated_pipelined_executor.h
mzient: It will be too late - some stuff will get destroyed which we want to be kept around.

dali/pipeline/executor/executor.cc
mzient: True.

dali/python/backend_impl.cc
mzient: It doens't seem so - there's no UseStream stuff there (though perhaps there should be...).

mzient: It doesn't seem so - the distinction between CPU and GPU input is done by type there, so there's no is_cpu_data.

dali/c_api/c_api.cc
mzient: Definitely.

dali/pipeline/data/buffer.cc
mzient: Note that this is a re-check - if someone has somehow increased the use count of this pointer, then something very wrong is happening. This _could_ be skipped, but I prefer to keep this check here just in case. After all, when we're _not_ throwing, it's just one extra if.

JanuszL: Understood.

dali/pipeline/executor/async_separated_pipelined_executor.h
JanuszL: Understood.

dali/python/backend_impl.cc
JanuszL: I was asking if the `stream` that is passed i L1530 is converted to `AccessOrder` automatically or you need to create `AccessOrder` there from it?

include/dali/core/access_order.h
```diff
@@ -0,0 +1,126 @@
...
+   */
+  cudaStream_t get() const noexcept { return has_value() ? stream_ : 0; }
+
+  /**
+   * @brief Returns stream handle, if any.
+   *
+   * Returns the underlying handle when it denotes a CUDA stream. If this order does not
+   * represent a valid CUDA stream, the return value is 0, but it should not be treated
+   * as a valid CUDA stream. This behaviour is to facilitate the interoperatibility with
+   * legacy functions that take cudaStream_t.
```
JanuszL, line 61:
```suggestion
   * legacy functions that take cudaStream_t. Use get() when possible.
```

mzient: It's not that simple :) In general, when passing a value to a function that has no notion of our special host pseudostream, it's safer to use `stream()` than `get()`. The legacy part more about functions that don't need a stream, but take one nonetheless.

dali/pipeline/executor/executor.cc
mzient: It will be gpu_op_stream_ - same as for mixed, since in pure CPU stage there's no usage of streams.

include/dali/core/access_order.h
JanuszL: Understood.

dali/python/backend_impl.cc
mzient: It is converted automatically.

JanuszL (approved):

stiepan (approved):

JanuszL (approved):

PR #3540, opened 2021-11-29
Enable tests for dynamically loaded cuda libraries

JanuszL:
- adds TL0_python-self-test-base-cuda test that test against
  CUDA 11.1 in CI
- makes DALI look for CUDA libs installed from wheel first,
  later inside /usr/local/cuda
- installs the latest CUDA wheel in tests
- adds a python warning when DALI cannot access CUDA driver, NPP, cuFFT or nvJPEG libraries

## Description

### What happened in this PR
- adds TL0_python-self-test-base-cuda test that test against
  CUDA 11.1 in CI
- makes DALI look for CUDA libs installed from wheel first,
  later inside /usr/local/cuda

#### Additional information

  - NA

### Tests
- [ ] Existing tests apply
- [x] New tests added
  - [x] Python tests
  - [ ] GTests
  - [ ] Benchmark
  - [ ] Other
- [ ] N/A

### Documentation
- [ ] Documentation updated
  - [ ] Docstring
  - [ ] Doxygen
  - [ ] RST
  - [ ] Jupyter
  - [ ] Other
- [x] N/A

#### Requirements
- [ ] N/A

**REQ IDs**: N/A

**JIRA TASK**: DALI-2470

### DIFF
--- a/dali/kernels/kernel.h
+++ b/dali/kernels/kernel.h
@@ -1,4 +1,4 @@
-// Copyright (c) 2018, NVIDIA CORPORATION. All rights reserved.
+// Copyright (c) 2018, 2021, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -26,6 +26,13 @@
 #include "dali/core/util.h"
 
 namespace dali {
+
+/**
+ * @brief  Returns version of available cuFFT library
+ *
+ * @returns MAJOR*1000 + MINOR*10 + PATH or -1 if not available
+ */
+int GetCufftVersion();
 
 /**
  * @brief Defines the DALI kernel API. See dali::kernels::examples::Kernel for details
--- a/dali/kernels/kernel.h
+++ b/dali/kernels/kernel.h
@@ -1,4 +1,4 @@
-// Copyright (c) 2018, 2021, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+// Copyright (c) 2018-2021, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
--- a/dali/kernels/signal/fft/cufft_helper.cc
+++ b/dali/kernels/signal/fft/cufft_helper.cc
@@ -13,6 +13,7 @@
 // limitations under the License.
 
 #include "dali/kernels/signal/fft/cufft_helper.h"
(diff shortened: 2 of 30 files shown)


dali/kernels/kernel.h
```diff
@@ -1,4 +1,4 @@
-// Copyright (c) 2018, NVIDIA CORPORATION. All rights reserved.
+// Copyright (c) 2018, 2021, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
awolant, line 1: It is just a question, but is there a particular reason to use `2018, 2021` instead of `2018-2021`? For me it is the some, but maybe there is some meaning behind this?

dali/kernels/signal/fft/cufft_helper.cc
```diff
@@ -0,0 +1,38 @@
...
+  }
+  ret = cufftGetProperty(MAJOR_VERSION, &minor);
+  if (ret != CUFFT_SUCCESS) {
+    return -1;
+  }
+  ret = cufftGetProperty(PATCH_LEVEL, &patch);
+  if (ret != CUFFT_SUCCESS) {
+    return -1;
+  }
+  return major*1000 + minor*10 + patch;
```
awolant, line 35: I've seen this bit of magic in few places in this PR. Maybe it would be good to extract in and call `GetVersionNumber` or whatever similar, so it is clear, why this happens like this?

For me it's ok how it is now, but I know how these versions are numbered.

awolant (approved):

dali/kernels/signal/fft/cufft_helper.cc
JanuszL: Done for cuFFT and nvJPEG. NPP follows different conventions to obtain major, minor, and patch values, still the function that combines them to one value is now common.

dali/python/backend_impl.cc
```diff
@@ -1158,6 +1159,43 @@ PYBIND11_MODULE(backend_impl, m) {
...
+      return version;
+    }
+  });
+
+  m.def("GetCufftVersion", [] {
+    int ret = -1;
+    try {
+      // we don't want to throw when it is not available, just return -1
+      ret = GetCufftVersion();
+    } catch (std::runtime_error) {}
```
szalpal, line 1177:
```suggestion
    } catch (const std::runtime_error&) {}
```
Shouldn't exceptions be caught be `const&`? I believe having them by value here creates additional copy, but I might be wrong.

szalpal (approved):

dali/python/backend_impl.cc
JanuszL: Good catch. Fixed.

PR #3572, opened 2021-12-15
GetProperty operator

szalpal:

## Description

### What happened in this PR

The PR add a new operator (CPU and GPU backend): `GetProperty`. This operator takes the existing tensor property (at the moment: source info and layout) and returns it in the Python API.

As for the implementation, it is pretty regular operator. Since I could assume that, once created, a particular operator will always return a particular property, I could abstract some of the required behaviors to the `Property` class. This makes adding new properties simple.

Additional changes:
- fixing typos

#### Additional information
`dali/operators`

nothing extraordinary

### Tests
- [ ] Existing tests apply
- [ ] New tests added
  - [x] Python tests
  - [ ] GTests
  - [ ] Benchmark
  - [ ] Other
- [ ] N/A

### Documentation
- [ ] Documentation updated
  - [x] Docstring
  - [ ] Doxygen
  - [ ] RST
  - [ ] Jupyter
  - [ ] Other
- [ ] N/A

#### Requirements
- [ ] N/A

**REQ IDs**: PROP.01-07

**JIRA TASK**: 
DALI-2499


szalpal:
!build

dali/operators/util/get_property.cc
```diff
@@ -0,0 +1,83 @@
...
+    auto& output = ws.template Output<Backend>(0);
+    for (int i = 0; i < input.num_samples(); i++) {
+      auto output_ptr = output[i].raw_mutable_data();
+      auto si = input[i].GetMeta().GetSourceInfo();
+      Memcpy(output_ptr, si.c_str(), si.length());
+    }
+  }
+
+ private:
+  void Memcpy(void* dst, const void* src, size_t len, cudaStream_t stream = nullptr) {
```
JanuszL, line 57: Why not to use `Copy` from types.h?

dali/operators/util/get_property.cc
```diff
@@ -0,0 +1,83 @@
...
+DALI_REGISTER_OPERATOR(GetProperty, GetProperty<CPUBackend>, CPU);
+
+namespace detail {
+
+template <typename Backend, typename BatchContainer = batch_container_t<Backend>>
+struct SourceInfo : public Property<Backend> {
+  TensorListShape<> GetShape(const BatchContainer& input) override {
+    TensorListShape<> ret{input.num_samples(), 1};
+    int max_length = 0;
+    for (int i = 0; i < ret.size(); i++) {
```
JanuszL, line 35: Do it need to be uniform? Maybe it can be optional?

dali/test/python/test_operator_get_property.py
```diff
@@ -0,0 +1,36 @@
...
+def pipe():
+    read, _ = fn.readers.file(file_root=images_dir)
+    return fn.get_property(read, key="source_info")
+
+def main():
+    p=pipe()
+    p.build()
+    out=p.run()
+    nda=out[0].as_array()
+    print(nda.tostring())
```
JanuszL, line 33: I'm not sure if this should be the way. Maybe rather:
```
for t in out[0]:
    print(np.array(t).tostring())
``?

Comment (szalpal) [reply dali/test/python/test_operator_get_property.py:33]: That's a draft test ;) I just wanted to check if the CI passes. This whole file is going to change

**Comment by szalpal** [comment]
!build

File: dali/test/python/test_operator_get_property.py (hunk 3 above)
Comment (JanuszL) [reply dali/test/python/test_operator_get_property.py:33]: I'm not commenting the test but the API and the recommended way to use it.
If we want the tensors returned by this OP uniform and the whole batch easily convertible to tensor? Or we think that the user will obtain the values per sample so `as_array()` won't be called.

File: dali/operators/util/get_property.cc [hunk 4]
```diff
@@ -0,0 +1,90 @@
...
+
+#include "dali/operators/util/get_property.h"
+#include "dali/pipeline/data/types.h"
+
+namespace dali {
+
+DALI_SCHEMA(GetProperty)
+    .DocStr(
+        R"code(Returns a property of the tensor passed as an input.
+The type of the output will depend on the ``key`` of the requested property.)code")
```
jantonguirao, line 23:
```suggestion
        R"code(Returns a data node property.

The output data type will depend on the property extracted. Supported properties are:
* ``"source_info"``: String-like (1D tensor of uint8) representing the source of the data node. Typically a file path associated with the data or other kind of identifier.
* ``"layout"``: String-like (1D tensor of uint8) representing the data layout. Each character names a dimension in the data shape. It can be used to understand the data layout (e.g. channel-first vs. channel-last).
)code")
```

jantonguirao: Note that the empty line between the first line and the rest is important for the operator table in the docs.

dali/operators/util/get_property.cc
```diff
@@ -0,0 +1,90 @@
...
+    .NumInput(1)
+    .NumOutput(1)
+    .AddArg("key",
+            R"code(Specifies, which property is requested. The following properties are supported:
```
jantonguirao, line 27:
```suggestion
            R"code(Specifies which property is requested. The following propertie keys are supported:
```

dali/operators/util/get_property.cc
```diff
@@ -0,0 +1,90 @@
...
+* ``"source_info"``:
+* ``"layout"``:
+)code",
+            DALI_STRING);
+
+
+DALI_REGISTER_OPERATOR(GetProperty, GetProperty<CPUBackend>, CPU);
+
+namespace detail {
+
```
jantonguirao, line 38: idea: I suggest you add a `StringLikeProperty` base class, that you could reuse for SourceInfo and Layout

dali/operators/util/get_property.cc
```diff
@@ -0,0 +1,158 @@
...
+  void FillOutput(workspace_t<Backend>& ws) override {
+    const auto& input = ws.template Input<Backend>(0);
+    auto& output = ws.template Output<Backend>(0);
+    for (size_t sample_id = 0; sample_id < input.num_samples(); sample_id++) {
+      auto si = GetSourceInfo(input, sample_id);
+      output[sample_id].Copy(span{si.c_str(), static_cast<int>(si.length())}, nullptr);
+    }
+  }
+
+  std::string GetSourceInfo(const BatchContainer& input, size_t idx) {
```
mzient, line 74:
```suggestion
  static std::string GetSourceInfo(const BatchContainer& input, size_t idx) {
```
Also, I think that GetSourceInfo() should be modified to returned `const std::string &` - this copy is quite pointless. I guess this is the right moment to make this change.

dali/operators/util/get_property.cc
szalpal: Done, it's not uniform anymore

szalpal:
!build

dali/operators/util/get_property.cc
szalpal: Done

dali/test/python/test_operator_get_property.py
szalpal: Done, I went with your suggestion

szalpal:
!build

dali/operators/util/get_property.cc
szalpal: I though about it, however I believe there wouldn't be too big of a gain with such abstraction:
1. `GetShape` functions are different
2. `GetType` is the same
3. `FillOutput` is partially the same, however I believe it would need so much abstraction magic (because `GetLayout` doesn't return `std::string`, but a `TensorLayout`), that the duplicated code is more readable. On top of that, this duplication is our beloved operator bolier-plate, so it's really almost an idiom in out case ;)

dali/operators/util/get_property.cc
```diff
@@ -0,0 +1,162 @@
...
+// limitations under the License.
+
+#include "dali/operators/util/get_property.h"
+#include "dali/pipeline/data/types.h"
+
+namespace dali {
+
+DALI_SCHEMA(GetProperty)
+    .DocStr(
+        R"code(Returns a property of the tensor passed as an input.
```
jantonguirao, line 22:
```suggestion
        R"code(Returns a property of the tensor passed as an input.

```
^^ The first line should be one simple sentence (the only visible in the operator table)

dali/operators/util/get_property.cc
```diff
@@ -0,0 +1,162 @@
...
+The type of the output will depend on the ``key`` of the requested property.
+
+Some of the properties are returned as a byte-array. The easiest way to convert them to string
+would be to )code")
```
jantonguirao, line 26: missing end of sentence?

dali/operators/util/get_property.cc
```diff
@@ -0,0 +1,162 @@
...
+    .AddArg("key",
+            R"code(Specifies, which property is requested. The following properties are supported:
+
+* ``"source_info"``: Returned type: byte-array.
+                     Information about the origin of the sample. The actual origin may differ according
+                     to the source of the data. For example, when the Tensor is read via :meth:`fn.readers.file`,
+                     the ``source_info`` property will contain full path of that file. When the Tensor
+                     is read by :meth:`fn.readers.webdataset`, the property will contain full path of the tar
+                     archive and the index of the component in that archive. Lastly, when the Tensor
+                     is loaded via :meth:`fn.external_source`, the ``source_info`` will be empty.
```
jantonguirao, line 38: If you meant to cover all cases we should add also TFRecord. You can group like "for readers dealing with containers (webdataset reading tar files, tf_record_reader reading *.tfrecord files, etc) ..."

dali/operators/util/get_property.cc
```diff
@@ -0,0 +1,162 @@
...
+#include "dali/pipeline/data/types.h"
+
+namespace dali {
+
+DALI_SCHEMA(GetProperty)
+    .DocStr(
+        R"code(Returns a property of the tensor passed as an input.
+The type of the output will depend on the ``key`` of the requested property.
+
+Some of the properties are returned as a byte-array. The easiest way to convert them to string
```
jantonguirao, line 25: A suggestion. I think that we should list the properties here in the body of the documentation. The argument documentation shows indented and later in the documentation so they might get hidden

dali/operators/util/get_property.cc
```diff
@@ -0,0 +1,162 @@
...
+
+namespace detail {
+namespace {
+
+template <typename Backend, typename BatchContainer = batch_container_t<Backend>>
+struct SourceInfo : public Property<Backend> {
+  TensorListShape<> GetShape(const BatchContainer& input) override {
+    TensorListShape<> ret{static_cast<int>(input.num_samples()), 1};
+    for (int i = 0; i < ret.size(); i++) {
+      ret.set_tensor_shape(i, {static_cast<int>(GetSourceInfo(input, i).length())});
```
jantonguirao, line 56:
```suggestion
      ret.set_tensor_shape(i, {static_cast<int64_t>(GetSourceInfo(input, i).length())});
```
as this is what it's used in TensorShape

dali/pipeline/data/meta.h
```diff
@@ -35,7 +35,7 @@ class DALIMeta {
     layout_ = layout;
   }
 
-  inline std::string GetSourceInfo() const {
+  inline const std::string &GetSourceInfo() const {
```
jantonguirao, line 38: :+1:

dali/test/python/test_dali_cpu_only.py
```diff
@@ -978,6 +978,24 @@ def test_tensor_subscript():
...
+    ref_paths = ["0/cat-1046544_640.png", "0/cat-111793_640.png", "0/cat-1245673_640.png", "0/cat-2184682_640.png",
+                 "0/cat-300572_640.png", "0/cat-3113513_640.png", "0/cat-3449999_640.png", "0/cat-3504008_640.png"]
+    p = file_properties(root_path, batch_size=8, num_threads=4, device_id=None)
+    p.build()
+    output = p.run()
+    for out in output:
+        for source_info, ref in zip(out, ref_paths):
+            assert np.array(source_info).tobytes().decode() == ref
+
+
```
jantonguirao, line 998: seems like this is not used. Remove?

dali/test/python/test_operator_get_property.py
```diff
@@ -0,0 +1,138 @@
...
+@pipeline_def
+def file_properties(root_path, device):
+    read, _ = fn.readers.file(file_root=root_path)
+    if device == 'gpu':
+        read = read.gpu()
+    return fn.get_property(read, key="source_info")
+
+
+def _test_file_properties(device):
+    root_path = os.path.join(test_data_root, 'db', 'single', 'png')
```
jantonguirao, line 37: nitpick: This is now making assumptions on what fn.readers.file will discover and in what order. Maybe it's more intuitive if we pass the file_list directly? The test would be more readable, I mean

dali/test/python/test_operator_get_property.py
```diff
@@ -0,0 +1,138 @@
...
+    for dev in ['cpu', 'gpu']:
+        yield _test_tfr_properties, dev
+
+
+@pipeline_def
+def es_properties(layouts, device):
+    num_outputs = len(layouts)
+
+    def gen_data():
+        from torch import rand
```
jantonguirao, line 116: this would make anyone running this test to have torch installed. Why not using `np.random` utils?

dali/test/python/test_operator_get_property.py
```diff
@@ -0,0 +1,138 @@
...
+def tfr_properties(root_path, index_path, device):
+    import nvidia.dali.tfrecord as tfrec
+    inputs = fn.readers.tfrecord(path=root_path, index_path=index_path,
+                                 features={"image/encoded": tfrec.FixedLenFeature((), tfrec.string, ""),
+                                           "image/class/label": tfrec.FixedLenFeature([1], tfrec.int64, -1)})
+    if device == 'gpu':
+        return fn.get_property(inputs["image/encoded"], key="source_info").gpu(), \
+               fn.get_property(inputs["image/class/label"], key="source_info").gpu()
+    return fn.get_property(inputs["image/encoded"], key="source_info"), \
+           fn.get_property(inputs["image/class/label"], key="source_info")
```
jantonguirao, line 90:
```suggestion
    out1 = fn.get_property(inputs["image/encoded"], key="source_info")
    out2 = fn.get_property(inputs["image/class/label"], key="source_info")
    if device == 'gpu':
      out1 = out1.gpu()
      out2 = out2.gpu()
    return out1, out2
```
^^ to avoid repetition

dali/test/python/test_operator_get_property.py
```diff
@@ -0,0 +1,138 @@
...
+    output = p.run()
+    for out, lt in zip(output, layouts):
+        out = out if device == 'cpu' else out.as_cpu()
+        for sample in out:
+            assert _uint8_tensor_to_string(sample), lt
+
+
+def test_es_properties():
+    for dev in ['cpu', 'gpu']:
+        yield _test_es_properties, dev
```
jantonguirao, line 138: maybe add a negative test? Unexpected key?

dali/operators/util/get_property.cc
```diff
@@ -0,0 +1,162 @@
...
+    auto& output = ws.template Output<Backend>(0);
+    for (size_t sample_id = 0; sample_id < input.num_samples(); sample_id++) {
+      auto layout = GetLayout(input, sample_id);
+      output[sample_id].Copy(make_cspan((const uint8_t*)layout.c_str(), layout.size()), nullptr);
+    }
+  }
+
+ private:
+  const TensorLayout& GetLayout(const BatchContainer& input, int idx) {
+    return input[idx].GetMeta().GetLayout();
```
jantonguirao, line 121: idea, make a 
```
template <typename Batch>
const TensorMeta& GetMeta(const Batch &batch, int idx);
```

and specialize for TensorVector and TensorList.

Then you can have a single definition for your properties instead of specializing GetLayout/GetSourceInfo for TensorList.

dali/operators/util/get_property.cc
```diff
@@ -0,0 +1,162 @@
...
+  DALIDataType GetType(const BatchContainer&) override {
+    return DALI_UINT8;
+  }
+
+  void FillOutput(workspace_t<Backend>& ws) override {
+    const auto& input = ws.template Input<Backend>(0);
+    auto& output = ws.template Output<Backend>(0);
+    for (size_t sample_id = 0; sample_id < input.num_samples(); sample_id++) {
+      auto si = GetSourceInfo(input, sample_id);
+      output[sample_id].Copy(make_cspan((const uint8_t*)si.c_str(), si.length()), nullptr);
```
mzient, line 70:
```suggestion
      output[sample_id].Copy(make_cspan((const uint8_t*)si.c_str(), si.length()), nullptr);  // NOLINT
```
or `reinterpret_cast` or 2x `static_cast`

dali/operators/util/get_property.cc
```diff
@@ -0,0 +1,162 @@
...
+    cudaMemcpyAsync(output_ptr, si.c_str(), si.length(), cudaMemcpyDefault, ws.stream());
+  }
+}
+
+
+template <typename Backend, typename BatchContainer = batch_container_t<Backend>>
+struct Layout : public Property<Backend> {
+  TensorListShape<> GetShape(const BatchContainer& input) override {
+    // Every tensor in the output has the same number of dimensions
+    return uniform_list_shape(input.num_samples(), {GetLayout(input, 0).size()});
```
jantonguirao, line 103: maybe we don't need to assume this here?

dali/operators/util/get_property.cc
```diff
@@ -0,0 +1,162 @@
...
+  DALIDataType GetType(const BatchContainer&) override {
+    return DALI_UINT8;
+  }
+
+  void FillOutput(workspace_t<Backend>& ws) override {
+    const auto& input = ws.template Input<Backend>(0);
+    auto& output = ws.template Output<Backend>(0);
+    for (size_t sample_id = 0; sample_id < input.num_samples(); sample_id++) {
+      auto layout = GetLayout(input, sample_id);
+      output[sample_id].Copy(make_cspan((const uint8_t*)layout.c_str(), layout.size()), nullptr);
```
jantonguirao, line 115: static_cast?

dali/operators/util/get_property.cc
```diff
@@ -0,0 +1,162 @@
...
+
+template <>
+void Layout<GPUBackend>::FillOutput(workspace_t<GPUBackend>& ws) {
+  const auto& input = ws.template Input<GPUBackend>(0);
+  auto& output = ws.template Output<GPUBackend>(0);
+  for (size_t sample_id = 0; sample_id < input.num_samples(); sample_id++) {
+    auto layout = GetLayout(input, sample_id);
+    auto output_ptr = output.raw_mutable_tensor(static_cast<int>(sample_id));
+    assert(ws.has_stream());  // GPU operator must have CUDA stream assigned
+    cudaMemcpyAsync(output_ptr, layout.c_str(), layout.size(), cudaMemcpyDefault, ws.stream());
```
jantonguirao, line 138: Copy does this for you, no need to specialize for GPU I think.

dali/operators/util/get_property.cc
```diff
@@ -0,0 +1,162 @@
...
+    .DocStr(
+        R"code(Returns a property of the tensor passed as an input.
+The type of the output will depend on the ``key`` of the requested property.
+
+Some of the properties are returned as a byte-array. The easiest way to convert them to string
+would be to )code")
+    .NumInput(1)
+    .NumOutput(1)
+    .AddArg("key",
+            R"code(Specifies, which property is requested. The following properties are supported:
```
JanuszL, line 30:
```suggestion
            R"code(Specifies, which property is requested.
            
The following properties are supported:
```

dali/operators/util/get_property.cc
```diff
@@ -0,0 +1,162 @@
...
+
+* ``"source_info"``: Returned type: byte-array.
+                     Information about the origin of the sample. The actual origin may differ according
```
JanuszL, line 33: I'm not sure if this formats correctly.

dali/operators/util/get_property.cc
```diff
@@ -0,0 +1,162 @@
...
+}
+
+template <>
+void SourceInfo<GPUBackend>::FillOutput(workspace_t<GPUBackend>& ws) {
+  const auto& input = ws.template Input<GPUBackend>(0);
+  auto& output = ws.template Output<GPUBackend>(0);
+  for (size_t sample_id = 0; sample_id < input.num_samples(); sample_id++) {
+    auto si = GetSourceInfo(input, sample_id);
+    auto output_ptr = output.raw_mutable_tensor(static_cast<int>(sample_id));
+    assert(ws.has_stream());  // GPU operator must have CUDA stream assigned
```
JanuszL, line 93: Do we need this? I think we can consider this as granted.

dali/operators/util/get_property.cc
JanuszL: I think we can do that. Every tensor should have the same rank and in effect the same layout. I cannot imagine when it is not the case.

dali/test/python/test_operator_get_property.py
JanuszL: In this case it should go to `test_pytorch.sh` while it doesn't need to.

dali/test/python/test_operator_get_property.py
```diff
@@ -0,0 +1,138 @@
...
+def es_properties(layouts, device):
+    num_outputs = len(layouts)
+
+    def gen_data():
+        from torch import rand
+        yield rand(num_outputs, 3, 4, 5)
+
+    inp = fn.external_source(source=gen_data, layout=layouts, num_outputs=num_outputs, batch=False, cycle=True)
+    if device == 'gpu':
+        return tuple(fn.get_property(i, key="layout").gpu() for i in inp)
```
JanuszL, line 121:
```suggestion
    inp = fn.external_source(source=gen_data, layout=layouts, num_outputs=num_outputs, batch=False, cycle=True, device = device )
```

dali/operators/util/get_property.cc
szalpal: We agreed earlier, that we shouldn't cover all the cases in the documentation, therefore these are just examples.

dali/operators/util/get_property.cc
szalpal: I'm not sure, what is your point here, but I like the way it's rendered right now:
image

dali/test/python/test_dali_cpu_only.py
szalpal: I'm not sure what you mean. It's a CPU-only test for `fn.get_property`.

dali/test/python/test_operator_get_property.py
szalpal: Good point. Done.

dali/test/python/test_operator_get_property.py
szalpal: Done

dali/test/python/test_operator_get_property.py
szalpal: The version with `.gpu()` at the end showed already a bug (or maybe not a bug, but a problem) in `Copy` function - see the ugly C-style cast. I'd keep the repetition to be guarded from that problem.

dali/test/python/test_operator_get_property.py
szalpal: Done

dali/operators/util/get_property.cc
szalpal: I'm not sure, which `Copy` you are referring to. I guess it is about `Tensor::Copy` function, which I used for the CPU variant. The `input` type in the GPU variant is a `TensorList` and it does not have such `Copy` function which would allow accessing tensors in a batch one by one. I'd gladly see some generalization of `FillOutput` function for both backend, but I'm afraid it is not possible. I mean, it is possible to some extent, but maybe it's not worth obfuscating the code just to save 3 LOC.

dali/operators/util/get_property.cc
szalpal: I went with `reinterpret_cast`

dali/operators/util/get_property.cc
szalpal: Done

dali/operators/util/get_property.cc
szalpal: Done

dali/operators/util/get_property.cc
szalpal: Done

dali/operators/util/get_property.cc
szalpal: Done. Although, linter didn't complain about the C-style cast

dali/test/python/test_operator_get_property.py
szalpal: Done

dali/operators/util/get_property.cc
szalpal: The way it's rendered right now looks nice. Is there a reason to split this?
https://user-images.githubusercontent.com/7101064/146656244-41c87a5c-e10a-44ff-be04-0462f7755402.png

dali/operators/util/get_property.cc
szalpal: It looks OK. Do you feel something better is necessary?
https://user-images.githubusercontent.com/7101064/146656244-41c87a5c-e10a-44ff-be04-0462f7755402.png

szalpal:
!build

szalpal:
!build

dali/operators/util/get_property.cc
szalpal: Removed

dali/operators/util/get_property.cc
szalpal: Fixed

dali/operators/util/get_property.cc
szalpal: Done

szalpal:
!build

dali/operators/util/get_property.cc
```diff
@@ -0,0 +1,45 @@
...
+    .AddArg("key",
+            R"code(Specifies, which property is requested. The following properties are supported:
+
+* ``"source_info"``: Returned type: byte-array.
+                     Information about the origin of the sample. The actual origin may differ according
+                     to the source of the data. For example, when the Tensor is read via :meth:`fn.readers.file`,
+                     the ``source_info`` property will contain full path of that file. When the Tensor
+                     is read by :meth:`fn.readers.webdataset`, the property will contain full path of the tar
+                     archive and the index of the component in that archive. When the Tensor
+                     is loaded via :meth:`~nvidia.dali.fn.external_source`, the ``source_info`` will be empty.
```
jantonguirao, line 36: Just a thought. After reading the rendered documents, I got the impression that this puts too much emphasis on webdataset or on readers in particular. It doesn't say that those are just examples so the reader might assume that it's only those cases that we support. I suggest that here we just describe what source info is, and add documentation on each reader (any operator that produces a source info property) with the particular format of source info used. What do you think?

dali/test/python/test_operator_get_property.py
jantonguirao: I don't quite understand what was the problem, but the code that I am suggesting is equivalent

dali/operators/util/get_property.cc
jantonguirao: Not fixed?

dali/operators/util/get_property.cc
jantonguirao: I see a c-style cast :)

dali/operators/util/get_property.cc
jantonguirao: OK

dali/operators/util/get_property.cc
```diff
@@ -0,0 +1,45 @@
...
+* ``"layout"``: Returned type: byte-array
+                Data layout in the given Tensor.
```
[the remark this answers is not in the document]
szalpal: How about adding a ref to https://docs.nvidia.com/deeplearning/dali/main-user-guide/docs/data_types.html#data-layouts ?

dali/operators/util/property.h
```diff
@@ -0,0 +1,112 @@
...
+namespace tensor_property {
+
+namespace detail {
+
+inline const DALIMeta& GetMeta(const TensorVector<CPUBackend>& batch, int tensor_idx) {
+  return batch[tensor_idx].GetMeta();
+}
+
+inline const DALIMeta& GetMeta(const TensorList<GPUBackend>& batch, int tensor_idx) {
+  return batch.GetMeta(static_cast<int>(tensor_idx));
```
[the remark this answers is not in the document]
szalpal: Done

dali/operators/util/get_property.h
```diff
@@ -0,0 +1,72 @@
...
+    output_desc[0].type = property_->GetType(input);
+    return true;
+  }
+
+  void RunImpl(workspace_t<Backend> &ws) override {
+    property_->FillOutput(ws);
+  }
+
+ private:
+  void PropertyFactory(const std::string &property_key) {
```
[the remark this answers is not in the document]
szalpal: Went with `std::unique_ptr<tensor_property::Property<Backend>> PropertyFactory(...)` since it does look more like a good practice to initialize members in the ctor initialize list

dali/operators/util/get_property.cc
szalpal: Removed the description of particular source_infos. Left only an example with `file_reader`.

jantonguirao (approved):

szalpal:
!build

dali/operators/util/get_property.cc
JanuszL: I think the rule that the parameter description should be one short sentence, and the follow up should be in the new line.
```
key - Specifies, which property is requested.
The following properties are supported:
....
```
We do that for other ops.

dali/test/python/test_operator_get_property.py
```diff
@@ -0,0 +1,156 @@
...
+def _test_improper_property(device):
+    root_path = os.path.join(get_dali_extra_path(), "db/webdataset/MNIST/devel-0.tar")
+    p = improper_property(root_path, device, batch_size=8, num_threads=4, device_id=0)
+    p.build()
+    output = p.run()
+
+
+@raises
+def test_improper_property():
+    for dev in ['cpu', 'gpu']:
```
JanuszL, line 155: Please use `assert_raises` from DALI nose_utils.py and check the exact message and error type.

dali/operators/util/property.cc
```diff
@@ -0,0 +1,65 @@
...
+}
+
+template <>
+void Layout<CPUBackend>::FillOutput(workspace_t<CPUBackend>& ws) {
+  const auto& input = ws.template Input<CPUBackend>(0);
+  auto& output = ws.template Output<CPUBackend>(0);
+  for (size_t sample_id = 0; sample_id < input.num_samples(); sample_id++) {
+    auto layout = GetLayout(input, sample_id);
+    output[sample_id].Copy(
+        make_cspan(reinterpret_cast<const uint8_t*>(layout.c_str()), layout.size()), nullptr);
```
JanuszL, line 38: Why you cannot use `.Copy` in other places?
In such case you can create one function for both backends and just use `has_stream` and `stream()` to set the right stream. Something like:
```
template <typename Backend>
void Layout<Backend>::FillOutput(workspace_t<Backend>& ws) {
  const auto& input = ws.template Input<Backend>(0);
  auto& output = ws.template Output<Backend>(0);
  for (size_t sample_id = 0; sample_id < input.num_samples(); sample_id++) {
    auto layout = GetLayout(input, sample_id);
    output[sample_id].Copy(
        make_cspan(reinterpret_cast<const uint8_t*>(layout.c_str()), layout.size()), ws.has_stream() ? ws.stream() : 0);
}

void Layout<CPUBackend>::FillOutput(workspace_t<CPUBackend>& ws);
void Layout<GPUBackend>::FillOutput(workspace_t<GPUBackend>& ws);
```

JanuszL: Ok, we cannot do it this way. My bad.

dali/test/python/test_operator_get_property.py
szalpal: Done

dali/operators/util/get_property.cc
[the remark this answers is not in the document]
jantonguirao: Yes, a link to that would be useful

dali/operators/util/get_property.cc
szalpal: Done

JanuszL (approved):

szalpal:
!build

PR #3576, opened 2021-12-16
Remove Buffer inheritence from TensorList

klecki:
## Description

### What happened in this PR
TensorList no longer inherits from Buffer.
TensorList has a Buffer member.
Repilicate necessary APIs.

Follow-ups:
Same treatment for Tensor.
Rename the TensorList APIs to CamelCase.

#### Additional information
TensorList, Buffer

### Tests
- [x] Existing tests apply
- [ ] New tests added
  - [ ] Python tests
  - [ ] GTests
  - [ ] Benchmark
  - [ ] Other
- [ ] N/A

### Documentation
- [x] Documentation updated
  - [x] Docstring
  - [ ] Doxygen
  - [ ] RST
  - [ ] Jupyter
  - [ ] Other
- [ ] N/A

#### Requirements
- [x] N/A

**REQ IDs**: N/A

**JIRA TASK**: N/A

### DIFF
--- a/dali/pipeline/data/buffer.h
+++ b/dali/pipeline/data/buffer.h
@@ -207,6 +207,10 @@ class DLL_PUBLIC Buffer {
     return !!data_;
   }
 
+  std::shared_ptr<void> get_data_ptr() const {
+    return data_;
+  }
+
   /**
    * @brief Sets the type of allocation (pinned/non-pinned) for CPU buffers
    */
@@ -307,7 +311,7 @@ class DLL_PUBLIC Buffer {
     return shares_data_;
   }
 
-  inline void SetBackingAllocation(const Buffer<Backend> &other) {
+  inline void set_backing_allocation(const Buffer<Backend> &other) {
     DALI_ENFORCE(IsValidType(other.type_), "Only valid type when creating an allocation");
     type_ = other.type_;
     data_ = other.data_;
@@ -318,15 +322,20 @@ class DLL_PUBLIC Buffer {
     device_ = other.device_;
   }
 
-  inline void SetBackingAllocation(const shared_ptr<void> &ptr, size_t bytes, DALIDataType type, size_t size = 0) {
-    DALI_ENFORCE(IsValidType(type), "Only valid type when creating an allocation");
+  inline void set_backing_allocation(const shared_ptr<void> &ptr, size_t bytes,
+                                   DALIDataType type = DALI_NO_TYPE, size_t size = 0) {
     type_ = TypeTable::GetTypeInfo(type);
     data_ = ptr;
     allocate_ = {};
     size_ = size;
     shares_data_ = true;
     num_bytes_ = bytes;
-    device_ = CPU_ONLY_DEVICE_ID; // TODO(klecki): Do we keep old, set new, or pass it here?
+    // setting the allocation, get the device
+    if (std::is_same<Backend, GPUBackend>::value) {
+      CUDA_CALL(cudaGetDevice(&device_));
+    } else {
+      device_ = CPU_ONLY_DEVICE_ID;
+    }
   }
 
   DISABLE_COPY_MOVE_ASSIGN(Buffer);
--- a/dali/pipeline/data/buffer.h
+++ b/dali/pipeline/data/buffer.h
@@ -332,7 +332,7 @@ class DLL_PUBLIC Buffer {
    * Current Buffer will be marked as sharing data, and reallocation of memory will be
    * prohibited until reset() is called.
    */
-  inline void set_backing_allocation(const Buffer<Backend> &other) {
+  inline void ShareData(const Buffer<Backend> &other) {
     type_ = other.type_;
     data_ = other.data_;
     allocate_ = {};
@@ -356,7 +356,7 @@ class DLL_PUBLIC Buffer {
     data_ = ptr;
     allocate_ = {};
     size_ = size;
-    shares_data_ = true;
+    shares_data_ = data_ != nullptr;
     num_bytes_ = bytes;
     pinned_ = pinned;
     // setting the allocation, get the device
--- a/dali/pipeline/data/buffer.h
+++ b/dali/pipeline/data/buffer.h
@@ -360,7 +360,7 @@ class DLL_PUBLIC Buffer {
     num_bytes_ = bytes;
     pinned_ = pinned;
     // setting the allocation, get the device
-    if ((std::is_same<Backend, GPUBackend>::value || pinned_) && device_ != CPU_ONLY_DEVICE_ID) {
+    if ((std::is_same<Backend, GPUBackend>::value || pinned_) && device_ == CPU_ONLY_DEVICE_ID) {
       CUDA_CALL(cudaGetDevice(&device_));
     } else {
       device_ = CPU_ONLY_DEVICE_ID;
--- a/dali/pipeline/data/buffer.h
+++ b/dali/pipeline/data/buffer.h
@@ -71,10 +71,10 @@ inline string ShapeString(vector<Index> shape) {
  * underlying storage is located (CPU or GPU).
  *
  * Buffers are untyped on construction, and don't receive a valid type until
- * 'set_type' or 'data<T>()' is called on a non-const buffer. Upon receiving
+ * 'set_type' is called on a non-const buffer. Upon receiving
  * a valid type, the underlying storage for the buffer is allocated. The type
  * of the underlying data can change over the lifetime of an object if
- * 'set_type' or 'data<T>()' is called again where the calling type does not
+ * 'set_type' is called again where the calling type does not
  * match the underlying type on the buffer. In this case, the Buffer swaps its
  * current type, but only re-allocates memory if it does not have enough bytes
  * of allocated storage to store the number of elements in the buffer with the
@@ -307,6 +307,28 @@ class DLL_PUBLIC Buffer {
     return shares_data_;
   }
 
+  inline void SetBackingAllocation(const Buffer<Backend> &other) {
+    DALI_ENFORCE(IsValidType(other.type_), "Only valid type when creating an allocation");
+    type_ = other.type_;
+    data_ = other.data_;
+    allocate_ = {};
+    size_ = other.size_;
+    shares_data_ = true;
+    num_bytes_ = other.num_bytes_;
+    device_ = other.device_;
+  }
+
+  inline void SetBackingAllocation(const shared_ptr<void> &ptr, size_t bytes, DALIDataType type, size_t size = 0) {
+    DALI_ENFORCE(IsValidType(type), "Only valid type when creating an allocation");
+    type_ = TypeTable::GetTypeInfo(type);
+    data_ = ptr;
+    allocate_ = {};
+    size_ = size;
+    shares_data_ = true;
+    num_bytes_ = bytes;
+    device_ = CPU_ONLY_DEVICE_ID; // TODO(klecki): Do we keep old, set new, or pass it here?
+  }
+
   DISABLE_COPY_MOVE_ASSIGN(Buffer);
 
   static void SetGrowthFactor(double factor) {
@@ -326,7 +348,6 @@ class DLL_PUBLIC Buffer {
 
   DLL_PUBLIC static constexpr double kMaxGrowthFactor = 4;
 
- protected:
   // Helper to resize the underlying allocation
   inline void ResizeHelper(Index new_size) {
     ResizeHelper(new_size, type_);
@@ -390,6 +411,7 @@ class DLL_PUBLIC Buffer {
     buffer.reset();
   }
 
+ protected:
   static double growth_factor_;
   static double shrink_threshold_;
 
--- a/dali/pipeline/data/buffer.h
+++ b/dali/pipeline/data/buffer.h
@@ -337,7 +337,7 @@ class DLL_PUBLIC Buffer {
     data_ = other.data_;
     allocate_ = {};
     size_ = other.size_;
-    shares_data_ = true;
+    shares_data_ = data_ != nullptr;
     num_bytes_ = other.num_bytes_;
     device_ = other.device_;
   }
--- a/dali/pipeline/data/buffer.h
+++ b/dali/pipeline/data/buffer.h
@@ -91,6 +91,21 @@ class DLL_PUBLIC Buffer {
    */
   inline Buffer() = default;
   virtual ~Buffer() = default;
+  Buffer(const Buffer &other) = delete;
+  Buffer& operator=(const Buffer &other) = delete;
+
+  Buffer(Buffer &&other) {
+    move_buffer(std::move(other));
+  }
+
+  Buffer& operator=(Buffer &&other) {
+    if (this != &other) {
+      move_buffer(std::move(other));
+    }
+    return *this;
+  }
+
+
 
   /**
    * @brief Returns a typed pointer to the underlying storage.
@@ -148,7 +163,7 @@ class DLL_PUBLIC Buffer {
    */
   inline size_t nbytes() const {
     // Note: This returns the number of bytes occupied by the current
-    // number of elements stored in the buffer. This is not neccessarily
+    // number of elements stored in the buffer. This is not necessarily
     // the number of bytes of the underlying allocation (num_bytes_)
     return size_ * type_.size();
   }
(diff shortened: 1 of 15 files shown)


klecki:
!build

klecki:
!build

dali/pipeline/data/buffer.h
```diff
@@ -377,6 +438,7 @@ class DLL_PUBLIC Buffer {
     }
   }
 
+ protected:
```
szalpal, line 441: Since you're removing inheritance from `Buffer`, I'd make the `move_buffer` method private. Whoever would like to inherit from `Buffer` in the future would need to write his own move routine anyway, so it's good idea to restrict this only for `Buffer` usage.

Or maybe remove the `move_buffer` and implement it in the idiomatic way?
```c++
operator=(Buffer&&) {
<what's currently in move_buffer(...)>
}

Buffer(Buffer&& b) {
*this = move(b);
}
```

dali/pipeline/data/tensor_list.h
```diff
@@ -626,7 +596,128 @@ class DLL_PUBLIC TensorList : private Buffer<Backend> {
...
+
+  /**
+   * @brief Sets a custom allocation function.
+   *
+   * Sets a custom allocation function. The allocation function returns
+   * a shared pointer with a matching deleter.
+   *
+   * @remarks Experimental - subject to change
+   */
+  inline void set_alloc_func(typename Buffer<Backend>::AllocFunc allocate) {
```
szalpal, line 704: Is it really experimental? Or just temporary?

klecki: I just propagated the docstring from what we have in Buffer.

dali/pipeline/data/buffer.h
klecki: I will probably do a similar PR just for Tensor, I can adjust it there.

dali/pipeline/data/tensor_list.h
```diff
@@ -626,7 +596,128 @@ class DLL_PUBLIC TensorList : private Buffer<Backend> {
...
+    data_.set_alloc_func(std::move(allocate));
+  }
```
mzient, line 706: Oh no... This was meant to be a temporary workaround... Please expose the buffer and let it be set there...

dali/pipeline/data/tensor_list.h
```diff
@@ -548,7 +518,7 @@ class DLL_PUBLIC TensorList : private Buffer<Backend> {
     // need to create a new view
     DALI_ENFORCE(num_samples() > 0,
                  "To create a view Tensor, the Tensor List must have at least 1 element.");
-    DALI_ENFORCE(IsValidType(type()),
+    DALI_ENFORCE(IsValidType(type_info()),
```
mzient, line 521:
```suggestion
    DALI_ENFORCE(IsValidType(type()),
```
Don't use type_info for such a simple check.

dali/pipeline/data/tensor_list.h
```diff
@@ -164,19 +146,22 @@ class DLL_PUBLIC TensorList : private Buffer<Backend> {
...
+    data_.reserve(bytes);
+  }
   /**
    * @brief Resize function to allocate a list of tensors. The input vector
    * contains a set of dimensions for each tensor to be allocated in the
    * list.
    */
   DLL_PUBLIC inline void Resize(const TensorListShape<> &new_shape) {
-    DALI_ENFORCE(IsValidType(type_),
+    DALI_ENFORCE(IsValidType(type_info()),
```
mzient, line 161:
```suggestion
    DALI_ENFORCE(IsValidType(type()),
```

dali/pipeline/data/tensor_list.h
```diff
@@ -227,25 +212,17 @@ class DLL_PUBLIC TensorList : private Buffer<Backend> {
    * Size can be set to 0 and type to NoType as intermediate step.
    */
   DLL_PUBLIC inline void ShareData(const TensorList<Backend> &other) {
-    DALI_ENFORCE(IsValidType(other.type_), "To share data, "
+    DALI_ENFORCE(IsValidType(other.type_info()), "To share data, "
```
mzient, line 215:
```suggestion
    DALI_ENFORCE(IsValidType(other.type()), "To share data, "
```

dali/pipeline/data/tensor_list.h
klecki: As discussed, I don't want to expose buffer as there might not be just one buffer in the future, but multiple samples.

dali/pipeline/data/tensor_list.h
klecki: Done

dali/pipeline/data/tensor_list.h
klecki: Done

dali/pipeline/data/tensor_list.h
klecki: done

klecki:
!build

dali/pipeline/data/buffer.h
```diff
@@ -91,6 +91,21 @@ class DLL_PUBLIC Buffer {
    */
   inline Buffer() = default;
   virtual ~Buffer() = default;
+  Buffer(const Buffer &other) = delete;
+  Buffer& operator=(const Buffer &other) = delete;
```
mzient, line 95: Isn't it automatic?

dali/pipeline/data/buffer.h
```diff
@@ -307,7 +326,45 @@ class DLL_PUBLIC Buffer {
...
+                                     DALIDataType type = DALI_NO_TYPE, size_t size = 0) {
+    type_ = TypeTable::GetTypeInfo(type);
+    data_ = ptr;
+    allocate_ = {};
+    size_ = size;
+    shares_data_ = true;
+    num_bytes_ = bytes;
+    // setting the allocation, get the device
+    if (std::is_same<Backend, GPUBackend>::value) {
+      CUDA_CALL(cudaGetDevice(&device_));
```
mzient, line 363: I don't like this logic. Assuming, that this buffer already has a device - why would we change it? We have multiple sources of device_id - the one currently used, current device (as obtained from cudaGetDevice) and the real one - associated with the pointer. Unfortunately, getting the real one is too expensive - and it can be ambiguous (managed buffers, cuMemMap, etc). Since we can't get the real one, who's to say that current context holds the right device and not the buffer object?

dali/pipeline/data/buffer.h
```diff
@@ -307,7 +326,45 @@ class DLL_PUBLIC Buffer {
...
+    } else {
+      device_ = CPU_ONLY_DEVICE_ID;
```
mzient, line 365: Managed buffers can have a device id and I've reworked them to do so in my PR.

dali/pipeline/data/buffer.h
```diff
@@ -307,7 +326,45 @@ class DLL_PUBLIC Buffer {
...
+  inline void set_backing_allocation(const shared_ptr<void> &ptr, size_t bytes,
+                                     DALIDataType type = DALI_NO_TYPE, size_t size = 0) {
+    type_ = TypeTable::GetTypeInfo(type);
+    data_ = ptr;
+    allocate_ = {};
+    size_ = size;
+    shares_data_ = true;
+    num_bytes_ = bytes;
+    // setting the allocation, get the device
+    if (std::is_same<Backend, GPUBackend>::value) {
```
mzient, line 362:
```suggestion
    if (std::is_same<Backend, GPUBackend>::value || is_pinned_) {
```

dali/pipeline/data/buffer.h
```diff
@@ -307,7 +326,45 @@ class DLL_PUBLIC Buffer {
...
+
+  /**
+   * @brief Set external memory as the backing memory for this Buffer.
+   *
+   * Current Buffer will be marked as sharing data, and reallocation of memory will be
+   * prohibited until reset() is called.
+   *
+   * For GPU memory, it is assumed to be associated with current device.
+   */
+  inline void set_backing_allocation(const shared_ptr<void> &ptr, size_t bytes,
```
mzient, line 353: I think at this point we should also specify whether the buffer is pinned or not
```suggestion
  inline void set_backing_allocation(const shared_ptr<void> &ptr, size_t bytes, bool pinned,
```

dali/pipeline/data/buffer.h
```diff
@@ -307,7 +326,45 @@ class DLL_PUBLIC Buffer {
...
+   * Current Buffer will be marked as sharing data, and reallocation of memory will be
+   * prohibited until reset() is called.
+   */
+  inline void set_backing_allocation(const Buffer<Backend> &other) {
+    type_ = other.type_;
+    data_ = other.data_;
+    allocate_ = {};
+    size_ = other.size_;
+    shares_data_ = true;
+    num_bytes_ = other.num_bytes_;
```
mzient, line 341: How about this?
```suggestion
    num_bytes_ = other.num_bytes_;
    shares_data_ = num_bytes_ > 0;
```
Do we want the buffer to be set to "sharing" even if it doesn't really share anything?

dali/pipeline/data/buffer.h
```diff
@@ -307,7 +326,45 @@ class DLL_PUBLIC Buffer {
...
   }
 
-  DISABLE_COPY_MOVE_ASSIGN(Buffer);
+  /**
+   * @brief Set another Buffer as the backing memory for this Buffer.
+   *
+   * Current Buffer will be marked as sharing data, and reallocation of memory will be
+   * prohibited until reset() is called.
+   */
+  inline void set_backing_allocation(const Buffer<Backend> &other) {
```
mzient, line 335: Why not `ShareData`? That's what we use in Tensor and TensorXxx - it's not that different here. After all, this does more than just setting the backing allocation - it copies other properties, too (type, logical size, device).

dali/pipeline/data/buffer.h
klecki: We had `DISABLE_COPY_MOVE_ASSIGN(Buffer);` previously, so I just replicated it manually while adding the move variants.

dali/pipeline/data/buffer.h
klecki: I think this names is more descriptive compared to ShareData - ShareData suggests that it can work kinda both ways, but maybe I'm reading too much into it. If you set some external allocation to this Buffer (be it some shared_ptr or other Buffer), we assume it's valid until you try to cause a reallocation.

dali/pipeline/data/buffer.h
klecki: Hmm, make sense, but should we allow sharing (basically setting some allocation for this buffer), that is empty and does not result in sharing anything?

dali/pipeline/data/buffer.h
klecki: Maybe it should be somehow passed with the pointer? But if you are setting a totally new allocation (that is potentially a "3rd-party" allocation as you are just passing a pointer), I would assume that it's unrelated to what the Buffer currently holds. 
Either way let's try to limit ways of interleaving the setup of our memory objects, so we don't get another type-shape-reserve situation with all possible permutations.

dali/pipeline/data/buffer.h
mzient: My point is: when we provide move variants and don't provide copy ones, they are implicitly deleted.

dali/pipeline/data/buffer.h
mzient: I'd keep it protected. This method is safe and self-contained. There's no need to make it public, since it has little merit outside of inheritance context (it's preferable to just move the entire thing), but when implementing move semantics in derived classes, it's much more readable to write `move_buffer(other)` than to resort to monstrosities like:
`static_cast<Buffer<Backend>&>(*this) = std::move(static_cast<Buffer<Backend>&>(other));`

dali/pipeline/data/tensor_list.h
```diff
@@ -154,8 +136,8 @@ class DLL_PUBLIC TensorList : private Buffer<Backend> {
     }
 
     use_copy_kernel &= (std::is_same<SrcBackend, GPUBackend>::value || other.is_pinned()) &&
-                       (std::is_same<Backend, GPUBackend>::value || pinned_);
-    type_.template Copy<SrcBackend, Backend>(dsts.data(), srcs.data(), sizes.data(),
+                       (std::is_same<Backend, GPUBackend>::value || is_pinned());
+    type_info().template Copy<SrcBackend, Backend>(dsts.data(), srcs.data(), sizes.data(),
```
mzient, line 140: Nitpick: type_info is no longer a dependent name, so template disambiguation is not necessary (unless some buggy compiler thinks otherwise, of course).
```suggestion
    type_info().Copy<SrcBackend, Backend>(dsts.data(), srcs.data(), sizes.data(),
```

dali/pipeline/data/tensor_list.h
```diff
@@ -112,9 +93,10 @@ class DLL_PUBLIC TensorList : private Buffer<Backend> {
     this->SetLayout(other.GetLayout());
 
     use_copy_kernel &= (std::is_same<SrcBackend, GPUBackend>::value || other.is_pinned()) &&
-                       (std::is_same<Backend, GPUBackend>::value || pinned_);
-    type_.template Copy<Backend, SrcBackend>(this->raw_mutable_data(), other.raw_data(),
-                                             this->size(), stream, use_copy_kernel);
+                       (std::is_same<Backend, GPUBackend>::value || is_pinned());
+    type_info().template Copy<Backend, SrcBackend>(unsafe_raw_mutable_data(*this),
```
mzient, line 97: Nitpick: type_info is no longer a dependent name, so template disambiguation is not necessary (unless some buggy compiler thinks otherwise, of course).
```suggestion
    type_info().Copy<SrcBackend, Backend>(dsts.data(), srcs.data(), sizes.data(),
```

dali/pipeline/data/buffer.h
klecki: I think I saw explicit deletes in the stream aware data objects PR?

dali/pipeline/data/buffer.h
klecki: Adjusted by doing only when necessary and we have the CPU_ONLY device set.

dali/pipeline/data/buffer.h
klecki: Done

dali/pipeline/data/buffer.h
klecki: Done

dali/pipeline/data/buffer.h
klecki: Reverted to ShareData for the Buffer overload.

dali/pipeline/data/buffer.h
klecki: Done

dali/pipeline/data/tensor_list.h
klecki: Not really, my compiler wasn't happy when I removed `template`.

dali/pipeline/data/buffer.h
mzient: True... I wonder why I did it... Whatever then.

klecki:
!build

klecki:
!build

dali/pipeline/data/buffer.h
mzient: Not done? I can still see hard `shares_data_ = true;`  there...

klecki:
!build

dali/pipeline/data/buffer.h
klecki: Argh, I thought I did it.

klecki:
!build

dali/pipeline/data/buffer.h
```diff
@@ -307,7 +326,46 @@ class DLL_PUBLIC Buffer {
...
+                                     DALIDataType type = DALI_NO_TYPE, size_t size = 0) {
+    type_ = TypeTable::GetTypeInfo(type);
+    data_ = ptr;
+    allocate_ = {};
+    size_ = size;
+    shares_data_ = data_ != nullptr;
+    num_bytes_ = bytes;
+    pinned_ = pinned;
+    // setting the allocation, get the device
+    if ((std::is_same<Backend, GPUBackend>::value || pinned_) && device_ != CPU_ONLY_DEVICE_ID) {
```
mzient, line 363:
```suggestion
    if ((std::is_same<Backend, GPUBackend>::value || pinned_) && device_ == CPU_ONLY_DEVICE_ID) {
```
?

klecki: damn, you're right.

mzient (approved):

szalpal (approved):

PR #3579, opened 2021-12-17
Fix coverity issues

stiepan:
Fixes following coverity issues:
* multiple complaints about uninitialized POD and arrays.
* remove unused member flags_ from ArgValue
* rename local variable so that it does not shadow the function parameter in cuda_vm_resource.h

None of the issues seems to be a bug.

## Description

#### Additional information

### Tests
- [x] Existing tests apply
- [ ] New tests added
  - [ ] Python tests
  - [ ] GTests
  - [ ] Benchmark
  - [ ] Other
- [ ] N/A

### Documentation
- [ ] Documentation updated
  - [ ] Docstring
  - [ ] Doxygen
  - [ ] RST
  - [ ] Jupyter
  - [ ] Other
- [ ] N/A

#### Requirements
- [x] N/A

**REQ IDs**: N/A

**JIRA TASK**: DALI-2501

### DIFF
--- a/dali/kernels/imgproc/convolution/laplacian_cpu_test.cc
+++ b/dali/kernels/imgproc/convolution/laplacian_cpu_test.cc
@@ -150,7 +150,7 @@ struct LaplacianCpuKernelTest : public ::testing::Test {
 
   void RunTest() {
     Kernel kernel;
-    KernelContext ctx;
+    KernelContext ctx = {};
 
     auto req = kernel.Setup(ctx, in_.shape, lapl_params_.window_sizes);
 
@@ -269,6 +269,7 @@ struct LaplacianCpuTest : public ::testing::Test {
   using Conv = SeparableConvolutionCpu<W, In, W, axes, has_channels>;
 
   static std::array<int, ndim> GetShape() {
+    static_assert(ndim == axes + has_channels);
     std::array<int, ndim> shape;
     for (int i = 0; i < axes; i++) {
       shape[i] = dim_sizes[i];
@@ -291,7 +292,7 @@ struct LaplacianCpuTest : public ::testing::Test {
 
   void RunBaseline() {
     Conv kernel;
-    KernelContext ctx;
+    KernelContext ctx = {};
     auto vol = volume(shape_);
     for (int axis = 0; axis < axes; axis++) {
       auto req = kernel.Setup(ctx, shape_, lapl_params_.window_sizes[axis]);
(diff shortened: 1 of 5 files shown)


JanuszL (approved):

szalpal (approved):

dali/operators/image/convolution/laplacian_params.h
```diff
@@ -131,7 +131,7 @@ class LaplacianWindows {
 template <int axes>
 std::array<float, axes> GetNormalizationFactors(
     const std::array<std::array<int, axes>, axes> &window_sizes) {
-  std::array<float, axes> factors;
+  std::array<float, axes> factors = {};
```
mzient, line 134: This is sad. I consider it a change for the worse. Now, if our initialization logic is indeed broken, a smarter tool won't be able to tell us wrong, because the array would be initialized. I'd actually report it as false positive and tell coverity to fix their s**t.

dali/kernels/imgproc/convolution/laplacian_cpu_test.cc
```diff
@@ -95,7 +95,7 @@ struct LaplacianWindows {
   static constexpr int window_size = window_size_;
   static constexpr bool use_smoothing = use_smoothing_;
 
-  LaplacianWindows() {
+  LaplacianWindows() : window_sizes{}, windows{} {
```
mzient, line 98: False positive, isn't it?

stiepan: It doesn't make much sense as it is not read anywhere. I'd say that all those issues are the same (not sure about KernelContext ctx; but I can see it is declared all over the tests the same way). On the other hand if there is an error of this kind one day maybe initialization those first makes the error more reproducible at least?

dali/operators/image/convolution/laplacian_params.h
stiepan: I wasn't sure if I should ignore those, but decided not to ignore them thinking that: if there is indeed a value that is not written with meaningful data one day, it at least will behave the same from run to run, so may be easier to catch?

stiepan: Though I mean it more in the context of class members rather than this one which is obviously over-cautious.

mzient: Well... at least with uninitialized data we can use sanitizers, like valgrind.
```
#include <iostream>
using namespace std;

template <int n, int m>
struct Mat {
    Mat(float constant = 0.0f) {
        for (int i = 0; i < n; i++)
            for (int j = 0; j < n; j++)
                data[i][j] = constant;
    }
    float data[n][m] = {};
};

int main() {
    Mat<2, 3> m = 42;

    for (int i = 0; i < 2; i++) {
        for (int j = 0; j < 3; j++)
            cout << m.data[i][j] << " ";
        cout << "\n";
    }

    return 0;
}
```
According to valgrind, the above program is correct. Removing the initializer `= {}` makes it clear that it isn't.

stiepan: That's a good point. Okay, so let me revert those changes and ignore them in the Coverity.

PR #3600, opened 2021-12-26
Explicitly coalesce stores in Slice for smaller output types

hugo213:
## Description

### What happened in this PR

This PR is an extension of the idea I proposed in https://github.com/NVIDIA/DALI/pull/3568 . That PR improved coalescing of L2 to Global Memory transfers for small ( < 32 bits) output types. However, L1 to L2 transfers still were a bottleneck. The L1 to L2 transfers are 128 bytes wide, which makes it impossible for a 32-thread warp to use more than 25% of the bandwidth when issuing single-byte stores, no matter what the access pattern is. 

In this PR I introduce a buffer, in which the data is cached until full 32-bit store (resulting in full 128 bytes wide memory request from the warp) is possible.

This visibly increases Slice's throughput, as you can see in the plots below, which present the throughput improvement in cropping 2000x2000, 1000x1000 and 500x500 uint8 RGB images to 1000x1000, 500x500 and 250x250 respectively on Titan V. The _Original_ is the Slice before my optimizations, _Coalesced (old)_ is the Slice with https://github.com/NVIDIA/DALI/pull/3568 applied, and _Coalesced (new)_ is the Slice with the proposed changes. Also, Slice's performance on 32-bit floats is drawn for reference, indicating the throughput that Slice can achieve for bigger data types when no coalescing problems occur.

coalesce2

#### Additional information
- Affected modules and functionalities: Slice's performance for smaller output types. Peformance for 32-bit and bigger types is unaffected as almost whole `PackedBuffer` logic gets optimized out.

### Tests
- [x] Existing tests apply
- [ ] New tests added
  - [ ] Python tests
  - [ ] GTests
  - [ ] Benchmark
  - [ ] Other
- [ ] N/A

### Documentation
- [ ] Documentation updated
  - [ ] Docstring
  - [ ] Doxygen
  - [ ] RST
  - [ ] Jupyter
  - [ ] Other
- [ ] N/A

#### Requirements
- [ ] N/A

**REQ IDs**: N/A

**JIRA TASK**: N/A

### DIFF
--- a/dali/kernels/slice/slice_gpu.cuh
+++ b/dali/kernels/slice/slice_gpu.cuh
@@ -84,7 +84,7 @@ union PackedBuffer {
   __device__ inline void store(T* mem, size_t count) {
     if (kCapacity == 1) {
       *mem = *values;
-    } else if (count == kCapacity && (uint64_t)mem % kCapacity == 0) {
+    } else if (count == kCapacity && (uint64_t)mem % sizeof(PackedType) == 0) {
       *reinterpret_cast<PackedType*>(mem) = raw;
     } else {
       #pragma unroll
--- a/dali/kernels/slice/slice_gpu.cuh
+++ b/dali/kernels/slice/slice_gpu.cuh
@@ -72,9 +72,29 @@ struct SliceBlockDesc {
   uint64_t size;
 };
 
-template<typename OutputType>
-constexpr int coalesced_values = sizeof(OutputType) >= 4 ? 1 : 4 / sizeof(OutputType);
+template<typename T>
+union PackedBuffer {
+  using PackedType = uint32_t;
+  static constexpr size_t capacity = sizeof(T) >= sizeof(PackedType) ? 1 : sizeof(PackedType) / sizeof(T);
 
+  T values[capacity];
+  PackedType raw;
+
+  __device__
+  inline void store(T* mem, size_t count) {
+    if (capacity == 1) {
+      *mem = *values;
+    }
+    else if (count == capacity) {
+      *(PackedType*)mem = raw;
+    } else {
+      #pragma unroll
+      for (size_t i = 0; i < count; i++) {
+        mem[i] = values[i];
+      }
+    }
+  }
+};
 
 /**
  * @brief Simplified algorithm when no padding is necessary
@@ -91,12 +111,14 @@ __device__ void SliceFuncNoPad(OutputType *__restrict__ out, const InputType *__
     return;
   }
 
-  for (; offset < block_end; offset += blockDim.x * coalesced_values<OutputType>) {
+  for (; offset < block_end; offset += blockDim.x * PackedBuffer<OutputType>::capacity) {
+    PackedBuffer<OutputType> result;
+
+    uint64_t i;
     #pragma unroll
-    for (uint64_t i = 0; i < coalesced_values<OutputType>; i++) {
+    for (i = 0; i < PackedBuffer<OutputType>::capacity; i++) {
       uint64_t idx = offset + i;
       if (idx >= block_end) break;
-      uint64_t out_idx = idx;
       uint64_t in_idx = 0;
 
       #pragma unroll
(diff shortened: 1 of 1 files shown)


JanuszL:
!build

JanuszL:
!build

jantonguirao (approved):

jantonguirao (approved):

dali/kernels/slice/slice_gpu.cuh
```diff
@@ -72,9 +72,28 @@ struct SliceBlockDesc {
...
+  static constexpr size_t kCapacity = sizeof(T) >= sizeof(PackedType) ?
+                                      1 : sizeof(PackedType) / sizeof(T);
+
+  T values[kCapacity];
+  PackedType raw;
+
+  __device__ inline void store(T* mem, size_t count) {
+    if (kCapacity == 1) {
+      *mem = *values;
+    } else if (count == kCapacity && (uint64_t)mem % sizeof(PackedType) == 0) {
```
JanuszL, line 87:
```suggestion
    } else if (count == kCapacity && (uintptr_t)mem % sizeof(PackedType) == 0) {
```

JanuszL: I guess that `std::align` would be an overkill here.

dali/kernels/slice/slice_gpu.cuh
```diff
@@ -72,9 +72,28 @@ struct SliceBlockDesc {
...
+      *reinterpret_cast<PackedType*>(mem) = raw;
```
JanuszL, line 88: In general this is UB, what would be the difference if you use `memcpy` instead here (which is the only recommended way according to the standard).

dali/kernels/slice/slice_gpu.cuh
```diff
@@ -72,9 +72,28 @@ struct SliceBlockDesc {
...
+    } else {
+      #pragma unroll
+      for (size_t i = 0; i < count; i++) {
+        mem[i] = values[i];
```
JanuszL, line 92: Same as above.

JanuszL (approved):

dali/kernels/slice/slice_gpu.cuh
hugo213: I've applied your suggestion, thanks! I also think that `std::align` would be an overkill, especially for this performance-critical part of code

dali/kernels/slice/slice_gpu.cuh
hugo213: I didn't know that was UB, thank you! I'll post benchmarks with `memcpy` in the evening.

dali/kernels/slice/slice_gpu.cuh
hugo213: Hmm, I think this is valid code and not UB, because both `mem` and `values` are arrays of `T`. Am I missing something?

dali/kernels/slice/slice_gpu.cuh
hugo213: Bad news – `memcpy` compiles to (unrolled) byte-by-byte copy:
```
 ST.E.U8.SYS [R12], R10 
 ST.E.U8.SYS [R12+0x1], R9 
 ST.E.U8.SYS [R12+0x2], R6 
 ST.E.U8.SYS [R12+0x3], R8 
```
which is exactly what happens without explicit coalescing.

dali/kernels/slice/slice_gpu.cuh
hugo213: EDIT: `static_cast` seems illegal here, changed to `reinterpret_cast` in https://github.com/NVIDIA/DALI/pull/3600/commits/ad25f59956ec68fedcd17f5bf265837d273f1a00

dali/kernels/slice/slice_gpu.cuh
JanuszL: So let us keep it as it is.

dali/kernels/slice/slice_gpu.cuh
JanuszL: My bad.

JanuszL:
!build

PR #3605, opened 2022-01-03
Debug mode direct ExternalSource

ksztenderski:
## Description

### What happened in this PR

Adds debug version of ExternalSource operator removing additional callback to the backend. Before that, in the debug mode for external source we created a separate pipeline like for any other operator, which seems like an overkill considering that all of the ExternalSource implementation (relevant for debug mode) is in Python.

Adds TensorList constructor from list of Tensors.

#### Additional information

### Tests
- [x] Existing tests apply
- [x] New tests added
  - [x] Python tests
  - [ ] GTests
  - [ ] Benchmark
  - [ ] Other
- [ ] N/A

### Documentation
- [x] Documentation updated
  - [x] Docstring
  - [ ] Doxygen
  - [ ] RST
  - [ ] Jupyter
  - [ ] Other
- [ ] N/A

#### Requirements
- [x] N/A

**REQ IDs**: N/A

**JIRA TASK**: N/A


dali/python/nvidia/dali/pipeline.py
```diff
@@ -55,6 +57,55 @@ def _get_default_stream_for_array(array):
     else:
         return None
 
+
+def _prep_data_for_feed_input(data, batch_size, layout):
```
ksztenderski, line 61: As I am using the same implementation in debug mode so I separated it from the feed_input.

dali/python/nvidia/dali/pipeline.py
```diff
@@ -711,67 +762,21 @@ def build(self, define_graph = None):
...
         self._built = True
 
-    def _feed_input(self, name, data, layout = None, cuda_stream = None, use_copy_kernel = False):
-        from nvidia.dali.external_source import _check_data_batch
-
-        infer_stream = False
+    def _feed_input(self, name, data, layout=None, cuda_stream=None, use_copy_kernel=False):
         if cuda_stream is None:
-            infer_stream = True
+            cuda_stream = _get_default_stream_for_array(data)
```
ksztenderski, line 767: I noticed that before when data was a list we called `_get_default_stream_for_array` multiple times. Was there any particular reason for that?

dali/python/backend_impl.cc
```diff
@@ -870,6 +896,19 @@ void ExposeTensorList(py::module &m) {
         }),
       "tl"_a,
       "layout"_a = py::none())
+    .def(py::init([](py::list &list_of_tensors, string layout = "") {
```
ksztenderski, line 899: While testing this constructor I experienced that when it was declared after the one from `py::object` the later was called when constructing TL with `TensorListGPU([TensorGPU])` and when I moved it up the problem was resolved. Is this ambiguity something we should be concerned about, especially considering that there is already a similar ambiguity with the `py::capsule` constructor?

ksztenderski:
!build

ksztenderski:
!build

dali/python/backend_impl.cc
```diff
@@ -585,6 +585,20 @@ std::unique_ptr<Tensor<Backend> > TensorListGetItemImpl(TensorList<Backend> &t,
...
+template <typename Backend>
+std::shared_ptr<TensorList<Backend>> TensorListFromListOfTensors(py::list &list_of_tensors,
+                                                                 string &layout) {
+  auto tl = std::make_shared<TensorList<Backend>>(list_of_tensors.size());
+  TensorVector<Backend> tv(list_of_tensors.size());
+  for (size_t i = 0; i < list_of_tensors.size(); ++i) {
+    auto &t = list_of_tensors[i].cast<Tensor<Backend>&>();
+    tv[i] = std::move(t);
+  }
+  tl->Copy(tv, 0);
```
klecki, line 597: I think we probably should have a sync. We might do a Copy on some other stream (we have this stream pool that creates stream per-device), and sync on that stream, as syncing stream 0 might be a bit over the top.

dali/python/backend_impl.cc
klecki: It's how pybind operates, it will try to match the constructors in succession, so the narrowest-matching ones should be higher than the more generic ones.

dali/test/python/test_backend_impl.py
```diff
@@ -142,3 +142,18 @@ def check_squeeze(shape, dim, in_layout, expected_out_layout):
              (0, (1, 5, 1), "ABC", "BC"),
              (None, (3, 5, 1), "ABC", "AB")]:
         yield check_squeeze, shape, dim, in_layout, expected_out_layout
+
+
+def test_tensorlist_constructor_from_list_of_tensors():
```
klecki, line 147: It might be nice to test non-uniformly shaped TensorList as well (where each tensor has different dimensions).

klecki (commented):
Comments only to TensorList constructor, I didn't read the rest yet.

dali/python/backend_impl.cc
ksztenderski: done

dali/test/python/test_backend_impl.py
ksztenderski: done

awolant (approved):

dali/python/nvidia/dali/pipeline.py
klecki, line 61: Just a nitpick, we can probably move it to external_source.py

dali/python/backend_impl.cc
```diff
@@ -594,6 +594,29 @@ std::unique_ptr<Tensor<Backend> > TensorListGetItemImpl(TensorList<Backend> &t,
...
 }
 
+template <typename Backend>
+std::shared_ptr<TensorList<Backend>> TensorListFromListOfTensors(py::list &list_of_tensors,
+                                                                 string &layout) {
+  auto tl = std::make_shared<TensorList<Backend>>(list_of_tensors.size());
+  TensorVector<Backend> tv(list_of_tensors.size());
+  for (size_t i = 0; i < list_of_tensors.size(); ++i) {
+    auto &t = list_of_tensors[i].cast<Tensor<Backend>&>();
+    tv[i] = std::move(t);
```
klecki, line 604: Doesn't this move the data out of the list? Won't this be destructive function?

dali/python/backend_impl.cc
```diff
@@ -594,6 +594,29 @@ std::unique_ptr<Tensor<Backend> > TensorListGetItemImpl(TensorList<Backend> &t,
...
+  if (!list_of_tensors.empty() && std::is_same<Backend, GPUBackend>::value) {
+    auto &t = list_of_tensors[0].cast<Tensor<GPUBackend>&>();
+    stream = UserStream::Get()->GetStream(t);
+  }
+
+  tl->Copy(tv, stream);
+  tl->SetLayout(layout);
+  CUDA_CALL(cudaStreamSynchronize(stream));
+
+  return tl;
```
klecki, line 617: Now that I'm thinking about it, I think other constructors just share the provided data with a special deleter to handle the Python refcount.

dali/python/nvidia/dali/_debug_mode.py
```diff
@@ -114,6 +118,110 @@ def __rxor__(self, other):
...
+    def _fetch(self, epoch_idx):
+        """Fetches data from callback or provided with feed_input."""
+
+        if self._callback is not None:
+            callback_out = self._get_batch(epoch_idx)
+            layout = self._layout
+            if self._num_outputs is not None:
+                for idx in range(self._num_outputs):
+                    if self._batch:
+                        raw_data = callback_out[idx]
```
klecki, line 210: Is this correct? We just save last output as `raw_data` here?

dali/python/nvidia/dali/_debug_mode.py
```diff
@@ -114,6 +118,110 @@ def __rxor__(self, other):
...
+                    else:
+                        raw_data = [callback_out[i][idx] for i in range(self._batch_size)]
```
klecki, line 212: Same here.

dali/python/nvidia/dali/_debug_mode.py
```diff
@@ -114,6 +118,110 @@ def __rxor__(self, other):
         return _PipelineDebug.current()._wrap_op_call(_arithm_op, ["bitxor", other, self], {})
 
 
+def _transform_data_to_tensorlist(data, batch_size, layout=None):
+    data = _pipeline._prep_data_for_feed_input(data, batch_size, layout)
```
klecki, line 122: Just checking, thanks to that the data is transformed to DALI types already, so we won't get here numpy for example?

dali/python/nvidia/dali/_debug_mode.py
ksztenderski: It was a mistake, fixed it.

dali/python/nvidia/dali/_debug_mode.py
ksztenderski: done

dali/python/nvidia/dali/pipeline.py
ksztenderski: done

dali/python/nvidia/dali/_debug_mode.py
ksztenderski: The idea here is to transform any data to TensorList (that includes [Tensor] -> TensorList). I've made it a separate function for the future implementation of the operator calls in the debug mode, where we would just use this function instead of an external source operator (supporting the debug mode feature of calling operators on data convertible to TensorList).

dali/python/backend_impl.cc
ksztenderski: changed to ShareData

ksztenderski:
!build

PR #3615, opened 2022-01-12
Frames decoder gpu

awolant:
## Description

### What happened in this PR

Adds new class `FramesDecoderGpu` capable of decoding video files using NVIDIA Video Codec SDK.

This class derives from `FramesDecoder` that uses FFMPEG to do the same thing.

### Tests
- [ ] Existing tests apply
- [x] New tests added
  - [ ] Python tests
  - [x] GTests
  - [ ] Benchmark
  - [ ] Other
- [ ] N/A

### Documentation
- [ ] Documentation updated
  - [ ] Docstring
  - [ ] Doxygen
  - [ ] RST
  - [ ] Jupyter
  - [ ] Other
- [ ] N/A

#### Requirements
- [ ] N/A

**REQ IDs**: N/A

**JIRA TASK**: 2252


awolant:
!build

awolant:
!build

dali/operators/reader/loader/video/nvdecode/ColorSpace.cu
```diff
@@ -0,0 +1,83 @@
...
+
+
+__constant__ float matYuv2Rgb[3][3] = {
+    1.164383f,  0.0f,       1.596027f,
+    1.164383f, -0.391762f, -0.812968f,
+    1.164383f,  2.017232f,  0.0f
+};
+
+__device__ static uint8_t Clamp(float x, float lower, float upper) {
+    return x < lower ? lower : (x > upper ? upper : x);
```
JanuszL, line 26: Can you check imgproc.cu?
Should `return fminf(fmaxf(x, 0.0f), max);` be faster?

dali/operators/reader/loader/video/nvdecode/ColorSpace.cu
```diff
@@ -0,0 +1,83 @@
...
+    float fy = (int)y - low, fu = (int)u - mid, fv = (int)v - mid;
+    const float maxf = (1 << sizeof(uint8_t) * 8) - 1.0f;
+
+    return Rgb { 
+        Clamp(matYuv2Rgb[0][0] * fy + matYuv2Rgb[0][1] * fu + matYuv2Rgb[0][2] * fv, 0.0f, maxf),
+        Clamp(matYuv2Rgb[1][0] * fy + matYuv2Rgb[1][1] * fu + matYuv2Rgb[1][2] * fv, 0.0f, maxf),
+        Clamp(matYuv2Rgb[2][0] * fy + matYuv2Rgb[2][1] * fu + matYuv2Rgb[2][2] * fv, 0.0f, maxf)};
+}
+
+__global__ static void YuvToRgbKernel(uint8_t *pYuv, int nYuvPitch, uint8_t *pRgb, int nRgbPitch, int nWidth, int nHeight) {
```
JanuszL, line 42: I wonder if this can be moved to color_space_conversion_impl.h.

dali/operators/reader/loader/video/frames_decoder_gpu.cc
```diff
@@ -0,0 +1,190 @@
...
+    decoder_info.ulNumDecodeSurfaces = 4;
+    decoder_info.ulNumOutputSurfaces = 1;
+
+    CUDA_CALL(cuvidCreateDecoder(&nvdecode_state_->decoder, &decoder_info));
+
+    // Create nv parser
+    CUVIDPARSERPARAMS parser_info;
+    memset(&parser_info, 0, sizeof(CUVIDPARSERPARAMS));
+    parser_info.CodecType = cudaVideoCodec_MPEG4;
+    parser_info.ulMaxNumDecodeSurfaces = 1;
```
JanuszL, line 112: I guess this is a ToDo, but with the current approach would increasing that number change the perf (NVDEC can decode next frame while we handle callback on the CPU side to receive the current one)?

dali/operators/reader/loader/video/frames_decoder_gpu.cc
```diff
@@ -0,0 +1,190 @@
...
+namespace detail {
+int process_video_sequence(void *user_data, CUVIDEOFORMAT *video_format) {
+  return video_format->min_num_decode_surfaces;
+}
+
+int process_picture_decode(void *user_data, CUVIDPICPARAMS *picture_params) {
+  FramesDecoderGpu *frames_decoder = static_cast<FramesDecoderGpu*>(user_data);
+  if (frames_decoder->flush_) {
+    frames_decoder->flush_ = false;
+    frames_decoder->decode_success_ = false;
```
JanuszL, line 39: Nitpick - I guess you can move this to L36 and remove L43.

dali/operators/reader/loader/video/frames_decoder_gpu.cc
```diff
@@ -0,0 +1,190 @@
...
+
+  // Copy decoded frame to output
+  CUVIDPROCPARAMS videoProcessingParameters = {};
+  videoProcessingParameters.progressive_frame = !picture_params->field_pic_flag;
+  videoProcessingParameters.second_field = 1;
+  videoProcessingParameters.top_field_first = picture_params->bottom_field_flag ^ 1;
+  videoProcessingParameters.unpaired_field = 0;
+  videoProcessingParameters.output_stream = 0;
+
+  CUdeviceptr frame = 0;
```
JanuszL, line 55:
```suggestion
  CUdeviceptr frame = {};
```

dali/operators/reader/loader/video/nvdecode/ColorSpace.cu
```diff
@@ -0,0 +1,83 @@
...
+    pDst2[1] = rgb2.g;
+    pDst2[2] = rgb2.b;
+
+    pDst2[3] = rgb2.r;
+    pDst2[4] = rgb2.g;
+    pDst2[5] = rgb2.b;
+
+}
+
+void Nv12ToColor32(uint8_t *dpNv12, int nNv12Pitch, uint8_t *dpBgra, int nBgraPitch, int nWidth, int nHeight, int iMatrix) {
```
JanuszL, line 79: I guess you need to accept stream here as well.

dali/operators/reader/loader/video/frames_decoder_gpu.cc
```diff
@@ -0,0 +1,190 @@
...
+      frames_decoder->Height(),
+      1);
+    CUDA_CALL(cudaDeviceSynchronize());
+  }
+
+  CUDA_CALL(cuStreamSynchronize(0));
+  CUDA_CALL(cuvidUnmapVideoFrame(frames_decoder->nvdecode_state_->decoder, frame));
+
+  frames_decoder->decode_success_ = true;
+  return 1;
```
JanuszL, line 82: Can you use the right enum here?

dali/operators/reader/loader/video/frames_decoder_gpu.cc
```diff
@@ -0,0 +1,190 @@
...
+int process_video_sequence(void *user_data, CUVIDEOFORMAT *video_format) {
+  return video_format->min_num_decode_surfaces;
+}
+
+int process_picture_decode(void *user_data, CUVIDPICPARAMS *picture_params) {
+  FramesDecoderGpu *frames_decoder = static_cast<FramesDecoderGpu*>(user_data);
+  if (frames_decoder->flush_) {
+    frames_decoder->flush_ = false;
+    frames_decoder->decode_success_ = false;
+    return 0;
```
JanuszL, line 40: Can you use the right enum here?

dali/operators/reader/loader/video/frames_decoder_gpu.cc
```diff
@@ -0,0 +1,190 @@
...
+    decoder_info.ulHeight = Height();
+    decoder_info.ulWidth = Width();
+    decoder_info.ulMaxHeight = Height();
+    decoder_info.ulMaxWidth = Width();
+    decoder_info.ulTargetHeight = Height();
+    decoder_info.ulTargetWidth = Width();
+    decoder_info.ulNumDecodeSurfaces = 4;
+    decoder_info.ulNumOutputSurfaces = 1;
+
+    CUDA_CALL(cuvidCreateDecoder(&nvdecode_state_->decoder, &decoder_info));
```
JanuszL, line 106: How expensive is that? What if we need to handle 100 000 videos, would that eat up whole RAM/GPU memory?

dali/operators/reader/loader/video/frames_decoder_gpu.cc
```diff
@@ -0,0 +1,190 @@
...
+
+    CUDA_CALL(cuvidParseVideoData(nvdecode_state_->parser, packet));
+
+    if (decode_success_) {
+      return true;
+    }
+  }
+
+  if (!last_frame_read_) {
+    CUVIDSOURCEDATAPACKET *packet = &nvdecode_state_->packet;
```
JanuszL, line 161: I guess this could be a function as can be called in L147 as well.

dali/operators/reader/loader/video/frames_decoder_gpu.cc
```diff
@@ -0,0 +1,190 @@
...
+}
+
+void FramesDecoderGpu::Reset() {
+  flush_ = true;
+  CUVIDSOURCEDATAPACKET *packet = &nvdecode_state_->packet;
+  memset(packet, 0, sizeof(CUVIDSOURCEDATAPACKET));
+  packet->payload = nullptr;
+  packet->payload_size = 0;
+  packet->flags = CUVID_PKT_ENDOFSTREAM;
+  CUDA_CALL(cuvidParseVideoData(nvdecode_state_->parser, packet));
```
JanuszL, line 184: This could be a function as well and reused in L123.

dali/operators/reader/loader/video/frames_decoder_test.cc
```diff
@@ -103,6 +107,140 @@ TEST_F(FramesDecoderTest, VariableFrameRate) {
     this->CompareFrames(frame.data(), this->GetVfrFrame(1, 0), file.FrameSize());
 }
 
+TEST_F(FramesDecoderTest, VariableFrameRateGpu) {
+    ASSERT_TRUE(cuInitChecked());
+    CUdevice device = { 0 };
+    CUDA_CALL(cuDeviceGet(&device, 0));
+    CUcontext context = { 0 };
+    CUDA_CALL(cuCtxCreate(&context, 0, device));
```
JanuszL, line 115: I assume you want to test on the first device not the one that the context was created for before calling that test?
```suggestion
    DeviceGuard(0);
```

dali/operators/reader/loader/video/frames_decoder_test.cc
```diff
@@ -103,6 +107,140 @@ TEST_F(FramesDecoderTest, VariableFrameRate) {
...
+    ASSERT_EQ(file.Height(), 600);
+    ASSERT_EQ(file.Width(), 800);
+    ASSERT_EQ(file.Channels(), 3);
+    ASSERT_EQ(file.NumFrames(), 60);
+
+    uint8_t *frame = nullptr;
+    cuMemAlloc(
+        reinterpret_cast<CUdeviceptr*>(&frame),
+        file.FrameSize());
+    std::vector<uint8_t> frame_cpu(file.FrameSize());
```
JanuszL, line 132: Why not use `DeviceBuffer` and just `resize` and `from_device`?

awolant:
!build

awolant:
!build

awolant:
!build

dali/operators/reader/loader/video/frames_decoder_gpu.cc
awolant: The only piece docs on this I could find says:
```
/* Return value from HandlePictureDecode() are interpreted as:
 *   0: fail
 *   >=1: suceeded
 */
```

dali/operators/reader/loader/video/frames_decoder_gpu.cc
awolant: Same as above

dali/operators/reader/loader/video/frames_decoder_test.cc
awolant: For this test context was not created before running it, so I did this manually. 

Maybe we can make some small utility in a separate PR and refactor it?

dali/operators/reader/loader/video/nvdecode/ColorSpace.cu
awolant: Not sure, but definitely it is cooler.

Done

dali/operators/reader/loader/video/nvdecode/ColorSpace.cu
awolant: I made this file for readable for now.

In the future, we need to extend this simple code to support more pixel formats. Then I think will be the right time to extract impl. Is that ok?

dali/operators/reader/loader/video/frames_decoder_gpu.cc
awolant: As far as I understood the available information, this parameter in the first place is about correct decoding. For now, I've set it to 5 which is enough to correctly decode formats we support in this code.

When we have proper benchmark, we can experiment with perf here. Increasing it will require more memory both for NVDECODE internals, but also for our buffer, as we will need to buffer more frames.

dali/operators/reader/loader/video/frames_decoder_gpu.cc
awolant: Done

dali/operators/reader/loader/video/frames_decoder_gpu.cc
awolant: Done

dali/operators/reader/loader/video/nvdecode/ColorSpace.cu
awolant: Done

dali/operators/reader/loader/video/frames_decoder_gpu.cc
awolant: The architectural idea for this is to have one `FramesDecoderGpu` for multiple video files and call `Reconfigure`, when we change it.

dali/operators/reader/loader/video/frames_decoder_gpu.cc
awolant: Done

dali/operators/reader/loader/video/frames_decoder_gpu.cc
awolant: Done

dali/operators/reader/loader/video/frames_decoder_test.cc
awolant: Done

awolant:
!build

dali/operators/reader/loader/video/frames_decoder_gpu.cc
JanuszL: Understood.

dali/operators/reader/loader/video/frames_decoder_test.cc
JanuszL: `DeviceGuard` does that. It saves the current context (to improve the interoperability with libraries using driver API instead of CUDA API), and calls `cudaSetDevice` which creates the context. So unless you need to access this context directly `DeviceGuard` should do.
I suspect that `cuDeviceGet` may also create the context, but I'm not sure.

dali/operators/reader/loader/video/frames_decoder.cc
```diff
@@ -169,11 +169,13 @@ bool FramesDecoder::ReadRegularFrame(uint8_t *data, bool copy_to_output) {
     }
 
     if (!copy_to_output) {
+      ++current_frame_;
```
JanuszL, line 172: In case of `if (ret == AVERROR_EOF) {` maybe it should set `current_frame_ = -1`?

dali/operators/reader/loader/video/frames_decoder.h
```diff
@@ -124,20 +124,38 @@ class DLL_PUBLIC FramesDecoder {
...
    */
-  void Reset();
+  virtual void Reset();
+
+  /**
+   * @brief Returns index of the frame that will be returned by the next call of ReadNextFrame
+   * 
+   * @return int Index of the next frame to be read
+   */
+  int CurrentFrame() { return current_frame_; }
```
JanuszL, line 147: Maybe it should be `NextFrameIdx`?

dali/operators/reader/loader/video/frames_decoder_gpu.h
```diff
@@ -0,0 +1,88 @@
...
+  DeviceBuffer<uint8_t> frame_;
+  int pts_;
+};
+
+class DLL_PUBLIC FramesDecoderGpu : public FramesDecoder {
+ public:
+  /**
+   * @brief Construct a new FramesDecoder object.
+   * 
+   * @param filename Path to a video file.
```
JanuszL, line 50:
```suggestion
   * @param filename Path to a video file.
   * @param stream Stream used for decode processing.
```

dali/operators/reader/loader/video/frames_decoder_gpu.h
```diff
@@ -0,0 +1,88 @@
...
+
+  bool ReadNextFrame(uint8_t *data, bool copy_to_output = true) override;
+
+  void SeekFrame(int frame_id) override;
+
+  void Reset() override;
+
+  int CurrentFramePts() { return index_[CurrentFrame()].pts; }
+
+  std::unique_ptr<NvDecodeState> nvdecode_state_;
```
JanuszL, line 62:
```suggestion
private:
  std::unique_ptr<NvDecodeState> nvdecode_state_;
```

dali/operators/reader/loader/video/frames_decoder_test.cc
```diff
@@ -78,31 +89,100 @@ TEST_F(FramesDecoderTest, VariableFrameRate) {
...
+    file.SeekFrame(25);
+    file.ReadNextFrame(frame);
+    copyD2H(frame_cpu.data(), frame.data(), frame.size());
+    this->CompareFramesAvgError(frame_cpu.data(), this->GetVfrFrame(1, 25), file.FrameSize());
+
+    // Seek back to frame
+    file.SeekFrame(12);
+    file.ReadNextFrame(frame);
+    copyD2H(frame_cpu.data(), frame.data(), frame.size());
+    this->CompareFramesAvgError(frame_cpu.data(), this->GetVfrFrame(1, 12), file.FrameSize());
```
JanuszL, line 168: Maybe you can seek to randomly selected frames from 1-58 range?

dali/operators/reader/loader/video/frames_decoder_gpu.cc
```diff
@@ -0,0 +1,261 @@
...
+  videoProcessingParameters.unpaired_field = 0;
+  videoProcessingParameters.output_stream = frames_decoder->stream_;
+
+  uint8_t *frame_output = nullptr;
+
+  // Take pts of the currently decoded frame
+  int current_pts = frames_decoder->piped_pts_.front();
+  frames_decoder->piped_pts_.pop();
+
+  if (current_pts == frames_decoder->CurrentFramePts()) {
```
JanuszL, line 60: As I understand `current_pts` is the result of the order how frames are extracted by libavutil and passed to the decoder, while `CurrentFramePts` is the expected order of the decoding (1, 2, 3 ...). Is that correct?
Can you add a comment here to make it cleared?

dali/operators/reader/loader/video/frames_decoder_gpu.cc
```diff
@@ -0,0 +1,261 @@
...
+    DALI_ENFORCE(avcodec_parameters_copy(
+      bsfc_->par_in, av_state_->ctx_->streams[0]->codecpar) >= 0);
+    DALI_ENFORCE(av_bsf_init(bsfc_) >= 0);
+
+    filtered_packet_ = av_packet_alloc();
+    DALI_ENFORCE(filtered_packet_, "Could not allocate av packet");
+
+    // Create nv decoder
+    CUVIDDECODECREATEINFO decoder_info = {};
+    memset(&decoder_info, 0, sizeof(CUVIDDECODECREATEINFO));
```
JanuszL, line 117: I'm not sure if you need `decoder_info = {}` when you call memset later.

dali/operators/reader/loader/video/frames_decoder_gpu.cc
```diff
@@ -0,0 +1,261 @@
...
+
+    decoder_info.bitDepthMinus8 = 0;
+    decoder_info.ChromaFormat = cudaVideoChromaFormat_420;
+    decoder_info.CodecType = cudaVideoCodec_H264;
```
JanuszL, line 121: You should not hardcode that...

dali/operators/reader/loader/video/frames_decoder_gpu.cc
```diff
@@ -0,0 +1,261 @@
...
+    decoder_info.ulTargetWidth = Width();
+    decoder_info.ulNumDecodeSurfaces = num_decode_surfaces_;
+    decoder_info.ulNumOutputSurfaces = 2;
+
+    CUDA_CALL(cuvidCreateDecoder(&nvdecode_state_->decoder_, &decoder_info));
+
+    // Create nv parser
+    CUVIDPARSERPARAMS parser_info;
+    memset(&parser_info, 0, sizeof(CUVIDPARSERPARAMS));
+    parser_info.CodecType = cudaVideoCodec_H264;
```
JanuszL, line 136: As above.

dali/operators/reader/loader/video/frames_decoder_gpu.h
```diff
@@ -0,0 +1,88 @@
...
+  uint8_t *current_frame_output_ = nullptr;
+  bool current_copy_to_output_ = false;
+  bool frame_returned_ = false;
+  bool flush_ = false;
+  bool last_frame_read_ = false;
+
+  AVBSFContext *bsfc_ = nullptr;
+  AVPacket *filtered_packet_ = nullptr;
+
+  const int num_decode_surfaces_ = 5;
```
JanuszL, line 73: Shouldn't be that stream/coded dependent?

dali/operators/reader/loader/video/frames_decoder_gpu.cc
```diff
@@ -0,0 +1,261 @@
...
+    parser_info.pfnSequenceCallback = detail::process_video_sequence;
+    parser_info.pfnDecodePicture = detail::process_picture_decode;
+    parser_info.pfnDisplayPicture = nullptr;
+
+    CUDA_CALL(cuvidCreateVideoParser(&nvdecode_state_->parser_, &parser_info));
+
+    // Init internal frame buffer
+    for (size_t i = 0; i < frame_buffer_.size(); ++i) {
+      frame_buffer_[i].frame_.resize(FrameSize());
+      frame_buffer_[i].pts_ = -1;
```
JanuszL, line 149: Maybe we should have one contiguous buffer instead of set of separate one. On the other hand we have pretty good allocator now so maybe it will just overcomplicate things.

dali/operators/reader/loader/video/frames_decoder_gpu.h
```diff
@@ -0,0 +1,88 @@
...
+
+  void SeekFrame(int frame_id) override;
+
+  void Reset() override;
+
+  int CurrentFramePts() { return index_[CurrentFrame()].pts; }
+
+  std::unique_ptr<NvDecodeState> nvdecode_state_;
+
+  uint8_t *current_frame_output_ = nullptr;
```
JanuszL, line 64: So this variable is used to extract decoded frame from the callback, right?

dali/operators/reader/loader/video/frames_decoder_gpu.h
```diff
@@ -0,0 +1,88 @@
...
+
+#include "dali/operators/reader/loader/video/nvdecode/cuviddec.h"
+#include "dali/operators/reader/loader/video/nvdecode/nvcuvid.h"
+
+#include "dali/core/dev_buffer.h"
+
+namespace dali {
+struct NvDecodeState {
+  CUvideodecoder decoder_;
+  CUvideoparser parser_;
```
JanuszL, line 33:
```suggestion
  CUvideodecoder decoder;
  CUvideoparser parser;
```
I think the idea is that struct have all fields public so we don't need to use naming as for internal fields.

dali/operators/reader/loader/video/frames_decoder_gpu.cc
```diff
@@ -0,0 +1,261 @@
...
+    frame_returned_ = false;
+    CUDA_CALL(cuvidParseVideoData(nvdecode_state_->parser_, packet));
+
+    if (frame_returned_) {
+      ++current_frame_;
+      return true;
+    }
+  }
+
+  if (!last_frame_read_) {
```
JanuszL, line 214: I don't fully understand that. `last_frame_read_` is always false. When do you set it to `true`?

dali/operators/reader/loader/video/frames_decoder_test.cc
awolant: This is a rare corner case where `DeviceGuard` on its own does not work, unfortunately.
That is because `cudaSetDevice` does not create the context. It sets primary device context as the current context. Primary context is created upon first call to any CUDA Runtime function. 
Here, the first call we make is to NVDECODE API and it fails due to lack of context.
That is why I needed to create it manually. 

I changed it to:
```
DeviceGuard(0);
CUDA_CALL(cudaDeviceSynchronize());
```
This way `cudaDeviceSynchronize` will create the context and the whole thing works.

dali/operators/reader/loader/video/frames_decoder.cc
awolant: Not really. When this condition triggers it means that we no longer have packages to read. But we still might have frames in internal FFMPEG buffers.
We set it when we have flushed all frames from FFMPEG internals.

dali/operators/reader/loader/video/frames_decoder.h
awolant: Done

dali/operators/reader/loader/video/frames_decoder_gpu.h
awolant: Done

dali/operators/reader/loader/video/frames_decoder_gpu.h
awolant: Can't do it, because I need to access it from the callback.

Alternative would be to make the callback to call a public method that accesses it but it seemed a bit convoluted. `FramesDecoderGpu` is not supposed to be used by the outside users, so I think maybe it is fine for now.

dali/operators/reader/loader/video/frames_decoder_test.cc
awolant: Done

dali/operators/reader/loader/video/frames_decoder_test.cc
```diff
@@ -78,31 +89,100 @@ TEST_F(FramesDecoderTest, VariableFrameRate) {
...
+    frame.resize(file.FrameSize());
+    std::vector<uint8_t> frame_cpu(file.FrameSize());
+
+    for (int i = 0; i < 60; ++i) {
+        ASSERT_EQ(file.CurrentFrame(), i);
+        file.ReadNextFrame(frame.data());
+        copyD2H(frame_cpu.data(), frame.data(), frame.size());
+        this->CompareFramesAvgError(frame_cpu.data(), this->GetVfrFrame(1, i), file.FrameSize());
+    }
+    ASSERT_EQ(file.CurrentFrame(), -1);
```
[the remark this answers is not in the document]
awolant: Done

dali/operators/reader/loader/video/frames_decoder_gpu.cc
awolant: Done

dali/operators/reader/loader/video/frames_decoder_gpu.cc
awolant: Done

dali/operators/reader/loader/video/frames_decoder_gpu.cc
awolant: The plan is that this PR will support h264 only. I'll add more formats support in a separate PR. I did not want to do it here, because it requires some technical code to get the capabilities of the decoder, get the codec from file and see, if they match. 
Also, we need to add them to DALI_extra to make proper tests, not only see, if it runs.
I did not want this to obfuscate the important thing here, which is how we communicate between FFMPEG, NVDECODE and DALI.

dali/operators/reader/loader/video/frames_decoder_gpu.cc
awolant: As above.

dali/operators/reader/loader/video/frames_decoder_gpu.h
awolant: Yes, as above.

dali/operators/reader/loader/video/frames_decoder_gpu.cc
awolant: Added comment to benchmark it, when we have proper benchmarks. It should be easy to test then.

dali/operators/reader/loader/video/frames_decoder_gpu.h
awolant: Yes

dali/operators/reader/loader/video/frames_decoder_gpu.h
awolant: Done

dali/operators/reader/loader/video/frames_decoder_gpu.cc
awolant: Leftover from before. Removed it, since now the same information is passed by `next_frame_idx`.

dali/operators/reader/loader/video/nvdecode/ColorSpace.cu
JanuszL: Sure

dali/operators/reader/loader/video/frames_decoder_gpu.cc
JanuszL: Can you add a ToDo comment here in such case? It would be easier to track places that require changes later.

dali/operators/reader/loader/video/frames_decoder_gpu.cc
JanuszL: Can you add a ToDo comment here in such case? It would be easier to track places that require changes later.

dali/operators/reader/loader/video/frames_decoder_gpu.h
JanuszL: How about making the callback to call the public method that corresponds to the callback?:
```
int process_picture_decode(void *user_data, CUVIDPICPARAMS *picture_params) {
  FramesDecoderGpu *frames_decoder = static_cast<FramesDecoderGpu*>(user_data);
  return frames_decoder->process_picture_decode_impl(picture_params);
}
```
I would still opt for making the data private.

dali/operators/reader/loader/video/frames_decoder_gpu.cc
JanuszL: Understood.

dali/operators/reader/loader/video/frames_decoder_gpu.cc
```diff
@@ -0,0 +1,262 @@
...
+  videoProcessingParameters.output_stream = frames_decoder->stream_;
+
+  uint8_t *frame_output = nullptr;
+
+  // Take pts of the currently decoded frame
+  int current_pts = frames_decoder->piped_pts_.front();
+  frames_decoder->piped_pts_.pop();
+
+  // current_pts is pts of frame that came from the decoder
+  // CurrentFramePts() is pts of the frame that we want to return
```
JanuszL, line 61:
```suggestion
  // NextFramePts() is pts of the frame that we want to return
```

dali/operators/reader/loader/video/frames_decoder_test.cc
JanuszL: I was thinking to replace `file.SeekFrame(25);` and `file.SeekFrame(12);` with that but that is fine.

JanuszL (approved):

dali/operators/reader/loader/video/frames_decoder_gpu.cc
```diff
@@ -0,0 +1,262 @@
...
+  // in this call to ReadNextFrame
+  // If they are the same, we just return this frame
+  // If not, we store it in the buffer for later
+
+  if (current_pts == frames_decoder->NextFramePts()) {
+    // Currently decoded frame is actually the one we wanted
+    frames_decoder->frame_returned_ = true;
+    if (frames_decoder->current_copy_to_output_ == false) {
+      return 1;
+  }
```
stiepan, line 71: nitpick: formatting - should be indented?

dali/operators/reader/loader/video/frames_decoder_gpu.cc
```diff
@@ -0,0 +1,262 @@
...
+
+    // Prepare nv packet
+    CUVIDSOURCEDATAPACKET *packet = &nvdecode_state_->packet;
+    memset(packet, 0, sizeof(CUVIDSOURCEDATAPACKET));
+    packet->payload = filtered_packet_->data;
+    packet->payload_size = filtered_packet_->size;
+    packet->flags = CUVID_PKT_TIMESTAMP;
+    packet->timestamp = filtered_packet_->pts;
+
+    // Send packet to the nv deocder
```
stiepan, line 210: nitpick: decoder

dali/operators/reader/loader/video/frames_decoder_gpu.cc
```diff
@@ -0,0 +1,262 @@
...
+    CUDA_CALL(cuvidParseVideoData(nvdecode_state_->parser, packet));
+
+    if (frame_returned_) {
+      ++next_frame_idx_;
+      return true;
+    }
+  }
+
+  SendLastPacket();
+  next_frame_idx_ = -1;
```
stiepan, line 221: Isn't it possible that `SendLastPacket(false)` would invoke the `process_picture_decode` callaback (due to cuvidParseVideoData empty packet call) multiple-times, populate the `frame_buffer_` and the 

```
if (next_frame_idx_ == -1) {
    return false;
  }
```
at the beginning of `ReadNextFrame` would prevent user from accessing the remaining frames?

dali/operators/reader/loader/video/frames_decoder_gpu.cc
```diff
@@ -0,0 +1,262 @@
...
+    CUVIDSOURCEDATAPACKET *packet = &nvdecode_state_->packet;
+    memset(packet, 0, sizeof(CUVIDSOURCEDATAPACKET));
+    packet->payload = filtered_packet_->data;
+    packet->payload_size = filtered_packet_->size;
+    packet->flags = CUVID_PKT_TIMESTAMP;
+    packet->timestamp = filtered_packet_->pts;
+
+    // Send packet to the nv deocder
+    frame_returned_ = false;
+    CUDA_CALL(cuvidParseVideoData(nvdecode_state_->parser, packet));
```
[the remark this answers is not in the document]
awolant: Yes

dali/operators/reader/loader/video/frames_decoder_gpu.cc
awolant: Done

dali/operators/reader/loader/video/frames_decoder_gpu.h
awolant: Done

awolant:
!build

dali/operators/reader/loader/video/frames_decoder_gpu.cc
stiepan: Okay, thanks! I was a bit confused how reading remaining frames compares to CPU decoder. So just for the record, please consider adding a comment (or an assert as you suggested) about exactly one remaining frame expected here.

stiepan (approved):

dali/operators/reader/loader/video/frames_decoder_gpu.cc
awolant: Added assert to make sure this is the case always.

awolant:
!build

awolant:
!build

PR #3618, opened 2022-01-13
Add Laplacian GPU kernel

stiepan:
Add Laplacian GPU kernel.

## Description

### What happened in this PR
* GPU Laplacian kernel is added along with cpp tests against CPU version as a baseline
* GPU convolution scale parameter is replaced with ConvEpilogue parameter to enable per sample scale and accumulation of result in the output buffer.
* Moved cuda cast kernel out from cast operator to separate file in kernels dir.

* Convolution GPU
* Cast operator

### Tests
- [ ] Existing tests apply
- [x] New tests added
  - [ ] Python tests
  - [x] GTests
  - [ ] Benchmark
  - [ ] Other
- [ ] N/A

### Documentation
- [x] Documentation updated
  - [ ] Docstring
  - [x] Doxygen
  - [ ] RST
  - [ ] Jupyter
  - [ ] Other
- [ ] N/A

#### Requirements
- [ ] N/A

**REQ IDs**: LAPL.01 - LAPL.17

**JIRA TASK**: DALI-2440


stiepan:
!build

klecki (commented):
Some comments.

dali/kernels/imgproc/convolution/convolution_gpu_test.cu
```diff
@@ -31,29 +31,94 @@
...
+  Scale05,
+  AccScaleOutput
+};
+
+template <int transform_case, typename OutType, typename WinType, int ndim>
+struct TransformParam;
+
+template <typename OutType, typename WinType, int ndim>
+struct TransformParam<Scale05, OutType, WinType, ndim> {
+  using TransformCpu = conv_transform::TransScaleSat<OutType, WinType>;
```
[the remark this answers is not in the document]
stiepan: `ConvEpilogue` matches what you could achieve with `TransScaleSat` and `TransScaleAddOutSat`, but does not allow specifying another accumulator (matrix C in cutlass gemm terms) - technically we could add that, but it still would be limited compared to its cpu counterpart, as the C and D have to be of the same type. This ConvEpilogue thingy is kind of just a package for arguments used to configure cutlass gemm operation, it does not perform any meaningful job on its own, so I decided not to go overboard with envisioning future uses cases apart from what I needed.

dali/kernels/imgproc/convolution/laplacian_gpu.cuh
```diff
@@ -0,0 +1,338 @@
...
+
+namespace detail {
+
+struct CastSampleDesc {
+  void* output;
+  const void* input;
+};
+
+template <typename Out, typename In>
+__global__ void BatchedCastKernel(const CastSampleDesc* samples,
```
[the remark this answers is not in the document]
stiepan: I did. Ideally, cast operator would have a dali kernel that I could call here as a subkernel exactly the way convolution is called, but cast op has the kernel fused in it, so I decided to just copy the cuda kernel for now. If you think it is good idea I be happy to split the cast operator into op and kernel and reuse kernel in both cases as a follow-up.

dali/kernels/imgproc/convolution/convolution_gpu.h
```diff
@@ -29,6 +30,29 @@
...
+    return scales_.empty() ? 1 : scales_.size();
+  }
+
+  inline float alpha(int sample_idx) const {
+    return scales_.empty() ? scale_ : scales_[sample_idx];
+  }
+
+  inline float beta() const {
+    return beta_;
+  }
```
mzient, line 48: I think all these functions should be marked as DALI_HOST_DEV

dali/kernels/imgproc/convolution/convolution_gpu.h
```diff
@@ -29,6 +30,29 @@
 namespace dali {
 namespace kernels {
 
+struct ConvEpilogue {
+  ConvEpilogue(span<const float> scales, float beta = 0.f)  // NOLINT
+      : scales_{scales}, scale_{}, beta_{beta} {}
+  ConvEpilogue(float scale, float beta = 0.f) : scales_{}, scale_{scale}, beta_{beta} {}  // NOLINT
+
+  inline float num_samples() const {
+    return scales_.empty() ? 1 : scales_.size();
```
mzient, line 39: Why so? I think some special value - or just 0 - would be preferable. Note that you can query for an arbitrary `alpha`  index when scales are empty, but the return value of this function suggests otherwise.

dali/kernels/imgproc/convolution/convolution_gpu.h
```diff
@@ -88,15 +115,21 @@ struct ConvolutionGpu {
...
     DALI_ENFORCE(
         window_anchors.size() == num_samples || window_anchors.size() == 0,
         make_string(
             "Unexpected number of window_anchors, expected either anchors for all samples ( ",
             num_samples,
             ") or no anchors for windows centered by default, got: ", window_anchors.size(), "."));
+
+    DALI_ENFORCE(num_scales == 1 || num_scales == num_samples,
+                 make_string("Scale argument must be either 1 or equal to batch size (",
+                             num_samples, "), got: ", num_scales, "."));
```
mzient, line 131: This is wrong - if num_scales is 1, but the epilogue was constructed with a 1-element span, this is an error and it will result in an out-of-bounds read when accessing `conv_epilogue.alpha(i)`.

dali/kernels/imgproc/convolution/laplacian_cpu_test.cc
```diff
@@ -36,38 +37,29 @@ inline std::enable_if_t<!std::is_integral<Out>::value> CompareOut(
...
- * to computing binominal coefficients: (1 + 1) ^ (2n). Derivative kernel of the order
- * 2 is [1, -2, 1] (in other words [1, -1] composed with itself) composed with
- * the smoothing window of appropriate size
- */
-template <int window_size>
-std::array<float, window_size> GetSobelWindow(int d_order) {
-  static_assert(window_size >= 3 && window_size % 2 == 1);
-  std::array<float, window_size> w = {0.};
-  w[0] = 1.;
+void FillSobelWindow(span<float> window, int d_order) {
```
mzient, line 40: What happened to the comment?

stiepan: It's in `laplacian_test.h`

dali/kernels/imgproc/convolution/separable_convolution_gpu.h
```diff
@@ -146,14 +149,15 @@ struct SeparableConvolutionGpu<Out, In, W, 3, has_channels, is_sequence> {
   static constexpr int channel_axes = static_cast<int>(has_channels);
   static constexpr int ndim = sequence_axes + axes + channel_axes;
   using Intermediate = decltype(std::declval<W>() * std::declval<In>());
-  static constexpr bool kUseOutAsIntermediate = sizeof(Intermediate) == sizeof(Out);
+  static constexpr bool outFitsIntermediate = sizeof(Intermediate) == sizeof(Out);
```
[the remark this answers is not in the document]
stiepan: Yeah, it should work I guess, but I am just reusing existing code here and `reinterpret` won't work for that, so it would possibly require some extra functionality that goes sample by sample to create new TensorListView. How about making it a follow up small task?

dali/kernels/imgproc/convolution/convolution_gpu.h
stiepan: Does it? Could you explain why?

dali/kernels/imgproc/convolution/convolution_gpu.h
stiepan: Good point. I am not sure if it is really necessary to validate if one passed empty span of floats (as opposed to span being empty because of the scalar constructor used), because it is kind-of internal functionality, so I just default so scalar =1.f in that case.

dali/kernels/imgproc/convolution/laplacian_gpu.cuh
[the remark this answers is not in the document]
mzient: Given how simple it is, there's probably no need to make it a fully-blown DALI kernel (with Setup and an interface with TensorListView) - but just extracting this function to a common header would still be appreciated.

dali/kernels/imgproc/convolution/separable_convolution_gpu.h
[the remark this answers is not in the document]
klecki: Sure

klecki (approved):

dali/kernels/imgproc/convolution/laplacian_gpu.cuh
```diff
@@ -0,0 +1,338 @@
...
+  static constexpr int sequence_axes = MultiDimConv::sequence_axes;
+  using SingleDimConv = ConvolutionGpu<Out, In, W, ndim, sequence_axes + deriv_axis, has_channels>;
+
+  KernelRequirements Setup(KernelContext& ctx, const TensorListShape<ndim>& in_shape,
+                           const std::array<TensorListShape<1>, axes>& window_sizes,
+                           bool has_smoothing) {
+    has_smoothing_ = has_smoothing;
+    if (has_smoothing_) {
+      if (!multi_dim_impl_) {
+        single_dim_impl_.reset();
```
mzient, line 82: I have my doubts about this - should we really free it? Is it possible that, due to varying parameters, `has_smoothing` changes from iteration to iteration? Perhaps `has_smoothing` should be a per-sample parameter and the batch split into smoothed and unsmoothed sub-batches? These are just wild thoughts.

dali/kernels/imgproc/convolution/laplacian_gpu.cuh
```diff
@@ -0,0 +1,338 @@
...
+    }
+    return single_dim_impl_->Setup(ctx, in_shape, window_sizes[deriv_axis]);
+  }
+
+  void Run(KernelContext& ctx, const TensorListView<StorageGPU, Out, ndim> out,
+           const TensorListView<StorageGPU, const In, ndim>& in,
+           const std::array<TensorListView<StorageCPU, const W, 1>, axes>& windows,
+           const ConvEpilogue& conv_epilogue) {
+    if (has_smoothing_) {
+      multi_dim_impl_->Run(ctx, out, in, windows, {}, conv_epilogue);
```
mzient, line 99:
```C++
assert(multi_dim_impl_);
```
would be nice

dali/kernels/imgproc/convolution/laplacian_gpu.cuh
stiepan: It is possible, and in that case it is reset and setup over again. I wasn't sure about that reset, if you think it's best not to have it, I can remove it.

I was thinking about that too. Actually, those kernels could be rewritten to reduce the number of kernel launches by increasing the memory consumption: on input one could in a one launch compute 

```
inter_left, inter_right = [dx(input), sx(input)]
```
then in the second launch

```
out_left, out_right = [sy(inter_left), dy(inter_right)]
```

And return `cast(out_left) + cast(out_right)`

And in that case it could be easier to exclude some samples from smoothing. But as the current approach seemed to performed reasonably when compared against gassuian_blur I decided to stick to it and apply the optimization with no smoothing only if the whole batch does not require it.

dali/kernels/imgproc/convolution/laplacian_gpu.cuh
stiepan: done

dali/kernels/imgproc/convolution/laplacian_gpu.cuh
```diff
@@ -0,0 +1,338 @@
...
+    } else {
+      single_dim_impl_->Run(ctx, out, in, windows[deriv_axis], {}, conv_epilogue);
```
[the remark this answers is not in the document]
stiepan: done

dali/kernels/imgproc/convolution/laplacian_gpu.cuh
stiepan: Added the asserts

stiepan:
!build

stiepan:
!build

klecki (approved):

dali/kernels/common/cast.cuh
```diff
@@ -0,0 +1,43 @@
...
+  void *output;
+  const void *input;
+};
+
+template <typename OType, typename IType>
+__global__ void BatchedCastKernel(const CastSampleDesc *samples, const BlockDesc<1> *blocks) {
+  const auto &block = blocks[blockIdx.x];
+  const auto &sample = samples[block.sample_idx];
+  auto *out = reinterpret_cast<OType *>(sample.output);
+  const auto *in = reinterpret_cast<const IType *>(sample.input);
```
mzient, line 33: Nitpick:
```suggestion
  auto *out = static_cast<OType *>(sample.output);
  const auto *in = static_cast<const IType *>(sample.input);
```

dali/kernels/imgproc/convolution/laplacian_gpu.cuh
```diff
@@ -0,0 +1,318 @@
...
+    }
+    if (!single_dim_impl_) {
+      single_dim_impl_ = std::make_unique<SingleDimConv>();
+    }
+    return single_dim_impl_->Setup(ctx, in_shape, window_sizes[deriv_axis]);
+  }
+
+  void Run(KernelContext& ctx, const TensorListView<StorageGPU, Out, ndim> out,
+           const TensorListView<StorageGPU, const In, ndim>& in,
+           const std::array<TensorListView<StorageCPU, const W, 1>, axes>& windows,
```
mzient, line 74: I find it a bit weird that it's called "PartialDeriv", but can actually calculate something else, depending on user-provided windows. I think that the utilities to fill these windows should be either a part of this class or a utility existing somewhere nearby.

stiepan: You can find the utility (LaplacianWindows) in `laplacian_params.h` that is used in CPU implementation too. The argument ``windows` is prepared by the operator and then passed to the kernel, hence the separation. As there is only limited number of possible laplacian windows, operator has a single instance of that class and computes them once, then only putting pointer in the appropriate tensorlistviews.

mzient: I think this can be done in a follow-up, but I do think that the function that populates these windows belongs in this header or one included by it, with clear indication int the Doxygen for `Run` how to populate these windows.

stiepan: Sure, I'll move LaplacianWindows to kernels as a separate header (to be included and used by cpu and gpu operator).

dali/kernels/common/cast.cuh
stiepan: done

dali/kernels/imgproc/convolution/laplacian_gpu.cuh
```diff
@@ -0,0 +1,318 @@
...
+
+  void Run(
+      KernelContext& ctx, const TensorListView<StorageGPU, Out, ndim> out,
+      const TensorListView<StorageGPU, const In, ndim>& in,
+      const std::array<std::array<TensorListView<StorageCPU, const W, 1>, axes>, axes>& windows,
+      std::array<span<const float>, axes> scale) {
+    return sobel_dx_.Run(ctx, out, in, windows[0], {}, {scale[0]});
+  }
+
+  DxKernel sobel_dx_;
```
mzient, line 255:
```suggestion
  DxKernel dx_kernel_;
```
While I can pass `PartialDeriv` (or `Laplacian`, for that matter) without built-in facilities for calculating the windows, calling this "Sobel" is a bit too much - there's nothing that makes it Sobel, nor does the implementation need to run sobel filter specifically - it can be any high-pass filter (or derivative).

dali/kernels/imgproc/convolution/laplacian_gpu.cuh
```diff
@@ -0,0 +1,318 @@
...
+          bool is_sequence>
+struct PartialDerivGpu {
+  using MultiDimConv = SeparableConvolutionGpu<Out, In, W, axes, has_channels, is_sequence>;
+  static constexpr int ndim = MultiDimConv::ndim;
+  static constexpr int sequence_axes = MultiDimConv::sequence_axes;
+  using SingleDimConv = ConvolutionGpu<Out, In, W, ndim, sequence_axes + deriv_axis, has_channels>;
+
+  KernelRequirements Setup(KernelContext& ctx, const TensorListShape<ndim>& in_shape,
+                           const std::array<TensorListShape<1>, axes>& window_sizes,
+                           bool has_smoothing) {
```
mzient, line 58: <discussed, offline; here for visibility>: I think that has_smoothing could be derived from `window_size` being empty. This would allow specifying this behavior per-axis from LaplacianGpu level.

stiepan: done

stiepan:
!build

stiepan:
!build

dali/kernels/imgproc/convolution/laplacian_gpu.cuh
```diff
@@ -0,0 +1,320 @@
...
+namespace kernels {
+
+namespace laplacian {
+
+/**
+ * @brief Computes convolution to obtain partial derivative in one of the dimensions.
+ * Convolution consits of `axes` windows, each to convolve along one dimension of the input data,
+ * where `deriv_axis`-th window is supposed to compute partial derivative along that axis,
+ * whereas the remaining windows should perform smoothing. If no smoothing is necessary in a whole
+ * batch, you can prevent smoothing convolutions form running by passing empty lists for
```
mzient, line 44: Nitpicks:
```suggestion
 * whereas the remaining windows should perform smoothing. If no smoothing is necessary in the whole
 * batch, you can prevent smoothing convolutions from running by passing empty lists for
```

dali/kernels/imgproc/convolution/laplacian_gpu.cuh
```diff
@@ -0,0 +1,320 @@
...
+    for (int axis = 0; axis < axes; axis++) {
+      if (axis != deriv_axis && window_sizes[axis].num_samples() > 0) {
+        return true;
+      }
+    }
+    return false;
+  }
+
+  KernelRequirements Setup(KernelContext& ctx, const TensorListShape<ndim>& in_shape,
+                           const std::array<TensorListShape<1>, axes>& window_sizes) {
```
mzient, line 65: Doxygen would be nice (you can add it in a follow-up).

mzient (approved):

PR #3625, opened 2022-01-17
Update optical flow operator to use the latest OF SDK capabilities

JanuszL:
- adds `hint_grid` argument to the optical flow operator so
  hint vector can have an arbitral resolution independent from
  the output
- renames `output_format` to `output_grid` argument name
- extends supported values for the output grid size
- adds more python tests
- renames files from Turing to just optical flow as Ampere supports
  it as well
- adds an ability to support sequence with different shapes inside
  the batch

## Description

### What happened in this PR
- adds `hint_format` argument to the optical flow operator so
  hint vector can have an arbitral resolution independent from
  the output
- extends supported values for the output grid size
- adds more python tests
- renames files from Turing to just optical flow as Ampere supports
  it as well

#### Additional information
  - optical flow operator

  - NA

### Tests
- [ ] Existing tests apply
- [x] New tests added
  - [x] Python tests
  - [ ] GTests
  - [ ] Benchmark
  - [ ] Other
- [ ] N/A

### Documentation
- [ ] Documentation updated
  - [x] Docstring
  - [ ] Doxygen
  - [ ] RST
  - [ ] Jupyter
  - [ ] Other
- [ ] N/A

#### Requirements
- [ ] N/A

**REQ IDs**: OFLOW.04, OFLOW.05, OFLOW.08

**JIRA TASK**: DALI-2556


dali/operators/sequence/optical_flow/optical_flow_adapter/optical_flow_adapter.h
```diff
@@ -26,13 +26,17 @@ namespace optical_flow {
 
 enum struct VectorGridSize {
   UNDEF,
+  SIZE_1 = 2,  /// 1x1 grid
+  SIZE_2 = 3,  /// 2x2 grid
   SIZE_4 = 4,  /// 4x4 grid
+  SIZE_8 = 8,  /// 8x8 grid
   MAX,
 };
```
mzient, line 34: Sorry, but this enum makes no sense and should go away in favour of a plain `int`.
1. The labels are just numbers and their numerical value matches the label (this should be enough, really, to get rid of this).
2. The MAX field is questionable at best in a sparse enumeration, such as this one
3. It doesn't enumerate anything. It _validates_ - but compile-time validation is of little use here.

If the value is supplied at run-time, we'll need run-time validation anyway. As such, we could have a function that checks whether given grid size is valid - but then again, doesn't optical flow API do just that for us?

dali/operators/sequence/optical_flow/optical_flow_adapter/optical_flow_adapter.h
```diff
@@ -26,13 +26,17 @@ namespace optical_flow {
...
 
 struct OpticalFlowParams {
   float perf_quality_factor;  /// 0..1, where 0 is best quality, lowest performance
-  VectorGridSize grid_size;
+  VectorGridSize out_grid_size;
+  VectorGridSize hint_grid_size;
```
mzient, line 39:
```suggestion
  int out_grid_size;
  int hint_grid_size;
```

dali/operators/sequence/optical_flow/optical_flow.h
```diff
@@ -105,12 +114,19 @@ class OpticalFlow : public Operator<Backend> {
...
+      case 4:
+        return optical_flow::VectorGridSize::SIZE_4;
+      case 8:
+        return optical_flow::VectorGridSize::SIZE_8;
+      default:
+        return optical_flow::VectorGridSize::UNDEF;
+        break;
     }
-    return optical_flow::VectorGridSize::MAX;
   }
```
mzient, line 130: Ekhm.... no. Let's get rid of this magic enum.

dali/operators/sequence/optical_flow/optical_flow_impl/optical_flow_impl.cc
```diff
@@ -14,28 +14,48 @@
...
+      break;
+    case VectorGridSize::SIZE_4:
+      out_width = (width + 3) / 4;
+      out_height = (height + 3) /4;
+      break;
+    default:
+      out_width = width;
+      out_height = height;
+      break;
+  }
```
mzient, line 56: Hmm... if VectorGridSize were just an `int`, as I insist it should, this would be reduced to:
```suggestion
  out_width = div_ceil(width, grid_size);
  out_height = div_ceil(height, grid_size);
```

JanuszL: Done

dali/operators/sequence/optical_flow/optical_flow.h
JanuszL: Done

dali/operators/sequence/optical_flow/optical_flow_adapter/optical_flow_adapter.h
JanuszL: Done

dali/operators/sequence/optical_flow/optical_flow_adapter/optical_flow_adapter.h
JanuszL: Done

dali/operators/sequence/optical_flow/optical_flow.h
```diff
@@ -41,6 +41,7 @@ struct backend_to_compute<GPUBackend> {
 
 const std::string kPresetArgName = "preset";                               // NOLINT
 const std::string kOutputFormatArgName = "output_format";                  // NOLINT
+const std::string kHintFormatArgName = "hint_format";                      // NOLINT
```
mzient, line 44: I don't think these fields were used a lot (if at all) given that previously there was just one legal value (4) - and even in the documentation this was referred to as "grid_size", not "output_format". I think we should use "output_grid" and "hint_grid", respectively or something along these lines. In any case, it's not "format", since the format is the same - a pair of numbers in x, y order.

dali/operators/sequence/optical_flow/optical_flow.h
```diff
@@ -41,6 +41,7 @@ struct backend_to_compute<GPUBackend> {
 
 const std::string kPresetArgName = "preset";                               // NOLINT
```
mzient, line 42: A general remark: `std::string` constants are not a good idea - but if they are used, they should be declared `static` or placed in an anonymous namespace - otherwise the program will fail to link if this header is included in multiple translation units.

dali/operators/sequence/optical_flow/optical_flow.h
JanuszL: Done

dali/operators/sequence/optical_flow/optical_flow.h
JanuszL: Done

dali/operators/sequence/optical_flow/optical_flow_impl/optical_flow_impl.cc
```diff
@@ -0,0 +1,347 @@
...
+  if (height != height_) {
+    height_ = height;
+    IsSizeSupported(height_min_, height_max_, height_, "Height");
+  }
+
+  init_params_.width = static_cast<uint32_t>(width_);
+  init_params_.height = static_cast<uint32_t>(height_);
+
+  DestroyBuffers();
```
JanuszL, line 174: Long term we can think about creating multiple of NV OF instances for requested resolutions, but this way we can easily run of out of the memory and it is hard to tell who can benefit from such optimization.

JanuszL:
!build

dali/operators/sequence/optical_flow/optical_flow.cc
```diff
@@ -109,34 +126,33 @@ void OpticalFlow<GPUBackend>::RunImpl(Workspace<GPUBackend> &ws) {
...
   }
+  const auto &input = ws.Input<GPUBackend>(0);
+  auto &output = ws.Output<GPUBackend>(0);
+  output.SetLayout("HWC");  // Channels represent the two flow vector components (x and y)
+
+  auto input_sh = input.shape();
+
+  // sort all sequences by size, to make sure that samples are processed grouped by shape to avoid
+  // NV OF reconfiguration as  much as possible
+  std::vector<DimsOrder> processing_order;
```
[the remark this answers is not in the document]
stiepan: Okay, I got the answer below.:)

dali/operators/sequence/optical_flow/optical_flow.h
```diff
@@ -79,11 +81,50 @@ class OpticalFlow : public Operator<Backend> {
...
+      {
+        int driver_cuda_version = 0;
+        CUDA_CALL(cuDriverGetVersion(&driver_cuda_version));
+        if (driver_cuda_version >= 11030 && driver_cuda_version < 11040)
+          of_stream = 0;
+      }
+    #endif
+
+    auto input_sh = input.shape();
+    of_lazy_init(input_sh[0][2], input_sh[0][1], depth_, image_type_, device_id_, of_stream);
```
stiepan, line 110: If `std::vector<DimsOrder> processing_order;` was created here, we could make sure this gets initialized with the actual shape of first sample, right?

dali/operators/sequence/optical_flow/optical_flow.cc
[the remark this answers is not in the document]
JanuszL: It was to avoid frequent configurations. I haven't compared perf compared to the order. I would expect rather negligible perf difference, and this code it to avoid expensive OF reconfiguration (deallocation of the buffers, destroying driver structures and recreation of it).

dali/operators/sequence/optical_flow/optical_flow.h
JanuszL: I can do it, but it will impact the perf only of the very first iteration as this function internals are called only once for the operators lifetime.

JanuszL: Done

dali/operators/sequence/optical_flow/optical_flow.cc
```diff
@@ -109,34 +117,23 @@ void OpticalFlow<GPUBackend>::RunImpl(Workspace<GPUBackend> &ws) {
     CUDA_CALL(cudaEventRecord(sync_, ws.stream()));
     CUDA_CALL(cudaStreamWaitEvent(of_stream, sync_, 0));
   }
+  const auto &input = ws.Input<GPUBackend>(0);
+  auto &output = ws.Output<GPUBackend>(0);
+  output.SetLayout("HWC");  // Channels represent the two flow vector components (x and y)
```
stiepan, line 122: I know it was already there, but:
1. Isn't the sample_dim == 4? Shouldn't we mention the leading `F` in the output layout?
2. What happens if somebody calls the op with layout FWHC?

dali/test/python/test_optical_flow.py
```diff
@@ -211,33 +224,52 @@ def flow_to_color(flow_uv, clip_flow=None, convert_to_bgr=False):
...
 
+def check_optflow(output_grid=1, hint_grid=1, use_temporal_hints=False):
+    batch_size = 3
+    pipe = of_pipeline(batch_size=batch_size, num_threads=3, device_id=0, output_grid=output_grid,
+                       hint_grid=hint_grid, use_temporal_hints=use_temporal_hints)
+    pipe.build()
+    if get_arch() < 8 and (output_grid != 4 or hint_grid != 4):
+        assert_raises(RuntimeError, pipe.run, "grid size : * is not supported, supported are:")
+
+    for _ in range(2):
```
stiepan, line 235: Shouldn't it be in `else:` clause?

dali/operators/sequence/optical_flow/optical_flow_impl/optical_flow_impl.cc
```diff
@@ -0,0 +1,347 @@
...
+std::vector<uint32_t> GetCapabilities(PFNNVOFGETCAPS get_cap, NvOFHandle handle, NV_OF_CAPS cap) {
+  uint32_t elm_num = 0;
+  CUDA_CALL(get_cap(handle, cap, nullptr, &elm_num));
+  std::vector<uint32_t> ret;
+  ret.resize(elm_num);
+  CUDA_CALL(get_cap(handle, cap, ret.data(), &elm_num));
+  return ret;
+}
+
+void IsGridSupported(std::vector<uint32_t> &caps, int grid_val, const std::string grid_kind) {
```
stiepan, line 47: nitpick: Can the caps be const&?

stiepan:
Left a few comments, apart from that it looks good to me.

dali/test/python/test_optical_flow.py
JanuszL: I thought it just ends up the test (early exit). I can add `else` if it reads better in your opinion.

dali/operators/sequence/optical_flow/optical_flow_impl/optical_flow_impl.cc
JanuszL: Done

dali/test/python/test_optical_flow.py
JanuszL: Done

stiepan: Sure, but I think it boils down to `pass` rather than `return` in that case.

dali/operators/sequence/optical_flow/optical_flow.cc
JanuszL: Added layout info to the op schema.
Fixed layout set in the op.

stiepan (approved):

banasraf (approved):

PR #3670, opened 2022-02-10
Use DynamicScratchpad in KernelManager.

mzient:

## Category:

## Description:
The dynamic scratchpad is now the primary scratchpad implementation.

### Tests
- [ ] Existing tests apply
- [ ] New tests added
  - [ ] Python tests
  - [ ] GTests
  - [ ] Benchmark
  - [ ] Other
- [x] N/A
**TODO**

### Documentation
- [ ] Documentation updated
  - [ ] Docstring
  - [ ] Doxygen
  - [ ] RST
  - [ ] Jupyter
  - [ ] Other
- [X] N/A

#### Requirements
- [X] N/A

**REQ IDs**: N/A

**JIRA TASK**: DALI-2449

### DIFF
--- a/dali/kernels/dynamic_scratchpad.h
+++ b/dali/kernels/dynamic_scratchpad.h
@@ -190,6 +190,9 @@ class DynamicScratchpad
 
   template <typename Kind>
   void *AllocImpl(size_t bytes, size_t alignment) {
+    if (bytes == 0)
+      return nullptr;  // do not initialize the resource in case of 0-sized allocation
+
     auto &r = resource<Kind>();
     if (!r.get_upstream()) {
       InitResource(type_tag<Kind>());
--- a/dali/kernels/dynamic_scratchpad.h
+++ b/dali/kernels/dynamic_scratchpad.h
@@ -60,6 +60,8 @@ class DynamicScratchpadImplT {
                              AccessOrder dealloc_order = {}) {
     static_assert(!std::is_same<Kind, mm::memory_kind::host>::value,
       "Cannot use a stream-ordered resource for plain host memory");
+    if (!dealloc_order.has_value())
+      dealloc_order = alloc_order;
     adapter<Kind>() = { rsrc, alloc_order, dealloc_order };
     set_upstream_resource<Kind>(&adapter<Kind>());
   }
--- a/dali/kernels/dynamic_scratchpad.h
+++ b/dali/kernels/dynamic_scratchpad.h
@@ -135,7 +135,7 @@ class DynamicScratchpad
     initial_sizes_ = initial_sizes;
     for (auto &s : initial_sizes_) {
       if (s == 0)
-        s = 4096;
+        s = 0x10000;  // 64k
     }
     if (!pinned_dealloc_order.has_value())
       pinned_dealloc_order = device_order;
--- a/dali/kernels/dynamic_scratchpad.h
+++ b/dali/kernels/dynamic_scratchpad.h
@@ -144,21 +144,9 @@ class DynamicScratchpad
     if (!managed_dealloc_order.has_value())
       managed_dealloc_order = device_order;
 
-    set_upstream_resource<mm::memory_kind::host>(mm::GetDefaultResource<mm::memory_kind::host>());
-
-    set_upstream_resource<mm::memory_kind::pinned>(
-        mm::GetDefaultResource<mm::memory_kind::pinned>(),
-        AccessOrder::host(),
-        pinned_dealloc_order);
-
-    set_upstream_resource<mm::memory_kind::device>(
-        mm::GetDefaultResource<mm::memory_kind::device>(),
-        device_order);
-
-    set_upstream_resource<mm::memory_kind::managed>(
-        mm::GetDefaultResource<mm::memory_kind::managed>(),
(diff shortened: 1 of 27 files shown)


dali/operators/image/resize/resize_base.cc
```diff
@@ -121,8 +121,8 @@ void ResizeBase<GPUBackend>::InitializeGPU(int minibatch_size, size_t temp_buffe
     minibatch_size_ = minibatch_size;
   }
   kmgr_.Resize(1, 0);
-  kmgr_.SetMemoryHint<mm::memory_kind::device>(temp_buffer_hint);
-  kmgr_.GetScratchpadAllocator(0).Reserve<mm::memory_kind::device>(temp_buffer_hint);
+  // kmgr_.SetMemoryHint<mm::memory_kind::device>(temp_buffer_hint);
+  // kmgr_.GetScratchpadAllocator(0).Reserve<mm::memory_kind::device>(temp_buffer_hint);
```
JanuszL, line 125: To remove, I guess.

dali/operators/image/resize/resize_op_impl_gpu.h
```diff
@@ -79,7 +79,7 @@ class ResizeOpImplGPU : public ResizeBase<GPUBackend>::Impl {
       kernels::KernelRequirements &req = kmgr_.Setup<Kernel>(mb_idx, ctx, mb.input, param_slice);
       mb.out_shape = req.output_shapes[0].to_static<frame_ndim>();
     }
-    kmgr_.ReserveMaxScratchpad(0);
+    // kmgr_.ReserveMaxScratchpad(0);
```
JanuszL, line 82: To remove, I guess.

dali/kernels/kernel_manager.h
```diff
@@ -218,9 +219,15 @@ class DLL_PUBLIC KernelManager {
...
   void Run(int thread_idx, int instance_idx, KernelContext &context, OutInArgs &&...out_in_args) {
-    assert(static_cast<size_t>(thread_idx) < scratchpads.size());
-    auto &sa = GetScratchpadAllocator(thread_idx);
-    Run<Kernel>(sa, instance_idx, context, std::forward<OutInArgs>(out_in_args)...);
+    assert(instance_idx >= 0 &&
+           static_cast<size_t>(instance_idx) < NumInstances() &&
+           "Kernel instance index (instance_idx) out of range");
+    auto &inst = instances[instance_idx];
+    DynamicScratchpad scratchpad({}, AccessOrder(context.gpu.stream));
+    auto *old_scratchpad = context.scratchpad;
```
JanuszL, line 227: Why do we need `context.scratchpad` for if we are not using it anyway here?

dali/kernels/dynamic_scratchpad.h
```diff
@@ -135,41 +137,71 @@ class DynamicScratchpad
...
-       mm::memory_kind::pinned,
-       mm::memory_kind::device,
-       mm::memory_kind::managed),
-      (ret = resource<Kind>().allocate(bytes, alignment)),
-      (assert(!"Incorrect memory kind id");));
-    return ret;
+  template <typename Kind>
+  void *AllocImpl(size_t bytes, size_t alignment) {
+    if (bytes == 0)
+      return nullptr;  // do not initialize the resource in case of 0-sized allocation
```
JanuszL, line 194: Do you envision cases where we want alloc 0 bytes?

mzient: I've already hit it in CPU Laplacian and GaussianBlur (and it made CPU-only test fail).

dali/kernels/kernel_manager.h
mzient: It's used by the `Run` method invoked afterwards.

JanuszL: I was rather asking about the ideas of replacing the old one with one created here.

dali/operators/image/resize/resize_base.cc
mzient: Removed in a follow-up.

dali/operators/image/resize/resize_op_impl_gpu.h
mzient: Removed in a follow-up.

dali/kernels/kernel_manager.h
mzient: Well, now the old one is always NULL, so we don't need to store it. I'll simplify it (again, possibly in a follow-up).

JanuszL: Ok, let do that in a follow-up.

stiepan: So kernel manager `scratchpads` are also to be removed in a follow-up, right?

stiepan (approved):

dali/kernels/kernel_manager.h
mzient: Yes, that's what I brought up in open floor. Since this involves an API change (we no longer need thread_idx), the change affects a large number of files - hence the decision to make the more important change that potentially affects performance (this one) first and then clean the rest.

JanuszL (approved):

PR #3673, opened 2022-02-11
Add proper handling of CUDA_ARCHITECTURES in CMake for 3.18 and newer

JanuszL:
- CMake 3.18 added a new way to handle the CUDA architecture the
  build is performed for. This PR sets USE_CMAKE_CUDA_ARCHITECTURES
  for this (and newer) CMake version to enable this new way

## Category:

## Description:
- CMake 3.18 added a new way to handle the CUDA architecture the
  build is performed for. This PR sets USE_CMAKE_CUDA_ARCHITECTURES
  for this (and newer) CMake version to enable this new way

- main CMake

- nA

### Tests
- [ ] Existing tests apply
- [ ] New tests added
  - [ ] Python tests
  - [ ] GTests
  - [ ] Benchmark
  - [ ] Other
- [x] N/A

### Documentation
- [ ] Documentation updated
  - [ ] Docstring
  - [ ] Doxygen
  - [ ] RST
  - [ ] Jupyter
  - [ ] Other
- [x] N/A

#### Requirements
- [x] N/A

**REQ IDs**: N/A

**JIRA TASK**: N/A

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,4 @@
-# Copyright (c) 2017-2021, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# Copyright (c) 2017-2022, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
@@ -341,6 +341,10 @@ if (DALI_CLANG_ONLY)
 endif()
 
 # Add ptx & bin flags for cuda compiler (nvcc)
+if(${CMAKE_VERSION} VERSION_GREATER_EQUAL "3.16.0")
+  set(USE_CMAKE_CUDA_ARCHITECTURES "ON")
+endif()
+
 if(USE_CMAKE_CUDA_ARCHITECTURES)
   CUDA_get_cmake_cuda_archs(CMAKE_CUDA_ARCHITECTURES ${CUDA_targeted_archs})
   message(STATUS "Generated CMAKE_CUDA_ARCHITECTURES: ${CMAKE_CUDA_ARCHITECTURES}")
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -321,15 +321,27 @@ message(STATUS "CUDA .cu files compiler: ${USED_CUDA_COMPILER}")
 CUDA_find_supported_arch_values(CUDA_supported_archs ${USED_CUDA_COMPILER} ${CUDA_known_archs})
 message(STATUS "CUDA supported archs: ${CUDA_supported_archs}")
 
-set(CUDA_TARGET_ARCHS_SORTED ${CUDA_TARGET_ARCHS})
-list(SORT CUDA_TARGET_ARCHS_SORTED)
-CUDA_find_supported_arch_values(CUDA_targeted_archs ${USED_CUDA_COMPILER} ${CUDA_TARGET_ARCHS_SORTED})
+if ("${CMAKE_VERSION}" VERSION_GREATER_EQUAL "3.16.0")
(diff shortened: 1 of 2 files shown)


JanuszL:
!build

CMakeLists.txt
```diff
@@ -341,6 +341,10 @@ if (DALI_CLANG_ONLY)
 endif()
 
 # Add ptx & bin flags for cuda compiler (nvcc)
+if(${CMAKE_VERSION} VERSION_GREATER_EQUAL "3.16.0")
```
klecki, line 344: One open question is: should we allow to provide `CMAKE_CUDA_ARCHITECTURES` directly? - as this is the official way of doing it, and the `CUDA_targeted_archs` was our workaround before `CMAKE_CUDA_ARCHITECTURES` was a thing. 

That depends on our philosophy if we want to be fully-CMake compliant or support this as another option, we can:
* detect if `CMAKE_CUDA_ARCHITECTURES` is set and use it with precedence over `CUDA_targeted_archs`
* keep `CUDA_targeted_archs` for all usages, and automatically propagate to  `CMAKE_CUDA_ARCHITECTURES` as in this PR - maybe we should warn the user when he tries to set it by hand?

What's with the documentation of this feature? The `USE_CMAKE_CUDA_ARCHITECTURES` was rather well hidden.

JanuszL: How about now?

JanuszL:
!build

cmake/CUDA_utils.cmake
```diff
@@ -48,10 +68,20 @@ else()
...
+    set(CUDA_TARGET_ARCHS_SORTED ${CUDA_TARGET_ARCHS})
+    list(SORT CUDA_TARGET_ARCHS_SORTED)
+    set(CMAKE_CUDA_ARCHITECTURES ${CUDA_TARGET_ARCHS_SORTED})
+  endif()
+else()
+  if ("${CUDA_TARGET_ARCHS}" STREQUAL "")
+    message("CUDA_TARGET_ARCHS cannot be empty, setting to the default")
+    set(CUDA_TARGET_ARCHS ${CUDA_known_archs} CACHE STRING "List of target CUDA architectures" FORCE)
+  endif()
+  set(CUDA_TARGET_ARCHS ${CUDA_known_archs} CACHE STRING "List of target CUDA architectures")
```
klecki, line 84: Erm, Why is this line duplicated here?

cmake/CUDA_utils.cmake
```diff
@@ -35,6 +35,26 @@ function(adjust_source_file_language_property SOURCES)
   endif()
 endfunction()
 
+# Generate list of xx-real for every specified supported architecture.
+# List should be sorted in increasing order.
+#
+# The last one will also be repeated as xx-virtual to ensure the generation of PTX for most recent
+# virtual architecture and maintain forward compatibility.
+function(CUDA_get_cmake_cuda_archs out_args_list arch_values)
```
klecki, line 43: I don't like the fact, that we have mixed global level code and function declarations here - I see I did it also myself when introducing `adjust_source_file_language_property`, but this `CUDA_get_cmake_cuda_archs` is somewhat connected to the code below the listing and setting of default architectures and the rest of code is below - it made the diff a bit surprising to read.

klecki, line 43: This function is not used as well.

klecki (commented):
Few issues:
1. CUDA_get_cmake_cuda_archs is not used, so we do not generate the -real, latest-virtual list

2. also we do not sanitize and sort the user's input - should we allow for any combination of -real, -virtual architectures to be provided in regular CMake way and just compile without adjusting it? Now there is this weird parsing, that leaves the -virtual intact, but removes the -real.

3. can we just move to CMAKE 3.16 and not entangle ourselves in layers of ifs?

JanuszL:
I don't think I have any good idea for a clean and reliable solution supporting old and new syntax.

PR #3679, opened 2022-02-15
Fix coverity issues

szalpal:

## Category:

## Description:
Fixing coverity issues

### Tests
- [x] Existing tests apply
- [ ] New tests added
  - [ ] Python tests
  - [ ] GTests
  - [ ] Benchmark
  - [ ] Other
- [ ] N/A

### Documentation
- [ ] Documentation updated
  - [ ] Docstring
  - [ ] Doxygen
  - [ ] RST
  - [ ] Jupyter
  - [ ] Other
- [ ] N/A

#### Requirements
- [x] N/A

**REQ IDs**: N/A

**JIRA TASK**: N/A

### DIFF
--- a/dali/pipeline/data/tensor_list.h
+++ b/dali/pipeline/data/tensor_list.h
@@ -755,7 +755,7 @@ class DLL_PUBLIC TensorList {
   }
 
  protected:
-  Buffer<Backend> data_;
+  Buffer<Backend> data_ = {};
   // We store a set of dimension for each tensor in the list.
   // We also pre-compute the offsets of each tensor in the
   // underlying allocation for random access
--- a/dali/test/dali_test_utils.cc
+++ b/dali/test/dali_test_utils.cc
@@ -67,7 +67,7 @@ void MakeRandomBatch(TensorList<CPUBackend> &data, int N,
   }
 }
 
-void CheckResults(DeviceWorkspace ws, int batch_size, int i,
+void CheckResults(const DeviceWorkspace& ws, int batch_size, int i,
                   TensorList<CPUBackend> &data, int output_idx) {
   TensorList<CPUBackend> res_cpu;
   if (ws.OutputIsType<GPUBackend>(output_idx)) {
--- a/dali/python/backend_impl.cc
+++ b/dali/python/backend_impl.cc
@@ -1291,7 +1291,6 @@ PYBIND11_MODULE(backend_impl, m) {
 
   m.def("IsDriverInitialized", [] {
     // we just want to check if cuda has been loaded already
-
     if (dlopen("libcuda.so", RTLD_NOLOAD | RTLD_NOW) ||
         dlopen("libcuda.so.1", RTLD_NOLOAD | RTLD_NOW)) {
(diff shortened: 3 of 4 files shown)


szalpal:
!build

dali/operators/reader/loader/video/nvdecode/color_space.cu
```diff
@@ -89,4 +90,5 @@ void yuv_to_rgb(uint8_t *yuv, int yuv_pitch, uint8_t *rgb, int rgb_pitch, int wi
     yuv_to_rgb_kernel
         <<<grid_layout, block_layout, 0, stream>>>
         (yuv, yuv_pitch, rgb, rgb_pitch, width, height);
+    CUDA_CALL(cudaGetLastError());
```
JanuszL, line 93: I think we do that at executor level.

jantonguirao (approved):

klecki (approved):

dali/pipeline/data/tensor_list.h
```diff
@@ -755,7 +755,7 @@ class DLL_PUBLIC TensorList {
   }
 
  protected:
-  Buffer<Backend> data_;
+  Buffer<Backend> data_ = {};
```
mzient, line 758:
```suggestion
  Buffer<Backend> data_{};
```
or leave it as-is - it's a false positive.

dali/test/dali_test_utils.cc
```diff
@@ -67,7 +67,7 @@ void MakeRandomBatch(TensorList<CPUBackend> &data, int N,
   }
 }
 
-void CheckResults(DeviceWorkspace ws, int batch_size, int i,
+void CheckResults(const DeviceWorkspace& ws, int batch_size, int i,
```
mzient, line 70: :+1:

dali/pipeline/data/tensor_list.h
szalpal: I believe, that in this case both
```c++
Buffer<Backend data_ = {};
Buffer<Backend> data_{};
```
are the same:

`Buffer<Backend> data_{}` performs [value initialization](https://en.cppreference.com/w/cpp/language/value_initialization) per syntax (5). `Buffer<Backend> data_ = {};` is a [copy-list-initialization](https://en.cppreference.com/w/cpp/language/list_initialization) per syntax (10) and since `Buffer<>` is not an aggregate type, the value initialization is performed (see `Explanation` in https://en.cppreference.com/w/cpp/language/list_initialization):
> Otherwise, if the braced-init-list is empty and T is a class type with a default constructor, [value-initialization](https://en.cppreference.com/w/cpp/language/value_initialization) is performed.

szalpal:
!build

dali/pipeline/data/tensor_list.h
mzient: Well.... theoretically. I specifically remember seeing bugs related to overspecifing initialization of an object having a default constructor (which resulted in an attempt to copy). I don't remember which compiler did this (clang, gcc or nvcc).

PR #3680, opened 2022-02-15
Remove `lineinfo` from RelWithDebInfo DALI builds

JanuszL:
- as CUDA doesn't provide a tool to strip debug info
  to avoid bloating release binary with the debug info
  that cannot be removed before the release remove
  `lineinfo` from RelWithDebInfo build configuration

## Category:

## Description:
- as CUDA doesn't provide a tool to strip debug info
  to avoid bloating release binary with the debug info
  that cannot be removed before the release remove
  `lineinfo` from RelWithDebInfo build configuration

- CMake

- NA

### Tests
- [x] Existing tests apply
- [ ] New tests added
  - [ ] Python tests
  - [ ] GTests
  - [ ] Benchmark
  - [ ] Other
- [ ] N/A

### Documentation
- [ ] Documentation updated
  - [ ] Docstring
  - [ ] Doxygen
  - [ ] RST
  - [ ] Jupyter
  - [ ] Other
- [x] N/A

#### Requirements
- [x] N/A

**REQ IDs**: N/A

**JIRA TASK**: DALI-2601

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -124,6 +124,7 @@ cmake_dependent_option(STATIC_LIBS "Build static libraries instead of shared-obj
 
 option(VERBOSE_LOGS "Adds verbose loging to DALI" OFF)
 option(WERROR "Treat all warnings as errors" OFF)
+option(RELWITHDEBINFO_CUDA_DEBUG "Add device side debug info for RelWithDebInfo build conifguration" OFF)
 
 cmake_dependent_option(DALI_CLANG_ONLY "Compile DALI using only Clang. Suitable only for developement."
     OFF "CMAKE_CXX_COMPILER_ID STREQUAL Clang" OFF)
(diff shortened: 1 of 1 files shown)


JanuszL:
It shrinks CUDA 11.x build from 700MB->400MB.

jantonguirao (approved):

klecki:
Can we split this into configurations that are suitable for building release wheels (without line info) and second one for the developer that want to keep the line info for the purpose of local debugging? Otherwise we would need to manually add the flag every time.

JanuszL:
> Can we split this into configurations that are suitable for building release wheels (without line info) and second one for the developer that want to keep the line info for the purpose of local debugging? Otherwise we would need to manually add the flag every time.

I have added an optional flag `RELWITHDEBINFO_CUDA_DEBUG` set to `OFF` by default. I don't think we use this limited debug info for the device code that often anyway. I think in most cases the developers use `Debug` build configuration which uses `-lineinfo`.

klecki (approved):

PR #3686, opened 2022-02-17
update license to GPLV3

libererer:


JanuszL:
@libererer  - thank you for your contribution.
However, as we don't want to prevent using DALI in closed source projects (and other legal reasons) we don't plan to change its license.

PR #3700, opened 2022-02-23
Introduce runtime typed TensorView

klecki:
### `TensorView<DynamicType>`
Introduce runtime typed TensorView.
The new specialization uses `DynamicType` tag
as the type parameter to TensorView.

The purpose is to have Tensor wrapper that is similar
to already existing DALI vocabulary types, and additionally
allows to keep the type as a runtime information.

### view<T, ndim> conversions

Extend the utility of the `view` function
template to allow for any conversions between
statically and dynamically typed TensorViews.

### Next steps
The target use of this type is to use it as the return
type of the TensorBatch::operator[] - using static type
is not possible due to the Tensor[List/Vector] not being
statically typed. This should allow for proper encapsulation
of the batch object - TensorView allow access to the data
but do not allow to adjust the allocation or changing
the metadata through the view - as opposed to the currently
used Tensor.

The current use pattern of obtaining statically TensorViews
in CPU operators using: `view<T>(tensor_vector[i])` 
will work without additional changes.

### To consider
I specifically added the specialization as small definition 
at the end of dynamic_tensor_view.h, and implemented
most of the operations as DynamicTensorView and ConstDynamicTensorView
if we find a good reason not to expose those dynamic views
as TensorView specialization.

Options for follow-up:
* extending the capabilities of `view` overload for TensorListView.
* adding `type()` method to TensorView that would not be callable
  for statically typed ones, but would ease the work in template contexts. 

### What this PR is not

This PR does not add similar capabilities to TensorListView
as they are not needed for the further work and TensorVector/List
and statically typed TensorListView are enough for current DALI
usage.

TensorViews, view<T, ndim>()

There is a lot of boilerplate related to move variants
for the conversions etc.

### Tests
- [ ] Existing tests apply
- [x] New tests added
  - [ ] Python tests
  - [x] GTests
  - [ ] Benchmark
  - [ ] Other
- [ ] N/A

### Documentation
- [x] Documentation updated
  - [ ] Docstring
  - [x] Doxygen
  - [ ] RST
  - [ ] Jupyter
  - [ ] Other
- [ ] N/A

#### Requirements
- [ ] N/A

**REQ IDs**: N/A

**JIRA TASK**: N/A

### DIFF
--- a/dali/pipeline/data/dynamic_tensor_view.h
+++ b/dali/pipeline/data/dynamic_tensor_view.h
@@ -1,0 +1,589 @@
+// Copyright (c) 2022, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef DALI_PIPELINE_DATA_DYNAMIC_TENSOR_VIEW_H_
+#define DALI_PIPELINE_DATA_DYNAMIC_TENSOR_VIEW_H_
+
+#include <stdexcept>
+#include <type_traits>
+#include <utility>
+#include <vector>
+#include "dali/core/tensor_shape.h"
+#include "dali/core/tensor_view.h"
+#include "dali/pipeline/data/types.h"
+
+namespace dali {
+
+/**
+ * @brief Tag type indicating that the TensorView is keeping the type information in runtime.
+ *
+ * It can be converted to the statically-typed TensorView using `view<T, ndim>()` function
+ * or by `TensorView::to_static_type<T, ndim>()` member function.
+ */
+struct DynamicType {
+  DynamicType() = delete;
+};
+
+template <typename Backend, typename DataType, int ndim = DynamicDimensions>
+struct DynamicTensorViewBase {
+  static_assert(std::is_same<std::remove_const_t<DataType>, void>::value,
+                "The underlying type must either be void or const void");
+  using element_type = DataType;
+  int dim() const {
+    return shape.sample_dim();
+  }
+  DALIDataType type() const {
+    return type_id;
+  }
+
+  ptrdiff_t num_elements() const {
+    return shape.num_elements();
+  }
+
+  /**
+   * @brief Utility to calculate pointer to element at given coordinates
+   */
+  template <typename... Indices>
+  DataType *operator()(int64_t idx0, Indices &&...idx) const {
+    return data + CalcOffset(shape, std::array<ptrdiff_t, sizeof...(Indices) + 1>{
+                                        idx0, (ptrdiff_t{idx})...});
+  }
+
+  /**
+   * @brief Utility to calculate pointer to element at given coordinates
+   */
+  template <typename Offset>
+  DataType *operator()(const Offset &pos) const {
+    return data + CalcOffset(shape, pos);
+  }
+
+  DataType *data = nullptr;
+  TensorShape<ndim> shape = {};
+  DALIDataType type_id = DALI_NO_TYPE;
+
+ protected:
+  DynamicTensorViewBase() = default;
+  DynamicTensorViewBase(const DynamicTensorViewBase &) = default;
+  DynamicTensorViewBase(DataType *data, const TensorShape<ndim> &shape, DALIDataType type_id)
+      : data(data), shape(shape), type_id(type_id) {}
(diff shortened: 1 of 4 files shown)


klecki:
!build

klecki:
!build

mzient (changes requested):
Inclusion of headers from dali/pipeline in stuff from dali/core is a show-stopper.

dali/pipeline/data/dynamic_tensor_view.h
```diff
@@ -0,0 +1,623 @@
...
+#ifndef DALI_PIPELINE_DATA_DYNAMIC_TENSOR_VIEW_H_
+#define DALI_PIPELINE_DATA_DYNAMIC_TENSOR_VIEW_H_
+
+#include <stdexcept>
+#include <type_traits>
+#include <utility>
+#include <vector>
+#include "dali/core/tensor_shape.h"
+#include "dali/core/tensor_view.h"
+#include "dali/pipeline/data/types.h"
```
mzient, line 24: That's a no-go.

mzient, line 24: I don't like this design. It will end up leaking into kernels and possibly `dali/core` and stuff from `dali/pipeline` shouldn't be used there.

dali/pipeline/data/dynamic_tensor_view.h
```diff
@@ -0,0 +1,623 @@
...
+    return {data, shape.template to_static<other_ndim>(), type_id};
+  }
+
+  using Base::data;
+  using Base::shape;
+  using Base::type_id;
+};
+
+template <typename Backend, int ndim = DynamicDimensions>
+struct ConstDynamicTensorView : DynamicTensorViewBase<Backend, const void, ndim> {
```
mzient, line 342: Seeing this copy-paste here makes me reconsider using `void`. Alternatively, how about `const DynamicType`?

dali/pipeline/data/dynamic_tensor_view.h
```diff
@@ -0,0 +1,623 @@
...
+
+  DynamicTensorView() = default;
+
+  /**
+   * @name Construct the view inferring the type_id from the pointer value.
+   */
+  // @{
+  template <typename T, typename = std::enable_if_t<!std::is_const<T>::value>>
+  DynamicTensorView(T *data, const TensorShape<ndim> &shape)
+      : Base(data, shape, TypeTable::GetTypeId<T>()) {}
```
mzient, line 96: How about:
```suggestion
  template <typename T>
  DynamicTensorView(T *data, const TensorShape<ndim> &shape)
      : DynamicTensorView(static_cast<void*>(data,) shape, TypeTable::GetTypeId<T>()) {}
```
?
This will scream that a pointer of type `const something` cannot be converted to `void`. The error produced by SFINAE would be much harder to understand.

dali/pipeline/data/dynamic_tensor_view.h
```diff
@@ -0,0 +1,623 @@
...
+   * @brief Utility to calculate pointer to element at given coordinates
+   */
+  template <typename Offset>
+  DataType *operator()(const Offset &pos) const {
+    return data + CalcOffset(shape, pos);
+  }
+
+  DataType *data = nullptr;
+  TensorShape<ndim> shape = {};
+  DALIDataType type_id = DALI_NO_TYPE;
```
mzient, line 73: I'd really like to have an element size somewhere.

dali/pipeline/data/dynamic_tensor_view.h
```diff
@@ -0,0 +1,623 @@
...
+
+  /**
+   * @brief Utility to calculate pointer to element at given coordinates
+   */
+  template <typename... Indices>
+  DataType *operator()(int64_t idx0, Indices &&...idx) const {
+    return data + CalcOffset(shape, std::array<ptrdiff_t, sizeof...(Indices) + 1>{
+                                        idx0, (ptrdiff_t{idx})...});
+  }
+
```
mzient, line 62: This can't work. Your `DataType` is void, so compiling this function would result in `void*` arithmetic.

dali/pipeline/data/dynamic_tensor_view.h
```diff
@@ -0,0 +1,623 @@
...
+  /**
+   * @brief Utility to calculate pointer to element at given coordinates
+   */
+  template <typename Offset>
+  DataType *operator()(const Offset &pos) const {
+    return data + CalcOffset(shape, pos);
+  }
```
mzient, line 69: Likewise.

dali/pipeline/data/dynamic_tensor_view_test.cc
```diff
@@ -0,0 +1,289 @@
...
+#include "dali/pipeline/data/types.h"
+#include "dali/pipeline/data/views.h"
+
+namespace dali {
+
+using EBT = EmptyBackendTag;  // allow most of the checks to fit in one line
+
+template <typename DynamicTV, int ndim>
+void compare(const DynamicTV &dtv, const void *ptr, const TensorShape<ndim> &shape,
+             DALIDataType dtype) {
```
mzient, line 33: I think it should be a macro, not a function - this way you'll get meaningless line info when the test fails. GTest doesn't like this kind of helpers and I'm guilty of having created ones, too.

PR #3701, opened 2022-02-23
Ensure sample encapsulation in Tensor Vector

klecki:
## Description:
Add APIs matching TensorList to TensorVector:
* sample pointer accessors
* Set/GetMeta

Change operator[] to return [Const]SampleView.
Introduce UnsafeSetSample and UnsafeCopySample 
to replace TensorVector[i].ShareData(tensor) 
and TensorVector[i].Copy(tensor) - they work with current
code base, but for proper sample-based data structure
more checks should be introduced - intended for follow up.

Adjust code where necessary:
* where possible use data accessors directly on the TensorVector
    instead of the sample, as it should be faster than create temporary, so:
    tv[i].mutable_data<T>() -> tv.mutable_tensor<T>(i) etc.
* Using SampleViews is compatible with code that uses `view<T>`,
    as `view<T>(Tensor)` is equivalent to `view<T>(sample_view(Tensor))` 

Adjustments:
* allow views to work with scalar Tensors (they treated them as empty)
* introduce distinct SampleView and ConstSampleView as they need to
    be returned by value and we need sensible overloads for `view<>`.
* allow to access `capacity` and `nbytes` of individual samples,
    introduce total_capacity and total_nbytes for sum of them.

Next steps written as TODO in TensorVector dosctring.

Keep the SampleView accessors temporarily with leading underscore
so they can be easily adjusted during review.

Current naming:
The sample view has temporary leading underscores in the API (`_raw_data`), 
that would be removed before merging the PR - it's to facilitate renaming in case
it's requested and to help with verifying all calls are reworked to the
intended `tv.raw_tensor(i)` etc.

The `Unsafe` prefix in SetSample and CopySample is intended to temporary stay
there to discourage introduction of new use cases till the followup introduces
remaining checks.  

- [x] Remove prefix `_` from sample view naming before merge.
  
TensorVector, SampleView, SampleWorkspace-based operators, CPU operators,
C API, small bits of executor.

### Tests
- [x] Existing tests apply
- [x] New tests added
  - [ ] Python tests
  - [x] GTests
  - [ ] Benchmark
  - [ ] Other
- [ ] N/A

### Documentation
- [x] Documentation updated
  - [x] Docstring
  - [ ] Doxygen
  - [ ] RST
  - [ ] Jupyter
  - [ ] Other
- [ ] N/A

#### Requirements
- [ ] N/A

**REQ IDs**: N/A

**JIRA TASK**: N/A

### DIFF
--- a/dali/benchmark/operator_bench.h
+++ b/dali/benchmark/operator_bench.h
@@ -1,4 +1,4 @@
-// Copyright (c) 2019-2021, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+// Copyright (c) 2019-2022, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -54,16 +54,13 @@ class OperatorBench : public DALIBenchmark {
     auto op_ptr = InstantiateOperator(op_spec);
 
     auto data_in = std::make_shared<TensorVector<CPUBackend>>(batch_size);
-    for (auto &in_ptr : *data_in) {
-      in_ptr = std::make_shared<Tensor<CPUBackend>>();
-      in_ptr->set_type<T>();
-      in_ptr->Resize({H, W, C});
-      in_ptr->SetLayout("HWC");
-    }
+    data_in->set_type<T>();
+    data_in->Resize(uniform_list_shape(batch_size, TensorShape<>{H, W, C}));
+    data_in->SetLayout("HWC");
 
     if (fill_in_data) {
-      for (auto &in_ptr : *data_in) {
-        auto *ptr = in_ptr->template mutable_data<T>();
+      for (int sample_id = 0; sample_id < batch_size; sample_id++) {
+        auto *ptr = data_in->template mutable_tensor<T>(sample_id);
         for (int i = 0; i < N; i++) {
           ptr[i] = static_cast<T>(i);
         }
--- a/dali/benchmark/operator_bench.h
+++ b/dali/benchmark/operator_bench.h
@@ -59,8 +59,8 @@ class OperatorBench : public DALIBenchmark {
     data_in->SetLayout("HWC");
 
     if (fill_in_data) {
-      for (int sample_id = 0; sample_id < batch_size; sample_id++) {
-        auto *ptr = data_in->template mutable_tensor<T>(sample_id);
+      for (int sample_idx = 0; sample_idx < batch_size; sample_idx++) {
+        auto *ptr = data_in->template mutable_tensor<T>(sample_idx);
         for (int i = 0; i < N; i++) {
           ptr[i] = static_cast<T>(i);
         }
--- a/dali/c_api/c_api.cc
+++ b/dali/c_api/c_api.cc
@@ -129,8 +129,6 @@ void SetExternalInputTensors(daliPipelineHandle *pipe_handle, const char *name,
   if (layout_str != nullptr) {
     layout = dali::TensorLayout(layout_str);
   }
-  dali::TensorVector<Backend> data(curr_batch_size);
-  data.set_pinned(flags & DALI_ext_pinned);
   auto type_id = static_cast<dali::DALIDataType>(data_type);
   auto elem_sizeof = dali::TypeTable::GetTypeInfo(type_id).size();
 
@@ -140,19 +138,20 @@ void SetExternalInputTensors(daliPipelineHandle *pipe_handle, const char *name,
   else
     order = AccessOrder::host();
 
+  dali::TensorVector<Backend> data(curr_batch_size);
+  data.set_pinned(flags & DALI_ext_pinned);
+  data.set_sample_dim(sample_dim);
+  data.set_type(type_id);
+  data.set_order(order);
+  data.SetLayout(layout);
 
   for (int i = 0; i < curr_batch_size; i++) {
     // We cast away the const from data_ptr, as there is no other way of passing it to the
     // Tensor as we must also set the shape and type metadata.
     // The vector that we pass to pipeline is const.
-    dali::Tensor<Backend> tmp;
-    tmp.set_order(order);
     std::shared_ptr<void> ptr(const_cast<void *>(data_ptr[i]), [](void *){});  // no deleter
-    tmp.set_backing_allocation(ptr,
-                               tl_shape[i].num_elements() * elem_sizeof, flags & DALI_ext_pinned,
-                               type_id, tl_shape[i].num_elements());
-    tmp.Resize(tl_shape[i], type_id);
-    data.UnsafeSetSample(i, tmp);
+    data.UnsafeSetSample(i, ptr, tl_shape[i].num_elements() * elem_sizeof, flags & DALI_ext_pinned,
+                         tl_shape[i], type_id, order, layout);
   }
   pipeline->SetExternalInput(name, data, order,
                              flags & DALI_ext_force_sync,
@@ -434,7 +433,7 @@ size_t daliNumElements(daliPipelineHandle* pipe_handle, int n) {
 
 template <typename T>
 static size_t daliTensorSizeHelper(dali::DeviceWorkspace* ws, int n) {
-  return ws->Output<T>(n).total_nbytes();
+  return ws->Output<T>(n).nbytes();
 }
 
 size_t daliTensorSize(daliPipelineHandle* pipe_handle, int n) {
--- a/dali/c_api/c_api.cc
+++ b/dali/c_api/c_api.cc
@@ -143,11 +143,14 @@ void SetExternalInputTensors(daliPipelineHandle *pipe_handle, const char *name,
     // We cast away the const from data_ptr, as there is no other way of passing it to the
     // Tensor as we must also set the shape and type metadata.
     // The vector that we pass to pipeline is const.
-    data[i].set_pinned(flags & DALI_ext_pinned);
-    data[i].set_order(order);
-    data[i].ShareData(const_cast<void *>(data_ptr[i]), tl_shape[i].num_elements() * elem_sizeof);
-    data[i].Resize(tl_shape[i], type_id);
-    data[i].SetLayout(layout);
+    dali::Tensor<Backend> tmp;
+    tmp.set_order(order);
+    std::shared_ptr<void> ptr(const_cast<void *>(data_ptr[i]), [](void *){});  // no deleter
+    tmp.set_backing_allocation(ptr,
+                               tl_shape[i].num_elements() * elem_sizeof, flags & DALI_ext_pinned,
+                               type_id, tl_shape[i].num_elements());
+    tmp.Resize(tl_shape[i], type_id);
+    data.UnsafeSetSample(i, tmp);
   }
   pipeline->SetExternalInput(name, data, order,
                              flags & DALI_ext_force_sync,
@@ -429,7 +432,7 @@ size_t daliNumElements(daliPipelineHandle* pipe_handle, int n) {
 
 template <typename T>
 static size_t daliTensorSizeHelper(dali::DeviceWorkspace* ws, int n) {
-  return ws->Output<T>(n).nbytes();
+  return ws->Output<T>(n).total_nbytes();
 }
 
 size_t daliTensorSize(daliPipelineHandle* pipe_handle, int n) {
--- a/dali/c_api/c_api.cc
+++ b/dali/c_api/c_api.cc
@@ -130,6 +130,7 @@ void SetExternalInputTensors(daliPipelineHandle *pipe_handle, const char *name,
     layout = dali::TensorLayout(layout_str);
   }
   dali::TensorVector<Backend> data(curr_batch_size);
+  data.set_pinned(flags & DALI_ext_pinned);
   auto type_id = static_cast<dali::DALIDataType>(data_type);
   auto elem_sizeof = dali::TypeTable::GetTypeInfo(type_id).size();
 
@@ -138,6 +139,7 @@ void SetExternalInputTensors(daliPipelineHandle *pipe_handle, const char *name,
     order = AccessOrder(stream);
   else
     order = AccessOrder::host();
+
 
   for (int i = 0; i < curr_batch_size; i++) {
     // We cast away the const from data_ptr, as there is no other way of passing it to the
--- a/dali/operators/audio/nonsilence_op.h
+++ b/dali/operators/audio/nonsilence_op.h
@@ -228,8 +228,8 @@ class NonsilenceOperatorCpu : public NonsilenceOperator<CPUBackend> {
                   args.reset_interval = reset_interval_;
 
                   auto res = DetectNonsilenceRegion(intermediate_buffers_[thread_id], args);
-                  auto beg_ptr = output_begin.mutable_tensor<int>(sample_id);
-                  auto len_ptr = output_length.mutable_tensor<int>(sample_id);
+                  auto *beg_ptr = output_begin.mutable_tensor<int>(sample_id);
+                  auto *len_ptr = output_length.mutable_tensor<int>(sample_id);
                   *beg_ptr = res.first;
                   *len_ptr = res.second;
               }, in_shape.tensor_size(sample_id));
--- a/dali/operators/audio/nonsilence_op.h
+++ b/dali/operators/audio/nonsilence_op.h
@@ -1,4 +1,4 @@
-// Copyright (c) 2020-2021, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+// Copyright (c) 2020-2022, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -228,8 +228,8 @@ class NonsilenceOperatorCpu : public NonsilenceOperator<CPUBackend> {
                   args.reset_interval = reset_interval_;
 
                   auto res = DetectNonsilenceRegion(intermediate_buffers_[thread_id], args);
-                  auto beg_ptr = output_begin[sample_id].mutable_data<int>();
-                  auto len_ptr = output_length[sample_id].mutable_data<int>();
+                  auto beg_ptr = output_begin.mutable_tensor<int>(sample_id);
+                  auto len_ptr = output_length.mutable_tensor<int>(sample_id);
                   *beg_ptr = res.first;
                   *len_ptr = res.second;
               }, in_shape.tensor_size(sample_id));
--- a/dali/operators/audio/preemphasis_filter_op.cc
+++ b/dali/operators/audio/preemphasis_filter_op.cc
@@ -1,4 +1,4 @@
-// Copyright (c) 2019-2021, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+// Copyright (c) 2019-2022, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -65,11 +65,11 @@ void PreemphasisFilterCPU::RunImplTyped(workspace_t<CPUBackend> &ws) {
   for (int sample_id = 0; sample_id < nsamples; sample_id++) {
     tp.AddWork(
       [this, &output, &input, sample_id](int thread_id) {
-        const auto in_ptr = input[sample_id].data<InputType>();
-        auto out_ptr = output[sample_id].mutable_data<OutputType>();
-        DALI_ENFORCE(input[sample_id].shape() == output[sample_id].shape(),
+        const auto in_ptr = input.tensor<InputType>(sample_id);
+        auto out_ptr = output.mutable_tensor<OutputType>(sample_id);
+        DALI_ENFORCE(input.tensor_shape(sample_id) == output.tensor_shape(sample_id),
                      "Input and output shapes don't match");
-        auto n = volume(output[sample_id].shape());
+        auto n = volume(output.tensor_shape(sample_id));
         auto coeff = preemph_coeff_[sample_id];
         if (coeff == 0.0f) {
           for (int64_t j = 0; j < n; j++) {
--- a/dali/operators/audio/preemphasis_filter_op.cc
+++ b/dali/operators/audio/preemphasis_filter_op.cc
@@ -65,8 +65,8 @@ void PreemphasisFilterCPU::RunImplTyped(workspace_t<CPUBackend> &ws) {
   for (int sample_id = 0; sample_id < nsamples; sample_id++) {
     tp.AddWork(
       [this, &output, &input, sample_id](int thread_id) {
-        const auto in_ptr = input.tensor<InputType>(sample_id);
-        auto out_ptr = output.mutable_tensor<OutputType>(sample_id);
+        const auto *in_ptr = input.tensor<InputType>(sample_id);
+        auto *out_ptr = output.mutable_tensor<OutputType>(sample_id);
         DALI_ENFORCE(input.tensor_shape(sample_id) == output.tensor_shape(sample_id),
                      "Input and output shapes don't match");
         auto n = volume(output.tensor_shape(sample_id));
--- a/dali/operators/decoder/nvjpeg/nvjpeg_decoder_decoupled_api.h
+++ b/dali/operators/decoder/nvjpeg/nvjpeg_decoder_decoupled_api.h
@@ -852,6 +852,7 @@ class nvJPEGDecoder : public Operator<MixedBackend>, CachedDecoderImpl {
       TensorVector<CPUBackend> tv(samples_hw_batched_.size());
 
       const auto &input = ws.Input<CPUBackend>(0);
+      tv.SetupLike(input);
       for (auto *sample : samples_hw_batched_) {
         int i = sample->sample_idx;
         const auto &out_shape = output_shape_.tensor_shape(i);
--- a/dali/operators/decoder/nvjpeg/nvjpeg_decoder_decoupled_api.h
+++ b/dali/operators/decoder/nvjpeg/nvjpeg_decoder_decoupled_api.h
@@ -1,4 +1,4 @@
-// Copyright (c) 2019-2021, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+// Copyright (c) 2019-2022, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -554,15 +554,16 @@ class nvJPEGDecoder : public Operator<MixedBackend>, CachedDecoderImpl {
     samples_jpeg2k_.clear();
 #endif  // NVJPEG2K_ENABLED
 
+    const auto &input = ws.Input<CPUBackend>(0);
     for (int i = 0; i < curr_batch_size; i++) {
-      const auto &in = ws.Input<CPUBackend>(0)[i];
-      const auto in_size = in.size();
-      thread_pool_.AddWork([this, i, &in, in_size](int tid) {
-        auto *input_data = in.data<uint8_t>();
+      auto *input_data = input.tensor<uint8_t>(i);
+      const auto in_size = input.tensor_shape(i).num_elements();
+      const auto &source_info = input.GetMeta(i).GetSourceInfo();
+      thread_pool_.AddWork([this, i, input_data, in_size, source_info](int tid) {
         SampleData &data = sample_data_[i];
         data.clear();
         data.sample_idx = i;
-        data.file_name = in.GetSourceInfo();
+        data.file_name = source_info;
         data.encoded_length = in_size;
 
         auto cached_shape = CacheImageShape(data.file_name);
@@ -704,15 +705,17 @@ class nvJPEGDecoder : public Operator<MixedBackend>, CachedDecoderImpl {
 
   void ProcessImagesCuda(MixedWorkspace &ws) {
     auto& output = ws.Output<GPUBackend>(0);
+    const auto &input = ws.Input<CPUBackend>(0);
     for (auto *sample : samples_single_) {
       assert(sample);
       auto i = sample->sample_idx;
       auto *output_data = output.mutable_tensor<uint8_t>(i);
-      const auto &in = ws.Input<CPUBackend>(0)[i];
+      const auto *in_data = input.tensor<uint8_t>(i);
+      const auto in_size = input.tensor_shape(i).num_elements();
       thread_pool_.AddWork(
-        [this, sample, &in, output_data](int tid) {
-          SampleWorker(sample->sample_idx, sample->file_name, in.size(), tid,
-            in.data<uint8_t>(), output_data, streams_[tid]);
+        [this, sample, in_data, in_size, output_data](int tid) {
+          SampleWorker(sample->sample_idx, sample->file_name, in_size, tid,
+            in_data, output_data, streams_[tid]);
         }, task_priority_seq_--);  // FIFO order, since the samples were already ordered
     }
   }
@@ -808,15 +811,17 @@ class nvJPEGDecoder : public Operator<MixedBackend>, CachedDecoderImpl {
   }
 
   void ProcessImagesHost(MixedWorkspace &ws) {
+    const auto &input = ws.Input<CPUBackend>(0);
     auto& output = ws.Output<GPUBackend>(0);
     for (auto *sample : samples_host_) {
       auto i = sample->sample_idx;
+      const auto *input_data = input.tensor<uint8_t>(i);
+      auto in_size = input.tensor_shape(i).num_elements();
       auto *output_data = output.mutable_tensor<uint8_t>(i);
-      const auto &in = ws.Input<CPUBackend>(0)[i];
       ImageCache::ImageShape shape = output_shape_[i].to_static<3>();
       thread_pool_.AddWork(
-        [this, sample, &in, output_data, shape](int tid) {
-          HostFallback<StorageGPU>(in.data<uint8_t>(), in.size(), output_image_type_, output_data,
+        [this, sample, input_data, in_size, output_data, shape](int tid) {
+          HostFallback<StorageGPU>(input_data, in_size, output_image_type_, output_data,
                                    streams_[tid], sample->file_name, sample->roi, use_fast_idct_);
           CacheStore(sample->file_name, output_data, shape, streams_[tid]);
         }, task_priority_seq_--);  // FIFO order, since the samples were already ordered
@@ -846,13 +851,13 @@ class nvJPEGDecoder : public Operator<MixedBackend>, CachedDecoderImpl {
       int j = 0;
       TensorVector<CPUBackend> tv(samples_hw_batched_.size());
 
+      const auto &input = ws.Input<CPUBackend>(0);
       for (auto *sample : samples_hw_batched_) {
         int i = sample->sample_idx;
-        const auto &in = ws.Input<CPUBackend>(0)[i];
         const auto &out_shape = output_shape_.tensor_shape(i);
 
-        tv[j].ShareData(const_cast<Tensor<CPUBackend> &>(in));
-        in_lengths_[j] = in.size();
+        tv.UnsafeSetSample(j, input, i);
+        in_lengths_[j] = input.tensor_shape(i).num_elements();
         nvjpeg_destinations_[j].channel[0] = output.mutable_tensor<uint8_t>(i);
         nvjpeg_destinations_[j].pitch[0] = out_shape[1] * out_shape[2];
         nvjpeg_params_[j] = sample->params;
--- a/dali/operators/generic/constant.cc
+++ b/dali/operators/generic/constant.cc
@@ -1,4 +1,4 @@
-// Copyright (c) 2020-2021, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+// Copyright (c) 2020-2022, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -80,7 +80,7 @@ void FillTensorVector(
   assert(is_uniform(shape));
   int64_t n = shape[0].num_elements();
   assert(src.size() == static_cast<size_t>(n) || src.size() == 1);
-  Dst *out = dst[0].mutable_data<Dst>();
+  Dst *out = dst.mutable_tensor<Dst>(0);
   if (src.size() == 1) {
     Dst val = ConvertSat<Dst>(src[0]);
     for (int64_t i = 0; i < n; i++) {
@@ -92,7 +92,7 @@ void FillTensorVector(
     }
   }
   for (int i = 1; i < shape.num_samples(); i++) {
-    dst[i].ShareData(dst[0]);
+    dst.UnsafeSetSample(i, dst, 0);
   }
 }
 }  // namespace
@@ -116,7 +116,7 @@ void Constant<CPUBackend>::RunImpl(HostWorkspace &ws) {
   out.Resize(output_shape_);
   int N = output_shape_.num_samples();
   for (int i = 0; i < N; i++) {
-    assert(out[i].raw_data() == output_[i].raw_data());
+    assert(out.raw_tensor(i) == output_.raw_tensor(i));
   }
   out.SetLayout(layout_);
 }
--- a/dali/operators/generic/erase/erase_utils.h
+++ b/dali/operators/generic/erase/erase_utils.h
@@ -95,14 +95,14 @@ std::vector<kernels::EraseArgs<T, Dims>> GetEraseArgs(const OpSpec &spec,
 
   for (int i = 0; i < nsamples; i++) {
     if (has_tensor_roi_anchor) {
-      const auto& anchor = view<const float>(ws.ArgumentInput("anchor")[i]);
+      auto anchor = view<const float>(ws.ArgumentInput("anchor")[i]);
       assert(anchor.shape.num_elements() > 0);
       roi_anchor.resize(anchor.shape.num_elements());
       std::memcpy(roi_anchor.data(), anchor.data, sizeof(float) * roi_anchor.size());
     }
 
     if (has_tensor_roi_shape) {
-      const auto& shape = view<const float>(ws.ArgumentInput("shape")[i]);
+      auto shape = view<const float>(ws.ArgumentInput("shape")[i]);
       assert(shape.shape.num_elements() > 0);
       roi_shape.resize(shape.num_elements());
       std::memcpy(roi_shape.data(), shape.data, sizeof(float) * roi_shape.size());
--- a/dali/operators/generic/erase/erase_utils.h
+++ b/dali/operators/generic/erase/erase_utils.h
@@ -1,4 +1,4 @@
-// Copyright (c) 2019, NVIDIA CORPORATION. All rights reserved.
+// Copyright (c) 2019-2022, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -95,17 +95,17 @@ std::vector<kernels::EraseArgs<T, Dims>> GetEraseArgs(const OpSpec &spec,
 
   for (int i = 0; i < nsamples; i++) {
     if (has_tensor_roi_anchor) {
-      const auto& anchor = ws.ArgumentInput("anchor")[i];
-      assert(anchor.size() > 0);
-      roi_anchor.resize(anchor.size());
-      std::memcpy(roi_anchor.data(), anchor.data<float>(), sizeof(float) * roi_anchor.size());
+      const auto& anchor = view<const float>(ws.ArgumentInput("anchor")[i]);
+      assert(anchor.shape.num_elements() > 0);
+      roi_anchor.resize(anchor.shape.num_elements());
+      std::memcpy(roi_anchor.data(), anchor.data, sizeof(float) * roi_anchor.size());
     }
 
     if (has_tensor_roi_shape) {
-      const auto& shape = ws.ArgumentInput("shape")[i];
-      assert(shape.size() > 0);
-      roi_shape.resize(shape.size());
-      std::memcpy(roi_shape.data(), shape.data<float>(), sizeof(float) * roi_shape.size());
+      const auto& shape = view<const float>(ws.ArgumentInput("shape")[i]);
+      assert(shape.shape.num_elements() > 0);
+      roi_shape.resize(shape.num_elements());
+      std::memcpy(roi_shape.data(), shape.data, sizeof(float) * roi_shape.size());
     }
 
     DALI_ENFORCE(roi_anchor.size() == roi_shape.size());
--- a/dali/operators/generic/permute_batch.cc
+++ b/dali/operators/generic/permute_batch.cc
@@ -1,4 +1,4 @@
-// Copyright (c) 2020-2021, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+// Copyright (c) 2020-2022, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -45,7 +45,8 @@ void PermuteBatch<CPUBackend>::RunImpl(HostWorkspace &ws) {
     int src = indices_[i];
     tp.AddWork([&, i, src](int tid) {
       output.SetMeta(i, input.GetMeta(i));
-      output[i].Copy(input[src]);
+      // TODO(klecki): SetSample
+      output.UnsafeCopySample(i, input, src);
     }, size);
   }
   tp.RunAll();
--- a/dali/operators/generic/slice/slice_base.cc
+++ b/dali/operators/generic/slice/slice_base.cc
@@ -1,4 +1,4 @@
-// Copyright (c) 2019-2021, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+// Copyright (c) 2019-2022, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -58,7 +58,6 @@ bool SliceBaseCpu<OutputType, InputType, Dims>::SetupImpl(std::vector<OutputDesc
   output_desc[0].shape.resize(nsamples, Dims);
 
   kernels::KernelContext ctx;
-  auto in_view = view<const InputType, Dims>(input);
   for (int i = 0; i < nsamples; i++) {
     auto in_view = view<const InputType, Dims>(input[i]);
     auto req = Kernel().Setup(ctx, in_view, args_[i]);
--- a/dali/operators/image/convolution/gaussian_blur.cc
+++ b/dali/operators/image/convolution/gaussian_blur.cc
@@ -108,10 +108,10 @@ class GaussianBlurOpCpu : public OpImplBase<CPUBackend> {
       params_[i] = ObtainSampleParams<axes>(i, spec_, ws);
       windows_[i].PrepareWindows(params_[i]);
       // We take only last `ndim` siginificant dimensions to handle sequences as well
-      auto elem_shape = input[i].shape().template last<ndim>();
+      auto elem_shape = input.tensor_shape(i).template last<ndim>();
       auto& req = kmgr_.Setup<Kernel>(i, ctx_, elem_shape, params_[i].window_sizes);
       // The shape of data stays untouched
-      output_desc[0].shape.set_tensor_shape(i, input[i].shape());
+      output_desc[0].shape.set_tensor_shape(i, input.tensor_shape(i));
     }
     return true;
   }
@@ -125,7 +125,7 @@ class GaussianBlurOpCpu : public OpImplBase<CPUBackend> {
 
     int nsamples = input.shape().num_samples();
     for (int sample_idx = 0; sample_idx < nsamples; sample_idx++) {
-      const auto& shape = input[sample_idx].shape();
+      const auto& shape = input.tensor_shape(sample_idx);
       auto elem_volume = volume(shape.begin() + dim_desc_.usable_axes_start, shape.end());
 
       int seq_elements = 1;
@@ -138,11 +138,11 @@ class GaussianBlurOpCpu : public OpImplBase<CPUBackend> {
         thread_pool.AddWork(
             [this, &input, &output, sample_idx, elem_idx, stride](int thread_id) {
               auto gaussian_windows = windows_[sample_idx].GetWindows();
-              auto elem_shape = input[sample_idx].shape().template last<ndim>();
+              auto elem_shape = input.tensor_shape(sample_idx).template last<ndim>();
               auto in_view = TensorView<StorageCPU, const In, ndim>{
-                  input[sample_idx].template data<In>() + stride * elem_idx, elem_shape};
+                  input.template tensor<In>(sample_idx) + stride * elem_idx, elem_shape};
               auto out_view = TensorView<StorageCPU, Out, ndim>{
-                  output[sample_idx].template mutable_data<Out>() + stride * elem_idx, elem_shape};
+                  output.template mutable_tensor<Out>(sample_idx) + stride * elem_idx, elem_shape};
               // I need a context for that particular run (or rather matching the thread &
               // scratchpad)
               auto ctx = ctx_;
--- a/dali/operators/image/distortion/jpeg_compression_distortion_op_gpu.cu
+++ b/dali/operators/image/distortion/jpeg_compression_distortion_op_gpu.cu
@@ -71,7 +71,7 @@ void JpegCompressionDistortionGPU::RunImpl(workspace_t<GPUBackend> &ws) {
   // Set quality argument for an image from samples
   if (is_sequence) {
     for (int i = 0; i < nsamples; i++) {
-      auto nframes = input.tensor_shape_span(i)[0];
+      auto nframes = input.shape().tensor_shape_span(i)[0];
       for (int j = 0; j < nframes; ++j) {
         quality_.push_back(quality_arg_[i].data[0]);
       }
--- a/dali/operators/image/distortion/jpeg_compression_distortion_op_gpu.cu
+++ b/dali/operators/image/distortion/jpeg_compression_distortion_op_gpu.cu
@@ -70,8 +70,9 @@ void JpegCompressionDistortionGPU::RunImpl(workspace_t<GPUBackend> &ws) {
 
   // Set quality argument for an image from samples
   if (is_sequence) {
+    const auto &in_shape = input.shape();
     for (int i = 0; i < nsamples; i++) {
-      auto nframes = input.shape().tensor_shape_span(i)[0];
+      int nframes = in_shape.tensor_shape_span(i)[0];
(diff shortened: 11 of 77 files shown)


klecki:
!build

klecki:
!build

klecki:
!build

klecki:
!build

klecki:
!build

klecki:
!build

klecki:
There is some issue with CPU-only test, I will debug it tomorrow, the rest works fine.

dali/benchmark/operator_bench.h
```diff
@@ -54,16 +54,13 @@ class OperatorBench : public DALIBenchmark {
...
-      in_ptr->SetLayout("HWC");
-    }
+    data_in->set_type<T>();
+    data_in->Resize(uniform_list_shape(batch_size, TensorShape<>{H, W, C}));
+    data_in->SetLayout("HWC");
 
     if (fill_in_data) {
-      for (auto &in_ptr : *data_in) {
-        auto *ptr = in_ptr->template mutable_data<T>();
+      for (int sample_id = 0; sample_id < batch_size; sample_id++) {
```
mzient, line 62:
```suggestion
      for (int sample_idx = 0; sample_idx < batch_size; sample_idx++) {
```
`id` suggest it's something more than a ordinal.

dali/pipeline/data/tensor_vector.cc
```diff
@@ -62,11 +63,112 @@ TensorVector<Backend>::TensorVector(TensorVector<Backend> &&other) noexcept {
...
-size_t TensorVector<Backend>::nbytes() const noexcept {
+void TensorVector<Backend>::UnsafeSetSample(int dst, const TensorVector<Backend> &owner, int src) {
+  // TODO(klecki): more consistency checks, contiguous -> non-contiguous removes shares_data from
+  // samples
+  if (type() == DALI_NO_TYPE && owner.type() != DALI_NO_TYPE) {
+    set_type(owner.type());
+  }
+  if (!order()) {
+    set_order(owner.order());
+  }
```
mzient, line 75: That's a footgun... The sequence `if (!order()) set_order(...)` is fine, but not in something that affects just one sample. I'm afraid it's another incarnation of a side-effect control flow, like allocation in `mutable_data`, which we got rid of.

dali/pipeline/data/tensor_vector.cc
```diff
@@ -62,11 +63,112 @@ TensorVector<Backend>::TensorVector(TensorVector<Backend> &&other) noexcept {
...
+
+  SetContiguous(false);
+  // Setting a new share overwrites the previous one - so we can safely assume that even if
+  // we had a sample sharing into TL, it will be overwritten
+  tensors_[dst]->ShareData(*owner.tensors_[src]);
+  tl_->Reset();
+}
+
+template <typename Backend>
+void TensorVector<Backend>::UnsafeSetSample(int dst, const Tensor<Backend> &owner) {
```
mzient, line 100:
```suggestion
void TensorVector<Backend>::UnsafeSetSample(int sample_idx, const Tensor<Backend> &owner) {
```

dali/pipeline/data/tensor_vector.cc
```diff
@@ -62,11 +63,112 @@ TensorVector<Backend>::TensorVector(TensorVector<Backend> &&other) noexcept {
   other.tensors_.clear();
 }
 
-
 template <typename Backend>
-size_t TensorVector<Backend>::nbytes() const noexcept {
+void TensorVector<Backend>::UnsafeSetSample(int dst, const TensorVector<Backend> &owner, int src) {
```
mzient, line 67:
```suggestion
void TensorVector<Backend>::UnsafeSetSample(int sample_idx, const TensorVector<Backend> &owner, int src_sample_idx) {
```

dali/pipeline/data/tensor_vector.cc
```diff
@@ -62,11 +63,112 @@ TensorVector<Backend>::TensorVector(TensorVector<Backend> &&other) noexcept {
...
+template <typename Backend>
+void TensorVector<Backend>::UnsafeSetSample(int dst, const Tensor<Backend> &owner) {
+  // TODO(klecki): more consistency checks, contiguous -> non-contiguous removes shares_data from
+  // samples
+  if (type() == DALI_NO_TYPE && owner.type() != DALI_NO_TYPE) {
+    set_type(owner.type());
+  }
+  if (!order()) {
+    set_order(owner.order());
+  }
```
mzient, line 108: Likewise - global side-effect on something that looks like it affects just 1 sample.

dali/pipeline/data/tensor_vector.cc
```diff
@@ -62,11 +63,112 @@ TensorVector<Backend>::TensorVector(TensorVector<Backend> &&other) noexcept {
...
+  // Bounds check
+  assert(dst >= 0 && dst < static_cast<int>(curr_tensors_size_));
```
mzient, line 77: Can we _please_ rename this into something that isn't misread quite as easily?
```suggestion
  assert(dst >= 0 && dst < static_cast<int>(curr_num_tensors_));
```
?
A bounds check should be self-explanatory... Also, `TensorVector` is the _only_ structure in our code base where `num_samples` is a `size_t`, not an `int`.

dali/pipeline/data/tensor_vector.cc
klecki: My intention was as following:
* we keep the source of through in batch object (for now type & order, next PR will move the shape, dimensionality etc)
* if the current order is empty (for the batch object), and someone sets a sample, we get the order from it and use it from now on.
* if we have some order already and we set a new sample, the order of new sample must match, so we don't have one order for batch and different one for samples.

So I can just make it hard error if the order that is currently set for batch doesn't match the one for the new sample?

dali/operators/audio/nonsilence_op.h
```diff
@@ -228,8 +228,8 @@ class NonsilenceOperatorCpu : public NonsilenceOperator<CPUBackend> {
                   args.reset_interval = reset_interval_;
 
                   auto res = DetectNonsilenceRegion(intermediate_buffers_[thread_id], args);
-                  auto beg_ptr = output_begin[sample_id].mutable_data<int>();
-                  auto len_ptr = output_length[sample_id].mutable_data<int>();
+                  auto beg_ptr = output_begin.mutable_tensor<int>(sample_id);
+                  auto len_ptr = output_length.mutable_tensor<int>(sample_id);
```
mzient, line 232: Nitpick:
```suggestion
                  int *beg_ptr = output_begin.mutable_tensor<int>(sample_id);
                  int *len_ptr = output_length.mutable_tensor<int>(sample_id);
```
It's not like we have a long or complicated type here.

dali/operators/audio/preemphasis_filter_op.cc
```diff
@@ -65,11 +65,11 @@ void PreemphasisFilterCPU::RunImplTyped(workspace_t<CPUBackend> &ws) {
   for (int sample_id = 0; sample_id < nsamples; sample_id++) {
     tp.AddWork(
       [this, &output, &input, sample_id](int thread_id) {
-        const auto in_ptr = input[sample_id].data<InputType>();
-        auto out_ptr = output[sample_id].mutable_data<OutputType>();
-        DALI_ENFORCE(input[sample_id].shape() == output[sample_id].shape(),
+        const auto in_ptr = input.tensor<InputType>(sample_id);
```
mzient, line 68: This is actually wrong. We don't want the pointer to be constant. We want a pointer to a constant object:
```suggestion
        const auto *in_ptr = input.tensor<InputType>(sample_id);
```

dali/operators/audio/preemphasis_filter_op.cc
```diff
@@ -65,11 +65,11 @@ void PreemphasisFilterCPU::RunImplTyped(workspace_t<CPUBackend> &ws) {
...
+        auto out_ptr = output.mutable_tensor<OutputType>(sample_id);
```
mzient, line 69: Nitpick:
```suggestion
        auto *out_ptr = output.mutable_tensor<OutputType>(sample_id);
```
(for completeness)

dali/operators/generic/permute_batch.cc
```diff
@@ -45,7 +45,8 @@ void PermuteBatch<CPUBackend>::RunImpl(HostWorkspace &ws) {
     int src = indices_[i];
     tp.AddWork([&, i, src](int tid) {
       output.SetMeta(i, input.GetMeta(i));
-      output[i].Copy(input[src]);
+      // TODO(klecki): SetSample
+      output.UnsafeCopySample(i, input, src);
```
mzient, line 49: Why does it need to be "unsafe"? At least in case of PermuteBatch, we know the shape, so we can actualy presize the output and safely copy the samples.

dali/operators/generic/slice/slice_base.cc
```diff
@@ -58,7 +58,6 @@ bool SliceBaseCpu<OutputType, InputType, Dims>::SetupImpl(std::vector<OutputDesc
   output_desc[0].shape.resize(nsamples, Dims);
 
   kernels::KernelContext ctx;
-  auto in_view = view<const InputType, Dims>(input);
```
mzient, line 61: :+1:

dali/operators/image/convolution/gaussian_blur.cc
```diff
@@ -138,11 +138,11 @@ class GaussianBlurOpCpu : public OpImplBase<CPUBackend> {
         thread_pool.AddWork(
             [this, &input, &output, sample_idx, elem_idx, stride](int thread_id) {
               auto gaussian_windows = windows_[sample_idx].GetWindows();
-              auto elem_shape = input[sample_idx].shape().template last<ndim>();
+              auto elem_shape = input.tensor_shape(sample_idx).template last<ndim>();
```
mzient, line 141: Pff... this will conflict with per-sample argument PR.

dali/operators/image/peek_shape/peek_image_shape.h
```diff
@@ -62,10 +64,9 @@ class PeekImageShape : public Operator<CPUBackend> {
   }
 
   template <typename type>
-  void WriteShape(Tensor<CPUBackend> &out, const TensorShape<3> &shape) {
-    type *data = out.mutable_data<type>();
+  void WriteShape(TensorView<StorageCPU, type, 1> out, const TensorShape<3> &shape) {
```
mzient, line 67: Not related to this PR, but it could be a template against `ndim` and use a run-time number of dimensions in the loop. This would help should we ever consider extending the functionality to volumes.

dali/operators/image/peek_shape/peek_image_shape.h
```diff
@@ -74,20 +75,22 @@ class PeekImageShape : public Operator<CPUBackend> {
...
         const auto& image = input[sample_id];
         // Verify input
-        DALI_ENFORCE(image.ndim() == 1,
+        // TODO(klecki): Move the checks to scope above
+        DALI_ENFORCE(image.shape().sample_dim() == 1,
                       "Input must be 1D encoded jpeg string.");
-        DALI_ENFORCE(IsType<uint8>(image.type()),
-                      "Input must be stored as uint8 data.");
-        auto img = ImageFactory::CreateImage(image.data<uint8>(), image.size(), {});
+        DALI_ENFORCE(image.type() == DALI_UINT8, "Input must be stored as uint8 data.");
```
mzient, line 87: While you're at it:
```suggestion
        DALI_ENFORCE(image.type() == DALI_UINT8, "The input must be raw, undecoded files stored as a flat uint8 array.");
```

dali/operators/generic/permute_batch.cc
klecki: I just wanted to discourage people from using it left and right. Especially with copy, there should be nothing wrong with it, but the full intended functionality is not there yet, we just handle what we already mostly do with `tv[i].Copy(tv[j])`

dali/operators/image/remap/displacement_filter_impl_gpu.cuh
```diff
@@ -325,7 +325,7 @@ class DisplacementFilter<GPUBackend, Displacement,
       sample.input = input.template tensor<T>(sample_idx);
       sample.raw_params = GetDisplacementParams(sample_idx);
       sample.shape = shape.tensor_shape<nDims>(sample_idx);
-      sample.mask = has_mask_ ? ws.ArgumentInput("mask")[sample_idx].data<int>()[0] : true;
+      sample.mask = has_mask_ ? (ws.ArgumentInput("mask").tensor<int>(sample_idx))[0] : true;
```
mzient, line 328: The extra parenthesis is confusing.
```suggestion
      sample.mask = has_mask_ ? ws.ArgumentInput("mask").tensor<int>(sample_idx)[0] : true;
```

dali/operators/python_function/dltensor_function.cc
```diff
@@ -77,9 +77,9 @@ py::list PrepareDLTensorInputs<CPUBackend>(HostWorkspace &ws) {
   py::list input_tuple;
   for (Index idx = 0; idx < ws.NumInput(); ++idx) {
     py::list dl_tensor_list;
+    auto &tvec = ws.UnsafeMutableInput<CPUBackend>(idx);
```
mzient, line 80:
```suggestion
    auto &input = ws.UnsafeMutableInput<CPUBackend>(idx);
```
Meaning over type.

dali/operators/reader/loader/nemo_asr_loader.h
```diff
@@ -158,7 +158,7 @@ class DLL_PUBLIC NemoAsrLoader : public Loader<CPUBackend, AsrSample> {
 
  private:
   template <typename OutputType>
-  void ReadAudio(Tensor<CPUBackend> &audio,
+  void ReadAudio(SampleView<CPUBackend> audio,
```
mzient, line 161: Const-ref? To consider.

dali/operators/decoder/nvjpeg/nvjpeg_decoder_decoupled_api.h
```diff
@@ -554,15 +554,16 @@ class nvJPEGDecoder : public Operator<MixedBackend>, CachedDecoderImpl {
...
 #endif  // NVJPEG2K_ENABLED
 
+    const auto &input = ws.Input<CPUBackend>(0);
     for (int i = 0; i < curr_batch_size; i++) {
-      const auto &in = ws.Input<CPUBackend>(0)[i];
-      const auto in_size = in.size();
-      thread_pool_.AddWork([this, i, &in, in_size](int tid) {
-        auto *input_data = in.data<uint8_t>();
+      auto *input_data = input.tensor<uint8_t>(i);
+      const auto in_size = input.tensor_shape(i).num_elements();
```
mzient, line 560: BEEP! Creating a temporary shape! Please restore a way to calculate a sample volume without creating a temporary shape object.

dali/operators/generic/erase/erase_utils.h
```diff
@@ -95,17 +95,17 @@ std::vector<kernels::EraseArgs<T, Dims>> GetEraseArgs(const OpSpec &spec,
 
   for (int i = 0; i < nsamples; i++) {
     if (has_tensor_roi_anchor) {
-      const auto& anchor = ws.ArgumentInput("anchor")[i];
-      assert(anchor.size() > 0);
-      roi_anchor.resize(anchor.size());
-      std::memcpy(roi_anchor.data(), anchor.data<float>(), sizeof(float) * roi_anchor.size());
+      const auto& anchor = view<const float>(ws.ArgumentInput("anchor")[i]);
```
mzient, line 98:
```suggestion
      auto anchor = view<const float>(ws.ArgumentInput("anchor")[i]);
```

dali/operators/reader/numpy_reader_gpu_op.cc
```diff
@@ -113,9 +113,9 @@ void NumpyReaderGPU::Prefetch() {
   }
   curr_tensor_list.Resize(tmp_shapes, ref_type);
 
-  size_t chunk_size = static_cast<size_t>( \
-                        div_ceil(static_cast<uint64_t>(curr_tensor_list.nbytes()),
-                                 static_cast<uint64_t>(thread_pool_.NumThreads())));
+  size_t chunk_size =
+      static_cast<size_t>(div_ceil(static_cast<uint64_t>(curr_tensor_list.total_nbytes()),
+                                   static_cast<uint64_t>(thread_pool_.NumThreads())));
```
mzient, line 118:
```suggestion
      div_ceil(static_cast<uint64_t>(curr_tensor_list.total_nbytes()),
               static_cast<uint64_t>(thread_pool_.NumThreads()));
```
Superfluous cast.

dali/operators/reader/numpy_reader_op.cc
```diff
@@ -14,20 +14,22 @@
...
 #include "dali/kernels/slice/slice_flip_normalize_permute_pad_cpu.h"
 #include "dali/kernels/transpose/transpose.h"
 #include "dali/core/static_switch.h"
 #include "dali/operators/reader/numpy_reader_op.h"
+#include "dali/pipeline/data/backend.h"
 
 namespace dali {
 
-static void CopyHelper(Tensor<CPUBackend> &output, const Tensor<CPUBackend> &input,
+static void CopyHelper(SampleView<CPUBackend> output, ConstSampleView<CPUBackend> input,
```
mzient, line 27: How about a meaningful name?
```suggestion
static void ParallelCopy(SampleView<CPUBackend> output, ConstSampleView<CPUBackend> input,
```

dali/operators/reader/numpy_reader_op.cc
```diff
@@ -45,7 +47,7 @@ static void CopyHelper(Tensor<CPUBackend> &output, const Tensor<CPUBackend> &inp
   }
 }
 
-static void TransposeHelper(Tensor<CPUBackend> &output, const Tensor<CPUBackend> &input) {
+static void TransposeHelper(SampleView<CPUBackend> output, ConstSampleView<CPUBackend> input) {
```
mzient, line 50: Helperpocalypse continued....
```suggestion
static void ReverseStorageOrder(SampleView<CPUBackend> output, ConstSampleView<CPUBackend> input) {
```

dali/operators/reader/numpy_reader_op.cc
```diff
@@ -77,7 +79,7 @@ static void SliceHelper(Tensor<CPUBackend> &output, const Tensor<CPUBackend> &in
   ), DALI_FAIL(make_string("Unsupported number of dimensions: ", ndim)););  // NOLINT
 }
 
-static void SlicePermuteHelper(Tensor<CPUBackend> &output, const Tensor<CPUBackend> &input,
+static void SlicePermuteHelper(SampleView<CPUBackend> output, ConstSampleView<CPUBackend> input,
```
mzient, line 82: Now that you're at it...
```suggestion
static void SliceTransposeHelper(SampleView<CPUBackend> output, ConstSampleView<CPUBackend> input,
```
We use "Tranpose" in the other function and it is a better name for a total reversal of axes.

dali/operators/segmentation/random_object_bbox.cc
```diff
@@ -533,7 +533,7 @@ void RandomObjectBBox::AllocateTempStorage(const TensorVector<CPUBackend> &input
   int64_t max_filtered_bytes = 0;
   int N = input.num_samples();
   for (int i = 0; i < N; i++) {
-    int64_t vol = input[i].size();
+    int64_t vol = input[i].shape().num_elements();
```
mzient, line 536:
```suggestion
    int64_t vol = input[i].shape_size(i);
```

dali/operators/segmentation/random_object_bbox.cc
```diff
@@ -592,10 +592,10 @@ void RandomObjectBBox::RunImpl(HostWorkspace &ws) {
       // We want to limit the size of this auxiliary storage to limit memory traffic.
       // To that end, when the indices fit in int32_t, we use that type for the labels,
       // otherwise we fall back to int64_t.
-      auto blob_label = (input[i].size() > 0x80000000) ? DALI_INT64 : DALI_INT32;
+      auto blob_label = (input[i].shape().num_elements() > 0x80000000) ? DALI_INT64 : DALI_INT32;
```
mzient, line 595:
```suggestion
      auto blob_label = (input.tensor_size(i) > 0x80000000) ? DALI_INT64 : DALI_INT32;
```

dali/operators/segmentation/random_object_bbox.h
```diff
@@ -192,7 +192,7 @@ class RandomObjectBBox : public Operator<CPUBackend> {
 
     ThreadPool *thread_pool = nullptr;
     TensorView<StorageCPU, int> out1, out2;
-    const Tensor<CPUBackend> *input = nullptr;
+    ConstSampleView<CPUBackend> input = {};
```
mzient, line 195:
```suggestion
    ConstSampleView<CPUBackend> input{};
```
I know it _should_ be the same, but, for some reason, I noticed that the compiler has a nasty habit of making a copy (in debug build).

dali/operators/signal/fft/spectrogram.cc
```diff
@@ -227,8 +227,8 @@ bool SpectrogramImplCpu<time_major>::SetupImpl(std::vector<OutputDesc> &out_desc
 
   auto view_window_fn = make_tensor_cpu<1>(window_fn_.data(), window_length_);
   for (int i = 0; i < nsamples; i++) {
-    auto view_signal_1d =
-        make_tensor_cpu<1>(input[i].template data<const InputType>(), {input[i].size()});
+    auto view_signal_1d = make_tensor_cpu<1>(input.template tensor<const InputType>(i),
+                                             {input.tensor_shape(i).num_elements()});
```
mzient, line 231:
```suggestion
                                             {input.tensor_size(i)});
```

dali/operators/signal/fft/spectrogram.cc
```diff
@@ -269,8 +269,8 @@ void SpectrogramImplCpu<time_major>::RunImpl(workspace_t<CPUBackend> &ws) {
         win_out.set_type<InputType>();
         win_out.Resize(window_out_desc_[0].shape.tensor_shape(i));
 
-        auto view_signal_1d =
-            make_tensor_cpu<1>(input[i].data<const InputType>(), {input[i].size()});
+        auto view_signal_1d = make_tensor_cpu<1>(input.tensor<const InputType>(i),
+                                                 {input.tensor_shape(i).num_elements()});
```
mzient, line 273:
```suggestion
                                                 {input.tensor_size(i)});
```

klecki:
!build

dali/pipeline/data/sample_view.h
```diff
@@ -125,24 +126,79 @@ class SampleView {
...
 
   /**
    * @brief Construct the view with explicitly provided type_id.
    */
-  SampleView(void *data, const TensorShape<> shape, DALIDataType type_id)
+  SampleViewBase(ptr_t data, const TensorShape<> shape, DALIDataType type_id)
       : data_(data), shape_(std::move(shape)), type_id_(type_id) {}
 
- private:
+ protected:
```
mzient, line 140: Please add an empty protected destructor.

dali/pipeline/data/sample_view_test.cc
```diff
@@ -53,13 +52,45 @@ TEST(SampleView, Constructors) {
...
+  const int32_t cdata{};
+  ConstSampleView<CPUBackend> const_from_ptr{&cdata, {1, 2, 3}};
+  compare(const_from_ptr, &cdata, {1, 2, 3}, DALI_INT32);
+
+  ConstSampleView<CPUBackend> const_from_void_ptr{
+      reinterpret_cast<const void *>(42), {1, 2, 3}, DALI_FLOAT};
+  compare(const_from_void_ptr, reinterpret_cast<void *>(42), {1, 2, 3}, DALI_FLOAT);
+
+  ConstSampleView<CPUBackend> const_from_nonconst{from_ptr};
+  compare(const_from_nonconst, &data, {1, 2, 3}, DALI_INT32);
```
mzient, line 65: For completeness - plain copy (for both const and non-const) and perhaps move.

dali/pipeline/data/tensor.h
```diff
@@ -422,6 +422,10 @@ class Tensor : public Buffer<Backend> {
     return *this;
   }
 
+  DALIMeta &GetMeta() {
+    return meta_;
+  }
```
mzient, line 427: Do we really need this? It seems quite dangerous. I know we do have a standalone DALIMeta object, but setting the metadata for a tensor/batch can benefit from extra validation (e.g. against dimensionality/size/type/...). Returning a non-const reference to DALIMeta makes such extra validation impossible.

dali/pipeline/data/tensor_vector.h
```diff
@@ -176,6 +258,7 @@ class DLL_PUBLIC TensorVector {
 
   TensorLayout GetLayout() const;
 
+  DALIMeta &GetMeta(int idx);
```
mzient, line 261: Looks dangerous.

dali/pipeline/data/tensor_vector.cc
mzient: > we keep the source of *truth* (?) in batch object (for now type & order, next PR will move the shape, dimensionality etc)

What would it look like? We'd require the parent object to remain alive and unchanged? This is a big requrimeent that would invalidate any and all attempts at having an owning SampleView. Chasing pointers is not a good idea performancewise. Also, this would make the view object potentially inconsistent - we keep a _copy_ of a data pointer, so we should keep a _copy_ of the metadata - at least the most essential, which is shape and type.

dali/pipeline/operator/op_spec_test.cc
```diff
@@ -74,7 +74,7 @@ TEST(OpSpecTest, GetArgumentTensorSet) {
     auto tv = std::make_shared<TensorVector<CPUBackend>>(2);
     tv->Resize(TensorListShape<0>(2), DALI_INT32);
     for (int i = 0; i < 2; i++) {
-      tv->tensor_handle(i)->mutable_data<int32_t>()[0] = 42 + i;
+      view<int32_t>((*tv)[i]).data[0] = 42 + i;
```
mzient, line 77:
```suggestion
      tv->mutable_tensor<int32_t>(i)[0] = 42 + i;
```
?

dali/pipeline/workspace/sample_workspace.h
```diff
@@ -134,6 +134,17 @@ class DLL_PUBLIC SampleWorkspace : public WorkspaceBase<SampleInputType, SampleO
...
+/**
+ * @brief After running sample-wise operator we need to fix the Tensor Vector guarantees that were
+ * broken by the legacy operators operating just on samples. We propagate the properties from
+ * samples and enforce they are consistent.
+ *
+ * TODO(klecki): Introduce RAII wrapper for MakeSampleView and EnforceCorrectness
+ * @param batch The workspace to update after executing samplewise operator
+ * @param contiguous If the operator infers outputs and thus uses contiguous allocations
+ */
+DLL_PUBLIC void EnforceCorrectness(HostWorkspace& ws, bool contiguous);
```
mzient, line 146: This name is non-descriptive. Does it _enforce_ or _propagate_? What is propagated? What is the result of this function?

dali/pipeline/workspace/sample_workspace.cc
```diff
@@ -48,4 +48,10 @@ void MakeSampleView(SampleWorkspace& sample, HostWorkspace& batch, int data_idx,
   }
 }
 
+void EnforceCorrectness(HostWorkspace& ws, bool contiguous) {
+  for (int i = 0; i < ws.NumOutput(); i++) {
+    ws.Output<CPUBackend>(i).PropagateUp(contiguous);
```
mzient, line 53: Propagate what?

dali/pipeline/data/tensor_vector.cc
```diff
@@ -440,10 +586,41 @@ void TensorVector<Backend>::resize_tensors(int new_size) {
...
     }
+    // TODO(klecki): Do not keep the invalidated tensors - this prevents memory hogging but
+    // also gets rid of reserved memory.
+    // tensors_.resize(new_size);
   }
   curr_tensors_size_ = new_size;
 }
 
+template <typename Backend>
+void TensorVector<Backend>::PropagateUp(bool contiguous) {
```
mzient, line 597: PropagateUp... what? What does it mean "up"? This calls for a better name.

dali/test/dali_test.h
```diff
@@ -186,7 +186,7 @@ class DALITest : public ::testing::Test {
     for (int i = 0; i < n; ++i) {
       std::memcpy(tl->template mutable_tensor<uint8>(i), data[i % nImgs],
                   data_sizes[i % nImgs]);
-      tl->SetSourceInfo(i, imgs.filenames_[i % nImgs]);
+      tl->GetMeta(i).SetSourceInfo(imgs.filenames_[i % nImgs]);
```
mzient, line 189: I have a baaaad feeling about this.

dali/test/dali_test_decoder.h
```diff
@@ -32,7 +32,8 @@ class GenericDecoderTest : public DALISingleOpTest<ImgType> {
     // single input - encoded images
     // single output - decoded images
 
-     TensorVector<CPUBackend> out(inputs[0]->num_samples());
+    TensorVector<CPUBackend> out(inputs[0]->num_samples());
+    std::vector<Tensor<CPUBackend>> tmp_out(inputs[0]->num_samples());
```
mzient, line 36: This doesn't look good...

klecki:
!build

dali/operators/image/convolution/gaussian_blur.cc
klecki: As will the changes in TensorVector - we will resolve.

dali/pipeline/data/tensor_vector.cc
```diff
@@ -90,6 +200,33 @@ size_t TensorVector<Backend>::capacity() const noexcept {
   return total_capacity;
 }
 
+template <typename Backend>
+std::vector<size_t> TensorVector<Backend>::nbytes() const noexcept {
```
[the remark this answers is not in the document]
klecki: It is useful in the memory statistics gathering, which I wanted to preserve, without giving access to the actual stored samples. It is rather rare operation so I don't think being expensive is a problem. 
You have to manually turn on gathering the statistics so I would not worry about that.

dali/pipeline/data/tensor_vector.cc
klecki: PropagateUpTheMetadataFromTheIndividuallySetSamplesToTheBatchObject, just like the [Constable Visit](https://wiki.lspace.org/Visit-The-Infidel-With-Explanatory-Pamphlets)

dali/benchmark/operator_bench.h
klecki: done

dali/pipeline/data/tensor_vector.cc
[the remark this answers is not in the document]
mzient: Could you point me to the usage? In any case - if it's only for statistics, then how about having a free function somewhere in the statistics code? I think it looks too inconspicuous for what it does - "nbytes" sounds like a simple query, possibly inviting misuse like calling it in a loop over samples.

dali/pipeline/data/tensor_vector.cc
mzient: Which direction is "Up"? Without your explanation, it's anything but obvious.
How about:
```suggestion
void TensorVector<Backend>::UpdateMetaFromSamples(bool contiguous) {
```
This clearly says that it's
a) metadata
b) from samples
c) that metadata is updated (presumably the main metadata as opposed to per-sample)

dali/operators/audio/preemphasis_filter_op.cc
klecki: I agree and I tend to add the `*`, just missed it.

dali/operators/audio/preemphasis_filter_op.cc
klecki: done

dali/pipeline/data/sample_view.h
```diff
@@ -43,23 +43,24 @@ class SampleView {
...
+  template <typename ptr_t_ = ptr_t>
+  std::enable_if_t<std::is_same<ptr_t_, void *>::value, void *> _raw_mutable_data() {
     return data_;
   }
 
   /**
    * @brief Return a const, un-typed pointer to the underlying storage.
    */
-  const void *raw_data() const {
+  const void *_raw_data() const {
```
stiepan, line 54: Why introducing the underscores?

stiepan, line 54: Why introducing the underscores?

klecki: I added a section about naming in the description.

stiepan: Oh, sure that's clever.

dali/pipeline/data/tensor_vector.cc
klecki: I moved away all of properties setup from the SetSample/CopySample

dali/pipeline/data/tensor_vector.cc
klecki: done

dali/pipeline/data/tensor_vector.cc
klecki: I prefer to keep the consistent for the documentation purposes.

dali/pipeline/data/tensor_vector.cc
klecki: Removed.

dali/operators/image/distortion/jpeg_compression_distortion_op_gpu.cu
```diff
@@ -71,7 +71,7 @@ void JpegCompressionDistortionGPU::RunImpl(workspace_t<GPUBackend> &ws) {
   // Set quality argument for an image from samples
   if (is_sequence) {
     for (int i = 0; i < nsamples; i++) {
-      auto nframes = input.tensor_shape_span(i)[0];
+      auto nframes = input.shape().tensor_shape_span(i)[0];
```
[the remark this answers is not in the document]
klecki: done

dali/operators/image/peek_shape/peek_image_shape.h
klecki: I hope this won't be the line that makes it 1k+

dali/operators/image/remap/displacement_filter_impl_gpu.cuh
klecki: done

dali/operators/python_function/dltensor_function.cc
klecki: done

dali/operators/python_function/dltensor_function.cc
```diff
@@ -106,8 +106,8 @@ py::list PrepareDLTensorInputsPerSample<CPUBackend>(HostWorkspace &ws) {
   for (Index s = 0; s < batch_size; ++s) {
     py::list tuple;
     for (Index idx = 0; idx < ws.NumInput(); ++idx) {
-      auto &t = ws.UnsafeMutableInput<CPUBackend>(idx)[s];
-      auto dl_capsule = TensorToDLPackView(t);
+      auto &tvec = ws.UnsafeMutableInput<CPUBackend>(idx);
+      auto dl_capsule = TensorToDLPackView(tvec[s], tvec.device_id());
```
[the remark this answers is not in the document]
klecki: Come on, `input`

dali/c_api/c_api.cc
```diff
@@ -143,11 +143,14 @@ void SetExternalInputTensors(daliPipelineHandle *pipe_handle, const char *name,
     // We cast away the const from data_ptr, as there is no other way of passing it to the
     // Tensor as we must also set the shape and type metadata.
     // The vector that we pass to pipeline is const.
-    data[i].set_pinned(flags & DALI_ext_pinned);
-    data[i].set_order(order);
-    data[i].ShareData(const_cast<void *>(data_ptr[i]), tl_shape[i].num_elements() * elem_sizeof);
-    data[i].Resize(tl_shape[i], type_id);
-    data[i].SetLayout(layout);
+    dali::Tensor<Backend> tmp;
+    tmp.set_order(order);
```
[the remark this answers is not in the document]
klecki: Done

dali/operators/generic/constant.cc
```diff
@@ -92,7 +92,7 @@ void FillTensorVector(
     }
   }
   for (int i = 1; i < shape.num_samples(); i++) {
-    dst[i].ShareData(dst[0]);
+    dst.UnsafeSetSample(i, dst, 0);
```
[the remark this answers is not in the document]
klecki: Added explanation it the PR description.

dali/operators/generic/erase/erase_utils.h
klecki: done

dali/operators/reader/numpy_reader_gpu_op.cc
klecki: done

dali/operators/reader/numpy_reader_op.cc
klecki: Do we really need to inflate the number of comments for this reason :upside_down_face:

dali/operators/reader/numpy_reader_op.cc
klecki: Does it mean that all the functions that also utilize the thread pool require a Parallel prefix instead of Helper suffix?

dali/operators/reader/numpy_reader_op.cc
klecki: I can see the name comes from the kernels used. I won't be applying as it seems out of scope to discuss the naming of those functions in this PR.

dali/operators/segmentation/random_object_bbox.h
klecki: Removed.

dali/pipeline/data/sample_view.h
klecki: done

dali/pipeline/data/sample_view.h
```diff
@@ -125,24 +126,79 @@ class SampleView {
...
+};
+
+
+template <typename Backend>
+class ConstSampleView : public SampleViewBase<Backend, const void *> {
+ public:
+  using Base = SampleViewBase<Backend, const void *>;
+  using Base::Base;
+
+  explicit ConstSampleView(const SampleView<Backend> &other)
```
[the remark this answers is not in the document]
klecki: done

dali/pipeline/data/sample_view_test.cc
klecki: Done

dali/pipeline/data/tensor_vector.cc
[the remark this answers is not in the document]
klecki: Executor::GetMaxSizesNonCont.

I can revert to the nbytes() being what it used to, and call this `individual_nbytes` or something.

dali/pipeline/operator/op_spec_test.cc
klecki: done

dali/pipeline/workspace/sample_workspace.h
klecki: I tried my best to fix it.

dali/pipeline/data/tensor_vector.cc
klecki: It's not just the `Meta`, this would be highly misleading. We have a lot of other, proper metadata/properties that are not a part of GetMeta/SetMeta.

dali/test/dali_test_decoder.h
klecki: already responded to that.

dali/test/dali_test.h
klecki: done

dali/pipeline/data/sample_view.h
```diff
@@ -33,8 +33,8 @@ namespace dali {
  * convenient `view<T, ndim>(SampleView)` conversion to TensorView, but doesn't break the batch
  * object encapsulation and doesn't allow to adjust the allocation.
  */
-template <typename Backend>
```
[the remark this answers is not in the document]
klecki: Yes, just a type tag that is the same thing as Tensor<Backend> etc.

dali/pipeline/data/tensor_list.h
```diff
@@ -605,6 +604,10 @@ class DLL_PUBLIC TensorList {
     return meta_[idx].ShouldSkipSample();
   }
 
+  inline DALIMeta &GetMeta(int idx) {
```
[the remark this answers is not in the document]
klecki: Right, but on the other hand I don't like the fact that we are replicating the DALIMeta API here. Don't see a better way of doing it right now, so I will revert. (Maybe a `ModifyMeta(<DALIMeta -> DALIMeta callback>)` ? :P)

dali/pipeline/data/tensor_vector.cc
klecki: 37 occurrences in main branch, I don't want to inflate the number of lines in this PR. It can be done separately.

dali/pipeline/data/tensor.h
klecki: Removed. As I said, I wanted to direct all calls through the DALIMeta api rather than repeat it in the TensorVector, but I am doing exactly the opposite with the access to tensors, so it doesn't make sense.

dali/pipeline/workspace/sample_workspace.cc
klecki: done

dali/operators/decoder/nvjpeg/nvjpeg_decoder_decoupled_api.h
klecki: There is no `tensor_size` in any of our batch objects now, and I am not sure about introducing it vs keeping the shape access explicit and dedicated to shape.

klecki:
!build

dali/operators/image/peek_shape/peek_image_shape.h
```diff
@@ -75,16 +75,13 @@ class PeekImageShape : public Operator<CPUBackend> {
     const auto &input = ws.template Input<CPUBackend>(0);
     auto &output = ws.template Output<CPUBackend>(0);
     size_t batch_size = input.num_samples();
-    DALI_ENFORCE(input.type() == DALI_UINT8, "Input must be stored as uint8 data.");
+    DALI_ENFORCE(input.type() == DALI_UINT8,
+                 "The input must be raw, undecoded files stored as a flat uint8 arrays.");
+    DALI_ENFORCE(input.sample_dim() == 1, "Input must be 1D encoded jpeg string.");
```
mzient, line 80:
```suggestion
    DALI_ENFORCE(input.sample_dim() == 1, "Input must be 1D encoded JPEG bit stream.");
```
?

dali/operators/image/peek_shape/peek_image_shape.h
```diff
@@ -75,16 +75,13 @@ class PeekImageShape : public Operator<CPUBackend> {
     const auto &input = ws.template Input<CPUBackend>(0);
     auto &output = ws.template Output<CPUBackend>(0);
     size_t batch_size = input.num_samples();
-    DALI_ENFORCE(input.type() == DALI_UINT8, "Input must be stored as uint8 data.");
+    DALI_ENFORCE(input.type() == DALI_UINT8,
+                 "The input must be raw, undecoded files stored as a flat uint8 arrays.");
```
mzient, line 79:
```suggestion
                 "The input must be raw, undecoded files stored as flat uint8 arrays.");
```
or
```suggestion
                 "The input must be a raw, undecoded file stored as a flat uint8 array.");
```

dali/operators/reader/numpy_reader_gpu_op.cc
```diff
@@ -97,10 +97,10 @@ void NumpyReaderGPU::Prefetch() {
 
   const size_t kGDSChunkGranularity = 1<<20;
 
-  size_t chunk_size = align_up(div_ceil(static_cast<uint64_t>(curr_tensor_list.total_nbytes()),
+  size_t chunk_size = align_up(div_ceil(static_cast<uint64_t>(curr_tensor_list.chunks_nbytes()),
```
mzient: chunks_nbytes returns a vector - how is that supposed to work?

klecki: This is from an in-between commit.

dali/operators/image/peek_shape/peek_image_shape.h
klecki: done.

dali/operators/image/peek_shape/peek_image_shape.h
klecki: Done, the second time :P

dali/pipeline/data/sample_view_test.cc
```diff
@@ -63,6 +63,18 @@ TEST(SampleView, Constructors) {
...
+  compare(const_from_const, &cdata, {1, 2, 3}, DALI_INT32);
+
+  SampleView<CPUBackend> nonconst_from_nonconst{from_ptr};
+  compare(nonconst_from_nonconst, &data, {1, 2, 3}, DALI_INT32);
+
+  SampleView<CPUBackend> nonconst_from_moved_nonconst{std::move(from_ptr)};
+  compare(nonconst_from_moved_nonconst, &data, {1, 2, 3}, DALI_INT32);
+
+  ConstSampleView<CPUBackend> const_from_moved_const{std::move(const_from_nonconst)};
+  compare(const_from_moved_const, &cdata, {1, 2, 3}, DALI_INT32);
```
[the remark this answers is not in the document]
klecki: done

klecki:
!build

dali/operators/reader/numpy_reader_gpu_op.cc
```diff
@@ -102,7 +102,6 @@ void NumpyReaderGPU::Prefetch() {
                                kGDSChunkGranularity);
   chunk_size = std::min(chunk_size, curr_tensor_list.nbytes());
 
-
```
mzient, line 105: Why touch this file at all?

dali/operators/reader/numpy_reader_op.cc
mzient: I do strongly believe that a function's name should explain what the function does (within reason). "Helper" adds very little if anything. Having taken another look at this function, it should probably be "ScheduleCopy", because the copy operations are scheduled and there's no `wait` at the end.

dali/operators/reader/numpy_reader_op.cc
mzient: Whatever.

dali/operators/reader/numpy_reader_gpu_op.cc
klecki: I had to adjust during rebase and then adjust it back, I guess the line was removed in the process. Do you miss it?

dali/pipeline/data/tensor_vector.cc
```diff
@@ -60,20 +65,170 @@ TensorVector<Backend>::TensorVector(TensorVector<Backend> &&other) noexcept {
...
+      "TensorVector and SetSize for desired number of samples in non-contiguous mode.");
+
+  tensors_[sample_idx]->Copy(*src.tensors_[src_sample_idx], order);
+}
+
+
+template <typename Backend>
+void TensorVector<Backend>::SetupLike(const Tensor<Backend> &sample) {
+  DALI_ENFORCE(!has_data(),
+               "Batch object can be initialized this way only when it doesn't have allocation.");
```
mzient, line 194:
```suggestion
               "Batch object can be initialized this way only when it doesn't have an allocation.");
```
or (probably better):
```suggestion
               "Batch object can be initialized this way only when it isn't allocated");
```

dali/pipeline/data/tensor_vector.cc
```diff
@@ -60,20 +65,170 @@ TensorVector<Backend>::TensorVector(TensorVector<Backend> &&other) noexcept {
...
+  set_type(sample.type());
+  set_sample_dim(sample.shape().sample_dim());
+  SetLayout(sample.GetLayout());
```
mzient, line 197: When changing the APIs, how about changing it to `set_layout`/`get_layout`?

dali/pipeline/data/tensor_vector.h
```diff
@@ -78,72 +88,122 @@ class DLL_PUBLIC TensorVector {
...
   int sample_dim() const {
-    return IsContiguous() ? tl_->sample_dim() : num_samples() ? tensors_[0]->shape().size() : 0;
+    return sample_dim_;
   }
 
   size_t nbytes() const noexcept;
 
   size_t capacity() const noexcept;
 
+  std::vector<size_t> chunks_nbytes() const noexcept;
```
mzient, line 113: Aren't those "chunks" mapped directly to samples? Why not `sample_nbytes`?

dali/pipeline/data/tensor_vector.h
```diff
@@ -78,72 +88,122 @@ class DLL_PUBLIC TensorVector {
...
+
+  std::vector<size_t> chunks_capacity() const noexcept;
```
mzient, line 115: Likewise.

dali/pipeline/data/tensor_vector.h
```diff
@@ -160,6 +220,16 @@ class DLL_PUBLIC TensorVector {
...
 
+  /**
+   * @brief Setup all the batch properties of this TensorVector the same way as the provided tensor:
+   *
+   * Precondition: the TensorVector should not have data.
+   * Configures: type, layout, pinned, order and dimensionality.
+   */
+  void SetupLike(const Tensor<Backend> &sample);
+
+  void SetupLike(const TensorVector<Backend> &other);
```
mzient, line 231: Why isn't it a template? The function body is the same (or almost so). How about TensorList? Do we really need the argument to be on the same backend?

dali/test/dali_test.h
```diff
@@ -1,4 +1,4 @@
-// Copyright (c) 2017-2021, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
mzient, line 1: Revert?

dali/pipeline/data/sample_view.h
```diff
@@ -125,24 +126,80 @@ class SampleView {
...
+      : Base(other._raw_data(), other.shape(), other.type()) {}
+
+  ConstSampleView &operator=(const SampleView<Backend> &other) {
+    data_ = other.data();
+    shape_ = other.shape();
+    type_id_ = other.type();
+    return *this;
+  }
+
+  explicit ConstSampleView(SampleView<Backend> &&other) {
```
stiepan, line 180: Why make this one `explict`?

dali/pipeline/data/tensor_list_test.cc
```diff
@@ -1,4 +1,4 @@
-// Copyright (c) 2017-2021, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+// Copyright (c) 2017-2022, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
stiepan, line 1: Looks like that update is not needed after all?

dali/pipeline/data/tensor_vector.h
klecki: They are mapped to allocations - if we have just one in contiguous mode I return that one.

dali/pipeline/data/tensor_vector.h
klecki: This is mainly used for the purpose of preparing the object explicitly for `SetSample` - you won't `SetSample` across backends, using the same pinned configuration across backends when copying also doesn't make sense.

dali/pipeline/data/tensor_list_test.cc
klecki: Done, but let me complain that the script that I use make it hard.

dali/test/dali_test.h
klecki: done, but only due to 1k loc threshold.

dali/pipeline/data/tensor_vector.cc
klecki: done

dali/pipeline/data/tensor_vector.cc
```diff
@@ -60,20 +65,170 @@ TensorVector<Backend>::TensorVector(TensorVector<Backend> &&other) noexcept {
...
+  set_order(sample.order());
+  set_pinned(sample.is_pinned());
+}
+
+template <typename Backend>
+void TensorVector<Backend>::SetupLike(const TensorVector<Backend> &other) {
+  DALI_ENFORCE(!has_data(),
+               "Batch object can be initialized this way only when it doesn't have allocation.");
```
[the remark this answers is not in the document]
klecki: done

dali/pipeline/data/tensor_vector.h
klecki: We can have one chunk as mentioned.

dali/pipeline/data/sample_view.h
klecki: done

dali/pipeline/data/sample_view.h
```diff
@@ -125,24 +126,80 @@ class SampleView {
...
+class ConstSampleView : public SampleViewBase<Backend, const void *> {
+ public:
+  using Base = SampleViewBase<Backend, const void *>;
+  using Base::Base;
+
+  ConstSampleView(const SampleView<Backend> &other)  // NOLINT
+      : Base(other._raw_data(), other.shape(), other.type()) {}
+
+  ConstSampleView &operator=(const SampleView<Backend> &other) {
+    data_ = other.data();
```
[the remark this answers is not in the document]
klecki: Good spot. Added a test.

dali/pipeline/operator/builtin/external_source.h
```diff
@@ -254,8 +254,10 @@ class ExternalSource : public Operator<Backend>, virtual public BatchSizeProvide
     DeviceGuard g(device_id_);
     DomainTimeRange tr("[DALI][ExternalSource] SetDataSource", DomainTimeRange::kViolet);
     TensorVector<SrcBackend> tv(vect_of_tensors.size());
+    assert(vect_of_tensors.size() > 0);
```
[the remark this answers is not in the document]
klecki: I think this function might be dead code. Let me check :upside_down_face:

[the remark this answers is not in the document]
klecki: Nope, looks like it's used only in tests. Not really user facing API.

klecki:
!build

klecki:
!build

dali/pipeline/executor/executor.h
```diff
@@ -170,14 +170,14 @@ class DLL_PUBLIC Executor : public ExecutorBase, public QueuePolicy {
     }
   }
 
-  template <typename Backend>
-  inline void GetMaxSizes(TensorList<Backend> &in, size_t &max_out_size,
+  template<typename backend>
```
mzient, line 173: Please revert. Type template parameters should start with a capital letter.

dali/pipeline/data/tensor_vector.h
mzient: OK with the backend - how about code duplication? I'd suggest this:

```C++
  void SetupLike(const Tensor<Backend> &sample) {
    SetupLikeImpl(sample);
  }
  void SetupLike(const TensorVector<Backend> &batch) {
    SetupLikeImpl(sample);
  }
  void SetupLike(const TensorList<Backend> &batch) {
    SetupLikeImpl(sample);
  }

 private:
  template <typename T>
  void SetupLikeImpl(const T &src) { do stuff }
```

klecki: I can look into that. Will this be all?

dali/pipeline/data/tensor_vector.h
```diff
@@ -78,72 +88,130 @@ class DLL_PUBLIC TensorVector {
...
 
   size_t nbytes() const noexcept;
 
   size_t capacity() const noexcept;
 
+  /**
+   * @brief Returns the size in bytes of the underlying data chunks
+   * TODO(klecki): Temporary API to be reworked, do not use.
+   */
+  std::vector<size_t> _chunks_nbytes() const noexcept;
```
mzient, line 117:
```suggestion
  std::vector<size_t> _chunks_nbytes() const;
```
After all, a `vector` is dynamically allocated...

dali/pipeline/data/tensor_vector.h
```diff
@@ -78,72 +88,130 @@ class DLL_PUBLIC TensorVector {
...
+
+  /**
+   * @brief Returns the real size of the underlying allocations
+   * TODO(klecki): Temporary API to be reworked, do not use.
+   */
+  std::vector<size_t> _chunks_capacity() const noexcept;
```
mzient, line 123: Likewise.

dali/pipeline/data/tensor_vector.cc
```diff
@@ -60,20 +65,147 @@ TensorVector<Backend>::TensorVector(TensorVector<Backend> &&other) noexcept {
...
+  DALI_ENFORCE(sample_dim() == shape.sample_dim(),
+               make_string("Sample must have the same dimensionality as a target batch, current: ",
+                           sample_dim(), " new: ", shape.sample_dim(), " for ", sample_idx, "."));
+  DALI_ENFORCE(this->order() == order, "Sample must have the same order as a target batch");
+  DALI_ENFORCE(GetLayout() == layout,
+               make_string("Sample must have the same layout as a target batch current: ",
+                           GetLayout(), " new: ", layout, " for ", sample_idx, "."));
+  DALI_ENFORCE(is_pinned() == pinned,
+               make_string("Sample must have the same pinned status as target batch, current: ",
+                           is_pinned(), " new: ", pinned, " for ", sample_idx, "."));
```
stiepan, line 149: Would it make sense to place it in a separate method and reuse across all Unsafe(...) methods? Kind of like with `SetupLikeImpl`?

mzient (approved):

dali/pipeline/data/tensor_vector.h
klecki: done

dali/pipeline/data/tensor_vector.h
klecki: done

dali/pipeline/operator/builtin/external_source.h
```diff
@@ -253,9 +254,11 @@ class ExternalSource : public Operator<Backend>, virtual public BatchSizeProvide
                             AccessOrder order = {}, ExtSrcSettingMode ext_src_setting_mode = {}) {
     DeviceGuard g(device_id_);
     DomainTimeRange tr("[DALI][ExternalSource] SetDataSource", DomainTimeRange::kViolet);
+    DALI_ENFORCE(vect_of_tensors.size() > 0, "Provided batch cannot be empty.");
```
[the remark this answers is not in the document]
klecki: I am not sure if DALI would even work with empty batch.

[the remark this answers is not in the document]
mzient: I think it would have broken big time with an empty batch before.

dali/operators/reader/nemo_asr_reader_op.cc
```diff
@@ -183,18 +185,21 @@ void NemoAsrReader::Prefetch() {
...
   const auto &sample = GetSample(ws.data_idx());
-  const auto &sample_audio = GetDecodedAudioSample(ws.data_idx());
+  auto sample_audio = GetDecodedAudioSample(ws.data_idx());
 
   auto &audio = ws.Output<CPUBackend>(0);
-  audio.Copy(sample_audio);
+  audio.Resize(sample_audio.shape(), sample_audio.type());
+  std::memcpy(
+      audio.raw_mutable_data(), sample_audio.raw_data(),
+      sample_audio.shape().num_elements() * TypeTable::GetTypeInfo(sample_audio.type()).size());
```
stiepan, line 202: Isn't it that GetTypeInfo introduces some locking while we could get this from, let sat GetDecodedAudioSample() for free?

klecki: Hmm, I can get it via the type info of the `audio` that we just resized.

PR #3713, opened 2022-03-01
Fix stream usage in C API

mzient:
## Category:

## Description:
- Stream value of 0 should no longer be converted to an AccessOrder in C API.
- Made C API tests that don't use CUDA usable without a CUDA driver (most CPU Backend tests - exclusing those that use copy kernel or both backends).
- Adjust cpplint to support C++17 structured bindings with space in `auto [a, b]`

### Tests
- [X] Existing tests apply (modified)
- [ ] New tests added
  - [ ] Python tests
  - [ ] GTests
  - [ ] Benchmark
  - [ ] Other
- [ ] N/A

### Documentation
- [ ] Documentation updated
  - [ ] Docstring
  - [ ] Doxygen
  - [ ] RST
  - [ ] Jupyter
  - [ ] Other
- [ ] N/A

#### Requirements
- [X] N/A

**REQ IDs**: N/A

**JIRA TASK**: N/A

### DIFF
--- a/dali/c_api/c_api_test.cc
+++ b/dali/c_api/c_api_test.cc
@@ -962,7 +962,7 @@ TYPED_TEST(CApiTest, daliOutputCopySamples) {
   daliDeletePipeline(&handle);
 }
 
-TYPED_TEST(CApiTest, CpuOnlyTest) {
+TEST(CApiTest, CpuOnlyTest) {
   dali::Pipeline pipe(1, 1, dali::CPU_ONLY_DEVICE_ID);
   pipe.AddExternalInput("dummy");
   std::vector<std::pair<std::string, std::string>> outputs = {{"dummy", "cpu"}};
--- a/dali/c_api/c_api_test.cc
+++ b/dali/c_api/c_api_test.cc
@@ -69,9 +69,27 @@ std::string GetDeviceStr(device_type_t dev) {
   return dev == CPU ? "cpu" : "gpu";
 }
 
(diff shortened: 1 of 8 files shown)


dali/c_api/c_api_test.cc
```diff
@@ -69,9 +69,27 @@ std::string GetDeviceStr(device_type_t dev) {
   return dev == CPU ? "cpu" : "gpu";
 }
 
+// Allocates a buffer on the specified backend and, if necessary, another one for the CPU
```
JanuszL, line 72: The only test we run from this file without GPU is CpuOnlyTest - see https://github.com/NVIDIA/DALI/blob/main/qa/TL0_cpu_only/test.sh#L34.
Maybe we should improve the pattern there - or change test naming here to run more CPU only test?

qa/TL0_cpu_only/test.sh
```diff
@@ -31,7 +31,7 @@ test_body() {
         exit 1
     fi
 
-    "$FULLPATH" --gtest_filter="*CpuOnlyTest*"
+    "$FULLPATH" --gtest_filter="*CpuOnly*:*CApi*/0.*-*0.UseCopyKernel:*ForceNoCopyFail:*daliOutputCopySamples"
```
JanuszL, line 34: Does `0` mean CPU?

mzient: Yes - unfortunately the name of the TypeParam is not a part of the name, so you can't match against that.

JanuszL (approved):

PR #3730, opened 2022-03-14
Optimization of tiled transposition algorithm on small data types

00max09:
## Category:

## Description:

This PR tries to pack continous reads and writes of small data types (smaller than 32bit) to smaller number of reads and writes using bigger data types to fill the memory bus.
Tiled transpose algorithm often has last row not being move from it's place. In old implementation if for example there was transposition of RGBA image (for example 1920x1080x4 with 8bit values), algorithm did 4 8bit reads and writes from global memory and 4 8bit writes and reads from shared memory for every pixel.
With this optimalization if data of picture is aligned to address that is divisible by 4, for picture from example above algorithm does 1 32bit read and write from global memory and 1 32bit write and read from shared memory for every pixel.
In detail if last dim size in tensor is divisable by 4 it's type is 8bit and is properly aligned then algorithm does 4 times less reads and writes, in bit worse case if lasst dim size in tensor is divisable by 2, it's type is at most 16bit and data is properly aligned algorithm does 2 times less reads and writes. 
For benchmark I did transposition of 500x500x4 tensor with uint8 type with batch_size of 256 and 100 iterations (all on Titan V).
- Before optimalizations : 3111 us
- After optimalization : 2552 us

To show value of optimalization I changed tensor to 500x500x3 (representing RGB picture) and got time of 2742us which shows that after optimalization if memory is not a problem user can change shape of tensor to optimalize execution time
  
- Changed TiledTranspose code in transpose kernel
- Added two tests transpose kernel

### Tests
- [ ] Existing tests apply
- [x] New tests added
  - [ ] Python tests
  - [x] GTests
  - [ ] Benchmark
  - [ ] Other
- [ ] N/A

### Documentation
- [ ] Documentation updated
  - [ ] Docstring
  - [ ] Doxygen
  - [ ] RST
  - [ ] Jupyter
  - [ ] Other
- [ ] N/A

#### Requirements
- [ ] N/A

**REQ IDs**: N/A

**JIRA TASK**: N/A

### DIFF
--- a/dali/kernels/transpose/transpose_gpu_impl.cuh
+++ b/dali/kernels/transpose/transpose_gpu_impl.cuh
@@ -20,7 +20,7 @@
 #include "dali/core/fast_div.h"
 #include "dali/kernels/transpose/transpose_gpu_def.h"
 #include "dali/core/static_switch.h"
-
+#include "dali/kernels/common/type_erasure.h"
 /**
  * @file
  *
@@ -100,18 +100,17 @@ namespace transpose_shared {
   extern __shared__ uint8_t shared_tmp[];
 }  // namespace transpose_shared
 
-
-template <int ndim, typename T>
-__device__ void TransposeTiledStatic(TiledTransposeDesc<T> desc) {
+template <int ndim, typename T, typename OldT>
+__device__ inline void TransposeTiledPacked(TiledTransposeDesc<OldT> desc, const unsigned pack_ratio)
+{
   unsigned start_tile = blockIdx.x * desc.tiles_per_block;
   unsigned end_tile = min(desc.total_tiles, start_tile + desc.tiles_per_block);
   if (start_tile >= end_tile)
     return;
-
   T (*tmp)[kTileSize][kTileSize+1] =
       reinterpret_cast<T (*)[kTileSize][kTileSize+1]>(transpose_shared::shared_tmp);
 
-  int lanes = desc.lanes;
+  int lanes = desc.lanes >> pack_ratio;
 
   unsigned tile_in_slice = start_tile;
   uint64_t fused_slice = ndim > 2
@@ -129,17 +128,16 @@ __device__ void TransposeTiledStatic(TiledTransposeDesc<T> desc) {
   pos[ndim - 1] = tile_x * kTileSize;
   pos[ndim - 2] = tile_y * kTileSize;
 
-  T *out = desc.out;
-  const T *in = desc.in;
+  T *out = reinterpret_cast<T*> (desc.out);
+  const T *in = reinterpret_cast<const T*>(desc.in);
 
   for (uint64_t tile = start_tile;;) {
     uint64_t in_ofs = 0, out_ofs = 0;
     #pragma unroll
     for (int d = 0; d < ndim - 2; d++) {
-      in_ofs  += desc.in_strides[d] * pos[d];
-      out_ofs += desc.out_strides[d] * pos[d];
+      in_ofs  += desc.in_strides[d]/pack_ratio * pos[d];
+      out_ofs += desc.out_strides[d]/pack_ratio * pos[d];
     }
-
     int64_t in_x  = pos[ndim-1] + threadIdx.x;
     int64_t in_y  = pos[ndim-2] + threadIdx.y;
     int64_t out_x = pos[ndim-1] + threadIdx.y;
@@ -148,10 +146,15 @@ __device__ void TransposeTiledStatic(TiledTransposeDesc<T> desc) {
     in_ofs  += desc.in_strides[ndim-2]  * in_y  + desc.in_strides[ndim-1]  * in_x;
     out_ofs += desc.out_strides[ndim-2] * out_y + desc.out_strides[ndim-1] * out_x;
 
+    in_ofs >>= pack_ratio;
+    out_ofs >>= pack_ratio;
+
     unsigned tile_w = min(static_cast<uint64_t>(kTileSize), desc.shape[ndim-1] - pos[ndim-1]);
     unsigned tile_h = min(static_cast<uint64_t>(kTileSize), desc.shape[ndim-2] - pos[ndim-2]);
+    unsigned jump = blockDim.y >> pack_ratio;
+    
     if (threadIdx.x < tile_w) {
-      for (unsigned ty = threadIdx.y, dy = 0; ty < tile_h; ty += blockDim.y, dy += blockDim.y) {
+      for (unsigned ty = threadIdx.y, dy = 0; ty < tile_h; ty += blockDim.y, dy += jump) {
         #pragma unroll 4
         for (int lane = 0; lane < lanes; lane++) {
           tmp[lane][ty][threadIdx.x] = __ldg(&in[in_ofs + desc.in_strides[ndim-2]*dy + lane]);
@@ -161,7 +164,7 @@ __device__ void TransposeTiledStatic(TiledTransposeDesc<T> desc) {
     __syncthreads();
 
     if (threadIdx.x < tile_h) {
-      for (unsigned ty = threadIdx.y, dy = 0; ty < tile_w; ty += blockDim.y, dy += blockDim.y) {
+      for (unsigned ty = threadIdx.y, dy = 0; ty < tile_w; ty += blockDim.y, dy += jump) {
         #pragma unroll 4
         for (int lane = 0; lane < lanes; lane++)
           out[out_ofs + desc.out_strides[ndim-1]*dy + lane] = tmp[lane][threadIdx.x][ty];
@@ -181,6 +184,23 @@ __device__ void TransposeTiledStatic(TiledTransposeDesc<T> desc) {
     }
     __syncthreads();
   }
+}
+
+template <int ndim, typename T>
+__device__ void TransposeTiledStatic(TiledTransposeDesc<T> desc) {
+  if(sizeof(T) == 1 && desc.lanes % 4 == 0 && reinterpret_cast<uintptr_t>(desc.out) % 4 == 0 && 
+      reinterpret_cast<uintptr_t>(desc.in) % 4 == 0 ) {
+    return TransposeTiledPacked<ndim, type_of_size<4>>(desc, 2);
+  }
+  if(sizeof(T) < 4 && desc.lanes % 2 == 0 && reinterpret_cast<uintptr_t>(desc.out) % 2 == 0 && 
+      reinterpret_cast<uintptr_t>(desc.in) % 2 == 0) {
+    if (sizeof(T) == 2) { 
+      return TransposeTiledPacked<ndim, type_of_size<4>>(desc, 1);
+    } else {
+      return TransposeTiledPacked<ndim, type_of_size<2>>(desc, 1);
+    }
+  }
+  return TransposeTiledPacked<ndim, T>(desc, 0);
 }
 
 template <typename T>
--- a/dali/kernels/transpose/transpose_gpu_impl.cuh
+++ b/dali/kernels/transpose/transpose_gpu_impl.cuh
@@ -101,8 +101,7 @@ namespace transpose_shared {
 }  // namespace transpose_shared
 
 template <int ndim, unsigned pack_ratio_log, typename T, typename OldT>
-__device__ inline void TransposeTiledPacked(TiledTransposeDesc<OldT> desc)
-{
+__device__ inline void TransposeTiledPacked(TiledTransposeDesc<OldT> desc) {
   unsigned start_tile = blockIdx.x * desc.tiles_per_block;
   unsigned end_tile = min(desc.total_tiles, start_tile + desc.tiles_per_block);
   if (start_tile >= end_tile)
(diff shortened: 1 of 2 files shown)


banasraf:
Formatting is a little bit off in many places. Can you run clang format on this code?

dali/kernels/transpose/transpose_gpu_impl.cuh
```diff
@@ -129,51 +127,55 @@ __device__ void TransposeTiledStatic(TiledTransposeDesc<T> desc) {
...
     __syncthreads();
 
     if (threadIdx.x < tile_h) {
-      for (unsigned ty = threadIdx.y, dy = 0; ty < tile_w; ty += blockDim.y, dy += blockDim.y) {
-        #pragma unroll 4
+      for (unsigned ty = threadIdx.y, dy = 0; ty < tile_w; ty += blockDim.y, dy += jump) {
+#pragma unroll 4
         for (int lane = 0; lane < lanes; lane++)
-          out[out_ofs + desc.out_strides[ndim-1]*dy + lane] = tmp[lane][threadIdx.x][ty];
+          out[out_ofs + desc.out_strides[ndim - 1] * dy + lane] = tmp[lane][threadIdx.x][ty];
```
mzient, line 169: Please don't reformat unchanged code.

dali/kernels/transpose/transpose_gpu_impl.cuh
```diff
@@ -129,51 +127,55 @@ __device__ void TransposeTiledStatic(TiledTransposeDesc<T> desc) {
...
-    int64_t out_x = pos[ndim-1] + threadIdx.y;
-    int64_t out_y = pos[ndim-2] + threadIdx.x;
+    in_ofs >>= pack_ratio;
+    out_ofs >>= pack_ratio;
 
-    in_ofs  += desc.in_strides[ndim-2]  * in_y  + desc.in_strides[ndim-1]  * in_x;
-    out_ofs += desc.out_strides[ndim-2] * out_y + desc.out_strides[ndim-1] * out_x;
+    unsigned tile_w = min(static_cast<uint64_t>(kTileSize), desc.shape[ndim - 1] - pos[ndim - 1]);
+    unsigned tile_h = min(static_cast<uint64_t>(kTileSize), desc.shape[ndim - 2] - pos[ndim - 2]);
+    unsigned jump = blockDim.y >> pack_ratio;
```
mzient, line 153: Why shift, not divide? I don't quite follow.

dali/kernels/transpose/transpose_gpu_impl_test.cu
```diff
@@ -32,94 +32,87 @@ using namespace transpose_impl;  // NOLINT
 
 
 TEST(SimplifyPermute, NoSimplification) {
-  int64_t shape[] = { 2, 3, 4, 5 };
-  int perm[] = { 0, 3, 2, 1 };
+  int64_t shape[] = {2, 3, 4, 5};
+  int perm[] = {0, 3, 2, 1};
```
mzient, line 36: Don't add noise by reformatting entire files.

dali/kernels/transpose/transpose_gpu_impl.cuh
```diff
@@ -129,51 +127,55 @@ __device__ void TransposeTiledStatic(TiledTransposeDesc<T> desc) {
...
 
   for (uint64_t tile = start_tile;;) {
     uint64_t in_ofs = 0, out_ofs = 0;
-    #pragma unroll
+#pragma unroll
     for (int d = 0; d < ndim - 2; d++) {
-      in_ofs  += desc.in_strides[d] * pos[d];
-      out_ofs += desc.out_strides[d] * pos[d];
+      in_ofs += desc.in_strides[d] / pack_ratio * pos[d];
+      out_ofs += desc.out_strides[d] / pack_ratio * pos[d];
```
mzient, line 138: Shouldn't it be:
```suggestion
      in_ofs += (desc.in_strides[d] >> pack_ratio) * pos[d];
      out_ofs += (desc.out_strides[d] >> pack_ratio) * pos[d];
```
?

dali/kernels/transpose/transpose_gpu_impl_test.cu
```diff
@@ -254,10 +318,9 @@ TEST(TransposeGeneric, AllPerm4D) {
   ASSERT_LT(grid_size * block_size, size) << "Weak test error: Grid too large to test grid loop";
 
   for (auto &perm : testing::Permutations4) {
-    std::cerr << "Testing permutation "
-      << perm[0] << " " << perm[1] << " " << perm[2] << " " << perm[3] << "  input shape "
-      << shape << "\n";
-      CUDA_CALL(cudaMemset(out_gpu, 0xff, size*sizeof(int)));
+    std::cerr << "Testing permutation " << perm[0] << " " << perm[1] << " " << perm[2] << " "
+              << perm[3] << "  input shape " << shape << "\n";
```
mzient, line 322: The arguments here were carefully grouped. Autoformatter obliterated it.

mzient (changes requested):
* Verify that the in_ofs / out_ofs are calculated properly (almost certainly they aren't) and re-run benchmarks on the corrected code if they aren't.
* Please revert autoformatting of the whole files and only format what you've changed.

00max09:
I ran microbenchmarks (if I understand correctly you mean benchmarks from test file that I changed that measure throughput) and after moving pack_ratio to template argument I don't see any loss in performance(before moving you could maybe see a 1% diffrence but that could just be rng), although I don't have any script to get mean from runs so I checked it by eye.

dali/kernels/transpose/transpose_gpu_impl.cuh
00max09: reverted

dali/kernels/transpose/transpose_gpu_impl_test.cu
00max09: reverted

dali/kernels/transpose/transpose_gpu_impl.cuh
00max09: repaired the lines, and added test for bigger number of dimensions, as I think current tests didn't cover this area of code (because there was dividing by 0)

dali/kernels/transpose/transpose_gpu_impl_test.cu
00max09: reverted

dali/kernels/transpose/transpose_gpu_impl.cuh
00max09: re-ran benchmarks and results didn't change

dali/kernels/transpose/transpose_gpu_impl.cuh
```diff
@@ -100,9 +100,9 @@ namespace transpose_shared {
   extern __shared__ uint8_t shared_tmp[];
 }  // namespace transpose_shared
 
-
-template <int ndim, typename T>
-__device__ void TransposeTiledStatic(TiledTransposeDesc<T> desc) {
+template <int ndim, unsigned pack_ratio, typename T, typename OldT>
```
mzient, line 103: Nitpick: pack_ratio is not really a ratio, but rather a pack_ratio_log - renaming or a commend would be nice (but it's not a must).

mzient:
> I ran microbenchmarks (if I understand correctly you mean benchmarks from test file that I changed that measure throughput) and after moving pack_ratio to template argument I don't see any loss in performance(before moving you could maybe see a 1% diffrence but that could just be rng), although I don't have any script to get mean from runs so I checked it by eye.

Yes, I meant the ones from that file. As long as it runs, doesn't slow down and doesn't crash, I'm happy. I was afraid of extra register pressure this change might inflict.

mzient (approved):

dali/kernels/transpose/transpose_gpu_impl.cuh
00max09: changed

dali/kernels/transpose/transpose_gpu_impl.cuh
```diff
@@ -305,4 +327,4 @@ __global__ void TransposeGenericBatch(const GenericTransposeDesc<T> *descs) {
 }  // namespace kernels
 }  // namespace dali
 
-#endif  // DALI_KERNELS_TRANSPOSE_TRANSPOSE_GPU_IMPL_CUH_
+#endif  // DALI_KERNELS_TRANSPOSE_TRANSPOSE_GPU_IMPL_CUH_
```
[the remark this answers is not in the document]
00max09: restored it

dali/kernels/transpose/transpose_gpu_impl.cuh
```diff
@@ -100,9 +100,9 @@ namespace transpose_shared {
   extern __shared__ uint8_t shared_tmp[];
 }  // namespace transpose_shared
 
-
-template <int ndim, typename T>
-__device__ void TransposeTiledStatic(TiledTransposeDesc<T> desc) {
+template <int ndim, unsigned pack_ratio_log, typename T, typename OldT>
+__device__ inline void TransposeTiledPacked(TiledTransposeDesc<OldT> desc)
+{
```
banasraf, line 105: { to the previous line

dali/kernels/transpose/transpose_gpu_impl_test.cu
```diff
@@ -190,6 +189,111 @@ TEST(TransposeTiled, BuildDescVectorized) {
...
+  in_gpu.resize(size);
+  out_gpu.resize(size);
+    
+  for(uintptr_t offset = 0; offset < 4; offset++)
+  {
+    shape = { 57, 37, 53, 4 };
+    size = volume(shape);
+    std::iota(in_cpu.begin(), in_cpu.end(), 0);
+    CUDA_CALL(cudaMemset(out_gpu, 0xff, size*sizeof(int)));
+    shape = { 57, 37, 52, 4 };
```
stiepan, line 208: Do we need two shapes that differ by 1 in one of the extents of just using `shape = { 57, 37, 53, 4 }` and the allocation size `volume({57, 37, 53, 4 }) + 4` would be ok?

dali/kernels/transpose/transpose_gpu_impl.cuh
```diff
@@ -148,10 +148,15 @@ __device__ void TransposeTiledStatic(TiledTransposeDesc<T> desc) {
     in_ofs  += desc.in_strides[ndim-2]  * in_y  + desc.in_strides[ndim-1]  * in_x;
     out_ofs += desc.out_strides[ndim-2] * out_y + desc.out_strides[ndim-1] * out_x;
 
+    in_ofs >>= pack_ratio_log;
+    out_ofs >>= pack_ratio_log;
+
     unsigned tile_w = min(static_cast<uint64_t>(kTileSize), desc.shape[ndim-1] - pos[ndim-1]);
     unsigned tile_h = min(static_cast<uint64_t>(kTileSize), desc.shape[ndim-2] - pos[ndim-2]);
+    unsigned jump = blockDim.y >> pack_ratio_log;
```
stiepan, line 156: Can we be sure that it is at least 4?

stiepan: In fact, would keeping `dy` jumps as they were (and scaling here the `desc.in_strides[ndim-2]` instead) make a (performance) difference? Personally, I was a bit surprised at first seeing that the `y` jump changes with this optimization.

00max09: I checked it now and i think we can't be sure, but from my understanding in this scenario old code wouldn't work as intended (the same values would be copied more than one time). I don't know if MaxThreadsPerBlock function can in real scenario return something less than 64 (or rather 32*numeber_of_lanes) , if no then the code should be safe, if yes then I added the critical bug, and there was performance bug before

00max09: About the performance I can change it to scaling the strides, it does not change the performance, only change is that i have to add two variables instead of one (doing it without variables loses 25us in benchmark time). 
There is also a third way, I can scale strides in the beginning(editing the desc), this could add minimal increase in performance, but I don't know if editing desc is a good idea

dali/kernels/transpose/transpose_gpu_impl_test.cu
00max09: changed

szalpal:
!build

stiepan (approved):

banasraf (approved):

dali/kernels/transpose/transpose_gpu_impl.cuh
00max09: changed for recalculating strides

00max09: > Can we be sure that it is at least 4 (or to be exact: `2 ** pack_ratio_log`)?

about that, if I calculate correctly this blockDim.y is bound by 1024/used_registers, where used registers in this case is less than 200 so for me code is safe

dali/kernels/transpose/transpose_gpu_impl.cuh
00max09: changed

PR #3734, opened 2022-03-15
Add direct operator calls in debug mode

ksztenderski:
## Category:

## Description:

It adds backend implementation of direct operators and replaces minipipelines in debug mode with direct operators.

Debug mode pipeline

WIP, no review needed yet.

### Tests
- [x] Existing tests apply
- [x] New tests added
  - [x] Python tests
  - [ ] GTests
  - [ ] Benchmark
  - [ ] Other
- [ ] N/A

### Documentation
- [ ] Documentation updated
  - [ ] Docstring
  - [ ] Doxygen
  - [ ] RST
  - [ ] Jupyter
  - [ ] Other
- [ ] N/A

#### Requirements
- [x] N/A

**REQ IDs**: N/A

**JIRA TASK**: N/A


dali/pipeline/operator/direct_operator.h
```diff
@@ -0,0 +1,223 @@
...
+}
+
+template <>
+template <>
+std::vector<std::shared_ptr<TensorList<GPUBackend>>> DirectOperator<GPUBackend>::Run(
+    const std::vector<std::shared_ptr<TensorList<GPUBackend>>> &inputs,
+    const std::unordered_map<std::string, std::shared_ptr<TensorList<CPUBackend>>> &kwargs,
+    cudaStream_t cuda_stream) {
+  ws.set_stream(cuda_stream);
+  CUDA_CALL(cudaStreamSynchronize(cuda_stream));
```
JanuszL, line 123: Long term we can make this synchronization optional.

dali/python/backend_impl.cc
```diff
@@ -1420,7 +1468,7 @@ PYBIND11_MODULE(backend_impl, m) {
 
   // DALIImageType
   py::enum_<DALIImageType>(types_m, "DALIImageType", "Image type\n<SPHINX_IGNORE>")
-    .value("RGB", DALI_RGB)
+    .value("`", DALI_RGB)
```
JanuszL, line 1471: ?

ksztenderski: Just kind of a typo error. Overall this PR is WIP, not really ready for review, as I am going to make some changes.

JanuszL: Such things tend to slip to the final PR unnoticed. Just wanted to highlight this.

ksztenderski:
!build

dali/pipeline/operator/direct_operator.h
```diff
@@ -0,0 +1,241 @@
...
+  auto tl = std::make_shared<TensorList<Backend>>();
+  tl->Copy(*input);
+  return tl;
+}
+
+/**
+ * @brief Direct operator providing eager execution of an operator in Run.
+ */
+template <typename Backend>
+class DLL_PUBLIC DirectOperator {
```
mzient, line 49:
```suggestion
class DLL_PUBLIC ImmediateOperator {
```
or
```suggestion
class DLL_PUBLIC EagerOperator {
```
?
"Direct" is doesn't really convey this meaning. It's an merely an abbreviation for "DirectlyCalledOperator" which is lengthy.

dali/pipeline/operator/direct_operator.h
mzient: Why would we ever need to synchronize _before_?

dali/pipeline/operator/direct_operator.h
```diff
@@ -0,0 +1,241 @@
...
+std::vector<std::shared_ptr<TensorList<GPUBackend>>> DirectOperator<GPUBackend>::Run(
+    const std::vector<std::shared_ptr<TensorList<GPUBackend>>> &inputs,
+    const std::unordered_map<std::string, std::shared_ptr<TensorList<CPUBackend>>> &kwargs,
+    cudaStream_t cuda_stream) {
+  ws.Clear();
+  ws.set_stream(cuda_stream);
+  CUDA_CALL(cudaStreamSynchronize(cuda_stream));
+  auto output = RunImpl<GPUBackend, GPUBackend, TensorList<GPUBackend>, TensorList<GPUBackend>>(
+      inputs, kwargs);
+  CUDA_CALL(cudaStreamSynchronize(cuda_stream));
```
mzient, line 136: We could get rid of this one, too - we could (and should) expose associated stream in TensorXxxGPU in Python and just tell the user that the data is available for that stream. We can (and already do) synchronize D2H copies.

dali/pipeline/operator/direct_operator.h
```diff
@@ -0,0 +1,241 @@
...
+  return input;
+}
+
+template <typename Backend>
+std::shared_ptr<TensorList<Backend>> AsTensorList(std::shared_ptr<TensorVector<Backend>> input) {
+  // TODO(ksztenderski): Remove copy.
+  auto tl = std::make_shared<TensorList<Backend>>();
+  tl->Copy(*input);
+  return tl;
+}
```
JanuszL, line 43: We have a very similar thing in `workspace_policy.h` - `PresentAsTensorList`. Maybe it should be unified and moved to a common untility.

dali/pipeline/operator/direct_operator.h
```diff
@@ -0,0 +1,241 @@
...
+
+  // Runs operator using shared thread pool and shared CUDA stream.
+  template <typename InBackend, typename OutBackend>
+  DLL_PUBLIC std::vector<std::shared_ptr<TensorList<OutBackend>>> Run(
+      const std::vector<std::shared_ptr<TensorList<InBackend>>> &inputs,
+      const std::unordered_map<std::string, std::shared_ptr<TensorList<CPUBackend>>> &kwargs) {
+    DALI_FAIL("Unsupported backends in DirectOperator.Run().");
+  }
+
+  // Runs operator using specified thread pool.
```
JanuszL, line 65:
```suggestion
  // Runs operator using specified thread pool and shared CUDA stream.
```

dali/pipeline/operator/direct_operator.h
```diff
@@ -0,0 +1,241 @@
...
+  template <typename InBackend, typename OutBackend>
+  DLL_PUBLIC std::vector<std::shared_ptr<TensorList<OutBackend>>> Run(
+      const std::vector<std::shared_ptr<TensorList<InBackend>>> &inputs,
+      const std::unordered_map<std::string, std::shared_ptr<TensorList<CPUBackend>>> &kwargs,
+      ThreadPool *tp) {
+    DALI_FAIL("Unsupported backends in DirectOperator.Run() with thread pool.");
+  }
+
+  // Runs operator using specified CUDA stream.
```
JanuszL, line 74:
```suggestion
  // Runs operator using shared thread and specified CUDA stream.
```

dali/pipeline/operator/direct_operator.h
```diff
@@ -0,0 +1,241 @@
...
+  template <typename InBackend, typename OutBackend>
+  DLL_PUBLIC std::vector<std::shared_ptr<TensorList<OutBackend>>> Run(
+      const std::vector<std::shared_ptr<TensorList<InBackend>>> &inputs,
+      const std::unordered_map<std::string, std::shared_ptr<TensorList<CPUBackend>>> &kwargs,
+      cudaStream_t cuda_stream) {
+    DALI_FAIL("Unsupported backends in DirectOperator.Run() with CUDA stream");
+  }
+
+  // Set shared thread pool used for all direct operators.
```
JanuszL, line 83:
```suggestion
  // Creates thread pool used for all direct operators.
```

dali/pipeline/operator/direct_operator.h
```diff
@@ -0,0 +1,241 @@
...
+  DLL_PUBLIC inline static void SetThreadPool(int num_threads, int device_id, bool set_affinity) {
+    shared_thread_pool = std::make_unique<ThreadPool>(num_threads, device_id, set_affinity);
+  }
+
+  // Set shared CUDA stream used for all direct operators.
```
JanuszL, line 88:
```suggestion
  // Creates shared CUDA stream used for all direct operators.
```
I would rather expect that `Set` function accepts the value it should set.

dali/pipeline/operator/direct_operator.h
```diff
@@ -0,0 +1,241 @@
...
+  template <typename InBackend, typename OutBackend>
+  DLL_PUBLIC std::vector<std::shared_ptr<TensorList<OutBackend>>> Run(
+      const std::vector<std::shared_ptr<TensorList<InBackend>>> &inputs,
+      const std::unordered_map<std::string, std::shared_ptr<TensorList<CPUBackend>>> &kwargs,
+      cudaStream_t cuda_stream) {
+    DALI_FAIL("Unsupported backends in DirectOperator.Run() with CUDA stream");
+  }
+
+  // Set shared thread pool used for all direct operators.
+  DLL_PUBLIC inline static void SetThreadPool(int num_threads, int device_id, bool set_affinity) {
```
JanuszL, line 84:
```suggestion
  DLL_PUBLIC inline static void CreateThreadPool(int num_threads, int device_id, bool set_affinity) {
```

dali/pipeline/operator/direct_operator.h
```diff
@@ -0,0 +1,241 @@
...
+    shared_thread_pool = std::make_unique<ThreadPool>(num_threads, device_id, set_affinity);
+  }
+
+  // Set shared CUDA stream used for all direct operators.
+  DLL_PUBLIC inline static void SetCudaStream(int device_id) {
```
JanuszL, line 89:
```suggestion
  DLL_PUBLIC inline static void CreateCudaStream(int device_id) {
```

dali/pipeline/operator/direct_operator.h
```diff
@@ -0,0 +1,241 @@
...
+      const std::vector<std::shared_ptr<TensorList<InBackend>>> &inputs,
+      const std::unordered_map<std::string, std::shared_ptr<TensorList<CPUBackend>>> &kwargs);
+
+  int batch_size;
+  size_t num_outputs;
+  workspace_t<Backend> ws;
+  OpSpec op_spec;
+  std::unique_ptr<OperatorBase> op;
+
+  static cudaStream_t shared_cuda_stream;
```
JanuszL, line 108: @mzient - I think it should be
```suggestion
  static CUDAStreamLease shared_cuda_stream;
```
I'm not sure if having it static works well with `SetCudaStream` that can change it for all instances of this class.

dali/pipeline/operator/direct_operator.h
```diff
@@ -0,0 +1,241 @@
...
+  template <typename InBackend, typename OutBackend, typename WSInputType, typename WSOutputType>
+  std::vector<std::shared_ptr<TensorList<OutBackend>>> RunImpl(
+      const std::vector<std::shared_ptr<TensorList<InBackend>>> &inputs,
+      const std::unordered_map<std::string, std::shared_ptr<TensorList<CPUBackend>>> &kwargs);
+
+  int batch_size;
+  size_t num_outputs;
+  workspace_t<Backend> ws;
+  OpSpec op_spec;
+  std::unique_ptr<OperatorBase> op;
```
JanuszL, line 106:
```suggestion
  int batch_size_;
  size_t num_outputs_;
  workspace_t<Backend> ws_;
  OpSpec op_spec_;
  std::unique_ptr<OperatorBase> op_;
```

dali/pipeline/operator/direct_operator.h
```diff
@@ -0,0 +1,241 @@
...
+    auto tensor_in = std::make_shared<WSInputType>();
+    tensor_in->ShareData(*inputs[in_idx]);
+
+    // Set default layout for input if not specified.
+    if (tensor_in->GetLayout().empty()) {
+      auto default_layout = op_spec.GetSchema().GetInputLayout(
+          in_idx, tensor_in->shape().sample_dim(), tensor_in->GetLayout());
+      if (!default_layout.empty()) {
+        tensor_in->SetLayout(default_layout);
+      }
```
JanuszL, line 195: We have `SetDefaultLayoutIfNeeded` in executor.h, maybe we can extract it to a common utility?

dali/pipeline/operator/direct_operator.h
```diff
@@ -0,0 +1,241 @@
...
+  outputs.reserve(num_outputs);
+
+  for (size_t i = 0; i < num_outputs; ++i) {
+    ws.AddOutput(std::make_shared<WSOutputType>(batch_size));
+  }
+
+  ws.SetBatchSizes(batch_size);
+
+  // Setup outputs.
+  if (op->Setup(output_desc, ws) && op->CanInferOutputs()) {
```
JanuszL, line 217:
```suggestion
  if (op->CanInferOutputs() && op->Setup(output_desc, ws)) {
```
If `op->CanInferOutputs()` return false we don't need to call `Setup`, I guess.

dali/pipeline/operator/direct_operator.h
```diff
@@ -0,0 +1,241 @@
...
+  for (size_t i = 0; i < num_outputs; ++i) {
+    outputs.push_back(AsTensorList<OutBackend>(ws.template OutputPtr<OutBackend>(i)));
+  }
+
+  return outputs;
+}
+
+template <typename Backend>
+std::unique_ptr<ThreadPool> DirectOperator<Backend>::shared_thread_pool =
+    std::make_unique<ThreadPool>(1, 0, false);
```
JanuszL, line 234: Can't we just have an nullptr here?

dali/pipeline/pipeline_debug.h
```diff
@@ -0,0 +1,117 @@
...
+
+template <>
+std::vector<std::shared_ptr<TensorList<CPUBackend>>> PipelineDebug::RunOperator(
+    int logical_id, const std::vector<std::shared_ptr<TensorList<CPUBackend>>> &inputs,
+    const std::unordered_map<std::string, std::shared_ptr<TensorList<CPUBackend>>> &kwargs) {
+  try {
+    return cpu_operators.at(logical_id).Run<CPUBackend, CPUBackend>(inputs, kwargs, &thread_pool);
+  } catch (std::out_of_range &e) {
+    DALI_FAIL(make_string("Failed to acquire CPU Operator in PipelineDebug. ", e.what()));
+  }
```
JanuszL, line 90:
```suggestion
  auto op = cpu_operators.find(logical_id);
  DALI_ENFORCE(op != cpu_operators.end(), "Failed to acquire CPU Operator in PipelineDebug.");
  return op->Run<CPUBackend, CPUBackend>(inputs, kwargs, &thread_pool);
```

dali/pipeline/pipeline_debug.h
```diff
@@ -0,0 +1,117 @@
...
+}
+
+template <>
+std::vector<std::shared_ptr<TensorList<GPUBackend>>> PipelineDebug::RunOperator(
+    int logical_id, const std::vector<std::shared_ptr<TensorList<GPUBackend>>> &inputs,
+    const std::unordered_map<std::string, std::shared_ptr<TensorList<CPUBackend>>> &kwargs) {
+  try {
```
JanuszL, line 97: As above.

dali/pipeline/pipeline_debug.h
```diff
@@ -0,0 +1,117 @@
...
+  } catch (std::out_of_range &e) {
+    DALI_FAIL(make_string("Failed to acquire GPU Operator in PipelineDebug. ", e.what()));
+  }
+}
+
+template <>
+std::vector<std::shared_ptr<TensorList<GPUBackend>>> PipelineDebug::RunOperator(
+    int logical_id, const std::vector<std::shared_ptr<TensorList<CPUBackend>>> &inputs,
+    const std::unordered_map<std::string, std::shared_ptr<TensorList<CPUBackend>>> &kwargs) {
+  try {
```
JanuszL, line 108: Same.

dali/python/nvidia/dali/_debug_mode.py
```diff
@@ -39,6 +40,9 @@ def __str__(self):
 
     __repr__ = __str__
 
+    def __len__(self):
```
JanuszL, line 43: Is it for the batch size?

dali/python/nvidia/dali/_debug_mode.py
```diff
@@ -358,53 +368,66 @@ def is_primitive_type(x):
 
         return False, 'cpu', data
 
-    def _create_subpipeline(self, op_wrapper, inputs, kwargs):
-        """Creates pipeline wrapper around operator call (only in the first run).
-        Each pipeline uses ExternalSource operators for processing inputs and kwargs.
-        Returns pipeline and classification of data as batch.
-        """
+    @staticmethod
```
JanuszL, line 371: I cannot add comment to L356-L369, so I will put it here. `_is_mxnet_array` should have the same dedicated case as `_is_torch_tensor`, and the `device` should be inferred from the array type, not set CPU for mxnet and GPU for torch.

dali/python/nvidia/dali/_debug_mode.py
```diff
@@ -419,57 +442,111 @@ def _external_source(self, name=None, **kwargs):
...
+        def check_classification(expected_is_batch, actual_is_batch, arg_type, value):
+            def classification_to_str(is_batch):
+                return 'batch' if is_batch else 'constant'
+
+            if expected_is_batch != actual_is_batch:
+                expected_str = classification_to_str(expected_is_batch)
+                actual_str = classification_to_str(actual_is_batch)
+
+                raise RuntimeError(f"In operator '{op_name}' {arg_type} {value} recognized as {actual_str} but built value"
+                                   f" in its place was recognized as {expected_str}.")
```
JanuszL, line 478:
```suggestion
                raise RuntimeError(f"The operator '{op_name}' {arg_type} {value} recognized as {actual_str} but it was created"
                                   f" with {expected_str}.")
```

dali/python/nvidia/dali/_debug_mode.py
```diff
@@ -419,57 +442,111 @@ def _external_source(self, name=None, **kwargs):
             raise RuntimeError(f"Unexpected operator 'ExternalSource'. Debug mode does not support"
                                " changing the order of operators executed within the pipeline.")
 
-    def _run_subpipeline(self, pipe_tuple, op_name, inputs, kwargs):
-        """Run pipeline wrapper of a single operator."""
-        self._debug_on = False
-        pipe, inputs_es, kwargs_es = pipe_tuple
+    def _run_op_on_device(self, op_name, device, inputs, kwargs):
```
JanuszL, line 445: Maybe I have missed that, but how do we enforce the regular pipeline flow that GPU op cannot be followed by the CPU one, or we don't?

dali/pipeline/operator/direct_operator.h
klecki: We are doing a copy here (I guess for a prototype) which will always work but the `PresentAsTensorList` requires contiguous TV to work.

JanuszL: It can be another flavor or option of that function.

dali/pipeline/operator/direct_operator.h
ksztenderski: Removed the synchronization before the run and left only synchronization after. For now (especially in terms of debug mode) it shouldn't bother us but in the future it seems unnecessary to synchronize immediately after the run and only synchronize when we actually need this data (that synchronization before the run was kind of supposed to show that concept).

dali/pipeline/operator/direct_operator.h
ksztenderski: Creating API for stream exposure seems like a good idea for a follow-up as it'll probably generate a lot of code and is not really that necessary for the debug mode. But fully agree with support for it in terms of eager operators.

dali/pipeline/operator/direct_operator.h
ksztenderski: done, went with `EagerOperator`

dali/pipeline/operator/direct_operator.h
ksztenderski: That's kind of the point, that it supports only CPU operators so the CUDA stream is not set.

dali/pipeline/operator/direct_operator.h
ksztenderski: Here the opposite, supports only GPU operators and thread pool is not set.

dali/pipeline/operator/direct_operator.h
ksztenderski: The idea was that we have some useable thread pool by default (used by all eager operators) and only if user wants to change it they can with `Set[Create]ThreadPool`. By making it `nullptr` here we would force users to explicitly create thread pool before using eager operators which could be inconvenient but at the same time would force them to understand how to set it and would avoid questions about how it is not multithreaded.
I guess we should discuss what should be the default behavior here and with the CUDA stream.

dali/pipeline/operator/direct_operator.h
ksztenderski: I think that "creates" might suggest that by default there is no thread pool, but I agree that "set" is not perfect either. Maybe "update"?

dali/pipeline/operator/direct_operator.h
ksztenderski: done

dali/pipeline/operator/direct_operator.h
ksztenderski: done

dali/pipeline/pipeline_debug.h
ksztenderski: done

dali/pipeline/operator/direct_operator.h
JanuszL: Understood.

JanuszL: On the other hand this is template. So it can be used to any kind of op, including mixed and GPU one.

dali/pipeline/operator/direct_operator.h
JanuszL: I don't think that the template prevent creating a CPU only `run` function with such signature.

dali/pipeline/operator/direct_operator.h
JanuszL: Understood.

dali/python/nvidia/dali/_debug_mode.py
ksztenderski: done

dali/pipeline/operator/direct_operator.h
ksztenderski: We do have to call `Setup`. Not calling it resulted in segfault (for `readers.File` but most likely same thing would happen for other operators).

dali/pipeline/operator/direct_operator.h
ksztenderski: done

dali/pipeline/operator/direct_operator.h
ksztenderski: done (changed to "update")

dali/pipeline/operator/direct_operator.h
ksztenderski: done, (`UpdateThreadPool`)

dali/pipeline/operator/direct_operator.h
ksztenderski: done, (`UpdateCudaStream`)

dali/pipeline/operator/direct_operator.h
ksztenderski: Changed type to `CUDAStreamLease`. Is having static `cuda_stream` a problem?

dali/pipeline/operator/direct_operator.h
JanuszL: Yes. I'm talking rubbish. Sorry.

dali/python/nvidia/dali/_debug_mode.py
ksztenderski: Were you thinking of e.g. `The operator 'random__CoinFlip' recognized argument 'probability' as batch but it was created with constant.`?
Because this gives `The operator 'random__CoinFlip' argument probability recognized as batch but it was created with constant.`

JanuszL: That sounds better. The original just sounds strange for me.

dali/pipeline/operator/direct_operator.h
ksztenderski: Well yes, but the thread pool won't be set anyway.

dali/python/nvidia/dali/_debug_mode.py
ksztenderski: Thanks for the catch, we didn't. Added check for that.

dali/python/nvidia/dali/ops.py
```diff
@@ -640,73 +707,79 @@ def __call__(self, *inputs, **kwargs):
...
+
+    # TODO(ksztenderski): Just a template for eager operator.
+    class DirectOperator(DirectOperatorBase):
+        def __init__(self, batch_size=-1, device_id=0, cuda_stream=None, **kwargs):
+            # Here all kwargs are supposed to be constants.
+            super().__init__(**kwargs)
+
+            self._spec.AddArg('max_batch_size', batch_size)
+            self._spec.AddArg('device_id', device_id)
+            # self._spec.AddArg('cuda_stream', cuda_stream)
```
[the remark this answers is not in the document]
ksztenderski: It's just a template for the future implementation of eager operator in python. Removed it from this PR.

dali/python/nvidia/dali/_debug_mode.py
ksztenderski: done

dali/pipeline/operator/direct_operator.h
ksztenderski: done

dali/pipeline/operator/direct_operator.h
ksztenderski: done

ksztenderski: done

ksztenderski:
!build

dali/pipeline/util/batch_utils.h
```diff
@@ -0,0 +1,79 @@
...
+std::shared_ptr<TensorList<Backend>> PresentAsTensorList(
+    const std::shared_ptr<TensorList<Backend>> &in) {
+  return in;
+}
+
+template <typename Backend>
+std::shared_ptr<TensorList<Backend>> PresentAsTensorList(
+    const std::shared_ptr<TensorVector<Backend>> &in) {
+  if (in->IsContiguous()) {
+    return in->AsTensorList(false);
```
JanuszL, line 69: I'm not sure if this going to work. `in->AsTensorList(false)` skips the continuity check.

dali/test/python/test_pipeline_debug.py
```diff
@@ -427,3 +427,43 @@ def test_seed_generation_base():
...
+    pipe.run()
+
+
+@pipeline_def(batch_size=8, num_threads=3, device_id=0, seed=47, debug=True)
+def cpu_after_gpu_pipeline():
+    jpegs, labels = fn.readers.file(
+        file_root=file_root, shard_id=0, num_shards=2, random_shuffle=True)
+    images = fn.decoders.image(jpegs, output_type=types.RGB, device='mixed')
+
+    output = fn.random_resized_crop(images, size=(224, 224), device='cpu')
```
JanuszL, line 461: What if we add:
```images = images.as_cpu()```?

File: dali/python/nvidia/dali/_debug_mode.py [hunk 27]
```diff
@@ -343,10 +354,12 @@ def is_primitive_type(x):
             return True, device, data_list
         else:
             if is_primitive_type(data) or _types._is_numpy_array(data) or \
-                    _types._is_mxnet_array(data) or isinstance(data, _tensors.TensorCPU):
+                    isinstance(data, _tensors.TensorCPU):
                 return False, 'cpu', data
             if _types._is_torch_tensor(data):
                 return False, 'gpu' if data.is_cuda else 'cpu', data
+            if _types._is_mxnet_array(data):
```
JanuszL, line 361: 👍

dali/python/nvidia/dali/_debug_mode.py
```diff
@@ -419,57 +446,122 @@ def _external_source(self, name=None, **kwargs):
...
+
+        return data_nodes
 
-        if key in self._subpipelines:
-            return self._run_subpipeline(self._subpipelines[key], op_wrapper.__name__, inputs, kwargs)
+    def _wrap_op_call(self, op_name, inputs, kwargs):
+        self._cur_op_id += 1
+        key = inspect.getframeinfo(
+            inspect.currentframe().f_back.f_back)[:3] + (self._cur_op_id,)
+        if not self._ops_built:
```
JanuszL, line 558: Maybe it is just only me, but having module called `_ops` and class property `_ops` leads to confusion.

dali/python/nvidia/dali/_debug_mode.py
JanuszL: Do you mean test itself or there is a check in the code added?
I was thinking about the following flow:
```
x = op1(devide="cpu")
y = op2(x.as_gpu(), devide="gpu")
z = op3(y.as_cpu(), devide="cpu")
```

dali/pipeline/util/batch_utils.h
ksztenderski: I don't think I understand, I skip the continuity check in `AsTensorList` because I just checked it in the if condition.

dali/python/nvidia/dali/_debug_mode.py
ksztenderski: Agreed to stay with the current implementation (allowing to have CPU operator after a GPU one as long as the data is on correct device) to support external data (like `numpy.ndarray`) in debug operators.

dali/pipeline/util/batch_utils.h
JanuszL: My point is, that in the original usage in workspace_policy.h this check was deliberately skipped, as the code was used to connect inputs and outputs in the graph before the actual TL were allocated, and they were not continuous at that time even if they were allocated as such later.

dali/python/nvidia/dali/_debug_mode.py
ksztenderski: fixed

dali/test/python/test_pipeline_debug.py
ksztenderski: We can force it with `images.get().as_cpu()` and it will work.

dali/pipeline/pipeline_debug.h
```diff
@@ -0,0 +1,111 @@
...
+
+ private:
+  int max_batch_size;
+  int device_id;
+  int num_threads;
+  CUDAStreamLease cuda_stream;
+  ThreadPool thread_pool;
+  std::unordered_map<int, EagerOperator<CPUBackend>> cpu_operators;
+  std::unordered_map<int, EagerOperator<GPUBackend>> gpu_operators;
+  std::unordered_map<int, EagerOperator<MixedBackend>> mixed_operators;
```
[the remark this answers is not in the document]
ksztenderski: done

dali/test/python/test_pipeline_debug.py
JanuszL: On the other hand, we can put any NumPy code in our pipeline so we should not block this. Let us leave it as it is.

ksztenderski:
!build

JanuszL (approved):

dali/pipeline/util/batch_utils.h
klecki: Yes.

klecki: To elaborate, the problem with this function is that it breaks the encapsulation of the TensorVector and uses that implementation detail to build the graph of inputs-outputs for the execution. We can probably use this as optimization for now in the eager ops (maybe add the comment mentioning it, that we know what we are doing - the TV is already finalized and filled with data, so we can use a shortcut), but the two usages of the function have distinct preconditions - @ksztenderski's usage requires valid and filled TensorVector, the current usage of PresentAsTensorList is a hack for executor to give the next operator the access to contiguous memory shown as TensorList that will be filled with data, before TensorVector is actually used. 

They should not have been generalized together.

ksztenderski: Reverted to separate implementations of `PresentAsTensorList` and `AsTensorList` for eager operator.

ksztenderski:
!build

JanuszL (approved):

ksztenderski:
!build

JanuszL (approved):

ksztenderski:
!build

dali/pipeline/operator/eager_operator.h
```diff
@@ -0,0 +1,238 @@
...
+  }
+
+  // Update shared thread pool used for all direct operators.
+  DLL_PUBLIC inline static void UpdateThreadPool(int num_threads, int device_id,
+                                                 bool set_affinity) {
+    shared_thread_pool = std::make_unique<ThreadPool>(num_threads, device_id, set_affinity);
+  }
+
+  // Update shared CUDA stream used for all direct operators.
+  DLL_PUBLIC inline static void UpdateCudaStream(int device_id) {
```
klecki, line 97: Hmm, the naming and semantics are a bit unfortunate here. We may need to think about how do we work on multi-gpu systems - the GPU operators should use a specific device and stream pair, for now we are using an internal stream, but the device should be assigned to the PipelineDebug instance probably.

dali/pipeline/operator/eager_operator.h
```diff
@@ -0,0 +1,238 @@
...
+
+/**
+ * @brief Direct operator providing eager execution of an operator in Run.
+ */
+template <typename Backend>
+class DLL_PUBLIC EagerOperator {
+ public:
+  DLL_PUBLIC inline EagerOperator(const OpSpec &spec)
+      : batch_size_(spec.GetArgument<int>("max_batch_size")),
+        num_outputs_(spec.GetSchema().NumOutput()),
```
klecki, line 60: As always is the case with DALI some ops have: CalculateOutputs and CalculateAdditionalOutputs to calculate the number of outputs. I think some of it is already used in Python layer in dali.ops, so maybe the spec rather than schema have a correct number of outputs.

dali/pipeline/operator/eager_operator.h
```diff
@@ -0,0 +1,238 @@
...
+  std::vector<OutputDesc> output_desc{};
+  std::vector<std::shared_ptr<TensorList<OutBackend>>> outputs{};
+
+  outputs.reserve(num_outputs_);
+
+  for (size_t i = 0; i < num_outputs_; ++i) {
+    ws_.AddOutput(std::make_shared<WSOutputType>(batch_size_));
+  }
+
+  ws_.SetBatchSizes(batch_size_);
```
klecki, line 211: Hmm, we probably should take the batch size from the inputs in this iteration, to support the iter-2-iter variable batch size. It is driven via the external source. Ofc user would be able to abuse it in DebugMode and run with different batch sizes among the operators in one iterations. 

The problem is that it infers the BS from external source, so we would need to run them ahead of other operators to infer the BS, store it in PipelineDebug and pass it here to be used, as it also affects generator operators that are inputless.

klecki (commented):
Looks nice, I think there is slight problem with the device/stream usage and we must think about how we can handle the batch size. If it turns out too complicated we can do it in a follow up, and just require max_batch_size_ in this mode for now.

I am not fully sure about our handling of multiple input sets.

dali/python/nvidia/dali/ops.py
```diff
@@ -528,20 +528,17 @@ def _check_arg_input(schema, op_name, name):
         raise TypeError("The argument `{}` for operator `{}` should not be a `DataNode` but a {}".format(
             name, op_name, _type_name_convert_to_string(schema.GetArgumentType(name), False)))
 
-def python_op_factory(name, schema_name = None, op_device = "cpu"):
+
+def python_op_factory(name, schema_name = None):
     class Operator(metaclass=_DaliOperatorMeta):
-        def __init__(self, **kwargs):
+        def __init__(self, device="cpu", **kwargs):
```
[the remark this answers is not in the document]
ksztenderski: done

dali/pipeline/operator/eager_operator.h
ksztenderski: Probably overlook, we already have a specific device and stream for a single `PipelineDebug` instance.

dali/pipeline/operator/eager_operator.h
ksztenderski: done

dali/python/nvidia/dali/_debug_mode.py
```diff
@@ -358,53 +369,67 @@ def is_primitive_type(x):
...
+
+        return data_nodes
+
+    def _create_op(self, op_class, inputs, kwargs):
+        """Creates direct operator."""
+
+        init_args, call_args, kwargs_classification = _PipelineDebug._separate_kwargs(kwargs)
+
+        if op_class.__name__ == 'ArithmeticGenericOp':
+            inputs = _PipelineDebug._init_arithm_op(init_args, *inputs)
```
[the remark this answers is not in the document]
ksztenderski: Done, passing `name` as a keyward argument.

dali/python/nvidia/dali/_debug_mode.py
```diff
@@ -358,53 +369,67 @@ def is_primitive_type(x):
...
+                init_args['device'] = _PipelineDebug._classify_data(inputs[0])[1]
+
+            if 'seed' not in init_args:
+                init_args['seed'] = self._seed_generator.integers(0, 2**32)
+
+        # Save inputs classification for later verification.
+        inputs_classification = [_Classification(
+            *_PipelineDebug._classify_data(input)[:2]) for input in inputs]
+
+        op_helper = op_class(**init_args)
```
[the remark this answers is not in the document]
ksztenderski: Done, created dedicated operator manager class.

dali/python/nvidia/dali/_debug_mode.py
```diff
@@ -419,57 +444,123 @@ def _external_source(self, name=None, **kwargs):
...
+    def _pack_to_data_node_debug(self, data, op_name):
+        if isinstance(data, (list, tuple)):
+            return [self._pack_to_data_node_debug(elem, op_name) for elem in data]
 
+        return DataNodeDebug(data, op_name, 'gpu' if isinstance(data, _tensors.TensorListGPU) else 'cpu', self)
+
+    def _run_op(self, op_tuple, inputs, kwargs):
+        """Run a single operator."""
+
+        op_helper, init_args, inputs_classification, kwargs_classification, expected_inputs_size = op_tuple
```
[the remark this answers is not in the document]
ksztenderski: done

dali/python/nvidia/dali/ops.py
```diff
@@ -701,12 +683,51 @@ def _repack_list(self, sets, fn):
...
+    from nvidia.dali._debug_mode import _transform_data_to_tensorlist
+    import nvidia.dali.tensors as tensors
+
+    inputs = list(inputs)
+
+    for i, input in enumerate(inputs):
+        if not isinstance(input, (tensors.TensorListCPU, tensors.TensorListGPU)) and \
+                not (isinstance(input, list) and
+                     all([isinstance(elem, (tensors.TensorListCPU, tensors.TensorListGPU)) for elem in input])):
+            inputs[i] = _transform_data_to_tensorlist(input, len(input))
```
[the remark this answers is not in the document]
ksztenderski: As we previously agreed the only way of using input sets in debug mode (and in eager operators) is with `[TensorList]` (or in debug mode `[DataNodeDebug]`). Added comment explaining this.

dali/pipeline/operator/eager_operator.h
ksztenderski: For now staying with current implementation, might add support for variable batch size in the future.

ksztenderski:
!build

ksztenderski:
!build

dali/python/nvidia/dali/_debug_mode.py
```diff
@@ -324,106 +512,11 @@ def feed_input(self, data_node, data, **kwargs):
...
-
-        for arg_name in call_args.keys():
-            # To use argument inputs OpSpec needs it specified (can be an empty placeholder).
-            op_helper._spec.AddArgumentInput(arg_name, '')
-
-        self._pipe.AddOperator(op_helper._spec, self._cur_logical_id)
-
-        return op_helper, init_args, inputs_classification, kwargs_classification, len(inputs)
+        self._operators[key] = _OperatorManager(
+            op_class, self._seed_generator.integers(0, 2**32), inputs, kwargs)
```
klecki, line 518: This is a change that calls the seed generator every time instead of running it when the op didn't have the argument. I guess it doesn't really matter, but just wanted to check if it's intended.

klecki (approved):
Looks ok, small nitpick.

I am also wandering if we can add a simple check for the current batch size and raise error that the variable batch size is not supported if we encounter smaller than max_batch_size batch until we start supporting it.

dali/pipeline/operator/eager_operator.h
JanuszL: OK. I see the check is there.

dali/test/python/test_pipeline_debug.py
```diff
@@ -427,3 +427,93 @@ def test_seed_generation_base():
...
+
+
+def test_input_sets():
+    pipe_standard = input_sets_pipeline()
+    pipe_debug = input_sets_pipeline(debug=True)
+    compare_pipelines(pipe_standard, pipe_debug, 8, 10)
+
+
+@pipeline_def(batch_size=8, num_threads=3, device_id=0, debug=True)
+def incorrect_input_Sets_pipeline():
```
JanuszL, line 507:
```suggestion
def incorrect_input_stes_pipeline():
```

dali/test/python/test_pipeline_debug.py
```diff
@@ -467,3 +467,53 @@ def test_cpu_operator_after_gpu():
...
+    jpegs, _ = fn.readers.file(file_root=file_root, seed=42, random_shuffle=True)
+    images = fn.decoders.image(jpegs, seed=42)
+    output = fn.cat([images, images, images], [images, images])
+
+    return tuple(output)
+
+
+@raises(ValueError, glob="All argument lists for Multpile Input Sets used with operator 'Cat' must have the same length")
```
JanuszL, line 515: Can the error message have the name of the operator consistent with the API used - so `Cat` for ops and `cat` for `fn`?

dali/python/nvidia/dali/fn.py
```diff
@@ -79,7 +79,7 @@ def fn_wrapper(*inputs, **kwargs):
         from nvidia.dali._debug_mode import _PipelineDebug
         current_pipeline = _PipelineDebug.current()
         if getattr(current_pipeline, '_debug_on', False):
-            return current_pipeline._wrap_op_call(op_wrapper, inputs, kwargs)
+            return current_pipeline._wrap_op_call(op_class, *inputs, **kwargs)
```
JanuszL, line 82:
```suggestion
            return current_pipeline._wrap_op_call(op_class, *inputs, name="_to_snake_case(op_class.__name__)", **kwargs)
```
or it doesn't make any sense?

dali/test/python/test_pipeline_debug.py
ksztenderski: As most of these error messages are for functionalities specifically for debug mode and the only way to use operators in debug mode is with `fn` API than I guess we can just change it to the snake_case.

dali/python/nvidia/dali/fn.py
ksztenderski: It makes sense, that's the way the names are created [here](https://github.com/ksztenderski/DALI/blob/main/dali/python/nvidia/dali/fn.py#L106), but I think it's even better to just pass the name to the `_wrap_op_class` and have the name generation in one place.

dali/test/python/test_pipeline_debug.py
JanuszL: So I would go for `fn` names.

dali/test/python/test_pipeline_debug.py
ksztenderski: done

dali/python/nvidia/dali/_debug_mode.py
```diff
@@ -419,57 +597,45 @@ def _external_source(self, name=None, **kwargs):
...
+        if not self._operators_built:
+            self._create_op(op_class, key, inputs, kwargs)
+
+        if key in self._operators:
+            if op_class.__name__ == 'ArithmeticGenericOp':
+                inputs = _PipelineDebug._extract_data_node_inputs(inputs)
+            return self._run_op(self._operators[key], inputs, kwargs)
         else:
-            raise RuntimeError(f"Unexpected operator '{op_wrapper.__name__}'. Debug mode does not support"
+            raise RuntimeError(f"Unexpected operator '{op_class}'. Debug mode does not support"
```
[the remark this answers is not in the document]
ksztenderski: Reverted to `fn` style names.

JanuszL (approved):

dali/python/nvidia/dali/_debug_mode.py
```diff
@@ -299,133 +339,167 @@ class _OperatorManager:
...
-        self.inputs_classification = [_Classification(input) for input in inputs]
+        self._inputs_classification = []
+
+        # When using input sets we have to create separate operators for each input.
+        input_set_len = 1
+        for i, input in enumerate(inputs):
+            classification = _Classification(input, f'Input {i}')
+
+            if isinstance(classification.is_batch, list):
+                if input_set_len == 1:
```
klecki, line 359: Wouldn't a -1 or something make more sense for the initial value?

klecki (commented):
One nitpick and I think there is small issue in MIS validation. Otherwise looks ok.

ksztenderski:
!build

dali/python/nvidia/dali/_debug_mode.py
```diff
@@ -52,85 +53,109 @@ def shape(self):
...
-        return _PipelineDebug.current()._wrap_op_call(_ops.ArithmeticGenericOp, self, other, name='bitxor')
+        return _PipelineDebug.current()._wrap_op_call(
+            _ops.ArithmeticGenericOp, DataNodeDebug.aritm_fn_name, self, other, name='bitxor')
 
     def __rxor__(self, other):
-        return _PipelineDebug.current()._wrap_op_call(_ops.ArithmeticGenericOp, other, self, name='bitxor')
+        return _PipelineDebug.current()._wrap_op_call(
+            _ops.ArithmeticGenericOp, DataNodeDebug.aritm_fn_name, other, self, name='bitxor')
 
+    aritm_fn_name = _to_snake_case(_ops.ArithmeticGenericOp.__name__)
```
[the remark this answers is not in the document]
ksztenderski: done (static member)

dali/python/nvidia/dali/_debug_mode.py
```diff
@@ -299,133 +339,167 @@ class _OperatorManager:
...
+        # When using input sets we have to create separate operators for each input.
+        input_set_len = 1
+        for i, input in enumerate(inputs):
+            classification = _Classification(input, f'Input {i}')
+
+            if isinstance(classification.is_batch, list):
+                if input_set_len == 1:
+                    input_set_len = len(classification.is_batch)
+                else:
+                    raise ValueError("All argument lists for Multipile Input Sets used "
```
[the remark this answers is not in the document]
ksztenderski: done

dali/python/nvidia/dali/_debug_mode.py
ksztenderski: done

klecki (approved):

JanuszL (approved):

PR #3736, opened 2022-03-15
Improve documentation for AccessOrder wait and set_order.

mzient:

## Category:

**Other** - Documentation

## Description:
Improves documentation of `AccessOrder::wait` and `Buffer::set_order`

### Tests
- [ ] Existing tests apply
- [ ] New tests added
  - [ ] Python tests
  - [ ] GTests
  - [ ] Benchmark
  - [ ] Other
- [X] N/A

### Documentation
- [ ] Documentation updated
  - [ ] Docstring
  - [X] Doxygen
  - [ ] RST
  - [ ] Jupyter
  - [ ] Other
- [ ] N/A

#### Requirements
- [X] N/A

**REQ IDs**: N/A

**JIRA TASK**: N/A

### DIFF
--- a/include/dali/core/access_order.h
+++ b/include/dali/core/access_order.h
@@ -103,9 +103,41 @@ class DLL_PUBLIC AccessOrder {
 
   /**
    * @brief Waits in `this` ordering context for the work scheduled in the `other` order.
+   *
+   * Waits for work scheduled in `other` order to complete. If `other` is host, this function
+   * is a no-op.
+   * If either `other` or `this` is a null order, this function has no efect.
+   *
+   * @note `wait` is transitive with a notable exception of null streams, for which the funcion
+   * is a no-op and can break transitivity.
+   *
+   * ```
+   * AccessOrder o1(stream1), o2(stream2), o3(stream3);
+   * o2.wait(o1);
+   * o3.wait(o2);  // o3 is now synchronized with o1
+   * ```
+   * ```
+   * AccessOrder o1(stream1), o2(AccessOrder::host()), o3(stream3);
+   * o2.wait(o1);  // waits for o1 on host
+   * o3.wait(o2);  // no-op, but host is now synchronized with o1, so all new work in o3 will
+   *               // be scheduled after o2.wait(o1);
+   * ```
+   * ```
+   * AccessOrder o1(stream1), o2{}, o3(stream3);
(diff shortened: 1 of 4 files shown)


mzient:
!build

include/dali/core/access_order.h
```diff
@@ -103,9 +103,41 @@ class DLL_PUBLIC AccessOrder {
...
+   * o2.wait(o1);  // no-op
+   * o3.wait(o2);  // no-op - o3 is not synchronized with o1!
+   * ```
    */
   void wait(const AccessOrder &other) const;
 
+  /**
+   * @brief Waits in `this` ordering context for a CUDA event
+   *
+   * This function executes cudaStreamWaitEvent in case `this` is a stream or
```
szalpal, line 136: To make things consistent, I'd propose this:
```suggestion
   * This function executes `cudaStreamWaitEvent` in case `this` is a stream, or
```
Or removing the backticks in the next line, around `cudaEventSynchronize`.

szalpal (approved):
Thanks for the clarifications in the documentation, that makes thing clearer.

One more food for thought, is it worth adding a note, that calling `set_order` is not no-cost and we should avoid calling it e.g. inside a loop (especially with runtime controlled iteration number)?

If it's too verbose, I'm fine with how the docs look right now.

include/dali/core/access_order.h
```diff
@@ -103,9 +103,41 @@ class DLL_PUBLIC AccessOrder {
...
+   * Waits for work scheduled in `other` order to complete. If `other` is host, this function
+   * is a no-op.
+   * If either `other` or `this` is a null order, this function has no efect.
+   *
+   * @note `wait` is transitive with a notable exception of null streams, for which the funcion
+   * is a no-op and can break transitivity.
+   *
+   * ```
+   * AccessOrder o1(stream1), o2(stream2), o3(stream3);
+   * o2.wait(o1);
```
JanuszL, line 116: This only means that all work scheduled on o2 will happen after o1, but when the call completes o1 can be still in progress (like `cudaStreamWaitEvent`), right?

mzient: Yes, it's exactly like `cudaStreamWaitEvent` because that's the underlying mechanism.

JanuszL (approved):

PR #3755, opened 2022-03-23
Add ndim argument to the external source operator

banasraf:
## Category:

New feature

## Description:

It adds a new feature needed for the Triton autofill. In future we might also make specifying the dimensionality of the data required.

external source - C++ and Python

external_source.py

### Tests
- [ ] Existing tests apply
- [x] New tests added
  - [x] Python tests
  - [ ] GTests
  - [ ] Benchmark
  - [ ] Other
- [ ] N/A

### Documentation
- [x] Documentation updated
  - [x] Docstring
  - [ ] Doxygen
  - [ ] RST
  - [ ] Jupyter
  - [ ] Other
- [ ] N/A

#### Requirements
- [ ] N/A

**REQ IDs**: EXTSRC.16

**JIRA TASK**: DALI-2521


banasraf:
!build

banasraf:
> I think that it deserves to be discussed whether we want just the `ndim` parameter or to introduce a `shape` with allowed placeholders (similar to what TF does). Especially would be nice, to maybe decide on one instead of adding `ndim` fist and later `shape` and than handling cases where one or both is specified.
> 
> `ndim=n` can map to a `shape=(None,) * n`, but `shape` can additionally have some sizes predefined, like `shape=(None, None, 3)` to specify HWC images with 3-channels only, but any height and width.

I feel like the topic of ndim already appeared in some of our discussions and there were no objections to that. 
For me, the justification for the ndim is that it's really easier to write ndim=4 rather than shape=(-1, -1, -1, -1).

dali/pipeline/operator/builtin/external_source.cc
```diff
@@ -97,6 +97,11 @@ The operator will validate that the fetched data is of the provided type.
 If the argument is omitted or ``DALIDataType.NO_TYPE`` is passed, the operator will infer
 the type based on the provided data.
 
-This argument will be required starting from DALI 2.0.)code", DALI_NO_TYPE);
+This argument will be required starting from DALI 2.0.)code", DALI_NO_TYPE)
+  .AddOptionalArg("ndim", R"code(Input dimensionality.
+
+The operator will validate that the fetched data has the provided dimensionality.
```
mzient, line 103:
```suggestion
The dimensionality of the data provided to the operator will be verified against this value.
```

dali/pipeline/operator/builtin/external_source.cc
```diff
@@ -97,6 +97,11 @@ The operator will validate that the fetched data is of the provided type.
...
+
+Specifying the input dimensionality will be required starting from DALI 2.0)code", -1);
```
mzient, line 105:
```suggestion
Specifying the input dimensionality will be required starting from DALI 2.0)code", nullptr);
```

dali/pipeline/operator/builtin/external_source.cc
```diff
@@ -97,6 +97,11 @@ The operator will validate that the fetched data is of the provided type.
 If the argument is omitted or ``DALIDataType.NO_TYPE`` is passed, the operator will infer
 the type based on the provided data.
 
-This argument will be required starting from DALI 2.0.)code", DALI_NO_TYPE);
+This argument will be required starting from DALI 2.0.)code", DALI_NO_TYPE)
+  .AddOptionalArg("ndim", R"code(Input dimensionality.
```
mzient, line 101:
```suggestion
  .AddOptionalArg<int>("ndim", R"code(Input dimensionality.
```
...and remove the default value (see below)

dali/pipeline/operator/builtin/external_source.h
```diff
@@ -220,6 +220,8 @@ class ExternalSource : public Operator<Backend>, virtual public BatchSizeProvide
         device_id_(spec.GetArgument<int>("device_id")),
         dtype_(spec.GetArgument<DALIDataType>("dtype")),
         previous_dtype_(DALIDataType::DALI_NO_TYPE),
+        ndim_(spec.GetArgument<int>("ndim")),
```
mzient, line 223: More elaborate logic will be required if default value is dropped, but I still think it's worth it.

dali/pipeline/operator/builtin/external_source.h
```diff
@@ -220,6 +220,8 @@ class ExternalSource : public Operator<Backend>, virtual public BatchSizeProvide
...
+        previous_ndim_(-1),
```
mzient, line 224: Do we really need "previous" now? We could just live with one value:
```C++
  if (spec.TryGetArgument(ndim_, "ndim"))
    DALI_ENFORCE(ndim >= 0, "Incorrect number of dimensions. Use positive values for tensors or 0 for scalars.);
  else
    ndim = -1;
```

dali/pipeline/operator/builtin/external_source.h
```diff
@@ -506,6 +508,23 @@ class ExternalSource : public Operator<Backend>, virtual public BatchSizeProvide
                   " and the current type is ", batch.type_info().name(), "."));
     previous_dtype_ = batch.type();
 
+    auto current_ndim = batch.shape().sample_dim();
+    DALI_ENFORCE(ndim_ == -1 || ndim_ == current_ndim,
+      make_string("ExternalSource expected data with ", ndim_, " dimensions and got ",
+                   current_ndim, " dimensions."));
+
+    DALI_ENFORCE(previous_ndim_ == -1 || previous_ndim_ == current_ndim,
```
mzient, line 516: Technically, we could use `ndim_` for previous and select the error message based `spec_.HasArgument("ndim")`

dali/pipeline/operator/builtin/external_source.cc
```diff
@@ -97,6 +97,11 @@ The operator will validate that the fetched data is of the provided type.
 If the argument is omitted or ``DALIDataType.NO_TYPE`` is passed, the operator will infer
 the type based on the provided data.
 
-This argument will be required starting from DALI 2.0.)code", DALI_NO_TYPE);
+This argument will be required starting from DALI 2.0.)code", DALI_NO_TYPE)
```
mzient, line 100: A high level remark:
How about making `layout` an argument, too? This way you could cross-validate ndim/layout here - and even infer ndim from the layout string.

dali/pipeline/operator/builtin/external_source.cc
banasraf: done

dali/pipeline/operator/builtin/external_source.cc
banasraf: done

dali/pipeline/operator/builtin/external_source.cc
banasraf: done

dali/pipeline/operator/builtin/external_source.h
banasraf: done

dali/pipeline/operator/builtin/external_source.h
banasraf: Got rid of the `previous`

dali/pipeline/operator/builtin/external_source.h
banasraf: done

dali/pipeline/operator/builtin/external_source.cc
banasraf: I agree that `layout` should be also a C++-level argument - even just because we tend to avoid Python-only features. Although, I would make this another task maybe. I want to move on with the rest of things needed for the Triton backend autofill. After doing that I can revisit the ES and make layout a C++ arg.

dali/test/python/test_operator_spectrogram.py
```diff
@@ -54,11 +56,9 @@ def define_graph(self):
 
     def iter_setup(self):
         data = self.iterator.next()
-        # randomly insert extra axis (channels?)
-        r = np.random.randint(-1, 2)
```
banasraf, line 58: This tested the changing ndim. We don't support that and now it causes an error, so I changed it to random but constant ndim.

dali/pipeline/operator/builtin/external_source.h
```diff
@@ -221,6 +223,12 @@ class ExternalSource : public Operator<Backend>, virtual public BatchSizeProvide
         dtype_(spec.GetArgument<DALIDataType>("dtype")),
         previous_dtype_(DALIDataType::DALI_NO_TYPE),
         sync_worker_(device_id_, false) {
+    if (spec.TryGetArgument(ndim_, "ndim")) {
+      DALI_ENFORCE(ndim_ >= 0, "Incorrect number of dimensions. "
+                   "Use positive values for tensors or 0 for scalars.");
+    } else {
+      ndim_ = -1;
+    }
```
mzient, line 231:
```suggestion
    }
    if (spec.TryGetArgument(layout_, "layout")) {
      if (ndim_ < 0)
        ndim = layout_.ndim();
      else
        DALI_ENFORCE(ndim_ == layout_.ndim(),
                     make_string("The layout '", layout, "' is not compatible with ndim ", ndim_, ".");
    }
```

dali/test/python/test_operator_spectrogram.py
mzient: :+1:

dali/pipeline/operator/builtin/external_source.cc
```diff
@@ -98,6 +98,12 @@ The operator will validate that the fetched data is of the provided type.
...
+  .AddOptionalArg<int>("ndim", R"code(Input dimensionality.
 
+The dimensionality of the data provided to the operator will be verified against this value.
+
+Specifying the input dimensionality will be required starting from DALI 2.0)code", nullptr)
+  .AddOptionalArg<TensorLayout>("layout",
+    "If provided, sets the layout of the data.", TensorLayout(""));
```
mzient, line 108:
```suggestion
  .AddOptionalArg<TensorLayout>("layout",
    "If provided, sets the layout of the data.", nullptr);
```

dali/pipeline/operator/builtin/external_source.h
```diff
@@ -221,7 +223,15 @@ class ExternalSource : public Operator<Backend>, virtual public BatchSizeProvide
...
         dtype_(spec.GetArgument<DALIDataType>("dtype")),
         previous_dtype_(DALIDataType::DALI_NO_TYPE),
+        ndim_(-1),
+        layout_(),
         sync_worker_(device_id_, false) {
+    if (spec.TryGetArgument(ndim_, "ndim")) {
+      DALI_ENFORCE(ndim_ >= 0, "Incorrect number of dimensions. "
+                   "Use positive values for tensors or 0 for scalars.");
+    }
+    layout_ = spec.GetArgument<TensorLayout>("layout");
```
mzient, line 233: I'd still suggest:
```suggestion
    spec.TryGetArgument(layout_, "layout");
```
and get rid of the default value. The result will be the same, but the documentation will be cleaner.

dali/pipeline/operator/builtin/external_source.h
```diff
@@ -287,6 +297,22 @@ class ExternalSource : public Operator<Backend>, virtual public BatchSizeProvide
   DISABLE_COPY_MOVE_ASSIGN(ExternalSource);
 
  protected:
+  inline bool HasNdim() {
+    return !layout_.empty() || spec_.HasArgument("ndim");
+  }
+
+  inline void InferNdim() {
+    if (!layout_.empty()) {
```
mzient, line 305: :+1: Keep it like that even if we get rid of the default - this way we can still use empty layouts, which has some merit at our Python interop level (namely, we tend to return an empty string rather than `None`).

dali/pipeline/operator/builtin/external_source.h
```diff
@@ -287,6 +297,22 @@ class ExternalSource : public Operator<Backend>, virtual public BatchSizeProvide
...
+      if (ndim_ != -1) {
+        DALI_ENFORCE(ndim_ == layout_.ndim(), make_string("Dimensionality of the provided "
+                     "layout does not match the ndim argument. The provided ndim: ", ndim_,
+                     ". Provided layout: ", layout_, "."));
```
mzient, line 309:
```suggestion
                     "layout does not match the ndim argument. The arguments provided:"
                     "\nndim    = ", ndim_,
                     "\nlayout = '", layout_, "'."));
```

dali/pipeline/operator/builtin/external_source.h
```diff
@@ -509,6 +533,31 @@ class ExternalSource : public Operator<Backend>, virtual public BatchSizeProvide
...
+                     input_ndim, " dimensions."));
+    } else if (ndim_ != -1) {
+      DALI_ENFORCE(input_ndim == ndim_,
+                   make_string("Dimensionality of the data fed to the external source has "
+                      "changed from previous iteration. Dimensionality in the previous "
+                      "iteration was ", ndim_, " and the current is ", input_ndim, "."));
+    }
+    ndim_ = input_ndim;
+
+    if (spec_.HasArgument("layout")) {
```
mzient, line 549: This is inconsistent with the behavior of InferNDim, which treats empty layout the same way as none at all. Which way is better is disputable, but we should be consistent.

mzient, line 549: This is inconsistent with the behavior of InferNDim, which treats empty layout the same way as none at all. Which way is better is disputable, but we should be consistent.
If we go with treating empty strings as no layout, then:
```suggestion
    if (!layout_.empty()) {
```

dali/pipeline/operator/builtin/external_source.h
banasraf: I guess there's no other way than treat empty as no layout.

dali/pipeline/operator/builtin/external_source.h
banasraf: done

mzient (approved):

dali/pipeline/operator/builtin/external_source.cc
```diff
@@ -98,6 +98,12 @@ The operator will validate that the fetched data is of the provided type.
 If the argument is omitted or ``DALIDataType.NO_TYPE`` is passed, the operator will infer
 the type based on the provided data.
 
-This argument will be required starting from DALI 2.0.)code", DALI_NO_TYPE);
+This argument will be required starting from DALI 2.0.)code", DALI_NO_TYPE)
+  .AddOptionalArg<int>("ndim", R"code(Input dimensionality.
```
szalpal, line 102: IMHO, this is way more understandable ;) Your call though.
```suggestion
  .AddOptionalArg<int>("ndim", R"code(Number of dimensions in the input.
```

dali/pipeline/operator/builtin/external_source.h
```diff
@@ -221,7 +223,15 @@ class ExternalSource : public Operator<Backend>, virtual public BatchSizeProvide
         device_id_(spec.GetArgument<int>("device_id")),
         dtype_(spec.GetArgument<DALIDataType>("dtype")),
         previous_dtype_(DALIDataType::DALI_NO_TYPE),
+        ndim_(-1),
+        layout_(),
         sync_worker_(device_id_, false) {
+    if (spec.TryGetArgument(ndim_, "ndim")) {
+      DALI_ENFORCE(ndim_ >= 0, "Incorrect number of dimensions. "
+                   "Use positive values for tensors or 0 for scalars.");
```
szalpal, line 231:
```suggestion
      DALI_ENFORCE(ndim_ >= 0, make_string("Incorrect number of dimensions (", ndim_,
                   "). Use positive values for tensors or 0 for scalars.");
```

dali/pipeline/operator/builtin/external_source.h
```diff
@@ -287,6 +297,23 @@ class ExternalSource : public Operator<Backend>, virtual public BatchSizeProvide
   DISABLE_COPY_MOVE_ASSIGN(ExternalSource);
 
  protected:
+  inline bool HasNdim() {
```
szalpal, line 300: What's the reason for making those `inline`? Per [standard specification](https://en.cppreference.com/w/cpp/language/inline)
> A function defined entirely inside a [class/struct/union definition](https://en.cppreference.com/w/cpp/language/classes), whether it's a member function or a non-member friend function, is implicitly an inline function

dali/pipeline/operator/builtin/external_source.h
```diff
@@ -287,6 +297,23 @@ class ExternalSource : public Operator<Backend>, virtual public BatchSizeProvide
...
+    return !layout_.empty() || spec_.HasArgument("ndim");
+  }
+
+  inline void InferNdim() {
+    if (!layout_.empty()) {
+      if (ndim_ != -1) {
+        DALI_ENFORCE(ndim_ == layout_.ndim(), make_string("Dimensionality of the provided "
```
szalpal, line 307: As above, I feel this would be easier to read. But, again, your call ;)
```suggestion
        DALI_ENFORCE(ndim_ == layout_.ndim(), make_string("Number of dimensions in the provided "
```

dali/pipeline/operator/builtin/external_source.h
```diff
@@ -509,6 +534,31 @@ class ExternalSource : public Operator<Backend>, virtual public BatchSizeProvide
...
     previous_dtype_ = batch.type();
 
+    auto input_ndim = batch.shape().sample_dim();
+    if (HasNdim()) {
+      DALI_ENFORCE(input_ndim == ndim_,
+                   make_string("ExternalSource expected data with ", ndim_, " dimensions and got ",
+                     input_ndim, " dimensions."));
+    } else if (ndim_ != -1) {
+      DALI_ENFORCE(input_ndim == ndim_,
+                   make_string("Dimensionality of the data fed to the external source has "
```
szalpal, line 544: Just a suggestion:
```suggestion
                   make_string("Number of dimensions of the data fed to the external source has "
```

dali/pipeline/operator/builtin/external_source.h
```diff
@@ -482,16 +507,16 @@ class ExternalSource : public Operator<Backend>, virtual public BatchSizeProvide
     }
   }
 
-  template<typename SrcBackend, template<typename> class SourceDataType>
-  inline void SetDataSourceHelper(const SourceDataType<SrcBackend> &batch, AccessOrder order = {},
-                                  ExtSrcSettingMode ext_src_setting_mode = {}) {
+  template <typename SrcBackend, template<typename> class SourceDataType>
+  inline void ValidateInputData(const SourceDataType<SrcBackend> &batch) {
```
szalpal, line 511: 👍🏻

dali/python/nvidia/dali/external_source.py
```diff
@@ -365,6 +365,14 @@ class ExternalSource():
 
     This argument will be required starting from DALI 2.0.
 
+`ndim` : int, optional
+    Input dimensionality.
```
szalpal, line 369:
```suggestion
    Number of dimensions in the input data.
```

dali/python/nvidia/dali/external_source.py
```diff
@@ -365,6 +365,14 @@ class ExternalSource():
...
+
+    The dimensionality of the data provided to the operator will be verified against this value.
+    Number of dimensions can be also inferred from the ``layout`` argument if provided.
+
```
szalpal, line 373: I'd also include here a sentence that if both are provided, they must match.

dali/python/nvidia/dali/external_source.py
```diff
@@ -666,20 +680,24 @@ def __call__(
                         kwargs['dtype'] = dtype[i] if i < len(dtype) else nvidia.dali.types.DALIDataType.NO_TYPE
                     else:
                         kwargs['dtype'] = dtype
+                if ndim is not None:
+                    if isinstance(ndim, (list, tuple)):
+                        kwargs['ndim'] = ndim[i] if i < len(ndim) else None
+                    else:
```
szalpal, line 686: I feel, that if the operator documentation specifies `ndim` as an int and user provides anything else, I'd just error out. I fear, that allowing 1-element lists or something like that may lead to convoluted type errors. How about going:
```python
if not isinstance(ndim, int):
   raise TypeError("ndim shall be an int")
```

dali/test/python/test_external_source_dali.py
```diff
@@ -255,3 +256,111 @@ def test_changing_dtype():
...
+
+def test_ndim_arg():
+    batch_size = 2
+    src_data = [
+        [np.ones((120, 120, 3), dtype=np.uint8)]*batch_size
+    ]
+    src_pipe = Pipeline(batch_size, 1, 0)
+    src_ext1 = fn.external_source(source=src_data, dtype=DALIDataType.UINT8, ndim=3)
+    src_ext2 = fn.external_source(source=src_data, dtype=DALIDataType.UINT8, layout="HWC")
+    src_pipe.set_outputs(src_ext1, src_ext2)
```
szalpal, line 269: How about:
```python
    @pipeline_def(batch_size, 1, 0)
    def pipe():
    return fn.external_source(source=src_data, dtype=DALIDataType.UINT8, ndim=3),
           fn.external_source(source=src_data, dtype=DALIDataType.UINT8, layout="HWC")
```
Your call

szalpal (commented):
Put some suggestions.

My concern is about allowing the `ndim` to be 1-elem list or tuple, while the doc says it's an int. Could you verify this?

dali/python/nvidia/dali/external_source.py
banasraf: This list of ints case is for multiple outputs

dali/pipeline/operator/builtin/external_source.h
banasraf: Every other method in this class had `inline`, so I followed. Yup, it doesn't make sense

dali/pipeline/operator/builtin/external_source.cc
banasraf: done

dali/pipeline/operator/builtin/external_source.h
banasraf: done

dali/pipeline/operator/builtin/external_source.h
banasraf: removed

dali/pipeline/operator/builtin/external_source.h
banasraf: done

dali/pipeline/operator/builtin/external_source.h
banasraf: done

dali/python/nvidia/dali/external_source.py
banasraf: done

dali/python/nvidia/dali/external_source.py
banasraf: done

dali/python/nvidia/dali/external_source.py
banasraf: added info to the docstring

szalpal (approved):

banasraf:
!build

banasraf:
!build

banasraf:
!build

banasraf:
!build

banasraf:
!build

PR #3768, opened 2022-03-29
Fix large mini-batch handling in parallel external source

stiepan:
## Category:

## Description:
Parallel external source passes data from workers to the main process through shared memory buffers.
If the capacity of a buffer exceeds `max_int`, worker process fails to serialize the minibatch meta-data (which includes the capacity) when it is written into C-like structure (python's `struct` package).

This PR:
1. Makes serialization error message more verbose, so that it contains values that the worker attempted to put in the struct.
2. Changes size-related members to be `unsigned long long int` instead of `int`.
3. Adds test for shared queue class if it handles the values properly
4. Adds a test that uses samples of 2GB in parallel external source.

1. `_multiproc` module and tests.

1. Maybe `unsigned long long int` is too much and simply unsigned `int` there will be good enough?

### Tests
- [ ] Existing tests apply
- [x] New tests added
  - [x] Python tests
  - [ ] GTests
  - [ ] Benchmark
  - [ ] Other
- [ ] N/A

### Documentation
- [ ] Documentation updated
  - [ ] Docstring
  - [ ] Doxygen
  - [ ] RST
  - [ ] Jupyter
  - [ ] Other
- [ ] N/A

#### Requirements
- [x] N/A

**REQ IDs**: N/A

**JIRA TASK**: DALI-2679

### DIFF
--- a/dali/test/python/test_external_source_parallel_shared_batch.py
+++ b/dali/test/python/test_external_source_parallel_shared_batch.py
@@ -1,4 +1,4 @@
-# Copyright (c) 2021, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# Copyright (c) 2021-2022, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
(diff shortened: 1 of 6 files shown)


JanuszL (approved):

dali/test/python/test_external_source_parallel_shared_batch.py
```diff
@@ -116,22 +116,24 @@ def next_i():
...
-            [recv_msg] = res_queue.get()
-            msg_values = all_msgs[received].get_values()
-            received += 1
-            recv_msg_values = recv_msg.get_values()
-            assert len(msg_values) == len(recv_msg_values)
-            assert all(msg_value == recv_msg_value for msg_value, recv_msg_value in zip(msg_values, recv_msg_values))
-    if not proc.exitcode:
-        task_queue.close()
-    proc.join()
+    try:
```
mzient, line 119: Q: no enter/exit to use `with proc`?

dali/test/python/test_external_source_parallel_shared_batch.py
```diff
@@ -144,3 +146,45 @@ def test_queue_recv():
...
+
+def test_queue_large_failure():
+    max_int32 = 2**31 - 1
+    max_uint32 = 2**32 - 1
+    error_message = "Failed to serialize object as C-like structure. Tried to populate following fields:"
+    for start_method in ("spawn", "fork"):
+        yield raises(RuntimeError, error_message)(_test_queue_large), \
+            start_method, [(max_int32 + 1, 0, max_uint32, max_uint32, max_uint32)]
+        yield raises(RuntimeError, error_message)(_test_queue_large), \
+            start_method, [(max_int32, max_int32, -1, 0, 0)]
```
mzient, line 190: Why should it fail now that there's a fix?

mzient (approved):

dali/test/python/test_external_source_parallel_shared_batch.py
stiepan: For plain proc not really, but good point that it looked like a good candidate for ctx manager.

stiepan:
!build

PR #3783, opened 2022-04-02
Improve cast operator performance

MirazSpecial:
Signed-off-by: Konrad Litwiński <klitwinski41418@gmail.com>

## Category:

**Refactoring** (*Redesign of existing code that doesn't affect functionality*) 

## Description:

Work's main motivation was to improve throughput for small batch sizes of data for **Cast**.

Originally running Cast kernel (BatchedCastKernel) required copying two arrays to GPU - 
- samples array (with descriptors of placement of samples in memory) - it used  8 * samples_number bytes.
- blocks array (with descriptors of what should be done for each threads block) - it used 20 * number_of_blocks bytes 
As there was a linear relationship between data size and number of blocks (number of blocks was around 1024 times smaller then data size) copying second array was a big cost of running Cast kernel. 

The idea of this optimization is to instead of copying block array, create an array with information of how big the samples are, and which block is the first one to parse every sample, copy it to GPU and then - in the kernel - calculate which sample should the block work on. To calculate that efficiently we use binary search over sample descriptors.   

## Additional information:

For image size 1000x1000 we achieved following improvement
image

### Affected modules and functionalities:

cast.cuh
cast.cu

### Key points relevant for the review:

The key changes are in the newly added BinSearchCastKernel kernel.

### Tests
- [x] Existing tests apply
- [ ] New tests added
  - [ ] Python tests
  - [ ] GTests
  - [ ] Benchmark
  - [ ] Other
- [ ] N/A

### Documentation
- [x] Existing documentation applies
- [ ] Documentation updated
  - [ ] Docstring
  - [ ] Doxygen
  - [ ] RST
  - [ ] Jupyter
  - [ ] Other
- [ ] N/A

#### Requirements
- [ ] Implements new requirements
- [ ] Affects existing requirements
- [ ] N/A

**REQ IDs**: N/A

**JIRA TASK**: N/A

PR #3784, opened 2022-04-03
Specialized Cast GPU kernel for smaller batch sizes

szkarpinski:
## Category:

## Description:

This PR is an extension of https://github.com/NVIDIA/DALI/pull/3783 and depends on it. Once https://github.com/NVIDIA/DALI/pull/3783 gets merged, I'll turn this PR from "draft" to "ready". By now, it's probably easiest to check out the diff between this PR and #3783 [here](https://github.com/szkarpinski/DALI/compare/d89def579edc92fd50bb3ff9f037197cd5910156...szkarpinski:cast-direct-experimental).

The https://github.com/NVIDIA/DALI/pull/3783 aims to improve Cast performance by reducing the size of host to device `cudaMemcpy` needed. For smaller batch sizes it is possible to remove *all* host to device `cudaMemcpy`'s, because all data needed fits in the 4KB parameter buffer. In this PR we introduce a specialized `DirectCastKernel` for batches of 8 samples and smaller, in which sample descriptor arrays are passed by value instead of by pointer.

We have mixed feelings about this idea, because it will add some complexity to DALI's code and we are not sure if it's worth it. The main purpose of this PR is to ask for your opinion.

### Why we **don't** need this PR
- Adds yet another kernel.
- Assumes that passing data parameter buffer via parameter buffer is (and will always be) faster.
- Assumes that the size description of single Cast sample will not exceed 512B.

### Why we need this PR
- Throughput gain, especially for single images, is huge! (see below)

### Masurements
The following plot shows the throughput improvement for casting 1000x1000x3 images from `uint8` to `float`. This was measured of Titan V GPU with `cast_bench.cc`.
image

- New `DirectCastKernel` added, together with `DirectCastKernelParamPack` structure to hold its parameters.

- As mentioned above, we are not certain if this idea fits "performance vs maintainablity" guidelines of DALI

### Tests
- [ ] Existing tests apply
- [ ] New tests added
  - [ ] Python tests
  - [ ] GTests
  - [ ] Benchmark
  - [ ] Other
- [ ] N/A

### Documentation
- [ ] Documentation updated
  - [ ] Docstring
  - [ ] Doxygen
  - [ ] RST
  - [ ] Jupyter
  - [ ] Other
- [ ] N/A

#### Requirements
- [ ] N/A

**REQ IDs**: N/A

**JIRA TASK**: N/A

### DIFF
--- a/dali/kernels/common/cast.cuh
+++ b/dali/kernels/common/cast.cuh
@@ -26,15 +26,46 @@ struct CastSampleDesc {
   const void *input;
 };
 
+struct CastSampleBlockDesc {
+  int first_block;
+  int sample_size;
+};
+
+template <typename OType, typename IType>
+__device__ __forceinline__ void CastKernelInternal(const CastSampleDesc& sample,
+                                                   int block_start, int block_end) {
+  auto *out = static_cast<OType *>(sample.output);
+  const auto *in = static_cast<const IType *>(sample.input);
+  for (int x = threadIdx.x + block_start; x < block_end; x += blockDim.x) {
+    out[x] = ConvertSat<OType>(in[x]);
+  }
+}
+
 template <typename OType, typename IType>
 __global__ void BatchedCastKernel(const CastSampleDesc *samples, const BlockDesc<1> *blocks) {
   const auto &block = blocks[blockIdx.x];
   const auto &sample = samples[block.sample_idx];
-  auto *out = static_cast<OType *>(sample.output);
-  const auto *in = static_cast<const IType *>(sample.input);
-  for (int x = threadIdx.x + block.start.x; x < block.end.x; x += blockDim.x) {
-    out[x] = ConvertSat<OType>(in[x]);
+  CastKernelInternal<OType, IType>(sample, block.start.x, block.end.x);
+}
+
+template <typename OType, typename IType>
+__global__ void BinSearchCastKernel(const CastSampleDesc *samples,
+                                    const CastSampleBlockDesc *params,
+                                    int nsamples, int block_volume_scale) {
+  int i = 0;
+  for (int jump = (1 << (32 - __clz(nsamples) - 1)); jump; jump /= 2) {
+    if (i + jump < nsamples && params[i + jump].first_block <= blockIdx.x)
+      i += jump;
   }
+  CastSampleDesc sample = samples[i];
+  int size = params[i].sample_size;
+  int block_offset = blockIdx.x - params[i].first_block;
+
+  int block_size = block_volume_scale * blockDim.x;
+  int block_start = block_offset * block_size;
+  int block_end = block_start + block_size <= size ? block_start + block_size : size;
+
+  CastKernelInternal<OType, IType>(sample, block_start, block_end);
 }
 
 }  // namespace kernels
--- a/dali/kernels/common/cast.cuh
+++ b/dali/kernels/common/cast.cuh
@@ -31,6 +31,13 @@ struct CastSampleBlockDesc {
   int sample_size;
 };
 
+const int DirectCastKernelMaxSamples = 8;
+struct DirectCastKernelParamPack {
+  CastSampleDesc samples[DirectCastKernelMaxSamples];
+  CastSampleBlockDesc blocks[DirectCastKernelMaxSamples];
+};
+
+
 template <typename OType, typename IType>
 __device__ __forceinline__ void CastKernelInternal(const CastSampleDesc& sample,
                                                    int block_start, int block_end) {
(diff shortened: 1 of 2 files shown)


klecki:
I wonder about feasibility of making the main implementation a device function taking the pointer to SampleDesc and adding thin layer of two `__gobal__` kernels where one just gets the SampleDesc via pointer and forwards it and the other gets it via the param buffer optionally unpacks it a bit and passes pointer to the data to the device function. Such pattern may be possible to generalize for other kernels if there is no much overhead from doing it that way.

If we manage to abstract the code complexity by adding some generic utility and still have gains similar to the one presented on the plot it would be great :)

I thought about alternative with dynamic parallelism and invoking other global functions from the kernel, but it would probably kill all the gains due to the overhead of the call.

klecki:
I wonder about feasibility of making the main implementation a device function taking the pointer to SampleDesc and adding thin layer of two `__gobal__` kernels where one just gets the SampleDesc via pointer and forwards it and the other gets it via the param buffer optionally unpacks it a bit and passes pointer to the data to the device function. Such pattern may be possible to generalize for other kernels if there is no much overhead from doing it that way.

If we manage to abstract the code complexity by adding some generic utility and still have gains similar to the one presented on the plot it would be great :)

I thought about alternative with dynamic parallelism and invoking other global functions from the kernel, but it would probably kill all the gains due to the overhead of the call.

szkarpinski:
Thank you for your comment! I thought about making this solution more general too, but I stumbled upon the following problem.

Parameter buffers reside in constant memory, which seems (at least on our Titan V with CC 7.0) not to support indexing with registers. What I mean is that accesses to constant memory compile to `mov ..., c[0x0][0x123]` in SASS, where both of the numbers are always constants.

The compiler seems to use the following workaround: when I try to access non-constant offset of the array passed as parameter, all the data from param buffer is copied to the local memory and the kernel continues to work on that copy, as local memory does support register indexing. However, copying to local memory results in a terrible performance drop.

This means that arrays passed to such kernels have to be accessed very carefully, without any variable indexing. In particular, this is the reason why I used unrolled loop to find the proper block instead of simply using binary search from @MirazSpecial's PR. This is also why I see no way of turning this idea into a more general "wrapper" for any kernel.

That being said, I couldn't find anything about limitations on addressing constant memory in  _Cuda Programming Guide_ and _PTX ISA_, so it's possible that I'm wrong. Maybe it's just NVCC that doesn't want to do it for some reason... I'll do some more research on the topic.

szkarpinski:
I tried to repeat the experiments to make some plots to illustrate the issues described above. This time, when I tried accessing the array in the parameter buffer with variable index, NVCC simply emmited a register-indexed access to constant memory, so there was no performance drop at all. I don't know whether my previous experimental setup was wrong or it is some nuance with NVCC. Anyway, sorry for the confusion.

Now, as the problem has magically disappeared, I will try to turn this idea into a general, thin wrapper, as you described.

klecki:
That's great to hear. If the problem reappears, we can follow up on that. Especially if you stumble upon a repro, it would be easier to ask around for clarification with example in hand.

szkarpinski:
Progress update: 

My dream solution would be to have some magic general wrapper for kernels like `copy_and_launch` that would copy the data from host to the device (either via parameter buffer or with `cudaMemcpy` for data bigger than 4KB) and launch the kernel.

I tried dynamic parallelism, but it's overhead is unacceptable in this usecase: dynamic kernel launch takes ~10x more time than Cast kernel itself for small image. I'll try other approaches and see ow close I can get to that ideal `copy_and_launch` wrapper. Once I get anything that works I'll open a new PR with it.

szkarpinski:
Hello, I'm sorry for the long silence. I finally found a repro for the strange issue I described above (i.e. NVCC making local copy of constant memory array).

Let's say we have a structure for passing array of numbers:
```cpp
struct Numbers {
  int data[256];
};
```
and now let's make a kernel that will iterate through those numbers and find two indices `a` and `b` in some random way:
```cpp
__global__ void goodKernel(int n, Numbers numbers, int* data) {
  int a, b;
  for (int i = 0; i < n; i++) {
    b = numbers.data[i];
    if (numbers.data[i] != 13) a = numbers.data[i];
  }
  data[b] = data[a];
}
```
So far so good: for `n=5` this kernel runs in ~8us and compiles to a sensible SASS. However, when I swap two lines:
```cpp
__global__ void badKernel(int n, Numbers numbers, int* data) {
  int a, b;
  for (int i = 0; i < n; i++) {
    if (numbers.data[i] != 13) a = numbers.data[i];
    b = numbers.data[i];
  }
  data[b] = data[a];
}
```
Boom! Tons of local memory stores and loads are generated and the execution time for `n=5` explodes to ~130us.

_This was compiled with NVCC 11.5 with `-O3` and `-arch=sm_70` and run on Titan V. I attached the source together with Makefile and SASS/PTX dumps as .zip to this comment._
[constant_memory_repro.zip](https://github.com/NVIDIA/DALI/files/8599620/constant_memory_repro.zip)

Sad conclusion from the above experiment is that using parameter space for passing bigger data, sample descriptors in particular, is not something NVCC would expect us to do and it results in unpredictable behaviour. 

I see no way of implementing the idea presented in this PR safely with the compiler behaving in such a strange way. So I think we can close this PR :(

klecki:
Thank you @szkarpinski for the great repro. I will follow up on that - I can indeed see the spill you are mentioning in the PTX and SASS.
Maybe we can get some guidance on how to proceed with that, but for now we are probably stuck with benchmarking individual solutions.

klecki:
I'm sorry for the delay in response, but I was waiting for the release of CUDA 11.7. There is new annotation for kernel parameters: https://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#grid-constant and I got a suggestion that it might help with this problem. I will try it out when I have some free cycles, if you want, you can give it a spin as well. Still, it won't be available in our CUDA 10 build, so this feature would probably need to wait for dropping that version.

szkarpinski:
Thank you very much! That `__grid_constant__` seems to be exactly what we need. I'll experiment with it next week to see how it works in this case.

JanuszL:
@szkarpinski - sure, no rush.

szkarpinski:
Closing, at least for now - this won't be practical (would require tons of ifdefs etc) until we drop support for CUDA<11.7

PR #3795, opened 2022-04-05
Features required for `autoserialization` in DALI Backend

szalpal:
## Category:

## Description:
This PR is related to https://github.com/triton-inference-server/dali_backend/pull/125. Please refer there for the overall feature description.

For the autoserialize feature in DALI Backend we need two enhancements in DALI:
1. API to check, if given pipeline is a formally valid DALI pipeline,
2. API that allows expressing `@pipeline_def` as a pipeline to be autoserialized (again, please refer to the https://github.com/triton-inference-server/dali_backend/pull/125 for details).

ad. 1, I'm adding a function to DALI's C API - `daliIsDeserializable`. Current implementation simply attempts to deserialize the pipeline and if it succeeds (i.e. pipeline is successfully deserialized), returns `true`. Naturally, this introduces a redundant serialization. However, I believe that this isn't a bottleneck in the processing. Additionally, I'm not aware of any other way of validating serialized object in protobuf.

ad. 2, I'm introducing a new module and a new decorator: `@nvidia.dali.triton.autoserialize`. The purpose of it is to mark the DALI pipeline in the Triton model repo, which will be a subject for autoserialization. Typical usage in reflected in the following example:
```python
# dali.py
import nvidia.dali as dali


@dali.triton.autoserialize
@dali.pipeline_def(batch_size=3, num_threads=1, device_id=0)
def pipe():
    data = dali.fn.external_source(device="cpu", name="DALI_INPUT_0")
    return data
```
The assumption is that whatever is marked by the `autoserialize`, shall be a `pipeline_def`. Misusing the decorator (e.g. importing a module with it) is a no-op.

## Additional information:
1. First commit is correcting an indentation is `pipline.h` file. You may want to turn it off during the review.

1. Please verify, if the `dali.triton.autoserialize` tests need any additional usage patterns to test.
2. 

### Tests
- [ ] Existing tests apply
- [ ] New tests added
  - [x] Python tests
  - [x] GTests
  - [ ] Benchmark
  - [x] Other
- [ ] N/A

### Documentation
- [ ] Documentation updated
  - [x] Docstring
  - [x] Doxygen
  - [ ] RST
  - [ ] Jupyter
  - [ ] Other
- [ ] N/A

#### Requirements
- [x] N/A

**REQ IDs**: N/A

**JIRA TASK**: N/A


szalpal:
!build

szalpal:
!build

szalpal:
!build

szalpal:
!build

szalpal:
!build

dali/python/nvidia/dali/triton.py
```diff
@@ -0,0 +1,70 @@
...
+
+import sys
+
+pipeline_autoserialized = False
+
+
+def autoserialize(dali_pipeline):
+    """
+    Decorator, that marks a DALI pipeline (represented by :meth:`nvidia.dali.pipeline_def`) for
+    autoserialization in DALI Backend's model repository.
```
klecki, line 23: I think this doc requires some links and a mention that it's DALI Backend for Triton.

dali/python/nvidia/dali/triton.py
```diff
@@ -0,0 +1,70 @@
...
+
+    For details about the autoserialization feature, please refer to the DALI Backend documentation.
```
klecki, line 25: I assume we will get a link here?

dali/python/nvidia/dali/triton.py
```diff
@@ -0,0 +1,70 @@
...
+        import nvidia.dali as dali
+
+        @dali.triton.autoserialize
+        @dali.pipeline_def(batch_size=3, num_threads=1, device_id=0)
+        def pipe():
+            data = dali.fn.external_source(device="cpu", name="DALI_INPUT_0")
+            return data
+
+
+        $ python identity.py /tmp/model.dali autoserialize.me
```
klecki, line 42: You sure about that usage?

As an alternative I would suggest trying to import that as a module (so a safeguard against someone putting actual `__main__` section), and looking for a function that would be marked for export by the autoserialize.

Another alternative would be to look how tools like nose does the test case discovery and doing something similar.

dali/python/nvidia/dali/triton.py
```diff
@@ -0,0 +1,70 @@
...
+    contains the pipeline marked with ``autoserialize``, pass the target file name
+    (i.e. the path to the file, where the serialized model will be saved) and the magical command
+    ``autoserialize.me``. For example:
+
+        # identity.py
+        import nvidia.dali as dali
+
+        @dali.triton.autoserialize
+        @dali.pipeline_def(batch_size=3, num_threads=1, device_id=0)
+        def pipe():
```
klecki, line 37: What about any arguments to the pipeline_def or the `pipe` function?

klecki (commented):
Some more or less general remarks.

I am thinking whether the user would always create a special file with just the pipeline definition or be allowed to slap an `@autoserialize` into a file with the definition for training and reference it in the triton config. The latter is much more sloppy, but possible scenario.

qa/TL0_python-triton/test.py
```diff
@@ -0,0 +1,67 @@
+# Copyright (c) 2022, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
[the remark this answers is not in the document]
szalpal: How would you propose to organise these tests?

szalpal:
> I am thinking whether the user would always create a special file with just the pipeline definition or be allowed to slap an `@autoserialize` into a file with the definition for training and reference it in the triton config. The latter is much more sloppy, but possible scenario.

Whatever is to be run within Triton, has to exist in the model repository, which has [a specified structure](https://github.com/triton-inference-server/server/blob/main/docs/model_repository.md#model-repository). I doubt that training routines will follow this structure - therefore the file is going to be copied (or created) anyway. So one of the usage scenarios I envision is that user will copy the file with a pipeline and add `@autoserialize(train_pipeline=False)` there. Or something along the lines.

dali/python/nvidia/dali/triton.py
szalpal: This is a great idea. I believe the usage should be something like:
```python
@autoserialize(train_pipeline=False)
@pipeline_def
def pipe(train_pipeline=True):
   ...
```
Anyway, I'd like to make it rather a follow-up to this PR.

dali/python/nvidia/dali/triton.py
szalpal: Added a link to DALI Backend. I didn't mention Triton here, since this file is named `triton.py`, which make the context obvious. Do you believe I shall mention it also in the doc?

dali/python/nvidia/dali/triton.py
```diff
@@ -0,0 +1,70 @@
+# Copyright (c) 2022, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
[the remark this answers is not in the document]
szalpal: Yeah, this could be also. @banasraf , what's your opinion?

Maybe we should think about the reversed situation? It seems to me that it'd be better to import frameworks like
```python
import nvidia.dali.mxnet
import nvidia.dali.tf
```

dali/python/nvidia/dali/triton.py
szalpal: Do I understand correctly that you propose to do something along:
```python
# mydali.py
@autoserialize
@pipeline_def
def mypipe():
   ...
```
```bash
$ python -c 'import mydali; f = discover_autoserialize(); f()'
```
This actually might work. I'll try to do it this way. And you're right we should secure it from actually putting the `__main__` magic.

About how does nose discover tests - it uses the name (i.e. when a function contains a `test` word) and the special attribute: `__test__`.

qa/TL0_python-triton/test.py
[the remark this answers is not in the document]
klecki: I would be ok with creating directory in the Python test location, and having one entry point there that would import the rest of the code. So I would be able to run it in dali/python/test using nose as all other tests.

dali/python/nvidia/dali/triton.py
klecki: I think it would do no harm to include the link. With the docs I think in most cases explicit is better than implicit. Some may not know what Triton is.

dali/python/nvidia/dali/triton.py
klecki: Something like this. You can probably look for whatever global symbol has the some magic member set by autoserialize or some similar mechanism.

dali/python/nvidia/dali/triton.py
klecki: I was thinking, that autoserialize could probably detect the arguments to `pipe` function, without the need to redeclare them in decorator and export (make available) them in the DALI backend as configuration? 
That would probably only work with builtins though.

dali/python/nvidia/dali/triton.py
```diff
@@ -0,0 +1,74 @@
...
+
+    Only a ``pipeline_def`` can be decorated with ``autoserialize``.
+
+    Only one ``pipeline_def`` may be decorated with ``autoserialize`` in given program.
+
+    :param dali_pipeline: DALI Python model definition (``pipeline_def``)
+    """
+    assert getattr(dali_pipeline, "is_pipeline_def",
+                   False), "Only `@pipeline_def` can be decorated with `@triton.autoserialize`."
+    dali_pipeline.autoserialize_me = True
```
klecki, line 73: Again, not sure if I am a fan of public member, also if we start doing such similar decorators, maybe we should follow some similar naming pattern and not `autoserialize_me` (just a nitpick).

dali/python/nvidia/dali/triton.py
```diff
@@ -0,0 +1,74 @@
...
+    """
+    TODO
+    :param module:
+    :return:
+    """
+    assert module is not None
+    ret = []
+    mems = inspect.getmembers(module)
+    modules = []
+    for mem in mems:
```
klecki, line 28: I suggest a bit more verbose naming rather than `mem[0]` and `mem[1]`
```suggestion
    for name, member in module_members:
```

dali/python/nvidia/dali/triton.py
```diff
@@ -0,0 +1,74 @@
...
+        obj = getattr(module, mem[0], None)
+        if inspect.ismodule(obj) and mem[1] not in visited:
+            modules.append(mem[0])
+            visited.append(mem[1])
+        elif inspect.isfunction(obj) and getattr(obj, 'autoserialize_me', False):
```
klecki, line 33: For clarity, maybe wrap the check `getattr(obj, 'autoserialize_me', False)` into function like `is_marked_autoserializable` or something?

dali/python/nvidia/dali/triton.py
```diff
@@ -0,0 +1,74 @@
...
+def invoke_autoserialize(module, filename):
+    """
+    TODO
+    :param module:
+    :param filename:
+    :return:
+    """
+    autoserialize_me_functions = _discover_autoserialize(module)
+    assert len(
+        autoserialize_me_functions) == 1, f"Precisely one autoserialize function must exist in the module. Discovered: {autoserialize_me_functions}"
```
klecki, line 49: Asserts shouldn't be used for error reporting I think. Please use proper exception type.

klecki: Also, probably we should have a nice error for case where we didn't find any, and where we found too many.

klecki (commented):
Some more comments

dali/python/nvidia/dali/triton.py
szalpal: This won't go along with autoconfiguration of DALI Backend.

Yeah, we have lot's of `auto` stuff...

I mean, it will unless any of the `def pipe()` args does not have a default value. Then it won't be possible to infer the argument value anyhow. Treating `def pipe()` args as `kwargs` for the decorator resolves this issue and allows nicer error checking. Anyway, that's the reason I'd like to do this in the followup PR - we still need to discuss the solution.

dali/python/nvidia/dali/triton.py
szalpal: Done! How about now?

dali/python/nvidia/dali/triton.py
szalpal: Done

dali/python/nvidia/dali/triton.py
szalpal: Done

dali/python/nvidia/dali/triton.py
szalpal: Done

dali/python/nvidia/dali/pipeline.py
```diff
@@ -1324,7 +1324,9 @@ def create_pipeline(*args, **kwargs):
                 pipe.set_outputs(*po)
             return pipe
         return create_pipeline
-    return actual_decorator(fn) if fn else actual_decorator
+    ret = actual_decorator(fn) if fn else actual_decorator
+    ret.is_pipeline_def = True
```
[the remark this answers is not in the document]
szalpal: I didn't do very extensive research in this topic, however this pattern seems to be popular enough. TBH, I can't imagine a scenario, where `is_pipeline_def` could be abused.

And if it should be private? I guess the most common way it could be used is outside the `pipeline_def` body, which rather gives the incentive to make it public. After all, why a regular user shouldn't be allowed to use it?

dali/python/nvidia/dali/triton.py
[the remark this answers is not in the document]
banasraf: For now, for the sake of consistency I would make it `nvidia.dali.plugin.triton`. If we decide to move framework integrations to the `nvidia.dali` level, we can do it later for all of them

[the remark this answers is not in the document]
szalpal: Done

qa/TL0_python-triton/test.py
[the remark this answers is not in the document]
szalpal: Rearranged the tests

dali/python/nvidia/dali/test/triton/improper_decorated_function.py
```diff
@@ -0,0 +1,6 @@
+import nvidia.dali as dali
+
+
+@dali.triton.autoserialize
+def func_under_test():
+    return 42
```
banasraf, line 6: \n

dali/python/nvidia/dali/pipeline.py
```diff
@@ -1323,6 +1323,7 @@ def create_pipeline(*args, **kwargs):
                     po = (pipe_outputs,)
                 pipe.set_outputs(*po)
             return pipe
+        create_pipeline.is_pipeline_def = True  # Add `is_pipeline_def` attribute to the function marked as `@pipeline_def`
```
klecki, line 1326: I suggest we should at least keep them private by adding a leading underscore. This will allow us to change it in the future and suggest that this is just implementation detail.

In the future we can add a function like `is_pipeline_def` that would check for the presence of this field.

BTW, what if someone makes it false by hand? :P

dali/python/nvidia/dali/triton.py
klecki: Yes, I agree that `_is_autoserialize` would be better. Note: With leading underscore. 

The `me` part is weird.

dali/python/nvidia/dali/plugin/triton.py
```diff
@@ -0,0 +1,36 @@
...
+    autoserialization in [DALI Backend's](https://github.com/triton-inference-server/dali_backend#dali-triton-backend) model repository.
+
+    For details about the autoserialization feature, please refer to the
+    [DALI Backend documentation](https://github.com/triton-inference-server/dali_backend#autoserialization).
+
+    Only a ``pipeline_def`` can be decorated with ``autoserialize``.
+
+    Only one ``pipeline_def`` may be decorated with ``autoserialize`` in given program.
+
+    To perform autoserialization, please refer to :meth:`nvidia.dali._utils.invoke_autoserialize`.
```
klecki, line 27: Why do we document something from a private module? Is it supposed to be use-facing functionality, or is it something internal?

dali/python/nvidia/dali/plugin/triton.py
```diff
@@ -0,0 +1,36 @@
...
+    """
+    Decorator, that marks a DALI pipeline (represented by :meth:`nvidia.dali.pipeline_def`) for
+    autoserialization in [DALI Backend's](https://github.com/triton-inference-server/dali_backend#dali-triton-backend) model repository.
+
+    For details about the autoserialization feature, please refer to the
+    [DALI Backend documentation](https://github.com/triton-inference-server/dali_backend#autoserialization).
+
+    Only a ``pipeline_def`` can be decorated with ``autoserialize``.
+
+    Only one ``pipeline_def`` may be decorated with ``autoserialize`` in given program.
```
klecki, line 25:
```suggestion
    Only one ``pipeline_def`` may be decorated with ``autoserialize`` in a given program.
```

dali/python/nvidia/dali/_utils/autoserialize.py
```diff
@@ -0,0 +1,70 @@
...
+            modules.append(name)
+            visited.append(path)
+        elif inspect.isfunction(obj) and _is_marked_autoserializable(obj):
+            ret.append(obj)
+    for mod in modules:
+        ret.extend(_discover_autoserialize(getattr(module, mod, None), visited=visited))
+    return ret
+
+
+def invoke_autoserialize(head_module, filename):
```
klecki, line 47: Just a nitpick, but maybe this whole file should also be present in triton module? It depends on the decorator that's already there.

dali/test/python/test_autoserialize.py
```diff
@@ -0,0 +1,40 @@
+# Copyright (c) 2022, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
klecki, line 1: Note that this file probably doesn't match any pattern form the qa/ scripts. Naming it test_pipeline_autoserialize would do. Or adding it explicitly.

dali/test/python/test_autoserialize.py
```diff
@@ -0,0 +1,40 @@
...
+    from .triton import decorated_function
+    invoke_autoserialize(decorated_function, serialized_filename)
+
+
+def test_indirect_import():
+    from .triton import imports_decorated_function
+    invoke_autoserialize(imports_decorated_function, serialized_filename)
+
+
+@raises(RuntimeError)
```
klecki, line 31: How about checking for the error message if the expected error is triggered?

klecki (commented):
Rather minor issues, I think it should be ok after those.

dali/test/python/test_autoserialize.py
szalpal: Done

dali/test/python/test_autoserialize.py
szalpal: Added

dali/python/nvidia/dali/plugin/triton.py
szalpal: User will face also a documentation in the Triton. That one will most probably be the first one (maybe the only one...) he sees. The docs here on the other hand targets the developer of DALI Backend, so that he should know, how to actually perform the autoserialization. That was my rationale

dali/python/nvidia/dali/plugin/triton.py
szalpal: Done

dali/python/nvidia/dali/_utils/autoserialize.py
szalpal: I could also see, that the `_is_autoserialize` is not the only attribute we might want to search for. In this case, adding a new one will be a matter of changing some names. In this case I believe it fits better here. Also, I believe it's better to have one `_utils` module, then several of them spread over. For these reasons, I'd go for keeping it here, but if you insist, I'll move it.

klecki: I'm ok, that was just a question.

klecki (approved):

szalpal:
!build

szalpal:
!build

klecki:
How about instead of adding `__init__` files and treating them as actual module, you can add the directory with tests to the path `sys.path.append('triton/')` and import as from current directory.

szalpal:
!build

PR #3840, opened 2022-04-22
Audio resampling operator

mzient:
## Category:

## Description:
Adds a standalone audio resampling operator.
The resampling can be parameterized with either:
- input and output sampling rates (scale and length are calculated from these)
- scale factor (the length is calculated from the scale)
- new length (scale is the ratio of lengths)

Audio decoder op (some minor refactoring).

### Tests
- [ ] Existing tests apply
- [X] New tests added
  - [X] Python tests
  - [ ] GTests
  - [ ] Benchmark
  - [ ] Other
- [ ] N/A

### Documentation
- [ ] Documentation updated
  - [X] Docstring
  - [ ] Doxygen
  - [ ] RST
  - [ ] Jupyter
  - [ ] Other
- [ ] N/A

#### Requirements
- [X] Implements new requirements
- [ ] N/A

**REQ IDs**: TODO

**JIRA TASK**: DALI-2750

### DIFF
--- a/dali/operators/audio/resample.cc
+++ b/dali/operators/audio/resample.cc
@@ -1,0 +1,168 @@
+// Copyright (c) 2022, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include <map>
+#include <vector>
+#include "dali/operators/audio/resample.h"
+#include "dali/operators/audio/resampling_params.h"
+#include "dali/kernels/kernel_params.h"
+
+namespace dali {
+
+DALI_SCHEMA(experimental__AudioResample)
+  .DocStr(R"(Resamples a signal with a different sampling rate.
+
+The resampling is achieved by applying a sinc filter with Hann window with an extent
+controlled by the ``quality`` argument.
+
+The resampling ratio can specified either directly or as a ratio of target to source sampling rate
+or calculated from the ratio of requested output length to input length.
+)")
+  .NumInput(1)
+  .NumOutput(1)
+  .AddOptionalArg<float>("in_rate", R"(Input sampling rate.
+
+The sampling rate of the input sample. This parameter must be specified together with ``out_rate``.
+The value is relative to ``out_rate`` and doesn't need to use any specific unit as long as the
+units of input and output rates match.
+
+The ``in_rate`` and ``out_rate`` parameters cannot be specified together with ``scale`` or
+``out_length``.)",
+    nullptr, true)
+  .AddOptionalArg<float>("out_rate", R"(Input sampling rate.
+
+The sampling rate of the input sample. This parameter must be specified together with ``out_rate``.
+The value is relative to ``out_rate`` and doesn't need to use any specific unit as long as the
+units of input and output rates match.
+
+The ``in_rate`` and ``out_rate`` parameters cannot be specified together with ``scale`` or
+``out_length``.)",
+    nullptr, true)
+  .AddOptionalArg<float>("scale", R"(The scaling factor.
+
+The scaling factor is the ratio of the target sampling rate to source sampling rate. For example,
+a ``scale=2`` will produce an output with twice as many samples as the input.
+
+This parameter cannot be specified together with ``in_rate`` and ``out_rate`` or ``out_length``.)",
+    nullptr, true)
+  .AddOptionalArg<int64_t>("out_length", R"(The requested output length, in samples.
+
+The scaling factor is the ratio of this output length to the input length. This parameter
+cannot be specified together with ``in_rate``, ``out_rate`` or ``scale``.)",
+    nullptr, true)
+  .AddOptionalArg("quality", R"(Resampling quality, where 0 is the lowest, and 100 is
+the highest.
+
+0 gives 3 lobes of the sinc filter, 50 gives 16 lobes, and 100 gives 64 lobes.)",
+    50.0f, false)
+  .AddOptionalArg<DALIDataType>("dtype", R"(The ouput type.
+
+If not specified, the output type is the same as the input type. When the type is changed,
+the values are normalized to fill the dynamic range of the target data type. When converting
+floating point inputs to integer types, the values are assumed to be in -1..1 range. When converting
+between signed and unsigned types, 0 translates to half-range of the unsigned type.)",
+    nullptr, false);
+
+namespace audio {
+
+using kernels::InTensorCPU;
+using kernels::OutTensorCPU;
+
+class ResampleCPU : public ResampleBase<CPUBackend> {
+ public:
+  using Base = ResampleBase<CPUBackend>;
+  explicit ResampleCPU(const OpSpec &spec) : Base(spec) {
+    auto params = ResamplingParams::FromQuality(quality_);
+    R.Initialize(params.lobes, params.lookup_size);
+  }
+
+  void RunImpl(HostWorkspace &ws) override {
+    auto &out = ws.template Output<CPUBackend>(0);
+    const auto &in = ws.template Input<CPUBackend>(0);
+
+    const auto &out_shape = out.shape();
+    const auto &in_shape = in.shape();
+    out.SetLayout(in.GetLayout());
+    int N = in.num_samples();
+    assert(N == static_cast<int>(scales_.size()));
+    assert(out.type() == dtype_);
+
+    auto &tp = ws.GetThreadPool();
+    in_fp32.resize(tp.NumThreads());
+    for (int s = 0; s < N; s++) {
+      tp.AddWork([&, this, s](int thread_idx) {
+        InTensorCPU<float> in_view;
+        TYPE_SWITCH(in.type(), type2id, T, (AUDIO_RESAMPLE_TYPES),
+          (in_view = ConvertInput(in_fp32[thread_idx], view<const T>(in[s]));),
+          (DALI_FAIL(
+              make_string("Unsupported output type: ", dtype_,
+                          "\nSupported types are : ", ListTypeNames<AUDIO_RESAMPLE_TYPES>()));));
+        TYPE_SWITCH(dtype_, type2id, T, (AUDIO_RESAMPLE_TYPES),
+          (ResampleTyped<T>(view<T>(out[s]), in_view, scales_[s]);),
+          (assert(!"Unreachable code.")));
+      });
+    }
+    tp.RunAll();
+  }
+
+  template <typename T>
+  void ResampleTyped(const OutTensorCPU<T> &out, const InTensorCPU<float> &in, double scale) {
+    int ch = out.shape.sample_dim() > 1 ? out.shape[1] : 1;
+    R.Resample(out.data, 0, out.shape[0], scale, in.data, in.shape[0], 1, ch);
+  }
+
+  template <typename T>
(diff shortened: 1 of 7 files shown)


dali/operators/audio/resample.h
```diff
@@ -0,0 +1,128 @@
...
+        int64_t in_length = shape.tensor_shape_span(s)[0];
+        int64_t out_length =
+          kernels::signal::resampling::resampled_length(in_length, in_rate, out_rate);
+        out_shape.tensor_shape_span(s)[0] = out_length;
+      }
+    } else if (scale_.HasValue()) {
+      scale_.Acquire(spec_, ws, N);
+      for (int s = 0; s < N; s++) {
+        double scale = scale_[s].data[0];
+        scales_[s] = scale;
```
mzient, line 92: TODO: Verify that the scale is positive.

dali/operators/audio/resample.h
```diff
@@ -0,0 +1,128 @@
...
+        int64_t out_length =
+          kernels::signal::resampling::resampled_length(in_length, 1, scale);
+        out_shape.tensor_shape_span(s)[0] = out_length;
+      }
+    } else if (out_length_.HasValue()) {
+      out_length_.Acquire(spec_, ws, N);
+      for (int s = 0; s < N; s++) {
+        int64_t in_length = shape.tensor_shape_span(s)[0];
+        int64_t out_length = out_length_[s].data[0];
+        scales_[s] = 1.0 * out_length / in_length;
```
mzient, line 103: TODO: Verify that the input is not empty if output isn't. Add special case for 0/0.

dali/operators/audio/resampling_params.h
```diff
@@ -0,0 +1,36 @@
...
+#define DALI_OPERATORS_AUDIO_RESAMPLING_PARAMS_H_
+
+#include <cmath>
+
+namespace dali {
+namespace audio {
+
+struct ResamplingParams {
+  int lobes = 16;
+  int lookup_size = 2048;
```
JanuszL, line 25:
```suggestion
  int lookup_size = 1025;
```
If we use `lobes * 64 + 1` formula then for `lobes = 16` we should provide 1025 by default, ... I guess.

dali/test/python/test_operator_audio_resample.py
```diff
@@ -0,0 +1,68 @@
+# Copyright (c) 2022, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
JanuszL, line 1: Please add this a test to test_dali_cpu_only.py and test_dali_variable_batch_size.py.

dali/operators/audio/resample.cc
```diff
@@ -0,0 +1,168 @@
...
+#include <map>
+#include <vector>
+#include "dali/operators/audio/resample.h"
+#include "dali/operators/audio/resampling_params.h"
+#include "dali/kernels/kernel_params.h"
+
+namespace dali {
+
+DALI_SCHEMA(experimental__AudioResample)
+  .DocStr(R"(Resamples a signal with a different sampling rate.
```
JanuszL, line 24:
```suggestion
  .DocStr(R"(Resamples a signal with a given sampling rate.
```

dali/operators/audio/resample.cc
```diff
@@ -0,0 +1,168 @@
...
+
+The resampling is achieved by applying a sinc filter with Hann window with an extent
+controlled by the ``quality`` argument.
+
+The resampling ratio can specified either directly or as a ratio of target to source sampling rate
```
JanuszL, line 29:
```suggestion
The resampling ratio can specified either directly or as a ratio of target to source sampling rate
```

dali/operators/audio/resample.cc
```diff
@@ -0,0 +1,168 @@
...
+or calculated from the ratio of requested output length to input length.
```
JanuszL, line 30:
```suggestion
The resampling ratio can specified either directly, or as a ratio of target to source sampling rate,
or calculated from the ratio of requested output length to input length.
```

dali/operators/audio/resample.cc
```diff
@@ -0,0 +1,168 @@
...
+  .AddOptionalArg<float>("in_rate", R"(Input sampling rate.
+
+The sampling rate of the input sample. This parameter must be specified together with ``out_rate``.
+The value is relative to ``out_rate`` and doesn't need to use any specific unit as long as the
+units of input and output rates match.
+
+The ``in_rate`` and ``out_rate`` parameters cannot be specified together with ``scale`` or
+``out_length``.)",
+    nullptr, true)
+  .AddOptionalArg<float>("out_rate", R"(Input sampling rate.
```
JanuszL, line 43:
```suggestion
  .AddOptionalArg<float>("out_rate", R"(Output sampling rate.
```

dali/operators/audio/resample.cc
```diff
@@ -0,0 +1,168 @@
...
+
+The sampling rate of the input sample. This parameter must be specified together with ``out_rate``.
```
JanuszL, line 45:
```suggestion
The sampling rate of the output sample. This parameter must be specified together with ``out_rate``.
```

dali/operators/audio/resample.cc
```diff
@@ -0,0 +1,168 @@
...
+The value is relative to ``out_rate`` and doesn't need to use any specific unit as long as the
```
JanuszL, line 46:
```suggestion
The value is relative to ``in_rate`` and doesn't need to use any specific unit as long as the
```

dali/operators/audio/resample.cc
```diff
@@ -0,0 +1,168 @@
...
+
+The scaling factor is the ratio of this output length to the input length. This parameter
+cannot be specified together with ``in_rate``, ``out_rate`` or ``scale``.)",
+    nullptr, true)
+  .AddOptionalArg("quality", R"(Resampling quality, where 0 is the lowest, and 100 is
+the highest.
+
+0 gives 3 lobes of the sinc filter, 50 gives 16 lobes, and 100 gives 64 lobes.)",
+    50.0f, false)
+  .AddOptionalArg<DALIDataType>("dtype", R"(The ouput type.
```
JanuszL, line 69: I don't know if this is clear enough here, that the input should be either -1, 1 for floats, or full range for integers?

dali/operators/audio/resample.cc
```diff
@@ -0,0 +1,168 @@
...
+    for (size_t i = 0; i < tmp.size(); i++) {
+      // squeeze to 0..1 range
+      tmp[i] = (ConvertSatNorm<float>(in.data[i]) + 1.0f) * 0.5f;
+    }
+    return make_tensor_cpu(tmp.data(), in.shape);
+  }
+
+ private:
+  kernels::signal::resampling::Resampler R;
+  std::vector<std::vector<float>> in_fp32;
```
JanuszL, line 160:
```suggestion
  kernels::signal::resampling::Resampler R_;
  std::vector<std::vector<float>> in_fp32_;
```

dali/operators/audio/resample.cc
```diff
@@ -0,0 +1,168 @@
...
+  }
+
+  template <typename T>
+  void ResampleTyped(const OutTensorCPU<T> &out, const InTensorCPU<float> &in, double scale) {
+    int ch = out.shape.sample_dim() > 1 ? out.shape[1] : 1;
+    R.Resample(out.data, 0, out.shape[0], scale, in.data, in.shape[0], 1, ch);
+  }
+
+  template <typename T>
+  InTensorCPU<float> ConvertInput(std::vector<float> &tmp, const InTensorCPU<T> &in) {
```
JanuszL, line 126: I'm not sure if `ConvertInput` needs to be a member. All it uses is just `dtype_`, other parameters are passed as arguments.

dali/operators/audio/resample.h
```diff
@@ -0,0 +1,153 @@
...
+    if (in_rate_.HasValue() + scale_.HasValue() + out_length_.HasValue() > 1)
+      DALI_FAIL("The sampling rates, ``scale`` and ``out_length`` cannot be used together.");
+    if (!in_rate_.HasValue() && !scale_.HasValue() && !out_length_.HasValue())
+      DALI_FAIL("No resampling factor specified! Please supply either the scale, "
+        "the output length or the input and output sampling rates.");
+    quality_ = spec_.template GetArgument<float>("quality");
+    DALI_ENFORCE(quality_ >= 0 && quality_ <= 100, make_string("``quality`` out of range: ",
+      quality_, "\nValid range is [0..100]."));
+    if (spec_.TryGetArgument(dtype_, "dtype")) {
+      // silence useless warning -----------------------------vvvvvvvvvvvvvvvv
```
JanuszL, line 44: ?

dali/operators/audio/resample.h
```diff
@@ -0,0 +1,153 @@
...
+      TYPE_SWITCH(dtype_, type2id, T, (AUDIO_RESAMPLE_TYPES), (T x; (void)x;),
+      (DALI_FAIL(make_string("Unsupported output type: ", dtype_,
```
JanuszL, line 46:
```suggestion
      TYPE_SWITCH(dtype_, type2id, T, (AUDIO_RESAMPLE_TYPES),
      (T x; (void)x;),
      (DALI_FAIL(make_string("Unsupported output type: ", dtype_,
```
I think this reads better for the `TYPE_SWITCH`.

dali/operators/audio/resample.h
```diff
@@ -0,0 +1,153 @@
...
+
+  DALIDataType dtype_ = DALI_NO_TYPE;
+  float quality_ = 50;
+
+  ArgValue<float> in_rate_{"in_rate", spec_};
+  ArgValue<float> out_rate_{"out_rate", spec_};
+  ArgValue<float> scale_{"scale", spec_};
+  ArgValue<int64_t> out_length_{"out_length", spec_};
+
+  std::vector<double> scales_;
```
JanuszL, line 147: Easy to confuse with `scale_`.

dali/operators/audio/resample.cc
mzient: Actually, neither is true. "given" suggests an absolute value and that's not how this operator works - you don't even need to supply sampling rates directly (just the ratio). Also, I could add a fast path for scale==1 (so if the rate isn't really different, there would be no perf hit).

dali/operators/audio/resample.cc
mzient: Actually, the rules for comma before "or" are more complex - there should be one if an _independent clause_ follows or.

> The resampling ratio can be specified directly, or it can be calculated as a ratio of input and output sampling rates.

Here, we have two independent clauses ("[ratio] can be be specified..." and "it can be calculated...").
In the original text, however, the clauses are dependent - they both refer to "be specified": "directly or as a ratio...".

Having said all that, the sentence is indeed incorrect, because:
- it's missing the verb
- it uses "either" in front of multiple choice

I think it should be:
> The resampling ratio can be specified directly or as a ratio of target to source sampling rate, or calculated...

but perhaps the second comma is not mandatory, as the last clause can be treated as dependent wrt "be" (specified or calculated).

dali/operators/audio/resample.cc
JanuszL: Up to you. I just wanted to point out that not necessarily the sampling rate need to be different.

dali/operators/audio/resample.cc
JanuszL: Up to you

dali/operators/audio/resample.cc
mzient: Well, -1..1 for integers makes _very_ little sense. I could add examples.

JanuszL: Up to you. On the other hand we don't do that for other ops....

dali/operators/audio/resample.h
mzient: Without this weird `T x; (void)x;` I get "unused local typedef" warning for `T`.

dali/test/python/test_dali_variable_batch_size.py
```diff
@@ -1020,6 +1020,7 @@ def test_subscript_dim_check():
                    operator_fn=fn.subscript_dim_check, num_subscripts=1)
 
 
+
```
JanuszL, line 1023: I think you need to add `fn.experimental.audio_resample` to the below as well.

JanuszL (approved):

stiepan (approved):

PR #3879, opened 2022-05-09
Add sequence processing to warp operator

stiepan:
## Category:

## Description:

Adds FHWC FDHWC processing to the warp affine operator. The transformation matrix can be specified per-frame and provided either as a positional or named argument.

If enabling sequence processing and per-frame named (CPU) matrix argument only, the PR would have 2 lines of code.
However, matrix can be specified as a positional argument that can reside in GPU memory. This required extending the SequenceOperator functionality with broadcasting of per-sample positional arguments. However, for now, we still require GPU batches to be contiguous, which further required the SequenceOperator to actually allocate some tensorlists and perform copies. Further, the workspace is not cleared between iterations now.

The utility for testing sequence processing ops with per-frame args required some changes to specify positional arguments and the intended device:
https://github.com/NVIDIA/DALI/pull/3901

To mark positional GPU argument as per-frame, per-frame operator was extended to work with GPU backend here:
https://github.com/NVIDIA/DALI/pull/3900

The order property of TensorVector is not reflected properly by move assignment/constructor for TensorVector:
https://github.com/NVIDIA/DALI/pull/3899

### Tests
- [ ] Existing tests apply
- [x] New tests added
  - [x] Python tests
  - [x] GTests
  - [ ] Benchmark
  - [ ] Other
- [ ] N/A

### Documentation
- [ ] Documentation updated
  - [ ] Docstring
  - [ ] Doxygen
  - [ ] RST
  - [ ] Jupyter
  - [ ] Other
- [ ] N/A

#### Requirements
- [ ] N/A

**REQ IDs**: WAFFINE.09

**JIRA TASK**: DALI-2506

### DIFF
--- a/dali/pipeline/operator/sequence_operator.h
+++ b/dali/pipeline/operator/sequence_operator.h
@@ -98,7 +98,7 @@ class SampleBroadcasting<GPUBackend> : public SampleBroadcasting<CPUBackend> {
  * resize the outputs.
  */
 template <typename Backend>
-class SequenceOperator : public Operator<Backend>, public SampleBroadcasting<Backend> {
+class SequenceOperator : public Operator<Backend>, protected SampleBroadcasting<Backend> {
  public:
   inline explicit SequenceOperator(const OpSpec &spec) : Operator<Backend>{spec} {}
 
--- a/dali/pipeline/operator/sequence_operator.h
+++ b/dali/pipeline/operator/sequence_operator.h
@@ -687,8 +687,8 @@ class SequenceOperator : public Operator<Backend>, public SampleBroadcasting<Bac
   void UnfoldOuterDims(const TensorVector<DataBackend> &batch,
                        TensorVector<DataBackend> &expanded_batch, const ExpandDesc &expand_desc) {
     auto ndims_to_unfold = expand_desc.NumDimsToExpand();
-    assert(expand_desc.NumSamples() == data.num_samples());
-    assert(batch.shape.first(ndims_to_unfold).num_elements() == expand_desc.NumExpanded());
+    assert(expand_desc.NumSamples() == batch.num_samples());
+    assert(batch.shape().first(ndims_to_unfold).num_elements() == expand_desc.NumExpanded());
     sequence_utils::unfold_outer_dims(batch, expanded_batch, expand_desc.NumDimsToExpand(),
                                       expand_desc.NumExpanded());
   }
--- a/dali/pipeline/operator/sequence_operator.h
+++ b/dali/pipeline/operator/sequence_operator.h
@@ -21,6 +21,7 @@
 #include <type_traits>
 #include <utility>
 #include <vector>
+#include "dali/kernels/common/scatter_gather.h"
 #include "dali/pipeline/data/tensor.h"
 #include "dali/pipeline/data/views.h"
 #include "dali/pipeline/operator/argument.h"
@@ -589,14 +590,45 @@ class SequenceOperator : public Operator<Backend> {
     return tv_builder.take();
   }
 
-  TensorList<CPUBackend> BroadcastSamples(const TensorList<CPUBackend> &arg_input,
-                                          const ExpandDesc &expand_desc) {
-    DALI_FAIL("Unsupported input container");
-  }
-
   TensorList<GPUBackend> BroadcastSamples(const TensorList<GPUBackend> &arg_input,
                                           const ExpandDesc &expand_desc) {
-    DALI_FAIL("Supported input container");
+    assert(expand_desc.NumSamples() == arg_input.num_samples());
+    TensorList<GPUBackend> broadcast(expand_desc.NumExpanded());
+    const auto &stream = expanded_.stream();
+    broadcast.set_order(AccessOrder(stream));
+    const auto &shape = arg_input.shape();
+    auto broadcast_shape = sequence_utils::broadcast_samples(shape, expand_desc);
+    broadcast.Resize(broadcast_shape, arg_input.type());
+    kernels::ScatterGatherGPU scatter_gather;
+    auto type_size = arg_input.type_info().size();
+    for (int sample_idx = 0, elem_idx = 0; sample_idx < expand_desc.NumSamples(); sample_idx++) {
+      auto sample_size = type_size * volume(shape[sample_idx]);
+      for (int i = 0; i < expand_desc.NumExpanded(sample_idx); i++) {
+        scatter_gather.AddCopy(broadcast.raw_mutable_tensor(elem_idx++),
+                               arg_input.raw_tensor(sample_idx), sample_size);
+      }
+    }
+    scatter_gather.Run(stream, true);
+    return broadcast;
+  }
+
+  TensorList<CPUBackend> BroadcastSamples(const TensorList<CPUBackend> &arg_input,
+                                          const ExpandDesc &expand_desc) {
+    assert(expand_desc.NumSamples() == arg_input.num_samples());
+    TensorList<CPUBackend> broadcast(expand_desc.NumExpanded());
+    broadcast.set_order(AccessOrder::host());
+    const auto &shape = arg_input.shape();
+    auto broadcast_shape = sequence_utils::broadcast_samples(shape, expand_desc);
+    broadcast.Resize(broadcast_shape, arg_input.type());
+    auto type_size = arg_input.type_info().size();
+    for (int sample_idx = 0, elem_idx = 0; sample_idx < expand_desc.NumSamples(); sample_idx++) {
+      auto sample_size = type_size * volume(shape[sample_idx]);
+      for (int i = 0; i < expand_desc.NumExpanded(sample_idx); i++) {
+        std::memcpy(broadcast.raw_mutable_tensor(elem_idx++), arg_input.raw_tensor(sample_idx),
+                    sample_size);
+      }
+    }
+    return broadcast;
   }
 
   template <typename DataBackend>
--- a/dali/pipeline/operator/sequence_operator.h
+++ b/dali/pipeline/operator/sequence_operator.h
@@ -38,11 +38,6 @@ inline bool is_per_frame(const TensorVector<CPUBackend> &arg_input) {
   return layout.size() > 0 && layout[0] == 'F';
 }
 
-template <typename Type>
-std::shared_ptr<Type> empty_like(const Type &batch) {
-  return std::make_shared<Type>();
-}
-
 template <typename BatchType>
 struct BatchBackend;
 
@@ -58,8 +53,8 @@ using batch_backend_t =
 }  // namespace detail
 
 
-// broadcasting on gpu requires some extra state,
-// so separate it from the sequence operator
+// broadcasting on gpu requires some extra state, so separate broadcasting from the
+// sequence operator and derive from appropriate specialization
 template <typename Backend>
 class SampleBroadcasting;
 
@@ -71,7 +66,8 @@ class SampleBroadcasting<CPUBackend> {
                         TensorVector<DataBackend> &expanded_batch, const ExpandDesc &expand_desc,
                         const ArgumentWorkspace &expanded) {
     (void)expanded;
-    sequence_utils::broadcast_samples(batch, expanded_batch, expand_desc);
+    sequence_utils::broadcast_samples(batch, expanded_batch, expand_desc.NumExpanded(),
+                                      expand_desc.DimsToExpand());
   }
 };
 
@@ -82,14 +78,16 @@ class SampleBroadcasting<GPUBackend> : public SampleBroadcasting<CPUBackend> {
 
   void BroadcastSamples(const TensorList<GPUBackend> &batch, TensorList<GPUBackend> &expanded_batch,
                         const ExpandDesc &expand_desc, const DeviceWorkspace &expanded) {
-    sequence_utils::broadcast_samples(batch, expanded_batch, expand_desc, scatter_gather_,
+    sequence_utils::broadcast_samples(batch, expanded_batch, expand_desc.NumExpanded(),
+                                      expand_desc.DimsToExpand(), scatter_gather_,
                                       expanded.stream());
   }
 
   void BroadcastSamples(const TensorList<CPUBackend> &batch, TensorList<CPUBackend> &expanded_batch,
                         const ExpandDesc &expand_desc, const DeviceWorkspace &expanded) {
     (void)expanded;
-    sequence_utils::broadcast_samples(batch, expanded_batch, expand_desc);
+    sequence_utils::broadcast_samples(batch, expanded_batch, expand_desc.NumExpanded(),
+                                      expand_desc.DimsToExpand());
   }
 
   kernels::ScatterGatherGPU scatter_gather_;
@@ -137,7 +135,12 @@ class SequenceOperator : public Operator<Backend>, public SampleBroadcasting<Bac
       DALI_ENFORCE(IsExpandable(),
                    "Operator requested to expand the sequence-like inputs, but no expandable input "
                    "was found");
-      SetupExpandedWorkspace(ws);
+      if (!is_expanded_input_set_up_) {
+        SetupExpandedWorkspace(expanded_, ws);
+        SetupExpandedInputs(ws);
+        SetupExpandedArguments(ws);
+        is_expanded_input_set_up_ = true;
+      }
       ExpandInputs(ws);
       ExpandArguments(ws);
       is_inferred = Operator<Backend>::Setup(output_desc, expanded_);
@@ -151,6 +154,10 @@ class SequenceOperator : public Operator<Backend>, public SampleBroadcasting<Bac
     if (!IsExpanding()) {
       Operator<Backend>::Run(ws);
     } else {
+      if (!is_expanded_output_set_up_) {
+        SetupExpandedOutputs(ws);
+        is_expanded_output_set_up_ = true;
+      }
       ExpandOutputs(ws);
       Operator<Backend>::Run(expanded_);
       PostprocessOutputs(ws);
@@ -234,6 +241,12 @@ class SequenceOperator : public Operator<Backend>, public SampleBroadcasting<Bac
     return GetReferenceInputIdx();
   }
 
+  virtual void SetupExpandedArgument(const workspace_t<Backend> &ws, const std::string &arg_name,
+                                     const TensorVector<CPUBackend> &arg_input) {
+    expanded_.AddArgumentInput(
+        arg_name, sequence_utils::expanded_like(arg_input, IsPerFrameArg(arg_name, arg_input)));
+  }
+
   /**
    * @brief Expands the tensor arguments from ``ws`` and adds them to expanded workspace.
    * Arguments marked as supporting per-frame tensors and with leading `F` in tensor layout are
@@ -253,6 +266,19 @@ class SequenceOperator : public Operator<Backend>, public SampleBroadcasting<Bac
     assert(IsExpandable());
     auto input_idx = GetArgExpandDescInputIdx(arg_name);
     ExpandArgumentLikeInput(arg_input, ExpandedArg(arg_name), arg_name, input_idx);
+  }
+
+  virtual void SetupExpandedInput(const workspace_t<Backend> &ws, int input_idx) {
+    ProcessInput(ws, input_idx, [&](const auto &batch) {
+      const auto &input_desc = GetInputExpandDesc(input_idx);
+      auto ndims_to_unfold = input_desc.NumDimsToExpand();
+      DALI_ENFORCE(batch.sample_dim() >= ndims_to_unfold,
+                   make_string("Cannot flatten the sequence-like batch for input ", input_idx,
+                               ". Samples cannot have less dimensions (got ", batch.sample_dim(),
+                               ") than the requested number of dimensions to unfold: ",
+                               ndims_to_unfold, "."));
+      expanded_.AddInput(sequence_utils::expanded_like(batch, ndims_to_unfold));
+    });
   }
 
   // TODO(ktokarski) Treat all but main input as arguments and make positional args expansion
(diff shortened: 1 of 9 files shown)


dali/pipeline/operator/sequence_operator.h
```diff
@@ -55,12 +110,13 @@ inline bool is_per_frame(const TensorVector<CPUBackend> &arg_input) {
  * resize the outputs.
  */
 template <typename Backend>
-class SequenceOperator : public Operator<Backend> {
+class SequenceOperator : public Operator<Backend>, public SampleBroadcasting<Backend> {
```
mzient, line 113: Wouldn't this suffice?
```suggestion
class SequenceOperator : public Operator<Backend>, protected SampleBroadcasting<Backend> {
```
You're re-exposing the function BroadcastSamples anyway.

dali/test/python/sequences_test_utils.py
```diff
@@ -15,6 +15,7 @@
 import os
 import random
 import numpy as np
+from typing import List, Union, Callable
```
klecki, line 18: Btw, I am not reviewing this file too much, as it's in the other PR.

klecki (commented):
Looks like there is copy-paste error in the ExpandInput when unfolding, otherwise mostly stylistic comments and nitpicks.

dali/pipeline/operator/sequence_operator.h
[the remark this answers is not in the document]
stiepan: I had in mind the empty base optimization "pattern" (as in CPU case we do not add any members), but it probably makes precisely no difference here and you have a good point. On the other hand, can this hurt us in this case? Maybe I'll just follow Michał's suggestion to make it protected inheritance.

dali/pipeline/operator/sequence_operator.h
```diff
@@ -195,11 +258,19 @@ class SequenceOperator : public Operator<Backend> {
                               const TensorVector<CPUBackend> &arg_input) {
     assert(IsExpandable());
     auto input_idx = GetArgExpandDescInputIdx(arg_name);
-    auto expanded_arg = ExpandArgumentLikeInput(arg_input, arg_name, input_idx);
-    auto expanded_handle = std::make_shared<TensorVector<CPUBackend>>(std::move(expanded_arg));
-    ExpandedAddArgument(arg_name, std::move(expanded_handle));
+    ExpandArgumentLikeInput(arg_input, ExpandedArg(arg_name), arg_name, input_idx);
```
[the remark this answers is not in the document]
stiepan: Yes, it is exactly that.

dali/pipeline/operator/sequence_operator.h
```diff
@@ -219,26 +290,38 @@ class SequenceOperator : public Operator<Backend> {
...
+      });
     } else {
       if (ref_input_idx != input_idx) {
         VerifyExpansionConsistency(ref_input_idx, ref_expand_desc, input_idx, input_desc);
       }
-      ExpandedAddProcessedInput(ws, input_idx,
-                                [&](const auto &input) { return UnfoldBatch(input, input_desc); });
+      ProcessInput(ws, input_idx, [&](const auto &input) {
+        auto &expanded_input = ExpandedInput<detail::batch_backend_t<decltype(input)>>(input_idx);
+        UnfoldBatch(input, expanded_input, ref_expand_desc);
```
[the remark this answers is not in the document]
stiepan: Spot-on

dali/pipeline/operator/sequence_shape.h
```diff
@@ -197,51 +198,61 @@ class UnfoldedSliceRange {
...
+template <typename Type>
+TensorLayout unfolded_sample_layout(const Type &batch, int ndims_to_unfold) {
+  assert(batch.sample_dim() >= ndims_to_unfold);
+  auto sample_dim = batch.sample_dim() - ndims_to_unfold;
+  const auto &initial_layout = batch.GetLayout();
+  return initial_layout.empty() ? "" : initial_layout.last(sample_dim);
+}
+
+template <typename Type>
+void setup_expanded_like(const Type &batch, Type &expanded_batch) {
```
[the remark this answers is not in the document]
stiepan: Okay, I even had it scribbled in my notes to change the order of arguments to match kernels/operators output/inputs, missed it.

dali/pipeline/operator/sequence_operator.h
```diff
@@ -36,8 +37,62 @@ inline bool is_per_frame(const TensorVector<CPUBackend> &arg_input) {
...
+struct BatchBackend;
+
+template <template <typename> class BatchContainer, typename Backend>
+struct BatchBackend<BatchContainer<Backend>> {
+  using Type = Backend;
+};
+
+template <typename BatchType>
+using batch_backend_t =
+    typename BatchBackend<std::remove_cv_t<std::remove_reference_t<BatchType>>>::Type;
```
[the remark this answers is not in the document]
stiepan: Does a particular suitable file come to your head?

dali/pipeline/operator/sequence_shape.h
```diff
@@ -266,19 +277,102 @@ inline TensorListShape<> unfold_outer_dims(const TensorListShape<> &shape, int n
...
+  assert(expand_extents.num_samples() == batch.num_samples());
+  const auto &shape = batch.shape();
+  auto broadcast_shape = broadcast_samples(shape, num_expanded_samples, expand_extents);
+  expanded_batch.SetLayout(batch.GetLayout());
+  expanded_batch.Resize(broadcast_shape, batch.type());
+  auto type_size = batch.type_info().size();
+  for (int sample_idx = 0, elem_idx = 0; sample_idx < batch.num_samples(); sample_idx++) {
+    auto sample_size = type_size * volume(shape[sample_idx]);
+    for (int i = 0; i < volume(expand_extents[sample_idx]); i++) {
+      std::memcpy(expanded_batch.raw_mutable_tensor(elem_idx++), batch.raw_tensor(sample_idx),
```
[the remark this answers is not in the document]
stiepan: I did, but the funny thing is that we don't have a thread pool when it is needed: we only need this coping broadcasting with device workspace which does not come with a thread pool.

dali/pipeline/operator/sequence_operator.h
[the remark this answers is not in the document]
klecki: I got composition suggested once or twice, but I'm not asking for a rewrite, just saying it's a bit artificial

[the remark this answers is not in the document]
mzient: Well, on the other hand, we could consider making the broadcasting functions `virtual`, which would let us customize it in the operators. Doing that with composition would be very artificial.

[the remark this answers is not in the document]
mzient: I'm fine either way, with inheritance having the benefit of less typing (and being already implemented).

dali/pipeline/operator/sequence_operator.h
```diff
@@ -406,6 +497,37 @@ class SequenceOperator : public Operator<Backend> {
...
+
+  void SetupExpandedArguments(const workspace_t<Backend> &ws) {
+    for (const auto &arg_input : ws) {
+      auto &shared_tvec = arg_input.second.tvec;
+      assert(shared_tvec);
+      SetupExpandedArgument(ws, arg_input.first, *shared_tvec);
+    }
+  }
+
+  void SetupExpandedWorkspace(const workspace_t<Backend> &ws) {
```
mzient, line 524: Since it's only run once, how about 
```suggestion
  void InitializeExpandedWorkspace(const workspace_t<Backend> &ws) {
```

dali/pipeline/operator/sequence_operator.h
```diff
@@ -607,6 +691,7 @@ class SequenceOperator : public Operator<Backend> {
  private:
   int expand_like_idx_ = -1;
   bool is_expanding_ = false;
+  bool is_expanded_set_up_ = false;
```
mzient, line 694: expanded what?
```suggestion
  bool is_expanded_ws_initialized_ = false;
```

dali/pipeline/operator/sequence_shape.h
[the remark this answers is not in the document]
mzient: Yup, memcpy order is preferred.

mzient, line 210: This isn't a generic function that can work with anything - it has very specific requirements for the target type and that should be somehow reflected in the template parameter name.
```suggestion
template <typename BufferObject>
void setup_expanded_like(const BufferObject &batch, BufferObject &expanded_batch) {
```

dali/pipeline/operator/sequence_shape.h
```diff
@@ -197,51 +198,61 @@ class UnfoldedSliceRange {
...
+}
+
+template <typename Type>
+void setup_expanded_like(const Type &batch, Type &expanded_batch) {
+  expanded_batch.set_pinned(batch.is_pinned());
+  expanded_batch.set_order(batch.order());
+}
+
+template <typename Type>
+std::shared_ptr<Type> expanded_like(const Type &batch) {
```
mzient, line 216:
```suggestion
template <typename BufferObject>
std::shared_ptr<BufferObject> expanded_like(const BufferObject &batch) {
```

dali/pipeline/operator/sequence_shape.h
```diff
@@ -197,51 +198,61 @@ class UnfoldedSliceRange {
...
-    tv_.set_type(type);
-    tv_.set_sample_dim(sample_dim);
-    tv_.SetLayout(layout);
-    tv_.set_pinned(is_pinned);
-    tv_.set_order(order);
-  }
+  TensorVectorBuilder(TensorVector<Backend> &tv) : tv_{tv} {}  // NOLINT
 
-  void push(const SliceView &view) {
+  void SetNext(const SliceView &view) {
```
mzient, line 230:
```suggestion
  void SetNext(const SliceView &view) {
    assert(size < tv_.num_samples());
```

dali/pipeline/operator/sequence_shape.h
```diff
@@ -197,51 +198,61 @@ class UnfoldedSliceRange {
...
   }
 
-  /**
-   * @brief Returns the created tensor vector: must be called after pusing exactly ``num_samples``.
-   * Calling ``take`` invalidates the ``TensorVectorBuilder`` instance.
-   */
-  TensorVector<Backend> take() {
-    assert(size == tv_.num_samples());
-    return std::move(tv_);
+  int Size() {
```
mzient, line 236:
```suggestion
  int NextFrameIdx() const {
```

dali/pipeline/operator/sequence_shape.h
```diff
@@ -197,51 +198,61 @@ class UnfoldedSliceRange {
...
+    return size;
   }
 
-  TensorVector<Backend> tv_;
+ private:
+  TensorVector<Backend> &tv_;
   int size = 0;
```
mzient, line 242:
```suggestion
  int next_ = 0;
```
?

dali/pipeline/workspace/workspace.h
```diff
@@ -72,6 +72,10 @@ class ArgumentWorkspace {
     return *it->second.tvec;
   }
 
+  TensorVector<CPUBackend>& UnsafeMutableArgumentInput(const std::string &arg_name) {
+    return const_cast<TensorVector<CPUBackend>&>(ArgumentInput(arg_name));
```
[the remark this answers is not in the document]
mzient: Yeah, I think we should rethink what constness means in workspaces. But it's beyond the scope of this PR and we already have a similar functions for regular inputs.

dali/pipeline/operator/sequence_operator.h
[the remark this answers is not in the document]
mzient: Actually, I have mixed feelings, because it can backfire terribly. Effectively, it's just a utility that extracts the type template argument from a single-type template. Just see:
```C++
    std::mutex m;
    std::lock_guard<std::mutex> g(m);
    static_assert(std::is_same<batch_backend_t<decltype(g)>, std::mutex>::value);
```
If we were to make it a general utility, we could make `Backend` a `typedef` inside classes templated against Backend (Buffer, TensorXxx, Operator) and then this trait could be defined as
```C++
template <typename X>
using backend_t = typename X::Backend;
```
or possibly
```C++
template <typename X>
using backend_t = typename std::remove_cv_t<std::remove_reference_t<X>>::Backend;
```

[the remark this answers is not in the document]
mzient: Having said all that, I'm fine with keeping it as is in this limited scope - perhaps this utility should even be moved inside the class, so it's not (ab)used elsewhere.

dali/pipeline/operator/sequence_operator.h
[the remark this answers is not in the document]
stiepan: changed it to protected inheritence

dali/pipeline/operator/sequence_operator.h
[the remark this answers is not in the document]
stiepan: Okay, so I moved it to the type_traits.h and added enable_if_t<is_batch_container<...>> guard, this way it can be reused and to abused.

dali/pipeline/operator/sequence_shape.h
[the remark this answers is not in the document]
stiepan: done

dali/pipeline/operator/sequence_operator.h
stiepan: done

dali/pipeline/operator/sequence_shape.h
stiepan: done

dali/pipeline/operator/sequence_shape.h
stiepan: done

dali/pipeline/operator/sequence_shape.h
stiepan: done

dali/pipeline/operator/sequence_shape.h
stiepan: done

dali/pipeline/operator/sequence_shape.h
```diff
@@ -266,19 +277,102 @@ inline TensorListShape<> unfold_outer_dims(const TensorListShape<> &shape, int n
...
+}
+
 template <typename Backend>
-TensorList<Backend> unfold_outer_dims(const TensorList<Backend> &data, int ndims_to_unfold) {
-  // TODO(ktokarski) TODO(klecki)
-  // Rework it when TensorList stops being contigious and supports "true sample" mode
-  const auto &shape = data.shape();
-  TensorList<Backend> tl;
-  tl.ShareData(data);
+void broadcast_samples(const TensorVector<Backend> &batch, TensorVector<Backend> &expanded_batch,
```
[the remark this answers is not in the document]
stiepan: As to the format: effectively it is as you say, more broadly the layout does not matter at this point: from batch (batch.num_samples() == expand_extents.num_samples()) we get expanded_batch.num_samples() == expanded_extents.num_elements() by repeating corresponding samples volume(expanded_extents[i]) times.

You are right that we could just compute the `num_expanded_samples`, it is just that it is already computed in the ExpandDesc, so I pass it here as an argument. Later on there is even relevant assert in broadcast shapes (`assert(num_expanded_samples == expand_extents.num_elements());`)

dali/pipeline/operator/sequence_shape.h
```diff
@@ -266,19 +277,102 @@ inline TensorListShape<> unfold_outer_dims(const TensorListShape<> &shape, int n
   }
 }
 
+inline TensorListShape<> broadcast_samples(const TensorListShape<> &shape, int num_expanded_samples,
```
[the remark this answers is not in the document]
stiepan: done

dali/test/python/sequences_test_utils.py
```diff
@@ -28,13 +29,55 @@
...
+    @property
+    def is_positional_arg(self):
+        return isinstance(self.name, int)
+
+    def __repr__(self):
+        return "ArgDesc{}".format((self.name, self.is_per_frame, self.dest_device),)
+
+
+class ArgCb:
+    def __init__(self, name: Union[str, int], cb: Callable[[SampleDesc], np.ndarray], is_per_frame: bool, dest_device: str = "cpu"):
```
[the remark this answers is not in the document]
stiepan: done

klecki (approved):

dali/pipeline/operator/sequence_shape.h
```diff
@@ -255,6 +267,103 @@ UnfoldedSliceRange unfolded_slice_range(const TensorVector<Backend> &data, int s
...
+
+inline void broadcast_samples(TensorList<GPUBackend> &expanded_batch,
+                              const TensorList<GPUBackend> &batch, int num_expanded_samples,
+                              const TensorListShape<> &expand_extents,
+                              kernels::ScatterGatherGPU &sg, cudaStream_t stream) {
+  assert(expand_extents.num_samples() == batch.num_samples());
+  const auto &shape = batch.shape();
+  auto broadcast_shape = broadcast_sample_shapes(shape, num_expanded_samples, expand_extents);
+  expanded_batch.SetLayout(batch.GetLayout());
+  expanded_batch.Resize(broadcast_shape, batch.type());
```
mzient, line 324:
```suggestion
  expanded_batch.set_order(stream);
  expanded_batch.Resize(broadcast_shape, batch.type());
```
Now that we deal with GPU stuff...

dali/pipeline/operator/sequence_shape.h
```diff
@@ -255,6 +267,103 @@ UnfoldedSliceRange unfolded_slice_range(const TensorVector<Backend> &data, int s
...
+  }
+  return broadcast;
+}
+
+template <typename Backend>
+void broadcast_samples(TensorVector<Backend> &expanded_batch, const TensorVector<Backend> &batch,
+                       int num_expanded_samples, const TensorListShape<> &expand_extents) {
+  assert(num_expanded_samples == expand_extents.num_elements());
+  assert(expand_extents.num_samples() == batch.num_samples());
+  auto expanded_builder = tv_builder_like(expanded_batch, batch, num_expanded_samples);
```
mzient, line 301:
```suggestion
  if (AccessOrder order = get_order())  // we use views, so they will have the same ordering as the input
    expanded_batch.set_order(order);
  auto expanded_builder = tv_builder_like(expanded_batch, batch, num_expanded_samples);
```

dali/pipeline/operator/sequence_shape_test.cc
```diff
@@ -12,17 +12,32 @@
...
+
+template <template <typename> class BatchContainer, typename Backend>
+struct BatchBackend<BatchContainer<Backend>> {
+  using Type = Backend;
+};
+
+template <typename BatchType>
+using batch_backend_t =
+    typename BatchBackend<std::remove_cv_t<std::remove_reference_t<BatchType>>>::Type;
+
```
mzient, line 40: This is a duplicate - please remove.

dali/pipeline/operator/sequence_shape_test.cc
```diff
@@ -135,155 +150,289 @@ void check_batch_props(ContainerT &batch, ContainerT &unfolded_batch, int num_sl
...
 
-TYPED_TEST(SequenceShapeUnfoldTVTest, Unfold2Extents) {
-  auto batch = this->CreateTestBatch(DALI_UINT8);
-  this->TestUnfolding(batch, 2);
+TYPED_TEST(SequenceShapeUnfoldTest, Unfold2Extents3Iters) {
+  auto [batch, expanded_batch] = this->CreateTestBatch(DALI_UINT8, false, "XYZ");
+  this->TestUnfolding(*expanded_batch, batch, 2);
+  batch.Resize({{2, 2, 6}, {13, 4, 11}}, DALI_FLOAT);
+  batch.SetLayout("XYZ");
+  this->TestUnfolding(*expanded_batch, batch, 2);
```
mzient, line 234:
```suggestion
  this->TestUnfolding(*expanded_batch, batch, 1);
  this->TestUnfolding(*expanded_batch, batch, 2);
```
Can you also check unfolding of just one dim? I think these tests are cheap and unfolding 1 dim is where we want the best possible coverage.

dali/pipeline/operator/sequence_shape_test.cc
```diff
@@ -135,155 +150,289 @@ void check_batch_props(ContainerT &batch, ContainerT &unfolded_batch, int num_sl
...
+  batch.Resize({{2, 2, 6}, {13, 4, 11}, {13, 4, 11}, {13, 4, 11}}, DALI_FLOAT);
+  batch.SetLayout("XYZ");
+  this->TestUnfolding(*expanded_batch, batch, 2);
```
mzient, line 237: Likewise.

dali/pipeline/operator/sequence_shape_test.cc
```diff
@@ -135,155 +150,289 @@ void check_batch_props(ContainerT &batch, ContainerT &unfolded_batch, int num_sl
...
+  auto [batch, expanded_batch] = this->template CreateTestBatch<uint8_t>(DALI_UINT8, true);
+  this->template TestBroadcasting<uint8_t>(*expanded_batch, batch, {{1, 1}, {7, 0}, {12, 4}});
 }
 
-TYPED_TEST(SequenceShapeUnfoldTLTest, Unfold3Extents) {
-  auto batch = this->CreateTestBatch(DALI_UINT8);
-  this->TestUnfolding(batch, 3);
+TYPED_TEST(SequenceShapeBroadcastTest, Broadcast2Extents3Iters) {
+  auto [batch, expanded_batch] = this->template CreateTestBatch<float>(DALI_FLOAT, true, "XYZ");
+  this->template TestBroadcasting<float>(*expanded_batch, batch, {{1, 1}, {7, 0}, {12, 4}});
```
mzient, line 427: Please add a test where you broadcast scalar values.

dali/pipeline/operator/sequence_shape_test.cc
```diff
@@ -135,155 +150,289 @@ void check_batch_props(ContainerT &batch, ContainerT &unfolded_batch, int num_sl
...
+  batch.Resize({{2, 2, 6}, {13, 4, 11}}, DALI_FLOAT);
+  batch.SetLayout("XYZ");
+  this->template FillBatch<float>(batch, 101);
+  this->template TestBroadcasting<float>(*expanded_batch, batch, {{5}, {1}});
+  batch.Resize({{2, 2, 6}, {13, 4, 11}, {13, 4, 11}, {13, 4, 11}}, DALI_FLOAT);
+  batch.SetLayout("XYZ");
+  this->template FillBatch<float>(batch, 7);
+  this->template TestBroadcasting<float>(*expanded_batch, batch, {{}, {}, {}, {}});
```
mzient, line 435: Please test broadcasting of scalars.

stiepan: done

dali/pipeline/operator/sequence_shape_test.cc
stiepan: There is a separate test case that just unfolds 1 dim (and the same for all three). I added two cases with multiple iters that test 1, 2 and 3 unfolds in a single test case between resizes.

dali/pipeline/operator/sequence_shape_test.cc
stiepan: Good catch!
done

dali/pipeline/operator/sequence_shape.h
stiepan: done

stiepan:
!build

stiepan:
!build

mzient (approved):

PR #3886, opened 2022-05-10
Fix coverity issues. Do not re-throw worker thread error in the destructor.

stiepan:

## Description:

Adds extra parameter to WaitForWork method to prevent it from re-throwing errors in the destructor.

Fixes possible deadlock on Shutdown if the worker thread quit with the execption

### Tests
- [ ] Existing tests apply
- [x] New tests added
  - [ ] Python tests
  - [x] GTests
  - [ ] Benchmark
  - [ ] Other
- [ ] N/A

### Documentation
- [ ] Documentation updated
  - [ ] Docstring
  - [ ] Doxygen
  - [ ] RST
  - [ ] Jupyter
  - [ ] Other
- [x] N/A

#### Requirements
- [x] N/A

**REQ IDs**: N/A

**JIRA TASK**: DALI-2736

### DIFF
--- a/dali/pipeline/util/worker_thread.h
+++ b/dali/pipeline/util/worker_thread.h
@@ -78,7 +78,7 @@ class WorkerThread {
   }
 
   inline ~WorkerThread() {
-    Shutdown(false);
+    Shutdown();
 #if NVML_ENABLED
     nvml::Shutdown();
 #endif
@@ -89,10 +89,10 @@ class WorkerThread {
    * When the destructor is called other things that work() is using may have been gone long
    * before causing a hang. Now when Shutdown is called we are sure that all things around still exist.
    */
-  inline void Shutdown(bool rethrow_worker_errors = true) {
+  inline void Shutdown() {
     // Wait for work to find errors
     if (running_) {
-      WaitForWork(rethrow_worker_errors);
+      WaitForWork(false);
 
       // Mark the thread as not running
       {
--- a/dali/pipeline/util/worker_thread.h
+++ b/dali/pipeline/util/worker_thread.h
@@ -121,7 +121,7 @@ class WorkerThread {
 
   inline void WaitForWork(bool rethrow_worker_errors = true) {
     std::unique_lock<std::mutex> lock(mutex_);
-    while (!work_complete_) {
+    while (!work_complete_ && running_) {
       completed_.wait(lock);
     }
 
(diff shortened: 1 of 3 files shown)


JanuszL (approved):

dali/pipeline/util/worker_thread.h
```diff
@@ -119,14 +119,14 @@ class WorkerThread {
...
-  inline void WaitForWork() {
+  inline void WaitForWork(bool rethrow_worker_errors = true) {
     std::unique_lock<std::mutex> lock(mutex_);
     while (!work_complete_) {
       completed_.wait(lock);
     }
 
     // Check for errors
-    if (!errors_.empty()) {
+    if (rethrow_worker_errors && !errors_.empty()) {
```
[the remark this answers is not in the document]
mzient: I think that was the idea, wasn't it? Ideally, we would break if there was something other than `runtime_error`.

dali/pipeline/util/worker_thread.h
```diff
@@ -89,10 +89,10 @@ class WorkerThread {
    * When the destructor is called other things that work() is using may have been gone long
    * before causing a hang. Now when Shutdown is called we are sure that all things around still exist.
    */
-  inline void Shutdown(void) {
+  inline void Shutdown(bool rethrow_worker_errors = true) {
```
mzient, line 92: Do we have a legitimate case for rethrowing in shutdown? Isn't it ultimately always called from some kind of destructor?

dali/pipeline/util/worker_thread.h
```diff
@@ -119,14 +119,14 @@ class WorkerThread {
     cv_.notify_one();
   }
 
-  inline void WaitForWork() {
+  inline void WaitForWork(bool rethrow_worker_errors = true) {
     std::unique_lock<std::mutex> lock(mutex_);
     while (!work_complete_) {
```
stiepan, line 124: Btw. it seems that calling WaitForWork while worker actually does something and then it raises the error, may deadlock as the worker thread does not set&notify completed_.

dali/pipeline/util/worker_thread.h
stiepan: Yeah, one way or another it is always called from destructors.

stiepan: done

mzient (approved):

JanuszL (approved):

PR #3915, opened 2022-05-18
Remove intermediate Tensor and use DynamicScratchpad for op tile descirptors.

mzient:
* Remove unused includes

## Category:

## Description:
  * Removed intermediate Tensor and use DynamicScratchpad for arithmetic op tile descirptors
  * Made Scratchpad::ToPinned/ToHost/ToManaged work with non-trivially copyable objects
  * Removed unused includes

Arithmetic operators, like many other, use auxiliary structures passed to the GPU. These structures were copied from plain host memory (non-pinned) to GPU, occasionally causing the CUDA driver to wait. This PR removes an intermediate tensor and instead uses DynamicScratchpad to allocate temporary pinned storage and then copies from that storage to device memory.

Also, since Arithmetc ops use rather complex types and pass them to the GPU, a fix was needed in Scratchpad to allow copying such objects (an alternative would be to use memcpy for host).

Additionally, there were some unused includes, which could be safely removed, reducing internal dependencies and possibly saving some rebuilds.

Arithmetic operator.

### Tests
- [X] Existing tests apply
- [ ] New tests added
  - [ ] Python tests
  - [ ] GTests
  - [ ] Benchmark
  - [ ] Other
- [ ] N/A

### Documentation
- [ ] Documentation updated
  - [ ] Docstring
  - [ ] Doxygen
  - [ ] RST
  - [ ] Jupyter
  - [ ] Other
- [ ] N/A

#### Requirements
- [X] N/A

**REQ IDs**: N/A

**JIRA TASK**: 2793

### DIFF
--- a/dali/kernels/context.h
+++ b/dali/kernels/context.h
@@ -146,7 +146,9 @@ class Scratchpad {
   if_iterable<Collection, T*>
   ToHost(const Collection &c) {
     T *ptr = AllocateHost<T>(dali::size(c));
-    std::copy(begin(c), end(c), ptr);
+    T *it = ptr;
+    for (const auto &x : c)
+      new(it++)T(x);
     return ptr;
   }
 
@@ -154,7 +156,9 @@ class Scratchpad {
   if_iterable<Collection, T*>
   ToPinned(const Collection &c) {
     T *ptr = AllocatePinned<T>(dali::size(c));
-    std::copy(begin(c), end(c), ptr);
+    T *it = ptr;
+    for (const auto &x : c)
+      new(it++)T(x);
     return ptr;
   }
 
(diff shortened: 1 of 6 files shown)


dali/kernels/context.h
```diff
@@ -146,23 +146,29 @@ class Scratchpad {
   if_iterable<Collection, T*>
   ToHost(const Collection &c) {
     T *ptr = AllocateHost<T>(dali::size(c));
-    std::copy(begin(c), end(c), ptr);
+    T *it = ptr;
+    for (const auto &x : c)
+      new(it++)T(x);
```
mzient, line 151: Bug fix: the previous code didn't work with objects that had non-trivial copy assignment. It's still a big of an abuse to use such objects with scratchpad at all, but here we go :\

JanuszL (approved):

jantonguirao (approved):

PR #3921, opened 2022-05-20
Add VFR detection

awolant:
## Category:

New Feature

## Description:

It adds detection of whether decoded video file is variable frame rate or not and ability to access this information. It is done based on the video file index. we go through the index and check time differences between the frames. If there are any differences decoded file is VFR. 

This check will be optimized in the future. Goal for this PR was to add a check that works in every scenario.

`FramesDecoder` and its tests.

### Tests
- [ ] Existing tests apply
- [x] New tests added
  - [ ] Python tests
  - [x] GTests
  - [ ] Benchmark
  - [ ] Other
- [ ] N/A

### Documentation
- [ ] Documentation updated
  - [ ] Docstring
  - [ ] Doxygen
  - [ ] RST
  - [ ] Jupyter
  - [ ] Other
- [ ] N/A

#### Requirements
- [ ] N/A

**REQ IDs**: N/A

**JIRA TASK**: DALI-2798


awolant:
!build

dali/operators/reader/loader/video/frames_decoder_test.cc
```diff
@@ -36,6 +36,7 @@ class FramesDecoderTestBase : public VideoTestBase {
     ASSERT_EQ(decoder.Width(), ground_truth.Width());
     ASSERT_EQ(decoder.Channels(), ground_truth.NumChannels());
     ASSERT_EQ(decoder.NumFrames(), ground_truth.NumFrames());
+    ASSERT_EQ(decoder.IsVfr(), ground_truth.IsVfr());
```
JanuszL, line 39: Can you add a test for a case with non-vfr video file?

dali/operators/reader/loader/video/frames_decoder.cc
```diff
@@ -134,6 +135,23 @@ void FramesDecoder::BuildIndex() {
...
 }
 
+void FramesDecoder::DetectVfr() {
+  if (NumFrames() < 3) {
+    is_vfr_ = false;
+    return;
+  }
+
+  int pts_step = index_[1].pts - index_[0].pts;
+  for (int frame_id = 2; frame_id < NumFrames(); ++frame_id) {
```
JanuszL, line 145: How fast is that? Can't we do this when the index is created?

dali/operators/reader/loader/video/frames_decoder.cc
```diff
@@ -134,6 +135,23 @@ void FramesDecoder::BuildIndex() {
...
+    if ((index_[frame_id].pts - index_[frame_id - 1].pts) != pts_step) {
```
JanuszL, line 146: Just to clarify. PTS is an integer and we are not affected by floating point operation precision.

dali/operators/reader/loader/video/frames_decoder.cc
awolant: We could mix it with index creation. For now, I decided to do it like this for readability (look PR teamplate). I added task to optimize it further, when we have proper benchmark. Part of this optimization will be to get this info without any decoding when possible.

dali/operators/reader/loader/video/frames_decoder.cc
awolant: Yes

dali/operators/reader/loader/video/frames_decoder_test.cc
awolant: This test covers both VFR and CFR files.

JanuszL: Ok. Missed that.

dali/operators/reader/loader/video/frames_decoder.cc
JanuszL: Understood.

mzient (approved):

PR #3924, opened 2022-05-21
Fix scalar handling in GPU cast

mzient:

## Category:

## Description:
Scalars were not properly handled in operator cast - there was an attempt to collapse the shape, which failed catastrophically (assertion in Debug build and access violation in Release) if the input shape is 0D, in which case it needs to be artificially expanded.
This PR adds special handling for that case.

Operator cast.

N/A

**TODO** (separate PR): Add proper tests for operator `cast`.

### Tests
- [ ] Existing tests apply
- [ ] New tests added
  - [ ] Python tests
  - [ ] GTests
  - [ ] Benchmark
  - [ ] Other
- [X] N/A

### Documentation
- [ ] Documentation updated
  - [ ] Docstring
  - [ ] Doxygen
  - [ ] RST
  - [ ] Jupyter
  - [ ] Other
- [X] N/A

#### Requirements
- [X] N/A

**REQ IDs**: N/A

**JIRA TASK**: N/A

### DIFF
--- a/dali/operators/generic/cast.cu
+++ b/dali/operators/generic/cast.cu
@@ -58,8 +58,13 @@ bool CastGPU::SetupImpl(std::vector<OutputDesc> &output_desc, const DeviceWorksp
 void CastGPU::PrepareBlocks(const DeviceWorkspace &ws) {
   const auto &input = ws.Input<GPUBackend>(0);
   const auto &input_shape = input.shape();
-  std::array<std::pair<int, int>, 1> collapse_groups = {{{0, input_shape.sample_dim()}}};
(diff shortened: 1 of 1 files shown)


mzient:
!build

stiepan (approved):

stiepan:
I get that adding thorough testing may be out of scope, but how about including here a single test case that would explode without the fix and passes otherwise?

mzient:
> I get that adding thorough testing may be out of scope, but how about including here a single test case that would explode without the fix and passes otherwise?

I don't think it's worth wasting a perfectly good CI run ;)

jantonguirao (approved):

PR #3928, opened 2022-05-23
Add more NVTX ranges to the executor

JanuszL:
- adds separate NVTX ranges to the setup, output allocation and run operator
  stages in the executor
image

## Category:

## Description:
- adds separate NVTX ranges to the setup and run operator
  stages in the executor

- executor

- NA

### Tests
- [ ] Existing tests apply
- [ ] New tests added
  - [ ] Python tests
  - [ ] GTests
  - [ ] Benchmark
  - [ ] Other
- [x] N/A

### Documentation
- [ ] Documentation updated
  - [ ] Docstring
  - [ ] Doxygen
  - [ ] RST
  - [ ] Jupyter
  - [ ] Other
- [x] N/A

#### Requirements
- [x] N/A

**REQ IDs**: N/A

**JIRA TASK**: N/A

### DIFF
--- a/dali/pipeline/executor/executor.cc
+++ b/dali/pipeline/executor/executor.cc
@@ -348,9 +348,14 @@ void Executor<WorkspacePolicy, QueuePolicy>::RunHelper(OpNode &op_node, Workspac
     if (had_empty_layout) empty_layout_in_idxs.push_back(i);
   }
 
+  bool ret = false;
   {
-    DomainTimeRange tr("Setup");
-    if (op.Setup(output_desc, ws)) {
+    DomainTimeRange tr("[DALI][Executor] Setup");
+    ret = op.Setup(output_desc, ws);
+  }
+  {
+    DomainTimeRange tr("[DALI][Executor] Allocate outputs");
+    if (ret) {
       DALI_ENFORCE(
           static_cast<size_t>(ws.NumOutput()) == output_desc.size(),
           "Operator::Setup returned shape and type information for mismatched number of outputs");
@@ -374,7 +379,7 @@ void Executor<WorkspacePolicy, QueuePolicy>::RunHelper(OpNode &op_node, Workspac
     }
   }
   {
-    DomainTimeRange tr("Run");
+    DomainTimeRange tr("[DALI][Executor] Run");
     op.Run(ws);
   }
 
--- a/dali/pipeline/executor/executor.cc
+++ b/dali/pipeline/executor/executor.cc
@@ -348,14 +348,14 @@ void Executor<WorkspacePolicy, QueuePolicy>::RunHelper(OpNode &op_node, Workspac
     if (had_empty_layout) empty_layout_in_idxs.push_back(i);
   }
 
-  bool ret = false;
+  bool should_allocate = false;
   {
     DomainTimeRange tr("[DALI][Executor] Setup");
-    ret = op.Setup(output_desc, ws);
+    should_allocate = op.Setup(output_desc, ws);
   }
   {
     DomainTimeRange tr("[DALI][Executor] Allocate outputs");
-    if (ret) {
+    if (should_allocate) {
       DALI_ENFORCE(
           static_cast<size_t>(ws.NumOutput()) == output_desc.size(),
           "Operator::Setup returned shape and type information for mismatched number of outputs");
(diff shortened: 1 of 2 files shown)


JanuszL:
!build

JanuszL:
!build

JanuszL:
!build

dali/pipeline/executor/executor.cc
```diff
@@ -348,30 +348,35 @@ void Executor<WorkspacePolicy, QueuePolicy>::RunHelper(OpNode &op_node, Workspac
...
-                 "and type information for Operator outputs. In that case CanInferOutputs should "
-                 "always return true.");
-    for (int i = 0; i < ws.NumOutput(); i++) {
-      auto &desc = output_desc[i];
-      if (ws.template OutputIsType<CPUBackend>(i)) {
-        ws.template Output<CPUBackend>(i).Resize(desc.shape, desc.type);
-      } else {
-        ws.template Output<GPUBackend>(i).Resize(desc.shape, desc.type);
+  {
+    DomainTimeRange tr("Setup");
```
mzient, line 352: This will also measure the allocation of the outputs - we could argue it's a part of setup, but is it really? Just asking to clarify.
Also - will this name be somehow connected to the specific operator which is being set up?

JanuszL: > This will also measure the allocation of the outputs - we could argue it's a part of setup, but is it really? Just asking to clarify.

No preferences. I can extract and measure only setup itself. The gap in the profile between setup and run would be the allocation.

> Also - will this name be somehow connected to the specific operator which is being set up?

This is how it looks like:
image
It is inside the Operator range. But I can add `[DALI]` to the name as we do in other ranges.

JanuszL: After changes 
image

dali/pipeline/executor/executor.cc
```diff
@@ -348,30 +348,40 @@ void Executor<WorkspacePolicy, QueuePolicy>::RunHelper(OpNode &op_node, Workspac
...
-        ws.template Output<CPUBackend>(i).Resize(desc.shape, desc.type);
-      } else {
-        ws.template Output<GPUBackend>(i).Resize(desc.shape, desc.type);
+  bool ret = false;
+  {
+    DomainTimeRange tr("[DALI][Executor] Setup");
+    ret = op.Setup(output_desc, ws);
+  }
+  {
+    DomainTimeRange tr("[DALI][Executor] Allocate outputs");
```
klecki, line 357: I think this might be a bit misleading. The ops that don't use setup and don't allocate at all would probably report a very short time here, which works ok, but so would something that allocates, but it doesn't use Setup.
The second case is the misleading one. It might raise question why some of the allocations are that fast and if we can do it in other places. But maybe I am overthinking this.

dali/pipeline/executor/executor.cc
```diff
@@ -348,30 +348,40 @@ void Executor<WorkspacePolicy, QueuePolicy>::RunHelper(OpNode &op_node, Workspac
...
-                 "Operator::Setup returned true indicating that it successfully calculated shape "
-                 "and type information for Operator outputs. In that case CanInferOutputs should "
-                 "always return true.");
-    for (int i = 0; i < ws.NumOutput(); i++) {
-      auto &desc = output_desc[i];
-      if (ws.template OutputIsType<CPUBackend>(i)) {
-        ws.template Output<CPUBackend>(i).Resize(desc.shape, desc.type);
-      } else {
-        ws.template Output<GPUBackend>(i).Resize(desc.shape, desc.type);
+  bool ret = false;
```
klecki, line 351: If I might nitpick a bit, maybe something like 
```suggestion
  bool should_allocate = false;
```
or something with setup in name?

klecki (approved):

dali/pipeline/executor/executor.cc
JanuszL: I think the main point is to highlight places where the setup and allocation take too much time. If it is fast (because is just fast or no-op) it means we don't have to look into these parts of the operator execution.

dali/pipeline/executor/executor.cc
JanuszL: Fixed

dali/pipeline/executor/executor.cc
klecki: What I mean by misleading is the operators that don't use setup just allocate within run, so this is the misleading part. But as I said, I'm probably overthinking.

PR #3952, opened 2022-06-02
Add eager mode operators coverage test

ksztenderski:
Signed-off-by: ksztenderski <ksztenderski@nvidia.com>

## Category:

## Description:

Adds test checking coverage of eager operators. In the future test will be extended by GPU operators.

eager mode tests

### Tests:

- [ ] Existing tests apply
- [x] New tests added
  - [x] Python tests
  - [ ] GTests
  - [ ] Benchmark
  - [ ] Other
- [ ] N/A

### Documentation
- [ ] Documentation updated
  - [ ] Docstring
  - [ ] Doxygen
  - [ ] RST
  - [ ] Jupyter
  - [ ] Other
- [x] N/A

#### Requirements
- [x] N/A

**REQ IDs**: N/A

**JIRA TASK**: N/A

### DIFF
--- a/dali/test/python/test_eager_coverage.py
+++ b/dali/test/python/test_eager_coverage.py
@@ -1023,7 +1023,7 @@ def test_coverage():
     """
 
     methods = module_functions(eager, remove_prefix="nvidia.dali.experimental.eager")
-    # TODO(ksztenderski): Add coverage for math module.
+    # TODO(ksztenderski): Add coverage for math module and for GPU operators.
     exclude = "|".join(
         ["(^" + x.replace(".", "\.").replace("*", ".*").replace("?", ".") + "$)" for x in excluded_methods])
     exclude = re.compile(exclude)
--- a/dali/test/python/test_eager_coverage.py
+++ b/dali/test/python/test_eager_coverage.py
@@ -21,7 +21,7 @@ import nvidia.dali.experimental.eager as eager
 import nvidia.dali.fn as fn
 import nvidia.dali.tensors as tensors
 import nvidia.dali.types as types
-from nvidia.dali import pipeline_def
+from nvidia.dali.pipeline import Pipeline, pipeline_def
 from nvidia.dali._utils.eager_utils import _slice_tensorlist
 from test_dali_cpu_only_utils import (pipeline_arithm_ops_cpu, setup_test_nemo_asr_reader_cpu,
                                       setup_test_numpy_reader_cpu)
@@ -781,8 +781,9 @@ def test_reduce_sum():
 
 @pipeline_def(batch_size=batch_size, num_threads=4, device_id=None)
 def segmentation_select_masks_pipeline(source):
+    device = 'cpu' if Pipeline.current().device_id is None else 'gpu'
     polygons, vertices, selected_masks = fn.external_source(
-        source=source, num_outputs=3, device='cpu')
+        source=source, num_outputs=3, device=device)
     out_polygons, out_vertices = fn.segmentation.select_masks(
         selected_masks, polygons, vertices, reindex_masks=False)
 
@@ -791,8 +792,9 @@ def segmentation_select_masks_pipeline(source):
 
 @pipeline_def(batch_size=batch_size, num_threads=4, device_id=None)
 def segmentation_select_masks_input_pipeline(source):
+    device = 'cpu' if Pipeline.current().device_id is None else 'gpu'
     polygons, vertices, selected_masks = fn.external_source(
-        source=source, num_outputs=3, device='cpu')
+        source=source, num_outputs=3, device=device)
 
     return selected_masks, polygons, vertices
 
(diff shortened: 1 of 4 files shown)


ksztenderski:
!build

dali/test/python/test_eager_cpu_only.py
```diff
@@ -214,6 +305,12 @@ def test_jpeg_compression_distortion_cpu():
     check_single_input('jpeg_compression_distortion', quality=10)
 
 
+def test_image_decoder_crop_device():
```
mzient, line 308: What is this name supposed to mean? Normally "device" means either "GPU" or, in DALI terms, a configurable device (cpu/gpu) - this can be neither in CPU-only tests.

dali/test/python/test_eager_cpu_only.py
```diff
@@ -14,6 +14,7 @@
 
 import numpy as np
```
mzient, line 15: A general remark: why is this test separate from the regular test_dali_cpu_only? This seems like a lot of duplication.

dali/test/python/test_eager_cpu_only.py
ksztenderski: Names are taken from `test_dali_cpu_only.py`, seemed like a good idea for them to be the same.

dali/test/python/test_eager_cpu_only.py
ksztenderski: How would you propose to reuse that code, without some python shenanigans?

ksztenderski:
!build

dali/test/python/test_eager_coverage.py
```diff
@@ -0,0 +1,985 @@
...
+    if source is None:
+        raise RuntimeError('No source for file reader.')
+    data, _ = fn.readers.file(file_root=source)
+    out = op(data, **kwargs)
+    if isinstance(out, list):
+        out = tuple(out)
+    return out
+
+
+def test_image_decoder_cpu():
```
mzient, line 229: Why is everything still suffixed with `_cpu`? I think we can easily make this run the same test for a parametric backend - even if now we use only one backend.

szalpal:
I'd like to double, what @mzient pointed out in the comment: https://github.com/NVIDIA/DALI/pull/3952#discussion_r887825352

There seems to be a lot of code duplication between CPU-only operators tests and eager mode operators. Could we somehow remove the duplication? The rationale would be, that this way we can write only eager test (or CPU-only test, whatever) and this would handle both cases.

Moreover, variable_batch_size tests might be even better suitable for such case. That's because in variable_batch_size tests we run multiple times a pipeline with various batch sizes and compare it to the same pipeline ran with single, bigger batch size. Please consider.

dali/test/python/test_eager_coverage.py
```diff
@@ -0,0 +1,985 @@
+# Copyright (c) 2022, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the 'License');
```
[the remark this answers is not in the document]
ksztenderski: I'll add documentation. Just some explanation here:
> - `data` -> what is it for? having `data`, `audio_data` and `flat_data` is somewhat ambiguous, since the .

It is a little ambiguous, but `data` is of sample "image" shape, and for the rest I guess the names more or less reflect the purpose. In particular they differ in shapes.

>- `GetData` -> here it is hard to generally grasp the purpose from the code only, why couldn't these be just a free functions?

They could be free functions, I just preferred them organized in a class.

dali/test/python/test_eager_coverage.py
ksztenderski: done, removed `_cpu` suffixes and added TODO to cover GPU operators.

klecki (commented):
I hope one day we can sort the functions that we test here :P

dali/test/python/test_eager_coverage.py
```diff
@@ -0,0 +1,1022 @@
...
+        raise RuntimeError('No source for file reader.')
+    data, _ = fn.readers.file(file_root=source)
+    out = op(data, **kwargs)
+    if isinstance(out, list):
+        out = tuple(out)
+    return out
+
+
+def test_image_decoder():
+    check_single_input('decoders.image', pipe_fun=reader_op_pipeline, fn_source=images_dir,
```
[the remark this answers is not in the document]
ksztenderski: I used the generic one as much as possible to save time and lines of code. But some were easier to write and in my opinion to read (when they are a bit more complicated) as individual pipelines. Additionally I think in the future when someone will be writing tests for their new operator they will most likely write their own pipelines, so it will come up anyway.

dali/test/python/test_eager_coverage.py
```diff
@@ -0,0 +1,1087 @@
...
+    out_polygons, out_vertices = fn.segmentation.select_masks(
+        selected_masks, polygons, vertices, reindex_masks=False)
+
+    return out_polygons, out_vertices
+
+
+@pipeline_def(batch_size=batch_size, num_threads=4, device_id=None)
+def segmentation_select_masks_input_pipeline(source):
+    polygons, vertices, selected_masks = fn.external_source(
+        source=source, num_outputs=3, device='cpu')
```
mzient, line 795: likewise
```suggestion
        source=source, num_outputs=3, device='cpu' if device_id is None else 'gpu')
```

mzient (approved):

dali/test/python/test_eager_coverage.py
ksztenderski: done

szalpal (approved):

PR #3958, opened 2022-06-03
Fix fn.coord_transform handling of a default matrix in variable batch case

stiepan:

## Category:

## Description:
If M is not specified, its value is assumed to be id matrix and it is repeated as a per-sample arg to match the current batch size. However, if the dimensionality of data does not change, it is done only once, so if the batch size increases in the following iterations, the matrix is read out of vector bounds.

### Tests:

- [ ] Existing tests apply
- [x] New tests added
  - [x] Python tests
  - [ ] GTests
  - [ ] Benchmark
  - [ ] Other
- [ ] N/A

### Documentation
- [ ] Documentation updated
  - [ ] Docstring
  - [ ] Doxygen
  - [ ] RST
  - [ ] Jupyter
  - [ ] Other
- [x] N/A

#### Requirements
- [x] N/A

**REQ IDs**: N/A

**JIRA TASK**: DALI-2816

### DIFF
--- a/dali/operators/geometry/mt_transform_attr.cc
+++ b/dali/operators/geometry/mt_transform_attr.cc
@@ -130,8 +130,7 @@ void MTTransformAttr::ProcessMatrixArg(const OpSpec &spec, const ArgumentWorkspa
     }
     Repeat(per_sample_mtx_, mtx_, N);
     if (is_fused) {
-      translation_.resize(output_pt_dim_, 0);
-      Repeat(per_sample_translation_, translation_, N);
+      per_sample_translation_.resize(output_pt_dim_ * N, 0);
     }
   }
 }
--- a/dali/operators/geometry/mt_transform_attr.cc
+++ b/dali/operators/geometry/mt_transform_attr.cc
@@ -1,4 +1,4 @@
-// Copyright (c) 2020, NVIDIA CORPORATION. All rights reserved.
+// Copyright (c) 2020-2022, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


stiepan:
!build

banasraf (approved):

dali/operators/geometry/mt_transform_attr.cc
```diff
@@ -127,11 +127,11 @@ void MTTransformAttr::ProcessMatrixArg(const OpSpec &spec, const ArgumentWorkspa
...
-      Repeat(per_sample_mtx_, mtx_, N);
-      if (is_fused) {
-        translation_.resize(output_pt_dim_, 0);
-        Repeat(per_sample_translation_, translation_, N);
-      }
+    }
+    Repeat(per_sample_mtx_, mtx_, N);
+    if (is_fused) {
+      translation_.resize(output_pt_dim_, 0);
+      Repeat(per_sample_translation_, translation_, N);
```
mzient, line 134: I guess this warrants another function than Repeat. It's pretty useless to overwrite the entire array if we only need to fill a few samples and that only in a now unlikely case (this will get more likely with conditional execution).

```C++
  template <typename OutRange, typename InRange>
  void ResizeBatch(OutRange &&out, InRange &&fill, int new_batch_size) {
  ssize_t n = dali::size(fill);
  assert(dali::size(out) % n == 0);
  ssize_t old_batch_size = dali::size(out) / n;
  resize_if_possible(out, n * new_batch_size);
  for (ssize_t i = old_batch_size, k = i * n; i < new_batch_size; i++)
      for (ssize_t j = 0; j < n; j++, k++)
      out[k] = fill[j];
  }
```

```suggestion
    ResizeBatch(per_sample_mtx_, mtx_, N);
    if (is_fused) {
      translation_.resize(output_pt_dim_, 0);
      ResizeBatch(per_sample_translation_, translation_, N);
```

mzient, line 134: Won't this always be zero-filled?
```suggestion
      per_sample_translation_.resize(N * output_pt_dim_, 0);
```

stiepan: Done.

stiepan:
!build

PR #3963, opened 2022-06-06
Change how samples are owned

klecki:
## Description:
Add to the TensorVector the set of APIs already
present in TensorList:
* IsContiguousTensor
* IsDenseTensor
* AsReshapedTensor
* AsTensor

The APIs allow to access the memory of the contiguous
batch through the API of the Tensor. They are exposed
through the Python APIs for TensorList and the conversion
is valid as long as the allocation of the source object
is not changed.
Caching the returned Tensors is not replicated to simplify
the API.

The atomic counter of current internal "view" Tensors 
(the elements of tensors_) is removed - it use was insignificant
and the deleter appearing in the Tensors created via
the new API could mess it up as it kept the old allocation
and current counter.

Either way, the followups also remove the counter.

TensorVector introducing TensorList APIs to prepare
for the replacement.

There is a difference with how the Tensors are returned 
to Python: instead of returning then via pointer to a cached
Tensor instance with `py::return_value_policy::reference_internal`
(which keeps the TL alive as long as we use Tensor)
we just return the object and rely on the allocation being
kept by the shared_ptr.

If we kept the pointer to the old instance of such Tensor
and DALI destroyed that Tensor (for example with ::Reset())
it would cause a crash. 

The lifetime of such Tensor is still problematic now
so I am open to any suggestion. This API may be really
problematic if we allow in-place modification of TL or similar.

### Tests:
- [ ] Existing tests apply
- [x] New tests added
  - [ ] Python tests
  - [x] GTests
  - [ ] Benchmark
  - [ ] Other
- [ ] N/A

### Documentation
- [x] Documentation updated
  - [x] Docstring
  - [ ] Doxygen
  - [ ] RST
  - [ ] Jupyter
  - [ ] Other
- [ ] N/A

#### Requirements
- [ ] N/A

**REQ IDs**: N/A

**JIRA TASK**: N/A

### DIFF
--- a/dali/pipeline/data/tensor_vector.cc
+++ b/dali/pipeline/data/tensor_vector.cc
@@ -21,13 +21,12 @@ namespace dali {
 
 template <typename Backend>
 TensorVector<Backend>::TensorVector()
-    : views_count_(0), curr_num_tensors_(0), tl_(std::make_shared<TensorList<Backend>>()) {}
+    : curr_num_tensors_(0), tl_(std::make_shared<TensorList<Backend>>()) {}
 
 
 template <typename Backend>
 TensorVector<Backend>::TensorVector(int batch_size)
-    : views_count_(0),
-      curr_num_tensors_(0),
+    : curr_num_tensors_(0),
       tl_(std::make_shared<TensorList<Backend>>(batch_size)) {
   resize_tensors(batch_size);
 }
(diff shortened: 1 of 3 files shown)


klecki:
!build

klecki:
!build

PR #3976, opened 2022-06-08
Fix and reenable optimized Cast kernel

klecki:
## Description:
The bug in binsearch Cast kernel manifested
on empty samples - there was mismatch between
sample that was assigned to a given block
and the block that was assumed by the kernel
due to BlockDescriptos not being generated
for samples with 0 volume.

Now, the empty samples are skipped at the
end of the batch and when assigning the
lowest blockIdx to a given sample.

The binsearch looks for the rightmost
sample when searching the descriptors.

Test was adjusted to not generate empty samples
and another was added to randomly make
selected ranges of samples empty.

The speedup compared to the old implementation
is between 10% and 25%, averaging 16%, depending
on the size of the input.

Additionally the collapsed shape was reused
for the data volume.

Cast

Can I assume that there is at least one non-empty sample?
Should I add a check?

### Tests:

- [x] Existing tests apply
	- Recently extended Python Cast test was adjusted to not generate 0 samples randomly
	  and another one was added to do that in controlled manner.
- [x] New tests added
  - [x] Python tests
  - [ ] GTests
  - [ ] Benchmark
  - [ ] Other
- [ ] N/A

### Documentation
- [ ] Documentation updated
  - [ ] Docstring
  - [ ] Doxygen
  - [ ] RST
  - [ ] Jupyter
  - [ ] Other
- [ ] N/A

#### Requirements
- [x] N/A

**REQ IDs**: N/A

**JIRA TASK**: N/A

### DIFF
--- a/dali/test/python/test_operator_cast.py
+++ b/dali/test/python/test_operator_cast.py
@@ -69,7 +69,7 @@ def replace_with_empty_volumes(rng, input, empty_volume_policy):
         start = 0
         end = rng.integers(1, len(input) // 3)
     elif empty_volume_policy == "right":
-        start = rng.integers(1, len(input) * 2 // 3)
+        start = rng.integers(len(input) * 2 // 3, len(input) - 1)
         end = len(input)
     elif empty_volume_policy == "middle":
         start = rng.integers(1 + len(input) // 3, len(input) * 2 // 3)
--- a/dali/test/python/test_operator_cast.py
+++ b/dali/test/python/test_operator_cast.py
@@ -50,8 +50,8 @@ def replace_with_empty_volumes(rng, input, empty_volume_policy):
     input : List of np.array
         Batch to process
     empty_volume_policy : str
-        one of "left", "right, "middle", "all", to indicate if the batch suffix, prefix, infix
-        or all of them should be randomly replaced with 0-volumed samples
+        one of "left", "right, "middle", "mixed", "all", to indicate if the batch suffix, prefix,
+        infix or all of them should be randomly replaced with 0-volumed samples
 
     Returns
     -------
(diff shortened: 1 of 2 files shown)


klecki:
!build

klecki:
CC @MirazSpecial

dali/test/python/test_operator_cast.py
```diff
@@ -38,10 +38,51 @@ def random_shape(rng, ndim: int, max_size: int):
...
+        return input
+    if empty_volume_policy == "all":
+        left = replace_with_empty_volumes(rng, input, "left")
+        left_and_mid = replace_with_empty_volumes(rng, left, "middle")
+        return replace_with_empty_volumes(rng, left_and_mid, "right")
+    if empty_volume_policy == "left":
+        start = 0
+        end = rng.integers(1, len(input) // 3)
+    elif empty_volume_policy == "right":
+        start = rng.integers(1, len(input) * 2 // 3)
```
mzient, line 72:
```suggestion
        start = rng.integers(1 + len(input), len(input) * 2 // 3)
```

dali/test/python/test_operator_cast.py
```diff
@@ -119,3 +161,15 @@ def test_operator_cast():
...
+
+def test_operator_cast_empty_volumes():
+    types = [np.uint8, np.int32, np.float32]
+    for device in ['cpu', 'gpu']:
+        for in_type in types:
+            for out_type in types:
+                ndim = rng.integers(0, 4)
+                batch_size = rng.integers(12, 64)
+                for empty_volume_policy in ["left", "right", "middle", "all"]:
+                    yield (_test_operator_cast, ndim, batch_size, in_type, out_type, device,
```
mzient, line 174: no need to check all combinations - we do have plenty of cases, we can do it round-robin (except `all`, which should be dirt cheap and would invalidate the test).

mzient (approved):

klecki:
!build

dali/test/python/test_operator_cast.py
klecki: done

prak-nv (approved):

PR #4023, opened 2022-06-29
Add warnings against resetting pipeline before end of epoch and test parallel ES with fw iterator

stiepan:
### Tests:

- [ ] Existing tests apply
- [ ] New tests added
  - [ ] Python tests
  - [ ] GTests
  - [ ] Benchmark
  - [ ] Other
- [ ] N/A

### Documentation
- [ ] Documentation updated
  - [ ] Docstring
  - [ ] Doxygen
  - [ ] RST
  - [ ] Jupyter
  - [ ] Other
- [ ] N/A

#### Requirements
- [ ] N/A

**REQ IDs**: N/A

**JIRA TASK**: N/A

### DIFF
--- a/dali/python/nvidia/dali/pipeline.py
+++ b/dali/python/nvidia/dali/pipeline.py
@@ -920,6 +920,8 @@ Parameters
         Should not be mixed with :meth:`run` in the same pipeline"""
         with self._check_api_type_scope(types.PipelineAPIType.SCHEDULED):
             if self._first_iter and self._exec_pipelined:
+                # For one, when prefetching, parallel external source will reuse buffers
+                # that might be still referenced by no_copy input fed to pipeline
                 if not self.empty():
                     warnings.warn(
                         "Prefetching data into a non-empty pipeline may result in corrupted "
--- a/dali/python/nvidia/dali/pipeline.py
+++ b/dali/python/nvidia/dali/pipeline.py
@@ -920,6 +920,12 @@ Parameters
         Should not be mixed with :meth:`run` in the same pipeline"""
         with self._check_api_type_scope(types.PipelineAPIType.SCHEDULED):
             if self._first_iter and self._exec_pipelined:
+                if not self.empty():
+                    warnings.warn(
+                        "Prefetching data into a non-empty pipeline may result in corrupted "
+                        "outputs. Please make sure all batches from previous epoch are consumed "
+                        "before scheduling work for a new epoch.",
+                        Warning)
                 self._prefetch()
             else:
                 self._run_once()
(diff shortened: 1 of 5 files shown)


stiepan:
!build

dali/python/nvidia/dali/pipeline.py
```diff
@@ -920,6 +920,14 @@ def schedule_run(self):
         Should not be mixed with :meth:`run` in the same pipeline"""
         with self._check_api_type_scope(types.PipelineAPIType.SCHEDULED):
             if self._first_iter and self._exec_pipelined:
+                # For one, when prefetching, parallel external source will reuse buffers
+                # that might be still referenced by no_copy input fed to pipeline
+                if not self.empty():
+                    warnings.warn(
```
JanuszL, line 926: I think it is possible, prefetch queue depth is 2, batches to consume is 1, we can still schedule one more run. The native part can overschedule - it will just wait for the empty output buffer, but the ES may fail (parallel mode with nocopy, but the regular ES does copy and have an internal queue).

dali/python/nvidia/dali/pipeline.py
```diff
@@ -1078,7 +1086,19 @@ def reset(self):
 
         If pipeline iterator reached the end then reset its state to the beginning.
         """
-        if self._last_iter:
+        if not self._last_iter:
+            # resetting before some external source raised StopIteration is a no-op
+            if self._input_callbacks:
```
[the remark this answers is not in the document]
JanuszL: 🤷 I guess it still can happen and we should behave consistently.

dali/python/nvidia/dali/pipeline.py
stiepan: Do you mean that it should be supported? By possible you mean that using fw iterators correctly you may still end up in such situation?

JanuszL: > Should we warn only if we have PES and treat it as extra limitation of schedule api that is not there otherwise.

I think so. You can still use this API without the FW iterator.

dali/python/nvidia/dali/pipeline.py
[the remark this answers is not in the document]
stiepan: 😟I thought that one would be useful if one has two or more pipelines with (P)ES that raise StopIteration and should be reset but for some reason the epochs diverge. Either because the number of iterations is really (but unintentionally) different in those two or because one uses .run API, with prefetch_queue_depth 1 and resets all pipelines when the first one raises, not letting the others actually reach end of epoch.

PR #3827, opened 2022-04-15
Stop exposing internal contiguous TV storage

klecki:
## Category: Refactoring

There are few todos, look at the description.

## Description:
Remove from TensorVector the AsTensorList method
and constructor from external shared_ptr<TensorList>

Both allowed to observe the internal state of TensorVector
from the outside, breaking the encapsulation.

Most of those changes can be removed with the replacement
of TensorVector and TensorList as one common TensorBatch
object.
This refactoring is needed, so that replacement can be
done in two separate steps: refactoring the TensorVector
into the targeted TensorBatch class (one of the steps
is reduce the number of nested shared_ptrs),
and next replacing the TensorList by the refactored
class.

Adjust few places that relied on the changes of internal
state to be externally visible:
* Initializing the data graph in workspaces.
  The TV -> TL conversion is done vie Mixed stage.
* ArgumentInputs that are produced as TensorList
  need to be resynced by ShareData instead.

This reverts some changes from #2165:
  - CPU stage cannot be used as direct outputs
    due to the TensorList/TensorVector mismatch,
    we can only share data downwards, but we
    cannot share and preemptivelly expect the allocation
    to be mirrored.
  - CPU-only stage still uses Mixed stage, but just
    with MakeContigous constrained to CPU outputs.

Workspace initialization now takes the CPU_ONLY_DEVICE_ID
into consideration and does not set the stream (resulting
in has_stream() being false, which in turn keeps the
AccessOrder used in Mixed ops `RunImpl` as HostOrder only).

Just for the time of tests two optimizations with
contiguous TensorVector -> TensorList were disabled.

TODO:
Sharing contiguous TV into TL - will be made obsolete by follow up PRs.   

### Tests
- [x] Existing tests apply
- [ ] New tests added
  - [ ] Python tests
  - [ ] GTests
  - [ ] Benchmark
  - [ ] Other
- [ ] N/A

### Documentation
- [x] Existing documentation applies
- [ ] Documentation updated
  - [ ] Docstring
  - [ ] Doxygen
  - [ ] RST
  - [ ] Jupyter
  - [ ] Other
- [ ] N/A

#### Requirements
- [ ] Implements new requirements
- [ ] Affects existing requirements
- [ ] N/A

**REQ IDs**: N/A

**JIRA TASK**: N/A

PR #4030, opened 2022-06-30
[RFC] Use TensorVector everywhere

klecki:
## Description:
The first commit comes from #3827 - see the description of changes there.
I stopped maintaining the chain of commits as it was getting hard to manage, so after the general changes here are reviewed, I can carve out separate commits for more thorough review one by one.

Changes (some of them could be separated into individual PRs):

1. `device_id` parameter was added to `ShareData` and `SetSample` APIs that take shared_ptr - they already accept order, but the device it holds doesn't necessarily indicate the allocation device. (Can be extracted to separate PR, and changes that just add some kind of `device_id` parameter are related to this) 

2. TensorVector APIs are adjusted: 
    * Now it has states: Contiguous and Noncontiguous - which can be either enforced or in "don't care" mode
        * When enforced, changing the mode will error out
        * when not enforced, setting individual samples would mark it as currently non-contiguous
        * Resize got option to coalesce or to split to individual allocation, if we fit, it will still keep the backing allocation - this is the point that can be adjusted 
    * Most metadata is kept in the batch as the source of truth, I tried to make sure they are set in the same order in every callsite, and added some tests for permuting the setters. 
    * Setting or copying in new samples checks against that metadata and must match.
    * When we convert the buffer to non-contiguous and it had a contiguous backing allocation, we mark the `tensors_` that were pointing to that allocation as not sharing, but still aliasing - so we can call Resize() and get a new allocation without the error that we try to reallocate something that shares data
    * The open question is how to treat the variable batch size - when we make the TV smaller, unless we are sharing, we do not reset buffer and we hog the memory. It kinda aligns with the reserve, but might be problematic because of the memory hogging.
    * We need to add the AsTensor and AsReshapedTensor APIs, I tried a separate PR but needed some adjustments on the way (#3963). 

 3. TensorList is removed - the current implementation is just replaced by an alias to TensorVector. Replacing all the uses would introduce a lot of noise, it will be done separately (we can decide on what name we want to use: TensorList, TensorVector, TensorBatch, etc).
     * Some overloads for TensorList were removed, as they now duplicate the signature of TensorVector and are no longer needed.
     * The TensorVector version was preferred for sample access (for example in SequenceOperator).
  
4. As the operators **may** return non-contiguous data, some changes to prevent that were necessary:
    * This part is mostly about changes in op_graph, graph_descr.cc, and pipeline.cc and make_contiguous.*
    * Now, every output gets its own MakeContiguous operator (unconditionally).
    * The MakeContiguous either passes the data through or copies if the input was non-contiguous
    * Additional pass through the graph was added, to see which operators infer the outputs (this currently means, that they return contiguous data), or pass through data (which means we must check if the input that was passed through is contiguous or not. This is a bit problematic, as PassThrough is static information in OpSchema, and the CanInferOutputs() is a part of Operator object API.
    * MakeContiguous is extended to utilize the inferred information.
    * This will require follow-up for Split&Merge et al - they will also pass through the data (part of the batches) but in different manner. 
    * This might be a bit messy, due to realizing pretty late how do we handle duplicated outputs out of the pipeline.

 Know issues:
* numpy reader for GPU triggers a problem with non-contiguous outputs, didn't debug it yet.
* Looks like I introduced a leak in one of the recent rebases which shows up in the L3 test.

* TensorVector/TensorList
* MakeContiguous
* Graph Creation
* Pipeline Outputs

* How the TV keeps the memory in case of the batch size change
* How does the TV::Resize work in contiguous/non-contiguous modes.
	* Do we want additional modes for that?
* Any alternative to adding MakeContiguous for every output?

### Tests:

- [x] Existing tests apply
	* TensorVector is now covered by both TV and TL tests
	* must be tested on the whole DALI test suite
	* perf to be verified on L3 tests
	* some tests that verify the number of ops in graph were adjusted to accommodate for more of MakeContiguous nodes.
- [x] New tests added
	* new tests that check the mixed order of setting batch properties, etc. 
  - [ ] Python tests
  - [ ] GTests
  - [ ] Benchmark
  - [ ] Other
- [ ] N/A

### Documentation
- [ ] Documentation updated
  - [x] Docstring
  - [ ] Doxygen
  - [ ] RST
  - [ ] Jupyter
  - [ ] Other
- [ ] N/A

#### Requirements
- [ ] N/A

**REQ IDs**: N/A

**JIRA TASK**: DALI-2689, DALI-2548


klecki:
!build

dali/c_api/c_api.cc
```diff
@@ -98,16 +98,18 @@ void SetExternalInput(daliPipelineHandle *pipe_handle, const char *name, const v
   // We cast away the const from data_ptr, as there is no other way of passing it to the
   // TensorList, as we must also set the shape and type metadata.
   // It is passed further as const TensorList, so it's data cannot be modified.
-  data.set_pinned(flags & DALI_ext_pinned);
+  // data.set_pinned(flags & DALI_ext_pinned);
```
mzient, line 101: remove

dali/c_api/c_api.cc
```diff
@@ -138,11 +140,14 @@ void SetExternalInputTensors(daliPipelineHandle *pipe_handle, const char *name,
   else
     order = AccessOrder::host();
 
+  int device_id = order.is_device() ? order.device_id() : dali::CPU_ONLY_DEVICE_ID;
```
mzient, line 143: I guess `device_id` should make it to the C API, too.

dali/c_api/c_api.cc
```diff
@@ -460,7 +465,7 @@ size_t daliNumTensors(daliPipelineHandle* pipe_handle, int n) {
 
 template <typename T>
 static size_t daliNumElementsHelper(dali::DeviceWorkspace* ws, int n) {
-  return ws->Output<T>(n)._num_elements();
+  return ws->Output<T>(n).shape().num_elements();
```
mzient, line 468: Why not have this forwarded from TV? I think it adds a lot of noise to this PR and will make usage of the new structure more cumbersome.

dali/pipeline/data/buffer.h
```diff
@@ -38,6 +38,15 @@ class GPUBackend;
 class CPUBackend;
 
 
+enum class BatchState {
+  Default = 0,
```
mzient, line 42: ...and what does this mean? We can have a `Default`, but it should be an alias for something meaningful - or, if the default behaviour is too complex, we can describe it in a comment at least.

dali/pipeline/data/buffer.h
```diff
@@ -567,8 +572,16 @@ class DLL_PUBLIC Buffer {
     num_bytes_ = 0;
   }
 
+  /**
+   * @brief Clear the ShareData flag of the buffer. It will still hold (and co-own) the data due
+   * to the shared_ptr semantics, but it is allowed to call resize() on it.
+   */
+  void mark_detached() {
```
mzient, line 579:
```suggestion
  void detach() {
```

mzient: I think this deserves some discussion.

dali/pipeline/data/tensor.h
```diff
@@ -275,8 +257,8 @@ class Tensor : public Buffer<Backend> {
    * in use by the Tensor.
    */
   inline void ShareData(void *ptr, size_t bytes, bool pinned, const TensorShape<> &shape,
-                        DALIDataType type = DALI_NO_TYPE, AccessOrder order = {}) {
-    ShareData(shared_ptr<void>(ptr, [](void *) {}), bytes, pinned, shape, type, order);
+                        DALIDataType type, int device_id, AccessOrder order = {}) {
```
mzient, line 260: :+1: for getting rid of DALI_NO_TYPE

dali/pipeline/data/tensor.h
```diff
@@ -123,25 +122,6 @@ class Tensor : public Buffer<Backend> {
     order_.wait(order);
   }
 
-  /**
-   * @brief Loads the Tensor at index idx from the input TensorList.
-   */
-  template <typename InBackend>
-  inline void Copy(const TensorList<InBackend> &other, int idx, AccessOrder order = {}) {
```
mzient, line 130: So what do we do now? SampleView? It doesn't have all bells and whistles that a Tensor used to have - and I'd prefer not having to haul the source TensorVector and index everywhere some tensor metadata might be needed...

dali/pipeline/data/tensor_list_test.cc
```diff
@@ -513,18 +543,50 @@ TYPED_TEST(TensorListTest, TestTypeChangeLarger) {
 
   // Change the data type to something larger
   tensor_list.template set_type<double>();
```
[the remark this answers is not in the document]
klecki: When setting samples to new batch, you need to configure it with the metadata first, I think we agreed on it, to not have the first sample be special and set the metadata in some cases.
So the workflow is: you set, as a bare minimum, the number of samples, type and ndim, and than you can set samples that have the correct type and dimensionality.

dali/c_api/c_api.cc
klecki: Merged already as a separate PR :)

dali/pipeline/data/tensor_list_test.cc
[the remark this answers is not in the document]
mzient: OK, according to the discussion, let's restrict set_type to the case where the volume of the tensor is 0. This way, we won't have surprise reallocations - and perhaps we'll even find some bugs.

dali/pipeline/data/tensor_list_test.cc
```diff
@@ -513,18 +543,50 @@ TYPED_TEST(TensorListTest, TestTypeChangeLarger) {
...
-    ASSERT_EQ(tensor_list.tensor_offset(i), offsets[i]);
+    ASSERT_EQ(tensor_list.raw_tensor(i), base_ptr + offsets[i] * sizeof(double));
   }
 
   // nbytes should have increased by a factor of 2
   ASSERT_EQ(nbytes / sizeof(float) * sizeof(double), tensor_list.nbytes());
 }
 
+
+TYPED_TEST(TensorListTest, TestTypeChangeLargerNoncontiguous) {
```
mzient, line 561: This should be forbidden.

dali/pipeline/data/tensor_vector.cc
```diff
@@ -12,268 +12,494 @@
...
+bool same_owner(const std::weak_ptr<void> &x, const std::shared_ptr<void> &y) {
+  if (x.owner_before(y) || y.owner_before(x))
+    return false;
+  return true;
+}
+
+
+// TODO(klecki): move this to the class?
+TensorShape<> empty_shape(int dim) {
+  TensorShape<> result;
```
mzient, line 49: If this is intended to produce a shape that has a volume of 0, then we should check that `dim` is not 0.
```suggestion
  assert(dim > 0);
  TensorShape<> result;
```

dali/pipeline/data/tensor_vector.cc
```diff
@@ -12,268 +12,494 @@
...
+/**
+ * @defgroup copy_impl Helper code for copying batches
+ * The functions used as scaffolding for the synchronization of order for source and destination
+ * buffers. Generic copy between contiguous and noncontiguous batches that assumes that the
+ * buffers are already resized.
+ *
+ * @{
+ */
+/**
+ * @brief Pick the order for Copy to be run on.
```
mzient, line 67: It does quite a bit more than that (it actually calls `wait`).

dali/pipeline/data/tensor_vector.cc
```diff
@@ -431,285 +666,354 @@ bool TensorVector<Backend>::is_pinned() const {
 
 
 template <typename Backend>
-int TensorVector<Backend>::device_id() const {
-  if (IsContiguous()) {
-    return tl_->device_id();
-  } else if (!tensors_.empty()) {
-    return tensors_[0]->device_id();
+void TensorVector<Backend>::set_device_id(int device_id) {
```
mzient, line 669: Not really about this line - but related:
Now that we require a (hopefully valid?) device_id in share_data, I think that Buffer::set_device_id should throw if the pointer is not null.

dali/pipeline/data/tensor_vector.cc
```diff
@@ -431,285 +666,354 @@ bool TensorVector<Backend>::is_pinned() const {
...
-  UpdateViews();
 }
 
 
 template <typename Backend>
 void TensorVector<Backend>::reserve(size_t bytes_per_sample, int batch_size) {
   assert(batch_size > 0);
-  state_ = State::noncontiguous;
+  state_.Setup(BatchState::Noncontiguous);
   resize_tensors(batch_size);
```
mzient, line 704: Technically, a "reserve" function can specify smaller batch size than current one - in which case we should not shrink...

dali/pipeline/data/tensor_vector.h
```diff
@@ -122,40 +122,40 @@ class DLL_PUBLIC TensorVector {
    */
   std::vector<size_t> _chunks_capacity() const;
 
-  TensorListShape<> shape() const;
+  const TensorListShape<> &shape() const;
```
mzient, line 125:
```suggestion
  const TensorListShape<> &shape() const &;
```
Just in case...

Also, you can add an overload with `&&` qualifier which returns the shape by value - or simply delete it.

dali/pipeline/data/tensor_vector.cc
```diff
@@ -12,268 +12,494 @@
...
+  dsts.reserve(num_samples);
+  SmallVector<Index, 256> sizes;
+  sizes.reserve(num_samples);
+  for (int i = 0; i < num_samples; i++) {
+    dsts.emplace_back(dst.raw_mutable_tensor(i));
+    srcs.emplace_back(src.raw_tensor(i));
+    sizes.emplace_back(src.shape()[i].num_elements());
+  }
+
+  type_info.template Copy<SrcBackend, DstBackend>(dsts.data(), srcs.data(), sizes.data(),
```
mzient, line 120:
```suggestion
  type_info.Copy<SrcBackend, DstBackend>(dsts.data(), srcs.data(), sizes.data(),
```
type_info is not a dependent name, so Copy shouldn't need a template disambiguation.

dali/pipeline/data/tensor_vector.cc
```diff
@@ -12,268 +12,494 @@
...
+TensorVector<Backend>::TensorVector() : curr_num_tensors_(0) {}
 
 
 template <typename Backend>
-TensorVector<Backend>::TensorVector(int batch_size)
-    : views_count_(0),
-      curr_num_tensors_(0),
-      tl_(std::make_shared<TensorList<Backend>>(batch_size)) {
+TensorVector<Backend>::TensorVector(int batch_size) : curr_num_tensors_(0) {
+  // This is why we can't have nice things as `dim = 0` is already occupied
```
mzient, line 171: That's hardly the reason. The real reason we "can't have nice things", as you call it, is the very fact that there can be such a thing as an empty Tensor(List) - [i]at all[/i].

dali/pipeline/data/tensor_vector.cc
```diff
@@ -12,268 +12,494 @@
...
+  // by competing functionality, we need to guard ourself from thinking we have some scalar
+  // allocation where in fact we just have empty samples.
+  // So instead we set the dim to 1, and the resize_tensor will cause the shape to be resized
+  // to appropriate number of samples and 0-initialized, so copy from batch to batch,
+  // using this as (empty) source still works. Maybe there is a better solution to this problem.
```
mzient, line 176: One tongue-in-cheek solution is to disallow empty Tensor(List) and use `optional`. Another option is to make shape_ optional (and perhaps make shape() throw).

dali/pipeline/data/tensor_vector.cc
```diff
@@ -12,268 +12,494 @@
...
+  // Checks in the order of class members
+  DALI_ENFORCE(this->type() == type,
+               make_string("Sample must have the same type as the target batch, current: ",
+                           this->type(), ", new: ", type, error_suffix));
+
+  DALI_ENFORCE(this->sample_dim() == sample_dim,
+               make_string("Sample must have the same sample dim as the target batch, current: ",
+                           this->sample_dim(), ", new: ", sample_dim, error_suffix));
+
+  DALI_ENFORCE(this->GetLayout() == layout,
```
mzient, line 289: I guess we could accept a sample without a layout and adopt whatever layout is currently set.

dali/pipeline/data/tensor_vector.cc
```diff
@@ -12,268 +12,494 @@
...
-  DALI_ENFORCE(this->order() == src.order(), "Sample must have the same order as a target batch");
-  DALI_ENFORCE(
-      GetLayout() == src.GetLayout(),
-      make_string("Sample must have the same layout as a target batch current: ", GetLayout(),
-                  " new: ", src.GetLayout(), " for ", sample_idx, " <- ", src_sample_idx, "."));
-  DALI_ENFORCE(
-      is_pinned() == src.is_pinned(),
-      make_string("Sample must have the same pinned status as target batch, current: ", is_pinned(),
-                  " new: ", src.is_pinned(), " for ", sample_idx, " <- ", src_sample_idx, "."));
+  // Setting any individual sample converts the batch to non-contiguous mode
```
mzient, line 312: I guess at least a self-assignment check would be nice.

klecki: added

dali/pipeline/data/tensor_vector.cc
klecki: Done.

dali/pipeline/data/buffer.h
klecki: Added a comment here.

klecki:
Changes were merged as a separate PRs

PR #4043, opened 2022-07-08
Use actual batch size instead of max bat chsize in cropattr.

mzient:
Signed-off-by: Michal Zientkiewicz <michalz@nvidia.com>

## Category:
**Bug fix** (*non-breaking change which fixes an issue*)

## Description:
Fixes out-of-range access in CropAttr when actual batch size is smaller than max. batch size.

### Tests:

- [ ] Existing tests apply
- [ ] New tests added
  - [ ] Python tests
  - [ ] GTests
  - [ ] Benchmark
  - [ ] Other
- [ ] N/A
TODO

### Documentation
- [ ] Existing documentation applies
- [ ] Documentation updated
  - [ ] Docstring
  - [ ] Doxygen
  - [ ] RST
  - [ ] Jupyter
  - [ ] Other
- [ ] N/A

#### Requirements
- [ ] Implements new requirements
- [ ] Affects existing requirements
- [X] N/A

**REQ IDs**: N/A

**JIRA TASK**: N/A

PR #4045, opened 2022-07-11
Use actual batch size instead of max batch size in `crop_attr.h`

szalpal:

## Category:

## Description:

Fixes out-of-range access in CropAttr when actual batch size is smaller than max. batch size.

I had problems with adding a test to #4043 , so I created this PR. May we close the previous one and continue from here?

### Tests:

- [ ] Existing tests apply
- [ ] New tests added
  - [x] Python tests
  - [ ] GTests
  - [ ] Benchmark
  - [ ] Other
- [ ] N/A

### Documentation
- [ ] Documentation updated
  - [ ] Docstring
  - [ ] Doxygen
  - [ ] RST
  - [ ] Jupyter
  - [ ] Other
- [ ] N/A

#### Requirements
- [ ] N/A

**REQ IDs**: N/A

**JIRA TASK**: N/A

### DIFF
--- a/dali/test/python/test_dali_variable_batch_size.py
+++ b/dali/test/python/test_dali_variable_batch_size.py
@@ -12,24 +12,26 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from nvidia.dali.pipeline import Pipeline
-from segmentation_test_utils import make_batch_select_masks
-from test_utils import module_functions
-from nose.tools import nottest
-import nvidia.dali.fn as fn
-import nvidia.dali.types as types
-import nvidia.dali.math as dmath
-from nvidia.dali.plugin.numba.fn.experimental import numba_function
+import inspect
+import nose
 import numpy as np
+import nvidia.dali.fn as fn
+import nvidia.dali.math as dmath
+import nvidia.dali.types as types
+import os
+import random
+import re
+from nose.plugins.attrib import attr
+from nose.tools import nottest
+from nvidia.dali.pipeline import Pipeline, pipeline_def
+from nvidia.dali.plugin.numba.fn.experimental import numba_function
+
 import test_utils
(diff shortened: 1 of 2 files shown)


mzient (approved):

dali/test/python/test_dali_variable_batch_size.py
```diff
@@ -1081,6 +1083,36 @@ def test_subscript_dim_check():
...
+    pipe.build()
+
+    image_data = np.fromfile(
+        os.path.join(test_utils.get_dali_extra_path(), "db", "single", "jpeg", "100",
+                     "swan-3584559_640.jpg"),
+        dtype=np.uint8)
+    pipe.feed_input("IMAGE", [image_data])
+    pipe.feed_input("CROP_X", [np.float32(0.5)])
+    pipe.feed_input("IMAGE", [image_data])
+    pipe.feed_input("CROP_X", [np.float32(0.4)])
```
klecki, line 1111: Wow, I didn't see feed_input in quite some time.
I assume you checked that this test triggers the problem without the fix?

klecki (approved):

dali/test/python/test_dali_variable_batch_size.py
szalpal: Yes, without the fix wild segfault appears ;)

dali/test/python/test_dali_variable_batch_size.py
```diff
@@ -1081,6 +1083,36 @@ def test_subscript_dim_check():
     check_pipeline(data, single_op_pipeline, operator_fn=fn.subscript_dim_check, num_subscripts=1)
 
 
+def test_crop_argument_from_external_source():
```
mzient, line 1086: I think we can remove the previous variable batch size test for this operator, because apparently it failed to detect a rather obvious issue.

PR #4048, opened 2022-07-11
Fix lack of reset when `iter()` is called on the DALI framework iterator

JanuszL:
- the DALI framework iterator when `iter()` is called checks if any
  data has been consumed yet. If it hasn't it doesn't reset to prevent
  improper operation when DALI FW iterator prefetches the first batch after
  creation, and invocation of something like `enumerate(iterator)`,
  would reset the iterator and extend the length of the first epoch
  (prefetched data in the FW iterator would be added to the freshly
  reset iterator). The flag that is set when any data has been consumed
  was set only in the `next()` method and not `__next__()`. This PR
  fixes this problem and adjusts the test

## Category:

## Description:
- the DALI framework iterator when `iter()` is called checks if any
  data has been consumed yet. If it hasn't it doesn't reset to prevent
  improper operation when DALI FW iterator prefetches the first batch after
  creation, and invocation of something like `enumerate(iterator)`,
  would reset the iterator and extend the length of the first epoch
  (prefetched data in the FW iterator would be added to the freshly
  reset iterator). The flag that is set when any data has been consumed
  was set only in the `next()` method and not `__next__()`. This PR
  fixes this problem and adjusts the test

### Tests:

- [x] Existing tests apply
  - test_fw_iterators.py:test_*_autoreset_iter()
- [ ] New tests added
  - [ ] Python tests
  - [ ] GTests
  - [ ] Benchmark
  - [ ] Other
- [ ] N/A

### Documentation
- [ ] Documentation updated
  - [ ] Docstring
  - [ ] Doxygen
  - [ ] RST
  - [ ] Jupyter
  - [ ] Other
- [ ] N/A

#### Requirements
- [x] N/A

**REQ IDs**: N/A

**JIRA TASK**: N/A

### DIFF
--- a/dali/test/python/test_fw_iterators.py
+++ b/dali/test/python/test_fw_iterators.py
@@ -1728,9 +1728,9 @@ def check_autoreset_iter(fw_iterator, extract_data, auto_reset_op):
 
     loader = fw_iterator(pipeline, reader_name="reader", auto_reset=auto_reset_op)
     for _ in range(2):
-        loader = iter(loader)
-        for i in range(len(loader)):
-            data = next(loader)
+        loader_iter = iter(loader)
+        for i in range(len(loader_iter)):
+            data = next(loader_iter)
             for j, d in enumerate(extract_data(data[0])):
                 assert d[0] == i * batch_size + j, f"{d[0]} { i * batch_size + j}"
 
--- a/dali/test/python/test_fw_iterators.py
+++ b/dali/test/python/test_fw_iterators.py
@@ -1728,7 +1728,9 @@ def check_autoreset_iter(fw_iterator, extract_data, auto_reset_op):
 
     loader = fw_iterator(pipeline, reader_name="reader", auto_reset=auto_reset_op)
     for _ in range(2):
-        for i, data in enumerate(loader):
+        loader = iter(loader)
+        for i in range(len(loader)):
+            data = next(loader)
             for j, d in enumerate(extract_data(data[0])):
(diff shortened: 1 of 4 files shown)


JanuszL:
!build

jantonguirao (approved):

dali/test/python/test_fw_iterators.py
```diff
@@ -1728,7 +1728,9 @@ def BoringPipeline():
 
     loader = fw_iterator(pipeline, reader_name="reader", auto_reset=auto_reset_op)
     for _ in range(2):
-        for i, data in enumerate(loader):
+        loader = iter(loader)
+        for i in range(len(loader)):
+            data = next(loader)
```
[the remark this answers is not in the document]
JanuszL: Auotreset kicked in and not call to `iter()`, as enumerate calls `iter()` and handles `StopIteration`.

dali/test/python/test_fw_iterators.py
```diff
@@ -1728,7 +1728,9 @@ def BoringPipeline():
 
     loader = fw_iterator(pipeline, reader_name="reader", auto_reset=auto_reset_op)
     for _ in range(2):
-        for i, data in enumerate(loader):
+        loader = iter(loader)
```
mzient, line 1731: Hmmm... at this point I'd suggest to store it into a separate variable, should we switch to returning something else than `self` in the future.

JanuszL: Fixed

mzient (approved):

PR #4049, opened 2022-07-11
Expose device_id in ShareData and SetSample APIs

klecki:
## Description:
The device_id is necessary metadata when accepting
external memory via pointers.

The device_id can differ from the one that is
present in the order. device_id represents the
location of memory, the order tracks the necessary
synchronization - the provided memory can in theory
depend on a work being done on another device.

Adjust the ShareData/SetSample APIs, so the device_id
is provided in one go with the allocation.

Assume that the C API memory is provided on the
device of the current Pipeline.

TODO:
* Add the above mention to docstrings
* Add a test that checks Python Tensor/TensorList
    constructors in their current state (conversion from CuPy?)
* Mention the requirements of device_id in C API.

Setting external memory.

If the CPU_ONLY_DEVICE_ID is not used instead
of the proper device.

### Tests:
TODO

- [ ] Existing tests apply
- [ ] New tests added
  - [ ] Python tests
  - [ ] GTests
  - [ ] Benchmark
  - [ ] Other
- [ ] N/A

### Documentation
TODO

- [ ] Documentation updated
  - [ ] Docstring
  - [ ] Doxygen
  - [ ] RST
  - [ ] Jupyter
  - [ ] Other
- [ ] N/A

#### Requirements
- [ ] N/A

**REQ IDs**: N/A

**JIRA TASK**: N/A

### DIFF
--- a/dali/pipeline/data/buffer.h
+++ b/dali/pipeline/data/buffer.h
@@ -458,11 +458,10 @@ class DLL_PUBLIC Buffer {
    *
    * Current Buffer will be marked as sharing data, and reallocation of memory will be
    * prohibited until reset() is called.
-   *
-   * For GPU memory, it is assumed to be associated with current device.
    */
   inline void set_backing_allocation(const shared_ptr<void> &ptr, size_t bytes, bool pinned,
-                                     DALIDataType type = DALI_NO_TYPE, size_t size = 0) {
+                                     DALIDataType type, size_t size, int device_id,
+                                     AccessOrder order = {}) {
     free_storage();
     type_ = TypeTable::GetTypeInfo(type);
     data_ = ptr;
@@ -471,12 +470,8 @@ class DLL_PUBLIC Buffer {
     shares_data_ = data_ != nullptr;
     num_bytes_ = bytes;
     pinned_ = pinned;
-    // setting the allocation, get the device
-    if ((std::is_same<Backend, GPUBackend>::value || pinned_) && device_ == CPU_ONLY_DEVICE_ID) {
-      CUDA_CALL(cudaGetDevice(&device_));
-    } else {
-      device_ = CPU_ONLY_DEVICE_ID;
-    }
+    device_ = device_id;
+    set_order(order);
   }
 
   static void SetGrowthFactor(double factor) {
--- a/dali/pipeline/data/tensor_list.h
+++ b/dali/pipeline/data/tensor_list.h
@@ -22,6 +22,7 @@
 #include <list>
 #include <memory>
 #include <utility>
+#include "dali/core/access_order.h"
 #include "dali/core/tensor_shape.h"
 #include "dali/pipeline/data/backend.h"
 #include "dali/pipeline/data/buffer.h"
@@ -256,7 +257,7 @@ class DLL_PUBLIC TensorList {
    * Size can be set to 0 and type to NoType as intermediate step.
    */
   inline void ShareData(const shared_ptr<void> &ptr, size_t bytes, bool pinned,
-                        const TensorListShape<> &shape, DALIDataType type = DALI_NO_TYPE,
+                        const TensorListShape<> &shape, DALIDataType type, int device_id,
                         AccessOrder order = {}) {
     // Free the underlying storage.
     data_.free_storage();
@@ -265,7 +266,7 @@ class DLL_PUBLIC TensorList {
     this->set_order(order);
 
     // Save our new pointer and bytes. Reset our type, shape, and size
-    data_.set_backing_allocation(ptr, bytes, pinned, type, shape.num_elements());
+    data_.set_backing_allocation(ptr, bytes, pinned, type, shape.num_elements(), device_id, order);
     shape_ = {};
     offsets_.clear();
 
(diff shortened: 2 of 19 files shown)


klecki:
!build

dali/pipeline/data/tensor_list.h
```diff
@@ -265,7 +270,7 @@ class DLL_PUBLIC TensorList {
     this->set_order(order);
 
     // Save our new pointer and bytes. Reset our type, shape, and size
-    data_.set_backing_allocation(ptr, bytes, pinned, type, shape.num_elements());
+    data_.set_backing_allocation(ptr, bytes, pinned, type, shape.num_elements(), device_id, order);
```
stiepan, line 273: So now we set the order twice, right? Do we care, should the `set_backing_allocation` do it alone?

dali/pipeline/data/buffer.h
```diff
@@ -471,12 +474,8 @@ class DLL_PUBLIC Buffer {
...
     num_bytes_ = bytes;
     pinned_ = pinned;
-    // setting the allocation, get the device
-    if ((std::is_same<Backend, GPUBackend>::value || pinned_) && device_ == CPU_ONLY_DEVICE_ID) {
-      CUDA_CALL(cudaGetDevice(&device_));
-    } else {
-      device_ = CPU_ONLY_DEVICE_ID;
-    }
+    device_ = device_id;
+    set_order(order);
```
stiepan, line 478: My understanding of `set_order` (for two streams) is that I had a buffer that may be somehow used by the work scheduled on stream1, now I want to repupose the same memory for the work on stream2 so anything scheduled from then on on the stream2 will wait for the work hitherto scheduled on the stream1 to complete. But if I replace the underlying memory itself? Is there any connection between the two? (Esp. given the free_storage) at the top. Shouldn't it be `set_order(order, false)`?

klecki: Actually, it should probably be moved just after the `free_storage`, as is already the case in ShareData for current TensorList implementation. That would make sense - `free_storage` might not free data (if we ware sharing), so maybe it's good idea to synchronize before we do the actual swap.

klecki:
!build

dali/pipeline/data/tensor_list.h
klecki: Good catch, removing

klecki: On the second thought, I kept it and expanded the comment, it will be a no op in set_backing_allocation, but maybe it's worth to keep things explicit.

include/dali/c_api.h
```diff
@@ -194,6 +194,9 @@ DLL_PUBLIC void daliSetExternalInputBatchSize(daliPipelineHandle *pipe_handle, c
  * copying data onto GPU. This function is asynchronous, so it's your responsibility to
  * synchronize on a provided CUDA stream.
  *
+ * If GPU memory is provided, it is assumed to reside on the same device that was used
+ * to create the pipeline that is used. See `device_id` parameter of the `daliCreatePipeline`.
```
banasraf, line 198:
```suggestion
 * If GPU memory is provided, it is assumed to reside on the same device that
 * the pipeline that is used. See `device_id` parameter of the `daliCreatePipeline`.
```
just a suggestion

klecki:
!build

dali/python/backend_impl.cc
```diff
@@ -188,21 +189,23 @@ void FillTensorFromDlPack(py::capsule capsule, SourceDataType<SrcBackend> *batch
...
-    batch->set_device_id(-1);
+    device_id = CPU_ONLY_DEVICE_ID;
   } else if (dl_tensor.device.device_type == kDLCUDA) {
-    batch->set_device_id(dl_tensor.device.device_id);
+    device_id = dl_tensor.device.device_id;
   } else {
     DALI_FAIL(make_string("Not supported DLPack device type: ", dl_tensor.device.device_type, "."));
   }
 
+  batch->ShareData(shared_ptr<void>(dl_tensor.data,
```
stiepan, line 204: kind of nitpick, but could you move the comment about empty lambda from L188 to L204? It departed quite far from the original target.

stiepan (approved):

include/dali/c_api.h
klecki: I wanted to mention the device_id that we passed to the pipeline that we are using now. Quite a mouthful, will change.

klecki: Done

klecki:
!build

banasraf (approved):

klecki:
!build

dali/pipeline/data/tensor_vector_test.cc
```diff
@@ -227,6 +232,43 @@ TYPED_TEST(TensorVectorSuite, MoveAssignmentMetaData) {
...
+  for (int device_id = 0; device_id < num_devices; device_id++) {
+    TensorVector<TypeParam> batch;
+    batch.SetSize(1);
+    DeviceGuard dg(device_id);
+    batch.set_device_id(device_id);
+    batch.set_pinned(is_pinned);
+    batch.set_type(DALI_UINT8);
+    batch.set_sample_dim(shape.sample_dim());
+    batch.set_order(order);
+    CUDA_CALL(cudaSetDevice(device_id));
```
stiepan, line 254: Isn't the DeviceGaurd calling the setdevice on its own?

klecki: Fixing.

klecki:
!build

PR #4053, opened 2022-07-12
Add WebP parser

staniewzki:
## Category:

## Description:
Implements parser interface for WebP format.

### Tests:

- [ ] Existing tests apply
- [x] New tests added
  - [ ] Python tests
  - [x] GTests
  - [ ] Benchmark
  - [ ] Other
- [ ] N/A

### Documentation
- [ ] Documentation updated
  - [ ] Docstring
  - [ ] Doxygen
  - [ ] RST
  - [ ] Jupyter
  - [ ] Other
- [ ] N/A

#### Requirements
- [ ] N/A

**REQ IDs**: N/A

**JIRA TASK**: N/A

### DIFF
--- a/dali/imgcodec/parsers/webp.cc
+++ b/dali/imgcodec/parsers/webp.cc
@@ -94,16 +94,15 @@ std::string sequence_of_integers(const std::array<uint8_t, N> &data) {
 
 ImageInfo WebpParser::GetInfo(ImageSource *encoded) const {
   auto stream = encoded->Open();
+  ImageInfo info;
 
   stream->Skip<RiffHeader>();
-
-  ImageInfo info;
   const auto vp8_header = stream->ReadOne<vp8_header_t>();
   if (is_simple_lossy_format(vp8_header)) {
     const auto lossy_header = stream->ReadOne<WebpLossyHeader>();
     if (!is_sync_code_valid(lossy_header)) {
       DALI_FAIL("Sync code 157 1 42 not found at expected position. Found " +
-                sequence_of_integers(lossy_header.sync_code));
+                sequence_of_integers(lossy_header.sync_code) + " instead");
     }
 
     const int w = ReadValueLE<uint16_t>(*stream) & 0x3FFF;
--- a/dali/imgcodec/parsers/webp.cc
+++ b/dali/imgcodec/parsers/webp.cc
@@ -13,6 +13,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+#include <algorithm>
 #include <string>
 #include "dali/imgcodec/parsers/webp.h"
 #include "dali/core/byte_io.h"
@@ -65,7 +66,7 @@ static_assert(sizeof(WebpLosslessHeader) == 5);
 
 template<size_t N>
 constexpr std::array<uint8_t, N - 1> tag(const char (&c)[N]) {
-  std::array<uint8_t, N - 1> a;
+  std::array<uint8_t, N - 1> a{};
   std::copy(&c[0], &c[N - 1], a.begin());
   return a;
 }
--- a/dali/imgcodec/parsers/webp.cc
+++ b/dali/imgcodec/parsers/webp.cc
@@ -117,6 +117,7 @@ ImageInfo WebpParser::GetInfo(ImageSource *encoded) const {
 
     const int w = ReadValueLE<uint16_t>(*stream) & 0x3FFF;
     const int h = ReadValueLE<uint16_t>(*stream) & 0x3FFF;
+
     // VP8 always uses RGB
     info.shape = {h, w, 3};
   } else if (is_simple_lossless_format(vp8_header)) {
--- a/dali/imgcodec/parsers/webp.cc
+++ b/dali/imgcodec/parsers/webp.cc
@@ -1,4 +1,3 @@
-// Copyright (c) 2021, Lennart Behme. All rights reserved.
 // Copyright (c) 2022, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
--- a/dali/imgcodec/parsers/webp.cc
+++ b/dali/imgcodec/parsers/webp.cc
@@ -32,9 +32,12 @@ namespace {
 // respectively "VP8 " and "VP8L" in vp8_header_t. "VP8X" in vp8_header_t signifies
 // WebP extended file format, which is not supported.
 
+// Struct must be packed for reading them to work correctly
+#pragma pack(push, 1)
+
 struct RiffHeader {
   std::array<uint8_t, 4> riff_text;
-  std::array<uint8_t, 4> file_size;
+  uint32_t file_size;
   std::array<uint8_t, 4> webp_text;
 };
 static_assert(sizeof(RiffHeader) == 12);
@@ -44,7 +47,7 @@ using vp8_header_t = std::array<uint8_t, 4>;
 // Simple file format (lossy)
 // https://datatracker.ietf.org/doc/html/rfc6386#section-9.1
 struct WebpLossyHeader {
-  std::array<uint8_t, 4> chunk_size;
+  uint32_t chunk_size;
   std::array<uint8_t, 3> frame_tag;
   std::array<uint8_t, 3> sync_code;
 };
@@ -53,39 +56,39 @@ static_assert(sizeof(WebpLossyHeader) == 10);
 // Simple file format (lossless)
 // https://developers.google.com/speed/webp/docs/webp_lossless_bitstream_specification#2_riff_header
 struct WebpLosslessHeader {
-  std::array<uint8_t, 4> chunk_size;
+  uint32_t chunk_size;
   uint8_t signature_byte;
 };
 static_assert(sizeof(WebpLosslessHeader) == 5);
 
-template<size_t N>
-bool is_pattern_matching(const std::array<uint8_t, N> data, const char pattern[N + 1]) {
-  for (size_t i = 0; i < N; i++)
-    if (data[i] != pattern[i])
-      return false;
-  return true;
+#pragma pack(pop)  // end of packing scope
+
+constexpr size_t TagSize = 4;
+
+std::array<uint8_t, TagSize> tag(const char (&c)[TagSize + 1]) {
+  std::array<uint8_t, TagSize> a;
+  std::copy(&c[0], &c[TagSize], a.begin());
+  return a;
 }
 
 bool is_valid_riff_header(const RiffHeader &header) {
-  return is_pattern_matching(header.riff_text, "RIFF") &&
-         is_pattern_matching(header.webp_text, "WEBP");
+  return header.riff_text == tag("RIFF") && header.webp_text == tag("WEBP");
 }
 
 bool is_simple_lossy_format(const vp8_header_t &header) {
-  return is_pattern_matching(header, "VP8 ");
+  return header == tag("VP8 ");
 }
 
 bool is_simple_lossless_format(const vp8_header_t &header) {
-  return is_pattern_matching(header, "VP8L");
+  return header == tag("VP8L");
 }
 
 bool is_extended_format(const vp8_header_t &header) {
-  return is_pattern_matching(header, "VP8X");
+  return header == tag("VP8X");
 }
 
 bool is_sync_code_valid(const WebpLossyHeader &header) {
-  const auto &s = header.sync_code;
-  return s[0] == 0x9D && s[1] == 0x01 && s[2] == 0x2A;
+  return header.sync_code == std::array<uint8_t, 3>{0x9D, 0x01, 0x2A};
 }
 
 bool is_sync_code_valid(const WebpLosslessHeader &header) {
--- a/dali/imgcodec/parsers/webp.cc
+++ b/dali/imgcodec/parsers/webp.cc
@@ -20,61 +20,100 @@ namespace dali {
 namespace imgcodec {
 
 namespace {
-  bool is_pattern_matching(const uint8_t *data, const std::string &pattern) {
-    for (size_t i = 0; i < pattern.size(); i++)
-      if (data[i] != pattern[i])
-        return false;
-    return true;
-  }
 
-  bool is_valid_sync_code_lossy(const uint8_t *data) {
-    return data[0] == 0x9D && data[1] == 0x01 && data[2] == 0x2A;
-  }
+struct RiffHeader {
+  std::array<uint8_t, 4> riff_text;
+  std::array<uint8_t, 4> file_size;
+  std::array<uint8_t, 4> webp_text;
+};
+static_assert(sizeof(RiffHeader) == 12);
 
-  const uint8_t SIGNATURE_BYTE_LOSTLESS = 0x2F;
+using vp8_header_t = std::array<uint8_t, 4>;
+
+// Simple file format (lossy)
+// https://datatracker.ietf.org/doc/html/rfc6386#section-9.1
+struct WebpLossyHeader {
+  std::array<uint8_t, 4> chunk_size;
+  std::array<uint8_t, 3> frame_tag;
+  std::array<uint8_t, 3> sync_code;
+};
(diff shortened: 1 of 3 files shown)


dali/imgcodec/parsers/webp.cc
```diff
@@ -12,19 +12,118 @@
...
+    // Skipping chunk size (4 bytes) and frame tag (3 bytes)
+    stream->SeekRead(7, SEEK_CUR);
+
+    // Verify sync code
+    const auto sync_code = stream->ReadOne<std::array<uint8_t, 3>>();
+    if (!is_valid_sync_code_lossy(sync_code.data())) {
+      DALI_FAIL("Sync code 157 1 42 not found at expected position. Found " +
+                std::to_string(sync_code[0]) + " " +
+                std::to_string(sync_code[1]) + " " +
+                std::to_string(sync_code[2]) + " instead.");
```
jantonguirao, line 61:
```suggestion
      DALI_FAIL(make_string("Sync code 157 1 42 not found at expected position. Found ",
                                             sync_code[0], " ", sync_code[1], " ",  sync_code[2], " instead."));
```
Is usually preferred.

dali/imgcodec/parsers/webp.cc
```diff
@@ -12,19 +12,118 @@
 // See the License for the specific language governing permissions and
```
klecki, line 12: Please add at the top of the file the copyright from the `dali/image/webp.cc`, as this is based on it
```// Copyright (c) 2021, Lennart Behme. All rights reserved.```

File: dali/imgcodec/parsers/webp.cc [hunk 3]
```diff
@@ -12,19 +12,118 @@
...
 
+namespace {
+  bool is_pattern_matching(const uint8_t *data, const std::string &pattern) {
+    for (size_t i = 0; i < pattern.size(); i++)
+      if (data[i] != pattern[i])
+        return false;
+    return true;
+  }
+
+  bool is_valid_sync_code_lossy(const uint8_t *data) {
```
klecki, line 30: And maybe this:
```suggestion
  bool is_valid_sync_code_lossy(const std::array<uint8_t, 3> &data) {
```

dali/imgcodec/parsers/webp.cc
```diff
@@ -12,19 +12,118 @@
...
 
+#include <string>
 #include "dali/imgcodec/parsers/webp.h"
+#include "dali/core/byte_io.h"
 
 namespace dali {
 namespace imgcodec {
 
+namespace {
+  bool is_pattern_matching(const uint8_t *data, const std::string &pattern) {
```
klecki, line 23: We can consider:
```suggestion
  template <size_t N>
  bool is_pattern_matching(const std::array<uint8_t, N> &data, const char(&pattern)[N + 1]) { 
```
(N +  1) for the `\0` terminator, we can get the string literal directly and have it match the size of the array.

dali/imgcodec/parsers/webp.cc
```diff
@@ -12,19 +12,118 @@
...
+    return false;
+
+  // Check the RIFF header
+  if (!is_pattern_matching(header, "RIFF"))
+    return false;
+  // 4 bytes of file size skipped
+  if (!is_pattern_matching(header + 8, "WEBP"))
+    return false;
+
+  if (is_pattern_matching(header + 12, "VP8 ")) {
```
klecki, line 115: After reading the first part that handles the seeks basically "under the hood", this kinda feels like repeating the old, worse patterns. But I have no idea how to remove the repetition on make it better, I guess this is so small that we can live with it.

staniewzki:
@klecki following your suggestions, rewrote the `CanParse`, extracting some common code from `GetInfo`. I like the code much better now.

dali/imgcodec/parsers/webp.cc
staniewzki: why `(&pattern)[N + 1]` instead of plain `pattern[N + 1]`?

dali/imgcodec/parsers/webp.cc
```diff
@@ -12,19 +13,144 @@
...
+};
+static_assert(sizeof(WebpLossyHeader) == 10);
+
+// Simple file format (lossless)
+// https://developers.google.com/speed/webp/docs/webp_lossless_bitstream_specification#2_riff_header
+struct WebpLosslessHeader {
+  std::array<uint8_t, 4> chunk_size;
+  uint8_t signature_byte;
+};
+static_assert(sizeof(WebpLosslessHeader) == 5);
```
klecki, line 49: Nice idea with the typed structures. Do you think we can (or maybe should) add a comment, that shows the layout of the expected data in the docstring somewhere? Specifically to signify that those structures are packed one after another.

Something like:
```
Header layout:

RiffHeader|vp8_header_t|<WebpLossyHeader/WebpLosslessHeader>

The WebpLossyHeader and WebpLosslessHeader are selected based on the "VP8 " and "VP8L" in vp8_header_t.
"VP8X" in vp8_header_t signifies WebP extended file format which is not supported.
```

klecki (approved):
Cool idea to extract the definitions and common code.

dali/imgcodec/parsers/webp.cc
staniewzki: Seems like a good idea

staniewzki:
!build

dali/imgcodec/parsers/webp_test.cc
```diff
@@ -0,0 +1,129 @@
...
+namespace imgcodec {
+namespace test {
+
+class WebpParserTest : public ::testing::Test {
+ private:
+  WebpParser parser_;
+  const std::string webp_directory_ = testing::dali_extra_path() + "/db/single/webp/";
+
+ public:
+  WebpParserTest() {}
```
szalpal, line 32: Is the explicit ctor necessary? Could we get rid of it?

dali/imgcodec/parsers/webp_test.cc
```diff
@@ -0,0 +1,129 @@
...
+
+  std::vector<uint8_t> ReadFile(const std::string &filename) {
+    std::ifstream stream(webp_directory_ + filename, std::ios::binary);
+    return std::vector<uint8_t>(std::istreambuf_iterator<char>(stream),
+                                std::istreambuf_iterator<char>());
```
szalpal, line 37: How about?
```suggestion
    return {std::istreambuf_iterator<char>(stream), std::istreambuf_iterator<char>()};
```

szalpal (commented):
I added couple of questions, please refer to them.

dali/imgcodec/parsers/webp.cc
```diff
@@ -12,19 +13,155 @@
...
+
+template<size_t N>
+bool is_pattern_matching(const std::array<uint8_t, N> data, const char pattern[N + 1]) {
```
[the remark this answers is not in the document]
staniewzki: The point is to have pattern as a valid C-string with length `N` so including the null byte we get `N + 1`

dali/imgcodec/parsers/webp.cc
```diff
@@ -1,3 +1,4 @@
+// Copyright (c) 2021, Lennart Behme. All rights reserved.
```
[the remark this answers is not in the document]
staniewzki: The code here is based on the [previous implementation](https://github.com/NVIDIA/DALI/blob/main/dali/image/webp.cc), with special copyright.

dali/imgcodec/parsers/webp_test.cc
staniewzki: No it is not, I can remove it.

dali/imgcodec/parsers/webp_test.cc
```diff
@@ -0,0 +1,129 @@
...
+    auto img = ImageSource::FromFilename(webp_directory_ + filename);
+    EXPECT_TRUE(parser_.CanParse(&img));
+    EXPECT_EQ(parser_.GetInfo(&img).shape, expected_shape);
+  }
+};
+
+TEST_F(WebpParserTest, ValidWebpLossy) {
+  TestValidImage("lossy/kitty-2948404_640.webp", {433, 640, 3});
+  TestValidImage("lossy/domestic-cat-726989_640.webp", {426, 640, 3});
+  TestValidImage("lossy/cat-300572_640.webp", {536, 640, 3});
```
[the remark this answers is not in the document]
staniewzki: Well it does seem like a nice idea, but then it should be probably implemented in some common parser test class. Not sure how that relates to https://github.com/NVIDIA/DALI/pull/4058.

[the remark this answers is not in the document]
szkarpinski: Or https://github.com/NVIDIA/DALI/pull/4057 , which tests the parser against the old implementation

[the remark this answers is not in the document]
mzient: Hardcoded reference is preferred over relying on a third party library. We've seen with librosa what can happen.

dali/imgcodec/parsers/webp.cc
```diff
@@ -12,19 +13,155 @@
...
+  uint8_t signature_byte;
+};
+static_assert(sizeof(WebpLosslessHeader) == 5);
+
+template<size_t N>
+bool is_pattern_matching(const std::array<uint8_t, N> data, const char pattern[N + 1]) {
+  for (size_t i = 0; i < N; i++)
+    if (data[i] != pattern[i])
+      return false;
+  return true;
```
mzient, line 66: return std::strncmp(data.data(), pattern, N) == 0;

dali/imgcodec/parsers/webp.cc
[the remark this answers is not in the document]
mzient: Alternatively, you can write a function that will populate an array with the context of a string literal:
```C++
template <size_t N>
std::array<char, N-1> tag(const char (&c)[N]) {
    std::array<char, N-1> a={};
    std::copy(&c[0], &c[N-1], a.begin());
    return a;
}
```

Now, you can just do:
```
  if (header.riff_text == tag("RIFF"))
```

dali/imgcodec/parsers/webp.cc
```diff
@@ -12,19 +13,155 @@
...
+  std::array<uint8_t, 4> webp_text;
+};
+static_assert(sizeof(RiffHeader) == 12);
+
+using vp8_header_t = std::array<uint8_t, 4>;
+
+// Simple file format (lossy)
+// https://datatracker.ietf.org/doc/html/rfc6386#section-9.1
+struct WebpLossyHeader {
+  std::array<uint8_t, 4> chunk_size;
```
mzient, line 47:
```suggestion
#pragma pack(push, 1)
struct WebpLossyHeader {
  uint32_t chunk_size;
```

and adjust endianness (if needed) upon reading?

dali/imgcodec/parsers/webp.cc
```diff
@@ -12,19 +13,155 @@
...
+// Header layout:
+// RiffHeader | vp8_header_t | <WebpLossyHeader/WebpLosslessHeader>
+//
+// The WebpLossyHeader and WebpLosslessHeader are selected based on a identifier,
+// respectively "VP8 " and "VP8L" in vp8_header_t. "VP8X" in vp8_header_t signifies
+// WebP extended file format, which is not supported.
+
+struct RiffHeader {
+  std::array<uint8_t, 4> riff_text;
+  std::array<uint8_t, 4> file_size;
```
mzient, line 37:
```suggestion
  uint32_t file_size;
```

dali/imgcodec/parsers/webp.cc
```diff
@@ -12,19 +13,157 @@
...
+  uint32_t chunk_size;
+  uint8_t signature_byte;
+};
+static_assert(sizeof(WebpLosslessHeader) == 5);
+
+#pragma pack(pop)  // end of packing scope
+
+template<size_t N>
+constexpr std::array<uint8_t, N - 1> tag(const char (&c)[N]) {
+  std::array<uint8_t, N - 1> a;
```
mzient, line 68:
```suggestion
  std::array<uint8_t, N - 1> a{};
```
Otherwise it's not constexpr according to gcc 7.3.

dali/imgcodec/parsers/webp.cc
```diff
@@ -12,19 +13,157 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+#include <string>
```
mzient, line 16: For `std::copy`:
```suggestion
#include <algorithm>
#include <string>
```

dali/imgcodec/parsers/webp.cc
```diff
@@ -12,19 +13,158 @@
...
+
 ImageInfo WebpParser::GetInfo(ImageSource *encoded) const {
+  auto stream = encoded->Open();
   ImageInfo info;
-  // info.shape = {h, w, c};   // TODO(janton)
+
+  stream->Skip<RiffHeader>();
+  const auto vp8_header = stream->ReadOne<vp8_header_t>();
+  if (is_simple_lossy_format(vp8_header)) {
+    const auto lossy_header = stream->ReadOne<WebpLossyHeader>();
```
mzient, line 115:
```suggestion
    auto lossy_header = stream->ReadOne<WebpLossyHeader>();
    from_little_endian(lossy_header);
```

dali/imgcodec/parsers/webp.cc
```diff
@@ -12,19 +13,158 @@
...
+                sequence_of_integers(lossy_header.sync_code) + " instead");
+    }
+
+    const int w = ReadValueLE<uint16_t>(*stream) & 0x3FFF;
+    const int h = ReadValueLE<uint16_t>(*stream) & 0x3FFF;
+
+    // VP8 always uses RGB
+    info.shape = {h, w, 3};
+  } else if (is_simple_lossless_format(vp8_header)) {
+    const auto lossless_header = stream->ReadOne<WebpLosslessHeader>();
```
mzient, line 127:
```suggestion
    auto lossless_header = stream->ReadOne<WebpLosslessHeader>();
    from_little_endian(lossless_header);
```

dali/imgcodec/parsers/webp.cc
```diff
@@ -12,19 +13,158 @@
...
 
 bool WebpParser::CanParse(ImageSource *encoded) const {
-  return false;  // TODO(janton)
+  static_assert(sizeof(WebpLossyHeader) > sizeof(WebpLosslessHeader));
+  uint8_t data[sizeof(RiffHeader) + sizeof(vp8_header_t) + sizeof(WebpLossyHeader)];
+  if (!ReadHeader(data, encoded, sizeof(data)))
+    return false;
+  MemInputStream stream(data, sizeof(data));
+
+  if (!is_valid_riff_header(stream.ReadOne<RiffHeader>()))
```
mzient, line 157: It's not perfectly kosher, but since we're checking only the tag, it should be OK without endianness fixup.

dali/imgcodec/parsers/webp.cc
staniewzki: I planned on adding this to the next [PR](https://github.com/staniewzki/DALI/tree/webp-exif) where I will actually use the chunk_size. Doing this now would involve a lot of rebasing and recompiling.

dali/imgcodec/parsers/webp.cc
```diff
@@ -12,19 +13,158 @@
...
+bool is_sync_code_valid(const WebpLossyHeader &header) {
+  return header.sync_code == std::array<uint8_t, 3>{0x9D, 0x01, 0x2A};
+}
+
+bool is_sync_code_valid(const WebpLosslessHeader &header) {
+  return header.signature_byte == 0x2F;
+}
+
+template<size_t N>
+std::string sequence_of_integers(const std::array<uint8_t, N> &data) {
```
mzient, line 99: Hmm.. for me this is "to_string" or something along these lines.

dali/imgcodec/parsers/webp.cc
mzient: ACK

mzient (approved):

staniewzki:
!build

szalpal (approved):

PR #4063, opened 2022-07-14
Extract EXIF parser from OpenCV

havaker:
## Category:

## Description:

Many image formats may contain EXIF metadata. To support parsing this metadata I've added an EXIF parser from OpenCV to our `third_party/` directory.

Added a new module to the `third_party` directory.

### Tests:

- [ ] Existing tests apply
- [ ] New tests added
  - [ ] Python tests
  - [ ] GTests
  - [ ] Benchmark
  - [ ] Other
- [x] N/A

### Documentation
- [ ] Documentation updated
  - [ ] Docstring
  - [ ] Doxygen
  - [ ] RST
  - [ ] Jupyter
  - [ ] Other
- [x] N/A

#### Requirements
- [x] N/A

**REQ IDs**: N/A

**JIRA TASK**: N/A

### DIFF
--- a/third_party/opencv/exif/exif.cc
+++ b/third_party/opencv/exif/exif.cc
@@ -1,0 +1,553 @@
+// Copyright (c) 2022, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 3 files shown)


third_party/opencv/exif/exif.cc
```diff
@@ -0,0 +1,553 @@
+// Copyright (c) 2022, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
JanuszL, line 1: I'm not sure if we can add our license on top. I would at most extend the copyright of the original file.

JanuszL: If it is more than just a copy paste then it should be fine to do it this way.

mzient: The libtiff decoder has been substantially altered. Here, we're barely introducing some mechanical changes. I think the changes we've introduced don't warrant sublicensing.

havaker: In that case, I will remove our license blurb.

havaker:
!build

jantonguirao (approved):

mzient (approved):

PR #4069, opened 2022-07-18
Add utilities for swapping endianness.

mzient:

## Category:

## Additional information:
Adds a set of functions for handling endiannes - swapping, converting fixed endiannes to/from host encoding.
The user can list struct fields to be swapped.
C Arrays, std::array, std::pair and std::tuple are handled out-of-box.

### Tests:

- [ ] Existing tests apply
- [X] New tests added
  - [ ] Python tests
  - [X] GTests
  - [ ] Benchmark
  - [ ] Other
- [ ] N/A

### Documentation
- [ ] Documentation updated
  - [ ] Docstring
  - [x] Doxygen
  - [ ] RST
  - [ ] Jupyter
  - [ ] Other
- [ ] N/A

#### Requirements
- [ ] N/A

**REQ IDs**: N/A

**JIRA TASK**: N/A

### DIFF
--- a/dali/core/endian_util_test.cc
+++ b/dali/core/endian_util_test.cc
@@ -1,0 +1,65 @@
+// Copyright (c) 2022, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include <gtest/gtest.h>
+#include <dali/core/endian_util.h>
+#include <dali/core/int_literals.h>
+
+namespace dali {
+
+TEST(EndianUtilTest, Fundamental) {
+  int32_t a = 0x12345678;
+  swap_endian(a);
+  EXPECT_EQ(a, 0x78563412);
+  uint64_t b = 0x0123456789abcdef_u64;
+  swap_endian(b);
+  EXPECT_EQ(b, 0xefcdab8967452301_u64);
+  char c = 'x';
+  EXPECT_EQ(c, 'x');
+  uint32_t f_ieee_swapped = 0x0000803f;
+  float f = *(const float *)&f_ieee_swapped;
+  swap_endian(f);
+  EXPECT_EQ(f, 1);
+}
+
+namespace {
+
+struct Inner {
+  int16_t x, y;
+};
+
+struct Outer {
+  Inner inner;
+  int z;
+};
+
+}  // namespace
+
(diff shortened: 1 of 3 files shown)


dali/core/endian_util_test.cc
```diff
@@ -0,0 +1,83 @@
...
+
+struct Outer {
+  Inner inner;
+  int a[2];
+  std::array<int, 2> b;
+};
+
+}  // namespace
+
+SWAP_ENDIAN_FIELDS(Inner, x, y);
```
szkarpinski, line 50: I understand that preprocessor magic has its own rules, but it was not obvious to me what this does. The name suggests that it will "swap endian fields", but it *creates a function which can be used to* swap endian fields. Maybe some short doxygen comment would help?

third_party/cpplint.py
```diff
@@ -2973,7 +2973,7 @@ def CheckSpacingForFunctionCall(filename, clean_lines, linenum, error):
   # Note that we assume the contents of [] to be short enough that
   # they'll never need to wrap.
   if (  # Ignore control structures.
-      not Search(r'\b(if|for|while|switch|return|new|delete|catch|sizeof)\b',
+      not Search(r'\b(if|for|while|switch|return|new|delete|catch|sizeof|elif)\b',
```
mzient, line 2976: Linter comlained about:
```
#elif (condition)
```
but not
```#if (condition)```
The downside of this change is that it will now be allowed to have a function called `elif` and a space before opening parenthesis - but I don't think it's very likely to ever happen.

File: dali/core/endian_util_test.cc (hunk 1 above)
Comment (mzient) [reply dali/core/endian_util_test.cc:50]: I've added documentation.

**Review by szkarpinski — APPROVED** [review]

File: include/dali/core/endian_util.h [hunk 3]
```diff
@@ -0,0 +1,214 @@
...
+/**
+ * @brief Converts (in place) the object from big endian
+ *
+ * If the host system is big endian, this function is a no-op.
+ */
+template <typename T>
+void from_big_endian(T &t) {
+  if (is_little_endian)
+    swap_endian(t);
+}
```
awolant, line 210: Can you explain a bit a rationale behind designing this API like this? 

As far as I understand, `from_little_endian` and `to_big_endian` are doing the same thing, but I need to know the device "endianness" to know, which of these I need to choose to work. And if I choose incorrectly, there will be a silent no-op, regardless of the "endianness" of the argument. Is this understanding correct?
It looks like it assumes, that the argument has the "endianness" of the system, but it is not clear to me why. Are there some intangibles I am missing? After all, this functionality enables you to change precisely that.

mzient: The rationale is as follows:
If you obtain a raw structure in a particular (known and fixed) byte order, you call `from_that_order`.
Conversely, if you have a structure in host order and want to send/write/... it in a fixed byte order, you call `to_that_order`.
Internally, `from_` and `to_` do the same thing - just like POSIX `htonl` and `ntohl`. This wouldn't hold for "middle endian", but I don't think we plan to target [PDP-11](https://en.wikipedia.org/wiki/PDP-11) ;)

awolant: Ok. I suggest adding this info to the documentation for these functions. Right now, they say "convert in place" but it looks like intention was different.

awolant (approved):

PR #4071, opened 2022-07-18
Add restrictions to set_type

klecki:
## Description:
Disallow using set_type to change the underlying
type of existing allocation in the buffer. 
It was possible to cause a new allocation to happen
this way (when changing to bigger type), and it is
now disallowed.
Type can still be changed when explicitly requesting
both new shape and new type via Resize.
The type changing behaviour can be found only in tests.

TensorVector, Buffer and subclasses.

### Tests:

- [x] Existing tests apply
    - the tests for changing types were adjusted to expect error.
- [x] New tests added
  - [ ] Python tests
  - [x] GTests
  - [ ] Benchmark
  - [ ] Other
- [ ] N/A

### Documentation
- [x] Documentation updated
  - [x] Docstring
  - [ ] Doxygen
  - [ ] RST
  - [ ] Jupyter
  - [ ] Other
- [ ] N/A

#### Requirements
- [ ] N/A

**REQ IDs**: N/A

**JIRA TASK**: N/A

### DIFF
--- a/dali/pipeline/data/buffer.h
+++ b/dali/pipeline/data/buffer.h
@@ -322,6 +322,12 @@ class DLL_PUBLIC Buffer {
    */
   inline void set_type(const DALIDataType new_type_id) {
     DALI_ENFORCE(new_type_id != DALI_NO_TYPE, "new_type must be valid type.");
+    DALI_ENFORCE(type_.id() == new_type_id || type_.id() == DALI_NO_TYPE,
+                 make_string("set_type cannot be used to change the current type - it is not "
+                             "allowed to cause allocations. Currently set type: '",
+                             type_.id(), "' trying to set: '", new_type_id,
+                             "'. You may change the current type using Resize or by"
+                             " calling Reset first."));
     if (new_type_id == type_.id()) return;
     const TypeInfo &new_type = TypeTable::GetTypeInfo(new_type_id);
 
--- a/dali/pipeline/data/buffer.h
+++ b/dali/pipeline/data/buffer.h
@@ -320,7 +320,7 @@ class DLL_PUBLIC Buffer {
    */
   inline void set_type(const DALIDataType new_type_id) {
     DALI_ENFORCE(new_type_id != DALI_NO_TYPE, "new_type must be valid type.");
-    DALI_ENFORCE(type_.id() == new_type_id || !has_data(),
+    DALI_ENFORCE(type_.id() == new_type_id || (!has_data() || type_.id() == DALI_NO_TYPE),
                  make_string("set_type cannot be used to change the current type - it is not "
                              "allowed to cause allocations. Currently set type: '",
                              type_.id(), "' trying to set: '", new_type_id,
--- a/dali/pipeline/data/buffer.h
+++ b/dali/pipeline/data/buffer.h
@@ -309,20 +309,18 @@ class DLL_PUBLIC Buffer {
     order_ = order;
   }
 
-  /**
-   * @brief Sets the type of the buffer. If the buffer has not been
-   * allocated because it does not yet have a type, the calling type
-   * is taken to be the type of the data and the memory is allocated.
+  /** @defgroup SetTypeFunctions Type setting functions.
    *
-   * If the buffer already has a valid type, and the calling type does
-   * not match, the type of the buffer is reset and the underlying
-   * storage is re-allocated if the buffer does not currently own
-   * enough memory to store the current number of elements with the
-   * new data type.
+   * Set the type of the current buffer. It must be set before calling resize(size).
+   * It cannot be used to change the type after allocation happened.
+   *
+   * resize(size, new_type) can be used without prior set_type call or to request a different type
+   * after allocation.
+   * @{
    */
   inline void set_type(const DALIDataType new_type_id) {
     DALI_ENFORCE(new_type_id != DALI_NO_TYPE, "new_type must be valid type.");
-    DALI_ENFORCE(type_.id() == new_type_id || type_.id() == DALI_NO_TYPE,
+    DALI_ENFORCE(type_.id() == new_type_id || !has_data(),
                  make_string("set_type cannot be used to change the current type - it is not "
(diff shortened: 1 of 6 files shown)


klecki:
!build

dali/pipeline/data/buffer.h
```diff
@@ -322,6 +322,12 @@ class DLL_PUBLIC Buffer {
    */
   inline void set_type(const DALIDataType new_type_id) {
     DALI_ENFORCE(new_type_id != DALI_NO_TYPE, "new_type must be valid type.");
+    DALI_ENFORCE(type_.id() == new_type_id || type_.id() == DALI_NO_TYPE,
```
mzient, line 325: Isn't it too strict? If it's about reallocations / changing the size in bytes, then the following would suffice:
```suggestion
    DALI_ENFORCE(type_.id() == new_type_id || size() == 0,
```

klecki: Maybe a bit. Still, with the fact that we keep the Resize(shape, type) allowed, removing the set_type doesn't change that much, but maybe with the resize is just a case of naming or introducing another member function for that purpose, which we probably won't do soon.

dali/pipeline/data/buffer.h
```diff
@@ -309,19 +309,23 @@ class DLL_PUBLIC Buffer {
...
    */
   inline void set_type(const DALIDataType new_type_id) {
     DALI_ENFORCE(new_type_id != DALI_NO_TYPE, "new_type must be valid type.");
+    DALI_ENFORCE(type_.id() == new_type_id || (!has_data() || type_.id() == DALI_NO_TYPE),
+                 make_string("set_type cannot be used to change the current type - it is not "
+                             "allowed to cause allocations. Currently set type: '",
+                             type_.id(), "' trying to set: '", new_type_id,
+                             "'. You may change the current type using Resize or by"
+                             " calling Reset first."));
     if (new_type_id == type_.id()) return;
```
mzient, line 329: How about that? Now you're checking `new_type_id == type_.id()` twice.
```suggestion
    if (new_type_id == type_.id()) return;
    DALI_ENFORCE(!has_data() || type_.id() == DALI_NO_TYPE,
                 make_string("set_type cannot be used to change the current type - it is not "
                             "allowed to cause allocations. Currently set type: '",
                             type_.id(), "' trying to set: '", new_type_id,
                             "'. You may change the current type using Resize or by"
                             " calling Reset first."));
```

klecki: Sure

klecki:
!build

dali/pipeline/data/tensor_list.h
```diff
@@ -646,24 +646,24 @@ class DLL_PUBLIC TensorList {
     return {data_.capacity()};
   }
 
-  /**
-   * @brief Set the type of the TensorList. The type needs to be set before calling
-   * the Resize function that gives the shape. Type can be changed, if the current storage
-   * is not big enough, the memory will be reallocated.
+  /** @defgroup SetTypeFunctions Type setting functions.
```
mzient, line 649: I'm terribly sorry to disappoint you, but this doesn't work - visit the generated documentation page and see that `set_type` is now totally undocumented.

mzient (approved):
+ nitpick about using `@defgroup` which apparently doesn't work.

dali/pipeline/data/tensor_list.h
```diff
@@ -646,24 +646,24 @@ class DLL_PUBLIC TensorList {
...
+   *
+   * Set the type of the current batch. It must be set before calling Resize(shape).
+   * It cannot be used to change the type after allocation happened.
+   *
+   * Resize(shape, new_type) can be used without prior set_type call or to request a different type
+   * after allocation.
+   * @{
    */
```
mzient, line 657: The proper way:
```suggestion
  ///@{
  /**
   * Set the type of the current batch. It must be set before calling Resize(shape).
   * It cannot be used to change the type after allocation happened.
   *
   * Resize(shape, new_type) can be used without prior set_type call or to request a different type
   * after allocation.
   */
```
AFAIK `///@{ ` needs to be on its own line or in its own comment block.

dali/pipeline/data/tensor_list.h
```diff
@@ -646,24 +646,24 @@ class DLL_PUBLIC TensorList {
...
-  /**
-   * @brief Set the type of the TensorList. The type needs to be set before calling
-   * the Resize function that gives the shape. Type can be changed, if the current storage
-   * is not big enough, the memory will be reallocated.
-   */
   template <typename T>
   inline void set_type() {
     data_.set_type(TypeTable::GetTypeId<T>());
   }
+  /** @} */  // end of SetTypeFunctions
```
mzient, line 666:
```suggestion
  ///@}
```
Block comments for single line are a retreat to the '70s...

szalpal (approved):

dali/pipeline/data/tensor_list.h
klecki: Fixed, should be ok now

dali/pipeline/data/buffer.h
```diff
@@ -309,20 +309,25 @@ class DLL_PUBLIC Buffer {
...
+   * It cannot be used to change the type after allocation happened.
    *
-   * If the buffer already has a valid type, and the calling type does
-   * not match, the type of the buffer is reset and the underlying
-   * storage is re-allocated if the buffer does not currently own
-   * enough memory to store the current number of elements with the
-   * new data type.
+   * resize(size, new_type) can be used without prior set_type call or to request a different type
+   * after allocation.
+   * @{
```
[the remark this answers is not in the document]
klecki: I had do adjust doxygen config, I did it as a separate PR

klecki:
!build

dali/pipeline/data/tensor_list.h
klecki: And consistent use of doxygen in our codebase.

PR #4083, opened 2022-07-21
Add boilerplate synchronization for batch copying

klecki:
## Description:
Add helper functions that organize the copying
of batches.
Add functions for 3-way sync on the orders
based on the TensorList implementation.
Add implementation for the 4 cases of 
[non]contiguous x [non]contiguous.
Rework the TensorVector::Copy to use the
introduced functions.
This change prepares the TensorVector
for the next step of removing the internal
usage of TensorList as a member.
Add tests.

TensorVector::Copy

### Tests:

- [ ] Existing tests apply
- [x] New tests added
  - [ ] Python tests
  - [x] GTests
  - [ ] Benchmark
  - [ ] Other
- [ ] N/A

### Documentation
- [x] Documentation updated
  - [x] Docstring
  - [ ] Doxygen
  - [ ] RST
  - [ ] Jupyter
  - [ ] Other
- [ ] N/A

#### Requirements
- [ ] N/A

**REQ IDs**: N/A

**JIRA TASK**: N/A


klecki:
!build

klecki:
!build

dali/pipeline/data/tensor_vector.cc
```diff
@@ -13,12 +13,163 @@
...
+
+/**
+ * @brief Copy from non-contiguous batch to contiguous destination.
+ * Assumes matching shapes and types.
+ */
+template <typename DstBackend, typename SrcBackend, template <typename> typename SrcBatch>
+void CopySamplewiseImpl(void *dst, const SrcBatch<SrcBackend> &src, const TypeInfo &type_info,
+                        AccessOrder order, bool use_copy_kernel = false) {
+  auto num_samples = src.num_samples();
+  SmallVector<const void *, 256> srcs;
```
prak-nv, line 132: Why 256?

klecki: Hmm, this is a constant that is used for such copies throughout of our codebase. Maybe I can make it some constant that is defined once.

klecki:
!build

klecki:
!build

klecki:
!build

dali/pipeline/data/tensor_vector.cc
```diff
@@ -507,26 +658,62 @@ void TensorVector<Backend>::Reset() {
...
+  copy_impl::SyncAfter(this->order(), copy_order);
 
   resize_tensors(tl_->num_samples());
   UpdateViews();
+
+  // Update the layout and other metadata
+  SetLayout(in_tl.GetLayout());
+  for (int i = 0; i < curr_num_tensors_; i++) {
+    SetMeta(i, in_tl.GetMeta(i));
+  }
```
mzient, line 690: Doesn't UpdateViews do it?

dali/pipeline/data/tensor_vector.cc
```diff
@@ -13,13 +13,164 @@
...
+  SmallVector<const void *, kMaxStaticCopyBatchSize> srcs;
+  srcs.reserve(num_samples);
+  SmallVector<void *, kMaxStaticCopyBatchSize> dsts;
+  dsts.reserve(num_samples);
+  SmallVector<Index, kMaxStaticCopyBatchSize> sizes;
+  sizes.reserve(num_samples);
+  for (int i = 0; i < num_samples; i++) {
+    dsts.emplace_back(dst.raw_mutable_tensor(i));
+    srcs.emplace_back(src.raw_tensor(i));
+    sizes.emplace_back(src.shape()[i].num_elements());
```
mzient, line 95:
```suggestion
    dsts.push_back(dst.raw_mutable_tensor(i));
    srcs.push_back(src.raw_tensor(i));
    sizes.push_back(src.shape()[i].num_elements());
```
We're not constructing anything.

dali/pipeline/data/types.h
```diff
@@ -305,6 +306,16 @@ struct id2type_helper<id> { using type = data_type; };
...
 struct NoType {};
 
+
+/**
+ * @brief Size of stack-based array used to prepare the pointers and other parameters for
+ * the TypeInfo::Copy.
+ * For bigger batches, the list of pointers/sizes would be stored as dynamic allocation
+ * (SmallVector), used in a common pattern where we have a copy from or to a batch of samples.
+ */
+constexpr size_t kMaxStaticCopyBatchSize = 256;
```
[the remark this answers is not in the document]
klecki: Does it really fit into core?

dali/pipeline/data/tensor_vector.cc
klecki: Not really, it just copies over the metadata from internal tl_ to the tensors_.

dali/pipeline/data/tensor_vector.cc
```diff
@@ -507,26 +658,62 @@ void TensorVector<Backend>::Reset() {
...
+  // Update the metadata
   type_ = in_tl.type_info();
   sample_dim_ = in_tl.shape().sample_dim();
-  tl_->Copy(in_tl, order);
+
+  // Here both batches are contiguous
+  type_info().template Copy<Backend, SrcBackend>(
+      unsafe_raw_mutable_data(*tl_), unsafe_raw_data(in_tl), in_tl.shape().num_elements(),
+      copy_order.stream(), false);
+  copy_impl::SyncAfter(this->order(), copy_order);
```
mzient, line 681: What's the point of this exercise? Is it to get rid of `tl_->Copy` because there will be no more tl_ soon?

klecki: Yes. Also, this overload will be gone, so I can probably keep it in the old state, I just used similar code in both variants. I can also call the `CopyImpl` here to hide the details if you prefer.

dali/pipeline/data/types.h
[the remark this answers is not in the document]
prak-nv: How about go deeper - with template alias ``template <typename T> using BatchVector = SmallVector<T, kMaxStaticCopyBatchSize>``. Dunno about core;

dali/pipeline/data/tensor_vector.cc
```diff
@@ -13,13 +13,164 @@
...
+  dsts.reserve(num_samples);
+  SmallVector<Index, kMaxStaticCopyBatchSize> sizes;
+  sizes.reserve(num_samples);
+  for (int i = 0; i < num_samples; i++) {
+    dsts.emplace_back(dst.raw_mutable_tensor(i));
+    srcs.emplace_back(src.raw_tensor(i));
+    sizes.emplace_back(src.shape()[i].num_elements());
+  }
+
+  type_info.Copy<SrcBackend, DstBackend>(dsts.data(), srcs.data(), sizes.data(), num_samples,
```
prak-nv, line 98: This really could use ``struct CopyOperation<SrcBackend, DstBackend>`` aggregating all those vectors elements; ie. `SOA -> AOS`.
-would improve data locality
- cut number of allocations;
- also 'srcContinous(); dstContinous()` are 2 bits that could be easily converted to `enum`;
It'd be easy to add Factory of Commands for those 4 cases.

dali/pipeline/data/types.h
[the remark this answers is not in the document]
klecki: I was also thinking about the preconfigured alias, that seems like a good idea.

dali/pipeline/data/tensor_vector.cc
mzient: So, it _does_ it, and, again, you just want to pretend that tl_ doesn't exist....?

dali/pipeline/data/tensor_vector.cc
klecki: We already had it that way in TensorList, but I can adjust here.

dali/pipeline/data/tensor_vector.cc
klecki: Ok, whatever, I will revert this to the previous variant and remove it in the next PR.

dali/pipeline/data/tensor_vector.cc
klecki: reverted.

klecki:
!build

dali/pipeline/data/tensor_vector.cc
klecki: On the second thought, I've got a bit too forward with how it will look.  As there will be intermediate step before removing TL, this change is useful as it prepares the code for the lack of TL as a member of TV

klecki:
!build

dali/pipeline/data/tensor_vector.cc
prak-nv: After OTR discussion, I will create task to refactor this code to use Command design pattern and add it to the backlog.

include/dali/core/small_vector.h
```diff
@@ -730,6 +730,23 @@ class SmallVector : SmallVectorAlloc<T, allocator>, SmallVectorBase<T> {
...
+
+/**
+ * @brief SmallVector alias used when dealing with batches of data in common operations
+ *
+ * The static stack allocation size is adjusted for that purpose.
+ */
+template <typename T>
+using BatchVector = SmallVector<T, kMaxStaticCopyBatchSize>;
+
+
```
mzient, line 749: Sorry, but this violates separation of concerns. How about putting it in `tensor_list.h` or `tensor_vector.h`? So far these are only used in TensorList and CopyToExternal (which uses TensorList).

dali/pipeline/data/tensor_list_as_tensor_vector_test.cc
```diff
@@ -0,0 +1,182 @@
...
+    return shape;
+  }
+
+  /**
+   * Initialize & check a TensorVector based on an input shape
+   * Allocate it as float
+   */
+  void SetupTensorVector(TensorVector<Backend> *tensor_list,
+                       const TensorListShape<>& shape,
+                       vector<Index> *offsets) {
```
mzient, line 92: Nitpick
```suggestion
                         const TensorListShape<>& shape,
                         vector<Index> *offsets) {
```

mzient, line 92: I think we've abandoned using pointers instead of references for non-nullable output parameters.
```suggestion
  void SetupTensorVector(TensorVector<Backend> &tensor_list,
                         vector<Index> &offsets,
                         const TensorListShape<> &shape) {
```

klecki: As the comment states:
>  The section below is copied from tensor_list_test.cc with the only changes being:
>  * Rename all occurrences of TensorList to TensorVector
I have the full test file ported in another PR, but here I just copied the relevant tests.

klecki: aligned

klecki:
!build

include/dali/core/small_vector.h
klecki: done

prak-nv (approved):

mzient (approved):

klecki:
!build

PR #3574, opened 2021-12-16
Check when using ES source and feed_input

ksztenderski:
## Description
- [ ] **Bug fix** (*non-breaking change which fixes an issue*)
- [ ] **New feature** (*non-breaking change which adds functionality*)
- [ ] **Breaking change** (*fix or feature that would cause existing functionality to not work as expected*)
- [ ] **Refactoring** (*Redesign of existing code that doesn't affect functionality*)
- [ ] **Other** (*e.g. Documentation, Tests, Configuration*)

### What happened in this PR

Add raising exception when using feed_input on an ExternalSource operator that has been
initialized with source parameter.

#### Additional information
- Affected modules and functionalities:

- Key points relevant for the review:

### Tests
- [ ] Existing tests apply
- [ ] New tests added
  - [ ] Python tests
  - [ ] GTests
  - [ ] Benchmark
  - [ ] Other
- [ ] N/A

### Documentation
- [ ] Existing documentation applies
- [ ] Documentation updated
  - [ ] Docstring
  - [ ] Doxygen
  - [ ] RST
  - [ ] Jupyter
  - [ ] Other
- [ ] N/A

#### Requirements
- [ ] Implements new requirements
- [ ] Affects existing requirements
- [ ] N/A

**REQ IDs**: N/A

**JIRA TASK**: N/A

PR #4096, opened 2022-07-25
Fix feed_input after deserialization

klecki:
## Description:
Guard against a check using the Python
graph that is not constructed during
deserialization. 

Added a private property that signalizes
deserialization which is used to skip the checks.

Check was added in #3574

pipeline.py

### Tests:

- [ ] Existing tests apply
- [x] New tests added
  - [x] Python tests
  - [ ] GTests
  - [ ] Benchmark
  - [ ] Other
- [ ] N/A

### Documentation
- [ ] Documentation updated
  - [ ] Docstring
  - [ ] Doxygen
  - [ ] RST
  - [ ] Jupyter
  - [ ] Other
- [x] N/A

#### Requirements
- [ ] N/A

**REQ IDs**: N/A

**JIRA TASK**: N/A

### DIFF
--- a/dali/python/nvidia/dali/pipeline.py
+++ b/dali/python/nvidia/dali/pipeline.py
@@ -209,6 +209,7 @@ Parameters
         self._py_pool_started = False
         self._backend_prepared = False
         self._built = False
+        self._deserialized = False  # Marked True when deserializing
         self._first_iter = True
         self._last_iter = False
         self._iter = 0
@@ -907,11 +908,13 @@ Parameters
             name = data_node.name
 
         # Check for use of feed_input on an external_source operator that was
-        # initialized with 'source'.
-        if next((op._callback is not None for op in self._ops if op.name == name), False):
-            raise RuntimeError(
-                f"Cannot use `feed_input` on the external source '{name}' with a `source`"
-                " argument specified.")
+        # initialized with 'source'. This check makes sense only for fully Python-based
+        # pipelines, and not deserialized ones
+        if not self._deserialized:
+            if next((op._callback is not None for op in self._ops if op.name == name), False):
+                raise RuntimeError(
+                    f"Cannot use `feed_input` on the external source '{name}' with a `source`"
+                    " argument specified.")
 
         self._feed_input(name, data, layout, cuda_stream, use_copy_kernel)
 
@@ -1226,6 +1229,7 @@ Parameters
         pipeline._backend_prepared = True
         pipeline._pipe.Build()
         pipeline._built = True
+        pipeline._deserialized = True
         return pipeline
 
     def deserialize_and_build(self, serialized_pipeline):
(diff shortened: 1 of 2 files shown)


klecki:
!build

dali/python/nvidia/dali/pipeline.py
```diff
@@ -209,6 +209,7 @@ def __init__(self,
         self._py_pool_started = False
         self._backend_prepared = False
         self._built = False
+        self._deserialized = False  # Marked True when deserializing
```
mzient, line 212: This is too much of an assumption. There can be a legitimate sourceless external source that is not deserialized (this is actually used in tests). Something like `has_source` or simply checking the original `source` would be more adequate.

klecki: I don't follow? If we don't deserialize, the check works as intended, as we build the pipeline via frontend, and all the members there are valid. And this is just a flag that signifies that this pipeline was deserialized, I added one, instead of relying on checking the `not self._py_graph_built and self._backend_prepared and self_built` to check if we are in deserialized pipeline - as that might not be enough.

dali/python/nvidia/dali/pipeline.py
```diff
@@ -907,11 +908,13 @@ def feed_input(self, data_node, data, layout=None, cuda_stream=None, use_copy_ke
...
 
         # Check for use of feed_input on an external_source operator that was
-        # initialized with 'source'.
-        if next((op._callback is not None for op in self._ops if op.name == name), False):
-            raise RuntimeError(
-                f"Cannot use `feed_input` on the external source '{name}' with a `source`"
-                " argument specified.")
+        # initialized with 'source'. This check makes sense only for fully Python-based
+        # pipelines, and not deserialized ones
+        if not self._deserialized:
```
mzient, line 913: I don't like the idea of making deserialized pipelines somehow worse and cutting them off some capabilities "because we know they can't" instead of checking for actual presence of absence of a given capability.
I'd go with:
```suggestion
        if self._ops is not None:
```
or even combining this line with the one below into one property `has_source` - because ultimately that's what we want to know. The fact that a serialized pipeline _currently_ doesn't store source callbacks seems coincidental - and we could change that with relative ease by pickling the source.

klecki: My intention here was to partially revert the #3574 for the deserialized pipelines - with such check we can basically mark the problematic places and adjust them when we have proper functionality in place - instead of going into implementation details for it. 
`has_source` would be nice, but again, that just pushes the check further into DALI.

banasraf: For me, both solutions make equal sense. 
 I understand the reasoning behind Michal's solution and `if self._ops is not None:` looks good. 
 Also, I like the argument about using `._deserialized` which kind of documents that the problem is connected to the deserialization and such place can be revisited whenever we change the way we deal with deserializing.
 I think the PR is ok the way it is now, so I don't request any changes.

banasraf (approved):

mzient (approved):
Much as I dislike introducing artificial state variable...

PR #4098, opened 2022-07-25
Add a framework for testing imgcodec's decoders

szkarpinski:
## Category:

## Description:

This PR adds a framework for testing imgcodec's decoders. It allows comparing results of decoding with reference image stored in [numpy format](https://numpy.org/doc/stable/reference/generated/numpy.lib.format.html). The general idea is that reference numpy files with decoded images will be stored in `DALI_extra`.

**Why numpy format?** Files in numpy format are universal, self-contained, easy to create and view, and we already have C++ reader code. Another alternative would be comparing against trusted (?) library such as OpenCV, but this would give us no control over the reference output and could lead to situations where tests stop working because of breaking OpenCV change.

### Missing features:
This PR is still a draft, because the following things are not yet done:
- Add doxygen comments to provided classes
- Add more verbose error messages about differences in results

### Future work:
The following things are out of scope of this PR:
- At some point we'd like to generalize it for GPU codecs

### Example:

As nothing (even a PR) uses this framework yet, I provided some example usage. This code will appear in TIFF codec PR in a few days. When writing a test for decoder, one should override two methods, which would instantiate the parser and decoder.
```cpp
constexpr std::string img_ref_path = ".../reference/tiff/image.tiff.npy";
constexpr std::string img_path = ".../tiff/image.tiff";

class LibTiffDecoderTest : public NumpyDecoderTestBase<uint8_t> {
 protected:
  std::shared_ptr<ImageDecoderInstance> CreateDecoder(ThreadPool &tp) override {
    LibTiffDecoder decoder;
    return decoder.Create(CPU_ONLY_DEVICE_ID, tp);
  }
  std::shared_ptr<ImageParser> GetParser() override {
    return std::make_shared<TiffParser>();
  }
};

TEST_F(LibTiffDecoderTest, Test) {
  auto ref = ReadReference(img_ref_path);
  auto src = ImageSource::FromFilename(img_path);
  auto img = Decode(&src);
  AssertEqual(img, ref);
}

TEST_F(LibTiffDecoderTest, TestROI) {
  auto ref = ReadReference(img_ref_path);
  auto src = ImageSource::FromFilename(img_path);
  auto info = GetInfo(&src);

  DecodeParams params;
  params.use_roi = true;
  params.roi.begin = {13, 17, 0};
  params.roi.end = {info.shape[0] - 55, info.shape[1] - 10, 3};
  auto img = Decode(&src, params);
  AssertEqual(img, Crop(ref, params.roi.begin, params.roi.shape()));
}
```

- The framework lives in `decoder_test_helper.h` and provides the `NumpyDecoderTestBase` class template
- Numpy reading stuff was stolen from `numpy_loader.cc` and placed in `numpy_helper.cc`

This PR is kind of a blocker for PRs with codecs. I'd be happy to add any missing features or improve the efficiency of the tests, but I'd prefer to do so in follow-up PRs, to letothers write their codecs.

### Tests:

- [ ] Existing tests apply
- [x] New tests added
  - [ ] Python tests
  - [ ] GTests
  - [ ] Benchmark
  - [ ] Other
- [ ] N/A

### Documentation
- [ ] Documentation updated
  - [ ] Docstring
  - [ ] Doxygen
  - [ ] RST
  - [ ] Jupyter
  - [ ] Other
- [ ] N/A

#### Requirements
- [x] N/A

**REQ IDs**: N/A

**JIRA TASK**: N/A


dali/imgcodec/decoders/test/numpy_helper.cc
```diff
@@ -0,0 +1,235 @@
...
+    DALI_ENFORCE(TrySkip(hdr, ",") || target.shape.size() > 1,
+                 "The first number in a tuple must be followed by a comma.");
+  }
+  if (target.fortran_order) {
+    // cheapest thing to do is to define the tensor in an reversed way
+    std::reverse(target.shape.begin(), target.shape.end());
+  }
+}
+
+void ParseHeader(FileStream *file, NumpyHeaderMeta& parsed_header) {
```
mzient, line 162: Can it be
```suggestion
void ParseHeader(InputStream *file, NumpyHeaderMeta& parsed_header) {
```
?

dali/imgcodec/decoders/test/numpy_helper.cc
```diff
@@ -0,0 +1,235 @@
...
+  perm.resize(n_dims);
+  for (int i = 0; i < n_dims; ++i)
+    perm[i] = n_dims - i - 1;
+  TYPE_SWITCH(input.type(), type2id, T, NUMPY_ALLOWED_TYPES, (
+    kernels::TransposeGrouped(view<T>(output), view<const T>(input), make_cspan(perm));
+  ), DALI_FAIL(make_string("Unsupported input type: ", input.type())));  // NOLINT
+}
+
+Tensor<CPUBackend> ReadNumpy(const std::string &path) {
+  auto file = FileStream::Open(path, false, false);
```
mzient, line 211: I'd suggest splitting this function in two:
```suggestion
Tensor<CPUBackend> ReadNumpy(const std::string &path) {
  auto file = FileStream::Open(path, false, false);
  ReadNumpy(file);
}

Tensor<CPUBackend> ReadNumpy(InputStream &stream) {
```

PR #4108, opened 2022-07-28
Pin Numba version to 0.55.2

awolant:

### Tests:

- [ ] Existing tests apply
- [ ] New tests added
  - [ ] Python tests
  - [ ] GTests
  - [ ] Benchmark
  - [ ] Other
- [ ] N/A

### Documentation
- [ ] Documentation updated
  - [ ] Docstring
  - [ ] Doxygen
  - [ ] RST
  - [ ] Jupyter
  - [ ] Other
- [ ] N/A

#### Requirements
- [ ] N/A

**REQ IDs**: N/A

**JIRA TASK**: N/A

### DIFF
--- a/qa/TL0_cpu_only/test_nofw.sh
+++ b/qa/TL0_cpu_only/test_nofw.sh
@@ -1,6 +1,6 @@
 #!/bin/bash -e
 # used pip packages
-pip_packages='${python_test_runner_package} numpy>=1.17 numba scipy librosa==0.8.1'
+pip_packages='${python_test_runner_package} numpy>=1.17 numba==0.55.2 scipy librosa==0.8.1'
 
 target_dir=./dali/test/python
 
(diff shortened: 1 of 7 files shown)


awolant:
!build

jantonguirao (approved):

klecki:
I think that we probably can add it in https://github.com/NVIDIA/DALI/blob/main/qa/setup_packages.py to have it restricted just for CUDA 10. If this helps with CI we can merge this one, and adjust to a proper solution in a followup.

awolant:
Agreed. I just wanted to see, if this is the thing that caused the issue with this PR. Probably it could have been a draft.

JanuszL:
> I think that we probably can add it in https://github.com/NVIDIA/DALI/blob/main/qa/setup_packages.py to have it restricted just for CUDA 10. If this helps with CI we can merge this one, and adjust to a proper solution in a follow-up.

I don't think we can limit it to only 10. If you specify the version for 10 and none for 11 all CUDA versions above 10 will use the version defined from 10 (that is the way it was designed so packages for 10 can be installed when CUDA 10.1 is present in the system).
Maybe we can limit this promotion only to minor versions. 10.2->10 but 11 won't be downgraded.

awolant:
We'll have to pin it for 11 as well, because current version (0.56.0) is the last that supports Python < 3.11.

klecki (approved):

PR #4111, opened 2022-07-28
Fix Numba versions

awolant:

## Category:

Bug fix

## Description:

New realese of Numba is not compatible with how we test DALI in CI. It strictly requires CUDA 10.2 and our CI does not guarantee it.
Next release of Numba will drop Python 3.7 support, so I pinned Numba version for CUDA 11 as well for now.

This is experimental feature, so I decided to not include any info about this in the docs for now - we might want to pin another version in the future and now we do it due to the tests in CI only, so I think it is ok to not committing to any Number version for now.

### Tests:

- [x] Existing tests apply
- [ ] New tests added
  - [ ] Python tests
  - [ ] GTests
  - [ ] Benchmark
  - [ ] Other
- [ ] N/A

### Documentation
- [ ] Documentation updated
  - [ ] Docstring
  - [ ] Doxygen
  - [ ] RST
  - [ ] Jupyter
  - [ ] Other
- [ ] N/A

#### Requirements
- [ ] N/A

**REQ IDs**: N/A

**JIRA TASK**: N/A

### DIFF
--- a/qa/setup_packages.py
+++ b/qa/setup_packages.py
@@ -455,8 +455,12 @@ all_packages = [PlainPackage("opencv-python", ["4.5.1.48"]),
                           "110" : [
                             PckgVer("2.2.0", dependencies=["protobuf<4"])]}),
                 CudaPackage("numba",
-                        { "100" : ["0.55.2"],
-                          "110" : ["0.56.0"]})
+                        { "100" : [
+                            PckgVer("0.55.2", python_min_ver="3.7"),
+                            PckgVer("0.53.1", python_max_ver="3.6")],
+                          "110" : [
+                            PckgVer("0.56.0", python_min_ver="3.7"),
+                            PckgVer("0.53.1", python_max_ver="3.6")]})
                ]
 
 all_packages_keys = [pckg.key for pckg in all_packages]
--- a/qa/setup_packages.py
+++ b/qa/setup_packages.py
@@ -453,7 +453,10 @@ all_packages = [PlainPackage("opencv-python", ["4.5.1.48"]),
                         { "100" : [
                             PckgVer("2.2.0", dependencies=["protobuf<4"])],
                           "110" : [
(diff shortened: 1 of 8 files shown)


klecki (approved):

qa/setup_packages.py
```diff
@@ -453,7 +453,14 @@ def get_pyvers_name(self, url, cuda_version):
                         { "100" : [
                             PckgVer("2.2.0", dependencies=["protobuf<4"])],
                           "110" : [
-                            PckgVer("2.2.0", dependencies=["protobuf<4"])]})
+                            PckgVer("2.2.0", dependencies=["protobuf<4"])]}),
+                CudaPackage("numba",
+                        { "100" : [
+                            PckgVer("0.55.2", python_min_ver="3.7"),
+                            PckgVer("0.53.1")],
```
JanuszL, line 460:
```suggestion
                            PckgVer("0.53.1",  python_max_ver="3.6")],
```

qa/setup_packages.py
```diff
@@ -453,7 +453,14 @@ def get_pyvers_name(self, url, cuda_version):
...
+                          "110" : [
+                            PckgVer("0.56.0", python_min_ver="3.7"),
+                            PckgVer("0.53.1")]})
```
JanuszL, line 463:
```suggestion
                            PckgVer("0.53.1",  python_max_ver="3.6")]})
```

qa/setup_packages.py
JanuszL: Otherwise we will test  0.55.2 and 0.53.1 for pythons 3.7 and above.

awolant: Done. I had that and removed it before :)

qa/setup_packages.py
awolant: Done

JanuszL (approved):

PR #4124, opened 2022-08-01
Add AsTensor and related APIs to Tensor Vector

klecki:
## Description:
Add to the TensorVector the set of APIs already
present in TensorList:
* IsContiguousTensor
* IsDenseTensor
* AsReshapedTensor
* AsTensor

The APIs allow to access the memory of the contiguous
batch through the API of the Tensor. They are exposed
through the Python APIs for TensorList and the conversion
is valid as long as the allocation of the source object
is not changed.
Caching the returned Tensors is not replicated to simplify
the API.

The atomic counter of current internal "view" Tensors 
(the elements of tensors_) is removed - it use was insignificant
and the propagation of samples with such counter might
have strange side effects.  

Either way, the followups also remove the counter.

TensorVector introducing TensorList APIs to prepare
for the replacement.

There is a difference with how the Tensors are returned 
to Python:
In the TensorList version the Tensor is kept by the
TensorList object. We pass to Python the pointer 
to the cached Tensor instance with 
`py::return_value_policy::reference_internal`
(which keeps the TL alive as long as we use Tensor).

If we kept the reference to the old instance of such Tensor
and the TensorList cleared the cache of the Tensor objects,
for example due to Resize or Reset call, accessing that
Tensor would cause a segfault.

In the TensorVector version, we return a regular Tensor object,
whose allocation has a shared pointer to the same allocation
as the TensorVector. The TensorVector may be destroyed
by Python, and the Tensor would survive.

The lifetime of such Tensor is still problematic now
so I am open to any suggestion. Both of those API mention
that those are only views, and will both be problematic
if we allow Resizing the batch from Python, 
and a bit weird if we allow in-place modification of 
batch object in Python.

### Tests:

- [ ] Existing tests apply
- [x] New tests added
  - [ ] Python tests
  - [x] GTests
  - [ ] Benchmark
  - [ ] Other
- [ ] N/A

### Documentation
- [x] Documentation updated
  - [x] Docstring
  - [ ] Doxygen
  - [ ] RST
  - [ ] Jupyter
  - [ ] Other
- [ ] N/A

#### Requirements
- [ ] N/A

**REQ IDs**: N/A

**JIRA TASK**: N/A

### DIFF
--- a/dali/pipeline/data/tensor_vector.cc
+++ b/dali/pipeline/data/tensor_vector.cc
@@ -765,9 +765,8 @@ TensorVector<Backend> &TensorVector<Backend>::operator=(TensorVector<Backend> &&
 
 
 // This is to check if we are actually laid down in contiguous memory
-// TODO(klecki): make this internal and name it something like: IsContiguouslyStored?
 template <typename Backend>
-bool TensorVector<Backend>::IsContiguousTensor() const {
+bool TensorVector<Backend>::IsContiguousInMemory() const {
   if (num_samples() == 0 || shape().num_elements() == 0) {
     return true;
   }
@@ -790,14 +789,14 @@ bool TensorVector<Backend>::IsContiguousTensor() const {
 
 template <typename Backend>
 bool TensorVector<Backend>::IsDenseTensor() const {
-  return IsContiguous() && is_uniform(shape());
+  return IsContiguousInMemory() && is_uniform(shape());
 }
 
 
 template <typename Backend>
 Tensor<Backend> TensorVector<Backend>::AsReshapedTensor(const TensorShape<> &new_shape) {
-  DALI_ENFORCE(num_samples() > 0,
-               "To create a view Tensor, the batch must have at least 1 element.");
+  DALI_ENFORCE(new_shape.num_elements() == 0 || num_samples() > 0,
+               "To create a non-empty view Tensor, the batch must not be empty.");
   DALI_ENFORCE(IsValidType(type()),
                "To create a view Tensor, the batch must have a valid data type.");
   DALI_ENFORCE(
@@ -806,13 +805,13 @@ Tensor<Backend> TensorVector<Backend>::AsReshapedTensor(const TensorShape<> &new
                   "batch, requested: ",
                   new_shape.num_elements(), " expected: ", shape().num_elements()));
   Tensor<Backend> result;
-  result.ShareData(unsafe_sample_owner(*tl_, 0), tl_->capacity(),
-                   tl_->is_pinned(), new_shape, type(), device_id(), order());
+  result.ShareData(unsafe_owner(*tl_), tl_->capacity(), tl_->is_pinned(), new_shape, type(),
+                   device_id(), order());
   auto result_layout = GetLayout();
-  if (!GetLayout().empty()) {
+  if (result_layout.ndim() + 1 == new_shape.sample_dim()) {
     result_layout = TensorLayout("N") + result_layout;
+    result.SetLayout(result_layout);
   }
-  result.SetLayout(result_layout);
   return result;
 }
 
@@ -822,8 +821,11 @@ Tensor<Backend> TensorVector<Backend>::AsTensor() {
   DALI_ENFORCE(IsDenseTensor(),
                "The batch must be representable tensor - it must has uniform shape and be "
                "allocated in contiguous memory.");
-  DALI_ENFORCE(shape().num_samples() > 0,
-               "To create a view Tensor, the batch must have at least 1 element.");
+  if (shape().num_samples() == 0) {
+    DALI_ENFORCE(sample_dim() > 0,
+                 "To convert empty batch to a Tensor, valid dimensionality must be set");
+    return AsReshapedTensor(TensorShape<>::empty_shape(sample_dim()));
+  }
   return AsReshapedTensor(shape_cat(shape().num_samples(), shape()[0]));
 }
 
--- a/dali/pipeline/data/tensor_vector.cc
+++ b/dali/pipeline/data/tensor_vector.cc
@@ -804,9 +804,21 @@ Tensor<Backend> TensorVector<Backend>::AsReshapedTensor(const TensorShape<> &new
       make_string("To create a view Tensor, requested shape need to have the same volume as the "
                   "batch, requested: ",
                   new_shape.num_elements(), " expected: ", shape().num_elements()));
+
   Tensor<Backend> result;
-  result.ShareData(unsafe_owner(*tl_), tl_->capacity(), tl_->is_pinned(), new_shape, type(),
+
+  shared_ptr<void> ptr;
+  if (num_samples() > 0) {
+    ptr = unsafe_sample_owner(*tl_, 0);
+  } else if (IsContiguous()) {
+    ptr = unsafe_owner(*tl_);
+  } else {
+    ptr = nullptr;
+  }
+
+  result.ShareData(ptr, capacity(), is_pinned(), new_shape, type(),
                    device_id(), order());
+
   auto result_layout = GetLayout();
   if (result_layout.ndim() + 1 == new_shape.sample_dim()) {
     result_layout = TensorLayout("N") + result_layout;
--- a/dali/pipeline/data/tensor_vector.cc
+++ b/dali/pipeline/data/tensor_vector.cc
@@ -816,8 +816,7 @@ Tensor<Backend> TensorVector<Backend>::AsReshapedTensor(const TensorShape<> &new
     ptr = nullptr;
   }
 
-  result.ShareData(ptr, capacity(), is_pinned(), new_shape, type(),
-                   device_id(), order());
+  result.ShareData(ptr, capacity(), is_pinned(), new_shape, type(), device_id(), order());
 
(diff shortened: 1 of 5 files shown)


klecki:
!build

klecki:
(just rebased to the current main, so I'm waiting for comments with another CI runs).

dali/pipeline/data/tensor_vector.cc
```diff
@@ -766,22 +755,79 @@ TensorVector<Backend> &TensorVector<Backend>::operator=(TensorVector<Backend> &&
...
+template <typename Backend>
+bool TensorVector<Backend>::IsDenseTensor() const {
+  return IsContiguous() && is_uniform(shape());
+}
+
+
+template <typename Backend>
+Tensor<Backend> TensorVector<Backend>::AsReshapedTensor(const TensorShape<> &new_shape) {
+  DALI_ENFORCE(num_samples() > 0,
+               "To create a view Tensor, the batch must have at least 1 element.");
```
mzient, line 800:
```suggestion
  DALI_ENFORCE(volume(new_shape) == 0 || num_samples() > 0,
               "To create a non-empty view Tensor, the batch must not be empty.");
```

dali/pipeline/data/tensor_vector.cc
```diff
@@ -766,22 +755,79 @@ TensorVector<Backend> &TensorVector<Backend>::operator=(TensorVector<Backend> &&
...
+  DALI_ENFORCE(IsValidType(type()),
+               "To create a view Tensor, the batch must have a valid data type.");
+  DALI_ENFORCE(
+      shape().num_elements() == new_shape.num_elements(),
+      make_string("To create a view Tensor, requested shape need to have the same volume as the "
+                  "batch, requested: ",
+                  new_shape.num_elements(), " expected: ", shape().num_elements()));
+  Tensor<Backend> result;
+  result.ShareData(unsafe_sample_owner(*tl_, 0), tl_->capacity(),
+                   tl_->is_pinned(), new_shape, type(), device_id(), order());
```
mzient, line 810: Do we want to share the buffer even if the volume is 0?

dali/pipeline/data/tensor_vector.cc
```diff
@@ -766,22 +755,79 @@ TensorVector<Backend> &TensorVector<Backend>::operator=(TensorVector<Backend> &&
...
+  auto result_layout = GetLayout();
+  if (!GetLayout().empty()) {
+    result_layout = TensorLayout("N") + result_layout;
+  }
+  result.SetLayout(result_layout);
```
mzient, line 815: If we reshape, we can only use the source layout if the output shape has one extra (sample) dim.
```suggestion
  auto result_layout = GetLayout();
  if (result_layout.ndim() + 1 == shape.sample_dim()) {
    result_layout = TensorLayout("N") + result_layout;
    result.SetLayout(result_layout);  
  }
```
We could possibly copy the sample layout to the destination when the target shape has the same number of dimensions as our sample, but that is guessing.

dali/pipeline/data/tensor_vector.cc
```diff
@@ -766,22 +755,79 @@ TensorVector<Backend> &TensorVector<Backend>::operator=(TensorVector<Backend> &&
...
+}
+
+
+template <typename Backend>
+Tensor<Backend> TensorVector<Backend>::AsTensor() {
+  DALI_ENFORCE(IsDenseTensor(),
+               "The batch must be representable tensor - it must has uniform shape and be "
+               "allocated in contiguous memory.");
+  DALI_ENFORCE(shape().num_samples() > 0,
+               "To create a view Tensor, the batch must have at least 1 element.");
```
mzient, line 826: Again, we can have a 0-volumed output, no problem.

dali/pipeline/data/tensor_vector.cc
```diff
@@ -766,22 +755,79 @@ TensorVector<Backend> &TensorVector<Backend>::operator=(TensorVector<Backend> &&
...
     other.tensors_.clear();
   }
   return *this;
 }
 
 
+// This is to check if we are actually laid down in contiguous memory
+// TODO(klecki): make this internal and name it something like: IsContiguouslyStored?
+template <typename Backend>
+bool TensorVector<Backend>::IsContiguousTensor() const {
```
mzient, line 770: Actually, I'd swap the names, like:
IsContiguous -> UsesContiguousAllocation
this function -> IsContiguous

This would be consistent with TensorListView::is_contiguous which checks for incidental contiguity (not least because it doesn't manage the storage).

dali/pipeline/data/tensor_vector.cc
```diff
@@ -766,22 +755,79 @@ TensorVector<Backend> &TensorVector<Backend>::operator=(TensorVector<Backend> &&
...
+    }
+    base_ptr += tensors_[i]->shape().num_elements() * size;
+  }
+  return true;
+}
+
+
+template <typename Backend>
+bool TensorVector<Backend>::IsDenseTensor() const {
+  return IsContiguous() && is_uniform(shape());
```
mzient, line 793:
```suggestion
  return IsContiguousTensor() && is_uniform(shape());
```
(however, as I've said before, I'd recommend swapping the names, so it would stay like this).

banasraf (approved):

dali/pipeline/data/tensor_vector.cc
klecki: We already have the SetContiguous/IsContiguous in several places, where it means the selected mode for the batch, making it UseContiguousAllocation is both quite a mouthful to type (and recently I reverted the change from `._num_elements()` to `.shape().num_elements()`) and changes the meaning of the API that is already used for different thing.

This is the only part of the ported API that actually isn't publicly exposed, so we can easily change though, hence my proposal.

klecki: Maybe IsContiguousInMemory?

dali/pipeline/data/tensor_vector.cc
klecki: Done.
Just note, that it is expanding the functionality compared to the TensorList's version - I don't think it makes much sense to port those changes back to remove them soon.

Will try to add a test for empty cases later.

dali/pipeline/data/tensor_vector.cc
klecki: Should this just return an empty Tensor with nullptr if the volume is 0?
This maybe is problematic if we reserved some memory and try to make a Tensor from this batch - still it would have a valid capacity and may be even usable with correct shapes in Resize, so maybe leave it as is?

dali/pipeline/data/tensor_vector.cc
klecki: Done. Good spot.

dali/pipeline/data/tensor_vector.cc
klecki: What number of dimension should we get from empty, uninitialized Batch converted to Tensor? I guess we should stick with the weird default = 1, but I am not sure if enabling this conversion here doesn't add one problem in place of existing one.

klecki: I added a requirement for the sample_dim to be valid in that case.

klecki:
!build

dali/pipeline/data/tensor_vector.cc
```diff
@@ -766,22 +755,81 @@ TensorVector<Backend> &TensorVector<Backend>::operator=(TensorVector<Backend> &&
...
+  DALI_ENFORCE(IsValidType(type()),
+               "To create a view Tensor, the batch must have a valid data type.");
+  DALI_ENFORCE(
+      shape().num_elements() == new_shape.num_elements(),
+      make_string("To create a view Tensor, requested shape need to have the same volume as the "
+                  "batch, requested: ",
+                  new_shape.num_elements(), " expected: ", shape().num_elements()));
+  Tensor<Backend> result;
+  result.ShareData(unsafe_owner(*tl_), tl_->capacity(), tl_->is_pinned(), new_shape, type(),
+                   device_id(), order());
```
mzient, line 809: Now it requires contiguous allocation, not just incidental contiguity :(
```suggestion
  shared_ptr<void> ptr = num_samples() > 0 ? unsafes_sample_owner(*tl_, 0) : IsContiguous() ? unsafe_owner(*tl_) : nullptr;
  result.ShareData(ptr, tl_->capacity(), tl_->is_pinned(), new_shape, type(),
                   device_id(), order());
```

klecki: Now it is even worse, as we are taking some parameters like capacity from the contiguous allocation and some from the accidentally contiguous case.

klecki: And I have conscientious objection to nested ternary.

klecki:
!build

dali/pipeline/data/tensor_vector.cc
klecki: done

klecki:
!build

klecki:
!build

mzient (approved):
OK. For future reference: the name xxx_owner for a pointer to data (as opposed to an object that actually owns something) is very confusing.

PR #4155, opened 2022-08-11
Add bytes_per_sample_hint parameter to parallel external source

stiepan:

## Category:

## Description:
Adds a new parameter to parallel external source to enable user to provide a hint as to the sample size. This way parallel external source can precompute more resonalbe estimation of the required shared memory size and prevent DALI from reallocating the memory during pipeline run. When adjusting shm size, DALI at least doubles the size of a given chunk, so providing more accurate size explicitly can result in less memory consumption.

### Tests:

- [ ] Existing tests apply
- [x] New tests added
  - [x] Python tests
  - [ ] GTests
  - [ ] Benchmark
  - [ ] Other
- [ ] N/A

### Documentation
- [x] Documentation updated
  - [x] Docstring
  - [ ] Doxygen
  - [ ] RST
  - [ ] Jupyter
  - [ ] Other
- [ ] N/A

#### Requirements
- [x] N/A

**REQ IDs**: N/A

**JIRA TASK**: DALI-2960

### DIFF
--- a/dali/python/nvidia/dali/_multiproc/pool.py
+++ b/dali/python/nvidia/dali/_multiproc/pool.py
@@ -559,6 +559,23 @@ class Observer:
             self.thread = None
 
 
+def create_shm_chunk_manager_for_group(group, shm_pool, keep_alive_queue_size,
+                                       min_initial_chunk_size, num_workers,
+                                       batch_size=None):
+    num_mini_batches = 1 if group.batch else num_workers
+    if group.bytes_per_sample_hint is None or batch_size is None:
+        initial_chunk_size = min_initial_chunk_size
+    else:
+        num_samples_per_mini_batch = (batch_size + num_mini_batches - 1) // num_mini_batches
+        initial_chunk_size = num_samples_per_mini_batch * group.bytes_per_sample_hint
+        initial_chunk_size = max(min_initial_chunk_size, initial_chunk_size)
+    return ShmChunkManager(
+        shm_pool,
+        keep_alive_queue_size + group.prefetch_queue_depth,
+        initial_chunk_size,
+        num_mini_batches)
+
+
 class WorkerPool:
     """"Combines worker processes pool with callback contexts, can be used to schedule batches
     to be run on the workers and to receive resulting batches from the workers."""
@@ -585,8 +602,8 @@ class WorkerPool:
 
     @classmethod
     def from_groups(
-            cls, groups, keep_alive_queue_size, start_method="fork", num_workers=1,
-            initial_chunk_size=1024 * 1024, py_callback_pickler=None):
+            cls, groups, keep_alive_queue_size, batch_size=None, start_method="fork",
+            num_workers=1, min_initial_chunk_size=1024 * 1024, py_callback_pickler=None):
         """Creates new WorkerPool instance for given list of ExternalSource groups.
 
         Parameters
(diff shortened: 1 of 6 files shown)


stiepan:
!build

dali/python/nvidia/dali/pipeline.py
```diff
@@ -434,6 +434,16 @@ def executor_statistics(self):
             raise RuntimeError("Pipeline must be built first.")
         return self._pipe.executor_statistics()
 
+    def external_source_shm_statistics(self):
```
mzient, line 437: These are not really "statistics" - rather, "capacities". If we plan to add more than just the capacity, how about returning a dictionary / named tuple?
```
{
"capacity":[1000, 223487, 1000, 1000],
}
```
A NamedTuple would probably be nicer for the end-user.

A bare minimum is an array of arrays - so we can add more info just by extending the output array - I don't like it very much, since now it relies on documentation to get any semantics.

stiepan: I changed it to be a dicionary, very good point.

stiepan:
!build

stiepan:
!build

dali/python/nvidia/dali/external_source.py
```diff
@@ -137,7 +137,7 @@ class _ExternalSourceGroup(object):
 
     def __init__(self, callback, source_desc, is_multioutput, instances=[], *,
                  cuda_stream=None, use_copy_kernel=None, batch=True, parallel=False,
-                 prefetch_queue_depth=None, batch_info=None):
+                 prefetch_queue_depth=None, bytes_per_sample_hint=None, batch_info=None):
```
mzient, line 140: I'd suggest erroring out or at least warning the when this parameter is specified for a non-parallel external source.

stiepan: @klecki has a good point that switching between parallel / non-parallel mode will be extra bothersome with this being an error. That is the case for PES prefetch_queue_depth, commenting it out every time I switch to sequential mode is infuriating as those things can be. :D

dali/python/nvidia/dali/_multiproc/pool.py
```diff
@@ -559,6 +559,23 @@ def close(self):
...
 
+def create_shm_chunk_manager_for_group(group, shm_pool, keep_alive_queue_size,
+                                       min_initial_chunk_size, num_workers,
+                                       batch_size=None):
+    num_mini_batches = 1 if group.batch else num_workers
+    if group.bytes_per_sample_hint is None or batch_size is None:
+        initial_chunk_size = min_initial_chunk_size
+    else:
+        num_samples_per_mini_batch = (batch_size + num_mini_batches - 1) // num_mini_batches
+        initial_chunk_size = num_samples_per_mini_batch * group.bytes_per_sample_hint
```
klecki, line 570: The user provides the hint per sample, but the statistics that we return are per minibatch. We also don't mention this fact in the documentation, I think I will probably be confusing and may result in confusion - I would reuse the statistics and got memory overallocation.

klecki: I would suggest introducing `bytes_per_chunk_hint`, preferably hidden in a `parallel_options` disctionary, so we start grouping those specific options.

dali/python/nvidia/dali/external_source.py
```diff
@@ -522,12 +523,30 @@ class ExternalSource():
 `prefetch_queue_depth` : int, option, default = 1
     When run in ``parallel=True`` mode, specifies the number of batches to be computed in
     advance and stored in the internal buffer, otherwise parameter is ignored.
+
+`bytes_per_sample_hint`: int, option, default = None
+    If specified in ``parallel=True`` mode, the value serves as a hint when
```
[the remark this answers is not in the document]
mzient: I'd at least say it's ignored in non-parallel mode and perhaps issue a warning.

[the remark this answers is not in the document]
stiepan: Added a statement in the docs that it is ignored in non parallel mode.

mzient (approved):

stiepan:
!build

PR #4183, opened 2022-08-22
Use std::exchange in move operator for Tensors

klecki:
## Description:
Use the std::exchange in move operators,
to have single statement:
```x_ = std::exchange(other.x_, 0);```
instead of multiple separate lines:
```
x_ = other.x_;
other.x_ = 0;
``` 

This adjusts Tensor and SampleView types,
so they both have the empty {0} shape after
move.
Previously it left the Tensor with a scalar shape
that could indicate existing allocation after
move.

As the use after move shouldn't be practised,
this change should not impact anything.

TensorViews and TensorShapes are left mostly
unaffected as the static_dim doesn't compose
nicely with forcing {0} shape or clearing
the dimensionality. For a TensorView, it is
valid to have a shape and a nullptr,
for the shapes they can have dimensionality
and no samples.

Tensor, SampleView.

### Tests:

- [x] Existing tests apply
Sample View already has std::move tests.
TensorListShape tests invoke the move constructor.
- [x] New tests added
  - [ ] Python tests
  - [ ] GTests
  - [ ] Benchmark
  - [ ] Other
- [ ] N/A

### Documentation
- [ ] Documentation updated
  - [ ] Docstring
  - [ ] Doxygen
  - [ ] RST
  - [ ] Jupyter
  - [ ] Other
- [ ] N/A

#### Requirements
- [ ] N/A

**REQ IDs**: N/A

**JIRA TASK**: DALI-2966

### DIFF
--- a/include/dali/core/tensor_shape.h
+++ b/include/dali/core/tensor_shape.h
@@ -795,9 +795,8 @@ struct TensorListShape<DynamicDimensions>
   TensorListShape &operator=(const TensorListShape &) = default;
   TensorListShape &operator=(TensorListShape &&other) {
     shapes = std::move(other.shapes);
-    nsamples = other.size();
-    ndim = other.sample_dim();
-    other.nsamples = 0;
+    nsamples = std::exchange(other.nsamples, 0);
+    ndim = other.sample_dim();  // Sample dim could be static, and it is valid to not zero it.
     return *this;
   }
 
@@ -812,9 +811,8 @@ struct TensorListShape<DynamicDimensions>
   template <int other_sample_ndim>
   TensorListShape &operator=(TensorListShape<other_sample_ndim> &&other) {
     shapes = std::move(other.shapes);
-    nsamples = other.size();
-    ndim = other.sample_dim();
-    other.nsamples = 0;
+    nsamples = std::exchange(other.nsamples, 0);
+    ndim = other.sample_dim();  // Sample dim could be static, and it is valid to not zero it.
     return *this;
   }
 
@@ -894,8 +892,7 @@ struct TensorListShape : TensorListShapeBase<TensorListShape<sample_ndim>, sampl
   TensorListShape &operator=(const TensorListShape &) = default;
   TensorListShape &operator=(TensorListShape &&other) {
     shapes = std::move(other.shapes);
-    nsamples = other.size();
-    other.nsamples = 0;
+    nsamples = std::exchange(other.nsamples, 0);
     return *this;
   }
 
--- a/dali/pipeline/data/sample_view.h
+++ b/dali/pipeline/data/sample_view.h
@@ -112,12 +112,9 @@ class SampleViewBase {
 
   SampleViewBase &operator=(SampleViewBase &&other) {
     if (this != &other) {
-      data_ = other.data_;
-      other.data_ = nullptr;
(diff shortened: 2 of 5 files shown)

### CODE REVIEW

**Comment by klecki** [comment]
!build

**Review by prak-nv — APPROVED** [review]

File: include/dali/core/tensor_shape.h [hunk 1]
```diff
@@ -795,9 +795,8 @@ struct TensorListShape<DynamicDimensions>
   TensorListShape &operator=(const TensorListShape &) = default;
   TensorListShape &operator=(TensorListShape &&other) {
     shapes = std::move(other.shapes);
-    nsamples = other.size();
-    ndim = other.sample_dim();
-    other.nsamples = 0;
+    nsamples = std::exchange(other.nsamples, 0);
+    ndim = other.sample_dim();  // Sample dim could be static, and it is valid to not zero it.
```
szalpal, line 799: IIUC, previously the `nsamples` wasn't moved, but passed by value, so this changes the behaviour. Is that intentional?

include/dali/core/tensor_shape.h
```diff
@@ -812,9 +811,8 @@ struct TensorListShape<DynamicDimensions>
   template <int other_sample_ndim>
   TensorListShape &operator=(TensorListShape<other_sample_ndim> &&other) {
     shapes = std::move(other.shapes);
-    nsamples = other.size();
-    ndim = other.sample_dim();
-    other.nsamples = 0;
+    nsamples = std::exchange(other.nsamples, 0);
+    ndim = other.sample_dim();  // Sample dim could be static, and it is valid to not zero it.
```
szalpal, line 815: IIUC, previously the nsamples wasn't moved, but passed by value, so this changes the behaviour. Is that intentional?

include/dali/core/tensor_shape.h
```diff
@@ -894,8 +892,7 @@ struct TensorListShape : TensorListShapeBase<TensorListShape<sample_ndim>, sampl
   TensorListShape &operator=(const TensorListShape &) = default;
   TensorListShape &operator=(TensorListShape &&other) {
     shapes = std::move(other.shapes);
-    nsamples = other.size();
-    other.nsamples = 0;
+    nsamples = std::exchange(other.nsamples, 0);
```
szalpal, line 895: IIUC, previously the nsamples wasn't moved, but passed by value, so this changes the behaviour. Is that intentional?

szalpal (approved):

include/dali/core/tensor_shape.h
klecki: It doesn't matter if it's moved, it's an int.

include/dali/core/tensor_shape.h
klecki: As above.

include/dali/core/tensor_shape.h
klecki: As above

include/dali/core/tensor_shape.h
prak-nv: it is basically `int nsamples; nsamples = int::int(static_cast<int&&>(other.nsamples)); other.nsamples = 0; ` where both int copy and move constructors are trivial (ie. `memcpy`).

PR #4189, opened 2022-08-23
Remove TensorList and use only TensorVector

klecki:
## Description:
Introduces an type alias that replaces TensorList with TensorVector.

Next step will be to rename all the batch object to uniform name
and update the naming of UnsafeCopy and UnsafeSet APIs.

Limit the PassThrough to only 1-1 mappings, as it is currently used.
It will be extended for the upcoming split and merge operations.

Introduce a pass in executor that detects when the input
to MakeContiguous is always contiguous to allow for PassThrough.

PassThrough for the MakeContiguous needs to be computed to allow
for correct sizing of prefetch buffer queues - otherwise we would waste
the memory when we need to do the copy by prefetching the inputs
needlessly, or we would ask executor to allocate the output
of MakeContiguous and than share data into it causing an immediate
deallocation.

MakeContiguous is now inserted before every output, so we make sure
that all outputs are contiguous. Previously it was guaranteed that 
the GPU outputs were contiguous due to the use of TensorList.

Possible followup is to enable outputs from CPU stage again.

ArgumentWorkspace uses TensorVector only now.

All redundant overloads taking TensorList were removed, leaving only
the more generic TensorVector ones.

Some tweaks to debug utilities.

TensorVector, TensorList, Executor, OpGraph,

Probably inserting those MakeContgiuous nodes, I tried to **not** break the construction of the backend graph, it is quite fragile as it relies on matching names for the Tensor Nodes, used to join inputs and outputs of operators with quite enigmatic metadata about it.

### Tests:
- [x] Existing tests apply
  - Keeping for now the tensor_list_test and the tensor_list_as_tensor_vector_test, will remove the former in a followup, but want to have a guarantee that both pass. The latter is more generic port of the former.
  - Tests were adjusted to detect the inclusion of the new MakeContgiuousGPU as additional op that is added before every output. 
- [ ] New tests added
  - [ ] Python tests
  - [ ] GTests
  - [ ] Benchmark
  - [ ] Other
- [ ] N/A

### Documentation
- [ ] Documentation updated
  - [ ] Docstring
  - [ ] Doxygen
  - [ ] RST
  - [ ] Jupyter
  - [ ] Other
- [ ] N/A

#### Requirements
- [x] N/A

**REQ IDs**: N/A

**JIRA TASK**: DALI-2548


klecki:
!build

dali/test/python/operator/test_one_hot.py
```diff
@@ -1,4 +1,4 @@
-# Copyright (c) 2017-2020, NVIDIA CORPORATION. All rights reserved.
+# Copyright (c) 2017-2022, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
szalpal, line 1: Yeah, but nothing changed in this file 😅

dali/pipeline/executor/pipelined_executor.h
```diff
@@ -84,7 +84,7 @@ void PipelinedExecutorImpl<WorkspacePolicy, QueuePolicy>::Build(OpGraph *graph,
 }
 
 template <typename WorkspacePolicy, typename QueuePolicy>
-void PipelinedExecutorImpl<WorkspacePolicy, QueuePolicy>::SetupOutputInfo(const OpGraph &graph) {
+void PipelinedExecutorImpl<WorkspacePolicy, QueuePolicy>::SetupOutputInfo(OpGraph &graph) {
```
szalpal, line 87: Is there a particular reason to remove `const` here (and above)?

dali/pipeline/data/tensor_vector.cc
```diff
@@ -1176,13 +1122,13 @@ bool TensorVector<Backend>::shares_data() const {
...
-template void TensorVector<GPUBackend>::Copy<CPUBackend>(const TensorList<CPUBackend>&, AccessOrder);  // NOLINT
-template void TensorVector<GPUBackend>::Copy<GPUBackend>(const TensorList<GPUBackend>&, AccessOrder);  // NOLINT
+template void TensorVector<CPUBackend>::Copy<CPUBackend>(const TensorVector<CPUBackend> &,
+                                                         AccessOrder, bool);
+template void TensorVector<CPUBackend>::Copy<GPUBackend>(const TensorVector<GPUBackend> &,
+                                                         AccessOrder, bool);
+template void TensorVector<GPUBackend>::Copy<CPUBackend>(const TensorVector<CPUBackend> &,
+                                                         AccessOrder, bool);
+template void TensorVector<GPUBackend>::Copy<GPUBackend>(const TensorVector<GPUBackend> &,
+                                                         AccessOrder, bool);
```
szalpal, line 1132: How about putting every instantiation in a separate line and adding NOLINT? Your call

dali/operators/sequence/optical_flow/optical_flow_test.cc
```diff
@@ -1,4 +1,4 @@
-// Copyright (c) 2019-2021, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+// Copyright (c) 2019-2022, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
szalpal, line 1: Nothing changed in this file as well 🤔

dali/pipeline/pipeline.h
```diff
@@ -199,25 +200,6 @@ class DLL_PUBLIC Pipeline {
...
-   * @param name name of the input
-   * @param tl data
-   * @param order synchronization order of the operation - a CUDA stream or host order
-   * @param sync If SetExternalInputHelper should be blocking - waits until provided data is copied
-   *             to the internal buffer
-   * @param no_copy_mode Select whether to use the parameter defined in the External Source or
-   *                     override the mode of operation forcing the copy or no-copy
-   */
-  template <typename Backend>
-  DLL_PUBLIC inline void SetExternalInput(
```
szalpal, line 214: Why is this removed? Looks like it is using `TensorVector`, not `TensorList`.

dali/operators/signal/decibel/to_decibels_op_gpu.cu
```diff
@@ -33,6 +33,7 @@ class ToDecibelsImpl : public OpImplBase<GPUBackend> {
 
   explicit ToDecibelsImpl(ToDecibelsArgs args)
       : args_(std::move(args)) {
+    max_out_.SetContiguity(BatchContiguity::Contiguous);
```
szalpal, line 36: This line seems to be disturbing. Does that mean, that it's possible that the change breaks code of existing operators?

dali/test/python/operator/test_one_hot.py
klecki: I was debugging with this file and it got smashed by the githook.

dali/pipeline/executor/pipelined_executor.h
klecki: SetupMakeContiguousPassThrough mutates the graph a bit.

dali/pipeline/data/tensor_vector.cc
klecki: I removed it as I felt the NOLINT was a bit of overkill here. It's aligned quite well.

dali/operators/signal/decibel/to_decibels_op_gpu.cu
klecki: Not really. To get access to a contiguous allocation, you need to use one of the API that will throw you an error if you don't meet the preconditions and your allocation is not contiguous. So it won't break silently, on the first use you will get the error, hence why this was adjusted.

dali/pipeline/pipeline.h
klecki: Right. But here it's basically the same overload, doesn't really matter which one we remove.

dali/pipeline/data/tensor_vector.cc
```diff
@@ -764,6 +764,7 @@ void TensorVector<Backend>::recreate_views() {
     std::shared_ptr<void> sample_alias(contiguous_buffer_.get_data_ptr(), sample_ptr);
     tensors_[i].ShareData(sample_alias, tensor_size * type_info().size(), is_pinned(), shape()[i],
                           type(), device_id(), order());
+    tensors_[i].SetLayout(GetLayout());
```
banasraf, line 767: Why is it added here? Was it a bug that the layout wasn't set?

dali/pipeline/graph/graph_descr.cc
```diff
@@ -189,10 +193,10 @@ void OpGraph::AddOp(const OpSpec &spec, const std::string &op_name) {
     new_node.children_tensors.push_back(new_tensor.id);
 
     auto it_inserted = tensor_name_to_id_.insert({name, new_tensor.id});
-    DALI_ENFORCE(it_inserted.second, "Operator '" + spec.name() +
+    DALI_ENFORCE(it_inserted.second, "Operator '" + new_node.instance_name +
         "' has output with name " + name + ", but output "
         "with this name already exists as output of op '" +
-        this->Node(TensorSourceID(name)).spec.name() + "'");
+        this->Node(TensorSourceID(name)).instance_name + "'");
```
szalpal, line 199: How about going with `make_string` here?

dali/pipeline/graph/graph_descr.cc
```diff
@@ -535,6 +549,34 @@ void OpGraph::GenerateDOTFromGraph(const TensorNode &current_node, std::ofstream
...
+  }
+
+  // If the input is inferred, the allocation is done by executor in contiguous fashion
+  // this means we can just pass through the data instead of copying them.
+  bool is_input_always_contiguous = producer_op_node.op->CanInferOutputs();
+  if (is_input_always_contiguous) {
+    return true;
+  }
+
+  // see, if we were passed through
```
szalpal, line 568:
```suggestion
  // check, if we were passed through
```

dali/pipeline/graph/graph_descr.cc
```diff
@@ -535,6 +549,34 @@ void OpGraph::GenerateDOTFromGraph(const TensorNode &current_node, std::ofstream
...
+  auto maybe_source = FollowPassThroughUp(producer_op_node_id, tensor_id);
+
+  // we were not passed through, we can assume that we are produced non-contiguous
```
szalpal, line 571: Something went wrong with this sentence...

dali/pipeline/data/tensor_vector.cc
banasraf: I don't think it's worth it unless it may be backported

dali/pipeline/data/type_traits.h
```diff
@@ -56,18 +48,14 @@ template <template <typename Backend_> class T, typename Backend>
 struct is_batch_container {
   static constexpr bool value =
       is_backend<Backend>::value &&
-      (is_tensor_vector<T, Backend>::value || is_tensor_list<T, Backend>::value);
+      is_tensor_vector<T, Backend>::value;
 };
 
 template <typename Backend = CPUBackend>
 struct BatchContainer {
```
banasraf, line 55: This struct no longer makes much sense, as it's just type alias. Probably to remove at some point?

klecki: Sure, I can simplify in this PR or in the next one.

dali/pipeline/data/tensor_vector.cc
klecki: Done as #4202.

dali/pipeline/operator/builtin/make_contiguous.cc
```diff
@@ -12,24 +12,62 @@
...
+    make_contiguous_cpu->MarkPassThrough();
+  }
+  auto *make_contiguous_mixed = dynamic_cast<MakeContiguousBase<MixedBackend> *>(&op);
+  if (make_contiguous_mixed) {
+    make_contiguous_mixed->MarkPassThrough();
+  }
+  auto *make_contiguous_gpu = dynamic_cast<MakeContiguousBase<GPUBackend> *>(&op);
+  if (make_contiguous_gpu) {
+    make_contiguous_gpu->MarkPassThrough();
+  }
```
banasraf, line 54: Do you want to fail here as well? (like in `IsPassThrough`)
Or, if calling those functions on operators that are not MakeContiguous would rather be a programmer's mistake, then maybe there should be assert instead of DALI_FAIL.

dali/operators/sequence/optical_flow/optical_flow_test.cc
klecki: reverted. and I swore to not care about automated header edits :(

dali/pipeline/executor/executor.h
```diff
@@ -518,10 +512,22 @@ void Executor<WorkspacePolicy, QueuePolicy>::ShareOutputs(DeviceWorkspace *ws) {
       ), DALI_FAIL("Invalid op type"));  // NOLINT(whitespace/parens)
     ), DALI_FAIL("Invalid storage device"));  // NOLINT(whitespace/parens)
   }
+
+  // Mostly a sanity check - we don't want to return a non-contiguous batch to Python.
+  for (int i = 0; i < ws->NumOutput(); i++) {
+    const char *error_msg =
+        "DALI internal error: all outputs from the Pipeline must be contiguous, after being "
+        "processed by MakeContiguous operator.";
```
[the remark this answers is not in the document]
klecki: done

dali/pipeline/graph/graph_descr.cc
klecki: done

dali/pipeline/data/type_traits.h
klecki: Removed half the stuff from here and the operator that used it.

dali/pipeline/operator/builtin/make_contiguous.cc
klecki: I guess I can for completeness. I was thinking about it, this way misusing it will be easily spotted.

klecki:
!build

dali/pipeline/operator/op_schema.h
```diff
@@ -656,12 +656,20 @@ used with DALIDataType, to avoid confusion with `AddOptionalArg<type>(name, doc,
...
+   * TODO(klecki): Introduce additional class of Pass Through that enables partial mappings
+   * (splitting one input into multiple outputs or merging several inputs to one output).
+   *
    * @param inout - tells which inputs are passed through to which outputs.
-   *                Multiple inputs can be passed through to one output (at
-   *                least potentially, e.g. when conditionally forwarding
-   *                one of inputs to the output), but not vice versa.
+   *                Only bijective mappings are allowed.
    */
   DLL_PUBLIC inline OpSchema &PassThrough(const std::map<int, int> &inout) {
```
banasraf, line 665: If it's mapping [1, n) -> [1, n), wouldn't vector of size n just be enough for representing it? Just a note. If it's not that or if you think map is better, ignore this comment.

dali/pipeline/pipeline.cc
```diff
@@ -486,6 +494,7 @@ void Pipeline::Build(std::vector<PipelineOutputDesc> output_descs) {
   vector<string> outputs;
   for (const auto &out_desc : output_descs_) {
     string name = out_desc.name;
+    // what device is it? Op placement, output placement, requested placement?
```
banasraf, line 497: Is this an open question or general note on uncertainty of life?

banasraf (approved):

dali/pipeline/operator/op_schema.h
klecki: I left the possibility for a partial mapping, maybe I should not call it a bijective mapping (partial bijective functions exist).

dali/pipeline/pipeline.cc
klecki: I think it was the latter (but a bit of a former), removed.

klecki:
!build

szalpal (approved):

PR #4213, opened 2022-08-31
Fix the indexing for newer PyTorch

klecki:
## Description:
In new PyTorch when using one tensor to do selection from another
tensor, the indices are required to be on the same device or on the
cpu. Previously, we indexed a CPU tensor with GPU indices to get a GPU
tensor as a result. This change transfers the indexed tensor to GPU
first, so both tensors are on the same device and still return a GPU
tensor.

SSD example

### Tests:

- [x] Existing tests apply
L1 and L3 SSD tests.
- [ ] New tests added
  - [ ] Python tests
  - [ ] GTests
  - [ ] Benchmark
  - [ ] Other
- [ ] N/A

### Documentation
- [ ] Documentation updated
  - [ ] Docstring
  - [ ] Doxygen
  - [ ] RST
  - [ ] Jupyter
  - [ ] Other
- [ ] N/A

#### Requirements
- [x] N/A

**REQ IDs**: N/A

**JIRA TASK**: N/A

### DIFF
--- a/docs/examples/use_cases/pytorch/single_stage_detector/src/utils.py
+++ b/docs/examples/use_cases/pytorch/single_stage_detector/src/utils.py
@@ -20,7 +20,7 @@ def calc_iou_tensor(box1, box2):
     """ Calculation of IoU based on two boxes tensor,
         Reference to https://github.com/kuangliu/pytorch-src
         input:
-            box1 (N, 4) 
+            box1 (N, 4)
             box2 (M, 4)
         output:
             IoU (N, M)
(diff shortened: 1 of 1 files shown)


docs/examples/use_cases/pytorch/single_stage_detector/src/utils.py
```diff
@@ -203,6 +203,7 @@ def decode_single(self, bboxes_in, scores_in, criteria, max_output, max_num=200)
                torch.tensor(labels_out, dtype=torch.long), \
                torch.cat(scores_out, dim=0)
 
+        labels_out = labels_out.to(scores_out.device)
```
klecki, line 206: The fix is here, the reset was the auto whitespace trimming.

klecki:
!build

stiepan (approved):

Issue #4215, opened 2022-09-01
Unrecognized image format when passing index file to fn.readers.webdataset


skull8888888:
I have the following error. It only occurs when I pass the index file to the reader. Once I remove the index_paths argument the error disappears.

```
Unrecognized image format. Supported formats are: JPEG, PNG, BMP, TIFF, PNM, JPEG2000 and WebP.
```

My pipeline
```python
@pipeline_def
def wds_pipeline(urls, index_urls):
    img_raw, json = fn.readers.webdataset(
        paths=urls,
        index_paths=index_urls,
        ext=["jpg", "json"],
        missing_component_behavior="error",
        name="Reader")
    img = fn.decoders.image(img_raw, device="mixed", output_type=types.RGB)
    resized = fn.resize(img, device="gpu", resize_shorter=256.)
    output = fn.crop_mirror_normalize(
        resized,
        dtype=types.FLOAT,
        crop=(224, 224),
        device="gpu",
        mean=[0.48145466 * 255, 0.4578275 * 255, 0.40821073 * 255],
        std=[0.26862954 * 255, 0.26130258 * 255, 0.27577711 * 255])
    return output, fn.pad(json, device="cpu")
```

How I call it
```python
urls = "..."
index_urls = "... .idx"
pipe = wds_pipeline(batch_size=16, num_threads=4, device_id=0, urls=urls, index_urls=index_urls)
pipe.build()
pipe.run()
```
I confirmed that index files contain correct information.
```
v1.2 9039
jpg 2560 41185 000000005.jpg json 45568 386 000000005.json txt 47616 7 000000005.txt
jpg 49664 18276 000000031.jpg json 69632 385 000000031.json txt 71680 7 000000031.txt
jpg 73728 50680 000000026.jpg json 125952 441 000000026.json txt 128000 7 000000026.txt
...

szalpal:
@skull8888888 ,

Could you tell, how did you generate the index file you're passing to `index_paths` argument? Did you use the `wds2idx.py` script?

skull8888888:
I used the terminal command `wds2idx ....tar` following [this](https://docs.nvidia.com/deeplearning/dali/user-guide/docs/examples/general/data_loading/dataloading_webdataset.html#Creating-an-index) guide

szalpal:
@skull8888888 ,

I tried to reproduce this issue on my end and everything worked for me. Could you put together small webdataset archive which manifests the issue and send it here, so I could reproduce the problem?

[...]

PR #4226, opened 2022-09-04
Fix webdataset python index creation script

JanuszL:
- fixes wrong offset calculation in webdataset python
  index creation script. The right formula for the data
  offset calculation should be the output of (`block-number` + 1) * 512
  while it was calculated as (next ``block-number`) * 512 - data size
  (what doesn't work when the first entry is preceded by additional data
  like in POSIX.1-2001 archive

## Category:

## Description:.
- fixes wrong offset calculation in webdataset python
  index creation script. The right formula for the data
  offset calculation should be the output of (`block-number` + 1) * 512
  while it was calculated as (next ``block-number`) * 512 - data size
  (what doesn't work when the first entry is preceded by additional data
  like in POSIX.1-2001 archive

- tools/webdataset index creation script

- NA

### Tests:

- [x] Existing tests apply
  - [X]
    - test_webdataset_requirements.py (but I don't know how to generate tar file with non-0 offset for the initial entry)
- [ ] New tests added
  - [ ] Python tests
  - [ ] GTests
  - [ ] Benchmark
  - [ ] Other
- [ ] N/A

### Documentation
- [ ] Documentation updated
  - [ ] Docstring
  - [ ] Doxygen
  - [ ] RST
  - [ ] Jupyter
  - [ ] Other
- [x] N/A

#### Requirements
- [x] N/A

**REQ IDs**: N/A

**JIRA TASK**: N/A

Relates to https://github.com/NVIDIA/DALI/issues/4215

### DIFF
--- a/tools/wds2idx.py
+++ b/tools/wds2idx.py
@@ -1,5 +1,5 @@
 #!/usr/bin/python3
-# Copyright (c) 2021, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# Copyright (c) 2021-2022, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
@@ -100,7 +100,7 @@ class IndexCreator:
         )  # <type>... <size> <date> <name>
 
         # Extracting
-        for blocks_line, next_blocks_line, types_sizes_line in zip(tar_blocks, tar_blocks[1:],
+        for blocks_line, next_blocks_line, types_sizes_line in zip(tar_blocks, tar_blocks,
                                                                    tar_types_sizes):
             if not blocks_line or not types_sizes_line:
                 continue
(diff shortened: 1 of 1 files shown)


JanuszL:
!build

JanuszL:
!build

JanuszL:
!build

tools/wds2idx.py
```diff
@@ -100,7 +100,7 @@ def _get_data_tar(self):
         )  # <type>... <size> <date> <name>
 
         # Extracting
-        for blocks_line, next_blocks_line, types_sizes_line in zip(tar_blocks, tar_blocks[1:],
+        for blocks_line, next_blocks_line, types_sizes_line in zip(tar_blocks, tar_blocks,
```
klecki, line 103: Do we need to zip the same thing twice? 
We should be using current blocks_line in the calculation I guess.

tools/wds2idx.py
```diff
@@ -111,14 +111,13 @@ def _get_data_tar(self):
...
-                next_blocks_line[next_blocks_line.find(b"block") + 6 : next_blocks_line.find(b":")]) * 512  # noqa: E203, E501
+            offset = int(next_blocks_line[next_blocks_line.find(b"block") + 6:
+                                          next_blocks_line.find(b":")])
+            offset = (offset + 1) * 512
 
             size = types_sizes_line[: -len(name)]
             size = size[: size.rfind(b"-") - 8]  # "... <size> 20yy-mm-...."
             size = int(size[size.rfind(b" "):])
-            offset -= size
-            offset -= offset % 512
```
klecki, line 121: I am not really sure what is calculated here and what is the actual layout? Are the metadata at the end, that we can backtrack by the size and lower it to the closes multiple of 512?

tools/wds2idx.py
JanuszL: Fixed

tools/wds2idx.py
JanuszL: Added a comment.

JanuszL:
!build

awolant (approved):

PR #4243, opened 2022-09-08
Fix AssertClose bug in Imgcodec tests

szkarpinski:
## Category:

## Description:

There were three versions of `AssertClose` in Imgcodec's decoder test helper:
https://github.com/NVIDIA/DALI/blob/edbac3ce6875fee4a405a252e802ce9cb9d2c274/dali/imgcodec/decoders/decoder_test_helper.h#L172-L194

Only one of those interpreted the `eps` as being in dynamic range of the output type (and that's what the doc promised). As a result, `eps=1`, which should mean "we allow +/-1 error in `uint8` image" was interpreted as "we allow +/-1.0 error in **float** image". And, as floats have dynamic range 0.0-1.0, this effectively meant "we allow **any** error".

In this PR I fix the behaviour of the buggy overloads.

The fix is simply making those two overloads call the correct versions of `AssertClose` instead of calling `Check` directly.

## Additional information:

This required adjusting some tests:
- In LibJpegTurbo adjusting the eps was required
- In NvJpeg fixing `AssertClose` revealed another problem: reference images were generated using libjpegturbo, resulting in high (but local) differences. I created a new method `AssertSimilar`, which has a very liberal `eps` (=30%), but also requires the mean error to be relatively small (<1%).

### Tests:

- [x] Existing tests apply
- [ ] New tests added
  - [ ] Python tests
  - [ ] GTests
  - [ ] Benchmark
  - [ ] Other
- [ ] N/A

### Documentation
- [ ] Documentation updated
  - [ ] Docstring
  - [ ] Doxygen
  - [ ] RST
  - [ ] Jupyter
  - [ ] Other
- [ ] N/A

#### Requirements
- [x] N/A

**REQ IDs**: N/A

**JIRA TASK**: N/A

### DIFF
--- a/dali/imgcodec/decoders/nvjpeg/nvjpeg_test.cc
+++ b/dali/imgcodec/decoders/nvjpeg/nvjpeg_test.cc
@@ -75,7 +75,7 @@ class NvJpegDecoderTest : public NumpyDecoderTestBase<GPUBackend, OutputType> {
   /**
    * @brief Checks if the image is similar to the reference.
    *
-   * Checks if 1) max error is lower than 30% 2) mean squared error is lower than 1%.
+   * Checks if 1) max error is lower than 30% 2) mean squared error is lower than 2%.
    */
   void AssertSimilar(const TensorView<StorageCPU, const OutputType> &img,
                      const Tensor<CPUBackend> &ref_tensor) {
@@ -85,16 +85,16 @@ class NvJpegDecoderTest : public NumpyDecoderTestBase<GPUBackend, OutputType> {
       float eps = ConvertSatNorm<OutputType>(0.3);
       Check(img, ref, EqualConvertNorm(eps));
 
-      double mean_error = 0;
+      double mean_square_error = 0;
       uint64_t size = volume(img.shape);
       for (size_t i = 0; i < size; i++) {
         double img_value = ConvertSatNorm<double>(img.data[i]);
         double ref_value = ConvertSatNorm<double>(ref.data[i]);
-        double error = abs(img_value - ref_value);
-        mean_error += error;
+        double error = img_value - ref_value;
+        mean_square_error += error * error;
       }
-      mean_error /= size;
-      EXPECT_LT(mean_error, 0.01);
+      mean_square_error = sqrt(mean_square_error / size);
+      EXPECT_LT(mean_square_error, 0.02);
     ), DALI_FAIL(make_string("Unsupported reference type: ", ref_tensor.type())));  // NOLINT
   }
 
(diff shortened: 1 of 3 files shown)


klecki (approved):

dali/imgcodec/decoders/nvjpeg/nvjpeg_test.cc
```diff
@@ -72,6 +72,32 @@ class NvJpegDecoderTest : public NumpyDecoderTestBase<GPUBackend, OutputType> {
     return std::make_shared<JpegParser>();
   }
 
+  /**
+   * @brief Checks if the image is similar to the reference.
+   *
+   * Checks if 1) max error is lower than 30% 2) mean squared error is lower than 1%.
```
mzient, line 78: The implementation has mean absolute value (L1 norm), not mean square (L2 norm).

dali/imgcodec/decoders/nvjpeg/nvjpeg_test.cc
```diff
@@ -72,6 +72,32 @@ class NvJpegDecoderTest : public NumpyDecoderTestBase<GPUBackend, OutputType> {
...
+      float eps = ConvertSatNorm<OutputType>(0.3);
+      Check(img, ref, EqualConvertNorm(eps));
+
+      double mean_error = 0;
+      uint64_t size = volume(img.shape);
+      for (size_t i = 0; i < size; i++) {
+        double img_value = ConvertSatNorm<double>(img.data[i]);
+        double ref_value = ConvertSatNorm<double>(ref.data[i]);
+        double error = abs(img_value - ref_value);
+        mean_error += error;
```
mzient, line 94: Mean square, as mentioned?
```suggestion
        double error = img_value - ref_value;
        mean_error += error * error;
```

dali/imgcodec/decoders/nvjpeg/nvjpeg_test.cc
```diff
@@ -72,6 +72,32 @@ class NvJpegDecoderTest : public NumpyDecoderTestBase<GPUBackend, OutputType> {
...
+      }
+      mean_error /= size;
```
mzient, line 96:
```suggestion
      mean_error = sqrt(mean_error / size);
```

dali/imgcodec/decoders/nvjpeg/nvjpeg_test.cc
szkarpinski: Oh, I forgot to change the comment. But I'll make it mean square in a moment as suggested

szkarpinski: Changed. Thanks for the catch

szkarpinski:
!build

mzient (approved):

PR #4247, opened 2022-09-09
Use GPU Convert in nvJPEG decoder

staniewzki:
## Category:

## Description:
This PR adds usage of GPU Convert in nvJPEG decoder. This enables the decoder to output data types other than `uint8_t`, and support orientation adjustment.

During my work on this PR me and @szkarpinski found out that a lot of the orientation-related code in imgcodec is bugged, and this PR fixes all the previous mistakes.

### Tests:

- [x] Existing tests apply
- [x] New tests added
  - [ ] Python tests
  - [x] GTests
  - [ ] Benchmark
  - [ ] Other
- [ ] N/A

### Documentation
- [ ] Documentation updated
  - [ ] Docstring
  - [ ] Doxygen
  - [ ] RST
  - [ ] Jupyter
  - [ ] Other
- [ ] N/A

#### Requirements
- [ ] N/A

**REQ IDs**: N/A

**JIRA TASK**: N/A


dali/imgcodec/decoders/nvjpeg/nvjpeg.cc
```diff
@@ -171,17 +173,52 @@ DecodeResult NvJpegDecoderInstance::DecodeImplTask(int thread_idx,
...
   try {
     ParseJpegSample(*in, opts, ctx);
+
     if (roi.use_roi()) {
       ctx.shape[0] = roi.shape()[0];
       ctx.shape[1] = roi.shape()[1];
     }
-    DecodeJpegSample(*in, out.mutable_data<uint8_t>(), opts, ctx);
+
+    // Synchronizing on the acces to immediate_buffer
```
jantonguirao, line 196:
```suggestion
    // Synchronizing on the access to immediate_buffer
```

mzient, line 196:
```suggestion
    // Synchronizing on the access to itermediate_buffer
```
?

dali/imgcodec/util/output_shape.h
```diff
@@ -45,9 +45,9 @@ void OutputShape(OutShape &&out_shape,
       in_d = 1 - d;
     }
     int extent = info.shape[in_d];
-    if (d < roi.end.size())
+    if (roi && d < roi.end.size())
       extent = roi.end[d];
-    if (d < roi.begin.size())
+    if (roi && d < roi.begin.size())
```
jantonguirao, line 50:
```
if (roi) {
  if (d  < ...)
    ...
  if (d < ...)
    ...
}

**Review by jantonguirao — APPROVED** [review]

File: dali/imgcodec/decoders/nvjpeg/nvjpeg.cc [hunk 3]
```diff
@@ -171,17 +173,52 @@ DecodeResult NvJpegDecoderInstance::DecodeImplTask(int thread_idx,
...
+    auto& intermediate_buffer = ctx.resources.intermediate_buffer;
+    CUDA_CALL(cudaEventSynchronize(ctx.resources.decode_event));
```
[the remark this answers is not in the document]
staniewzki: Only when the buffer is actually used. So I guess here It could be conditional, and later on when some other resources are used we can sync again? The second sync with be a no-op, right?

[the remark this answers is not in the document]
staniewzki: The synchronization would have to happen almost in the same place anyways (https://github.com/NVIDIA/DALI/pull/4247), I don't think it makes any difference.

jantonguirao (approved):

dali/imgcodec/decoders/nvjpeg/nvjpeg.cc
```diff
@@ -171,17 +173,52 @@ DecodeResult NvJpegDecoderInstance::DecodeImplTask(int thread_idx,
...
+  bool needs_processing = opts.format == DALI_YCbCr ||
+                          opts.dtype != DALI_UINT8 ||
+                          is_orientation_adjusted;
+
   try {
     ParseJpegSample(*in, opts, ctx);
+
     if (roi.use_roi()) {
       ctx.shape[0] = roi.shape()[0];
       ctx.shape[1] = roi.shape()[1];
```
mzient, line 193: We need to use un-rotated size here - (as we do in ParseJpegSample). This is most certainly a bug and needs a follow-up work (this PR or otherwise).

mzient (approved):
OKayish - but I'd rather not merge it before #4224 
Also, the orientation support with ROI looks broken.

jantonguirao (approved):

staniewzki:
!build

dali/imgcodec/util/output_shape.cc
```diff
@@ -0,0 +1,46 @@
...
+ROI PreOrientationRoi(const ImageInfo &info, ROI roi) {
+  bool swap_xy = info.orientation.rotate % 180 == 90;
+  bool flip_x = info.orientation.rotate == 180 || info.orientation.rotate == 270;
+  bool flip_y = info.orientation.rotate == 90 || info.orientation.rotate == 180;
+  flip_x ^= info.orientation.flip_x, flip_y ^= info.orientation.flip_y;
+
+  auto flip_axis = [&](int idx) {
+    std::swap(roi.begin[idx], roi.end[idx]);
+    roi.begin[idx] = info.shape[idx ^ swap_xy] - roi.begin[idx];
+    roi.end[idx] = info.shape[idx ^ swap_xy] - roi.end[idx];
```
mzient, line 29: Well.... this will only work for exactly 2 dimensions. If we had, say, MJPEG sequence, we'd have an extra outermost dimension and the index xoring hack would break.
Having said that - let's merge it as is and improve in a follow-up.

dali/imgcodec/util/output_shape.h
```diff
@@ -44,15 +44,24 @@ void OutputShape(OutShape &&out_shape,
     if (rotate) {
       in_d = 1 - d;
     }
+
     int extent = info.shape[in_d];
-    if (d < roi.end.size())
-      extent = roi.end[d];
-    if (d < roi.begin.size())
-      extent -= roi.begin[d];
+    if (roi) {
```
mzient, line 49: This check seems redundant...

dali/imgcodec/util/output_shape.h
```diff
@@ -21,7 +21,7 @@ namespace dali {
 namespace imgcodec {
 
 template <typename OutShape>
-void OutputShape(OutShape &&out_shape,
+void DLL_PUBLIC OutputShape(OutShape &&out_shape,
```
mzient, line 24: I don't understand it. A template shouldn't require external linkage - unless you're doing something very wrong.

staniewzki: Well I don't fully understand as well, I though that you told me to do that 😛 Reverting this then.

dali/imgcodec/util/output_shape.h
staniewzki: I will double check it, I think I had some error without it. But yes, seems redundant.

mzient (approved):

dali/imgcodec/util/output_shape.h
staniewzki: Okay, I think I know why I did this, but that was wrong.

PR #4248, opened 2022-09-12
Prevent unnecessary synchronization in MakeContiguous.

mzient:

## Category:

## Description:
MakeContiguous may introduce additional, unnecessary synchronization when:
a) copying - this is solved by using the workspace stream to submit the H2D copy
b) sharing the data (when already contiguous).

The latter results in multiple calls to set_order - the workaround is to set the order to that of the source TensorList for the duration of the call to ShareData and restoring the original output order. These orders are guaranteed by the executor to be already synchronized.

Additionally, when ShareData is called with the same data (for TensorList) or the same underlying managaged object (in Tensor and Buffer), there's no call to free_storage, avoiding the synchronization with the deallocation order.

This is just a workaround - having the synchronization isn't entirely wrong, however fixing stream semantics is out of scope of this quick fix.

### Tests:
Regression tests: TensorListTestSuite

New tests: Actual tests for the synchronization are not implementable within current framework.

- [X] Existing tests apply
- [ ] New tests added
  - [ ] Python tests
  - [ ] GTests
  - [ ] Benchmark
  - [ ] Other
- [ ] N/A

### Documentation
- [X] Documentation updated
  - [ ] Docstring
  - [X] Doxygen
  - [ ] RST
  - [ ] Jupyter
  - [ ] Other
- [ ] N/A

#### Requirements
- [X] N/A

**REQ IDs**: N/A

**JIRA TASK**: N/A

### DIFF
--- a/dali/pipeline/data/buffer.h
+++ b/dali/pipeline/data/buffer.h
@@ -89,6 +89,60 @@ inline bool set_deletion_order(const std::shared_ptr<T> &ptr, AccessOrder order)
 }
 
 /**
+ * @brief Checks whether the underlying managed object of the shared pointers is the same.
+ *
+ * This function may return true for two pointers that do not compare equal if they were created
+ * as aliasing different shared pointers.
+ */
+template <typename T, typename U>
+bool same_managed_object(const std::shared_ptr<T> &a, const std::shared_ptr<U> &b) {
+  return !a.owner_before(b) && !b.owner_before(a);
+}
+
+template <typename T, typename U>
+bool same_managed_object(const std::shared_ptr<T> &a, const std::weak_ptr<U> &b) {
+  return !a.owner_before(b) && !b.owner_before(a);
+}
+
+template <typename T, typename U>
+bool same_managed_object(const std::weak_ptr<T> &a, const std::shared_ptr<U> &b) {
+  return !a.owner_before(b) && !b.owner_before(a);
+}
+
+template <typename T, typename U>
+bool same_managed_object(const std::weak_ptr<T> &a, const std::weak_ptr<U> &b) {
+  return !a.owner_before(b) && !b.owner_before(a);
+}
+
+
+/**
+ * @brief Checks whether two shared pointers are the same
+ *
+ * Unlike `operator==`, this function doesn't just compare the value returned by `get`, but also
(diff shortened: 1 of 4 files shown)


dali/pipeline/data/buffer.h
```diff
@@ -88,6 +88,60 @@ inline bool set_deletion_order(const std::shared_ptr<T> &ptr, AccessOrder order)
...
+}
+
+template <typename T, typename U>
+bool same_managed_object(const std::shared_ptr<T> &a, const std::weak_ptr<U> &b) {
+  return !a.owner_before(b) && !b.owner_before(a);
+}
+
+template <typename T, typename U>
+bool same_managed_object(const std::weak_ptr<T> &a, const std::shared_ptr<U> &b) {
+  return !a.owner_before(b) && !b.owner_before(a);
```
JanuszL, line 109: Does it make sense to express it using L103
```suggestion
  return same_managed_object(b, a);
```

mzient: I don't think it's worth it - the savings in the number of letters typed are marginal and readability isn't improved - rather the contrary. Now it's quite obvious that these functions do the same thing.

dali/pipeline/data/buffer.h
```diff
@@ -88,6 +88,60 @@ inline bool set_deletion_order(const std::shared_ptr<T> &ptr, AccessOrder order)
...
+}
+
+template <typename T, typename U>
+bool same_shared_ptr(const std::shared_ptr<T> &a, const std::weak_ptr<U> &b) {
+  return a == b && same_managed_object(a, b);
+}
+
+template <typename T, typename U>
+bool same_shared_ptr(const std::weak_ptr<T> &a, const std::shared_ptr<U> &b) {
+  return a == b && same_managed_object(a, b);
```
JanuszL, line 136: Similarly:
```suggestion
  return same_shared_ptr(b, a);
```

dali/pipeline/data/buffer.h
JanuszL: Sure, just an idea.

JanuszL (approved):
I wonder if we can test if it really removes the synchronization.

stiepan (approved):

PR #4249, opened 2022-09-12
Separate nvjpeg lib wrapper from the decoder

stiepan:

## Category:

## Description:
Fixes builds with dynamic cuda linking complaining about lack of nvjpeg symbold in imgcodec module.
Moves dynamic nvjpeg wrapper and necessary stub to a separate top-level dir that may be used both by operator and imgcoded modules.

### Tests:

- [x] Existing tests apply
- [ ] New tests added
  - [ ] Python tests
  - [ ] GTests
  - [ ] Benchmark
  - [ ] Other
- [ ] N/A

### Documentation
- [ ] Documentation updated
  - [ ] Docstring
  - [ ] Doxygen
  - [ ] RST
  - [ ] Jupyter
  - [ ] Other
- [x] N/A

#### Requirements
- [ ] N/A

**REQ IDs**: N/A

**JIRA TASK**: N/A

### DIFF
--- a/dali/nvjpeg/nvjpeg_wrap.cc
+++ b/dali/nvjpeg/nvjpeg_wrap.cc
@@ -1,4 +1,4 @@
-// Copyright (c) 2021, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+// Copyright (c) 2021-2022, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -17,9 +17,6 @@
 #include <mutex>
 #include <string>
 #include <unordered_map>
-
-#include "dali/operators/decoder/nvjpeg/nvjpeg_helper.h"
-
 
 namespace {
 
--- a/dali/CMakeLists.txt
+++ b/dali/CMakeLists.txt
@@ -47,6 +47,10 @@ if (BUILD_DALI_PIPELINE)
   add_subdirectory(c_api)
 endif()
 
+if (BUILD_DALI_OPERATORS OR BUILD_DALI_IMGCODEC)
+  add_subdirectory(nvjpeg)
+endif()
+
 if (BUILD_DALI_OPERATORS)
   add_subdirectory(operators)
 endif()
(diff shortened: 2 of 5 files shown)


JanuszL (approved):

dali/operators/CMakeLists.txt
```diff
@@ -80,6 +79,8 @@ if (NOT OPERATOR_TEST_SRCS_PATTERN STREQUAL "" OR NOT OPERATOR_TEST_SRCS_PATTERN
                 OPERATOR_TEST_SRCS_PATTERN_EXCLUDE)
 endif()
 
+# Required for operators.cc to have all symbols
+list(APPEND DALI_OPERATOR_SRCS "${CMAKE_CURRENT_SOURCE_DIR}/../nvjpeg/nvjpeg_helper.cc")
```
stiepan, line 83: Is there a way to pritify that?

JanuszL: Why moving it away from L61?
I think it was put inside:
```
if (NOT OPERATOR_SRCS_PATTERN STREQUAL "" OR
    NOT OPERATOR_SRCS_PATTERN_EXCLUDE STREQUAL "")
```
for a purpose.

stiepan: I guess initially it was just a matter of convienience: not to have to pass it explicity when building subset of ops as it cannot be omitted. Now, it is needed because the file is no longer in any op subdir so it is not discovered with collect_sources.

dali/nvjpeg/nvjpeg_wrap.cc
```diff
@@ -18,7 +18,7 @@
 #include <string>
 #include <unordered_map>
 
-#include "dali/operators/decoder/nvjpeg/nvjpeg_helper.h"
+#include "dali/nvjpeg/nvjpeg_helper.h"
```
mzient, line 21: Is it necessary at all to include this file?

dali/nvjpeg/nvjpeg_helper.cc
```diff
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-#include "dali/operators/decoder/nvjpeg/nvjpeg_helper.h"
+#include "dali/nvjpeg/nvjpeg_helper.h"
```
mzient, line 15: I don't like the idea of this file being moved - there's a copy in imgcodec - it was intentional; now I'm afraid we might see some conflicts (it only works because we made nvjpeg_helper.h header-only).

stiepan: Done, I only move the wrapper and stub placement.

mzient (approved):

PR #4270, opened 2022-09-21
Add experimental.decoders.video for CPU

banasraf:
Signed-off-by: Rafal <Banas.Rafal97@gmail.com>

This PR adds the CPU video decoder operator. It decodes a video file loaded to a memory buffer. 
In this first iteration, only the simplest scenario is implemented - decoding whole video file (1 video = 1 sample). 

## Category:

New feature

## Description:

It adds a new feature needed for video inferencing.

New operator.

Decoding impl, potential optimizations

### Tests:

- [ ] Existing tests apply
- [X] New tests added
  - [X] Python tests
  - [ ] GTests
  - [ ] Benchmark
  - [ ] Other
- [ ] N/A

### Documentation
- [X] Documentation updated
  - [X] Docstring
  - [ ] Doxygen
  - [ ] RST
  - [ ] Jupyter
  - [ ] Other
- [ ] N/A

#### Requirements
- [X] Implements new requirements
- [ ] N/A

**REQ IDs**: DECVID

**JIRA TASK**: N/A

### DIFF
--- a/dali/operators/decoder/video/video_decoder.cc
+++ b/dali/operators/decoder/video/video_decoder.cc
@@ -1,0 +1,69 @@
+// Copyright (c) 2022, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "dali/operators/decoder/video/video_decoder.h"
+
+namespace dali {
+
+template <>
+bool VideoDecoder<CPUBackend>::SetupImpl(std::vector<OutputDesc> &output_desc,
+                                         const workspace_t<CPUBackend> &ws) {
+  ValidateInput(ws);
+  const auto &input = ws.Input<CPUBackend>(0);
+  int batch_size = input.num_samples();
+  frames_decoders_.resize(batch_size);
+  auto &thread_pool = ws.GetThreadPool();
+  for (int i = 0; i < batch_size; ++i) {
+    auto sample = input[i];
+    auto data = reinterpret_cast<const char *>(sample.data<uint8_t>());
+    size_t size = sample.shape().num_elements();
+    thread_pool.AddWork([this, i, data, size](int tid) {
+      frames_decoders_[i] = std::make_unique<FramesDecoder>(data, size);
+    }, volume(sample.shape()));
+  }
+  thread_pool.RunAll();
+  output_desc.resize(1);
+  output_desc[0].shape = ReadOutputShape();
+  output_desc[0].type = DALI_UINT8;
+  return true;
+}
+
+template <>
+void VideoDecoder<CPUBackend>::RunImpl(workspace_t<CPUBackend> &ws) {
+  auto &output = ws.Output<CPUBackend>(0);
+  const auto &input = ws.Input<CPUBackend>(0);
+  int batch_size = input.num_samples();
+  auto &thread_pool = ws.GetThreadPool();
+  for (int s = 0; s < batch_size; ++s) {
+    thread_pool.AddWork([this, s, &output](int tid) {
+      DecodeSample(output[s], *frames_decoders_[s]);
+    }, volume(input[s].shape()));
+  }
+  thread_pool.RunAll();
+}
+
+DALI_SCHEMA(experimental__decoders__Video)
+    .DocStr(
+        R"code(Decodes a video file from a memory buffer (e.g. provided by external source).
+
+The video streams can be in most of the container file formats. FFmpeg is used to parse video
+ containers and returns a batch of sequences of sequence_length frames with shape (N, F, H, W, C),
+ where N is the batch size, and F is the number of frames).)code")
+    .NumInput(1)
+    .NumOutput(1)
+    .InputDox(0, "data", "TensorList", "Data buffer with a loaded video file.");
+
+DALI_REGISTER_OPERATOR(experimental__decoders__Video, VideoDecoder<CPUBackend>, CPU);
+
+}  // namespace dali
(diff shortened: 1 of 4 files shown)


dali/operators/decoder/video/video_decoder.cc
```diff
@@ -0,0 +1,69 @@
...
+                                         const workspace_t<CPUBackend> &ws) {
+  ValidateInput(ws);
+  const auto &input = ws.Input<CPUBackend>(0);
+  int batch_size = input.num_samples();
+  frames_decoders_.resize(batch_size);
+  auto &thread_pool = ws.GetThreadPool();
+  for (int i = 0; i < batch_size; ++i) {
+    auto sample = input[i];
+    auto data = reinterpret_cast<const char *>(sample.data<uint8_t>());
+    size_t size = sample.shape().num_elements();
```
JanuszL, line 30: Isn't `num_elements` return 1 as this is 1D file stream?
```suggestion
    size_t size = volume(sample.shape());
```

dali/operators/decoder/video/video_decoder.cc
```diff
@@ -0,0 +1,69 @@
...
+    thread_pool.AddWork([this, i, data, size](int tid) {
+      frames_decoders_[i] = std::make_unique<FramesDecoder>(data, size);
+    }, volume(sample.shape()));
```
JanuszL, line 33: Do we need to set any priority here? I guess the execution time is the same for all videos (it mostly creation of the context and parsing of the header).

dali/operators/decoder/video/video_decoder.cc
```diff
@@ -0,0 +1,69 @@
...
+template <>
+void VideoDecoder<CPUBackend>::RunImpl(workspace_t<CPUBackend> &ws) {
+  auto &output = ws.Output<CPUBackend>(0);
+  const auto &input = ws.Input<CPUBackend>(0);
+  int batch_size = input.num_samples();
+  auto &thread_pool = ws.GetThreadPool();
+  for (int s = 0; s < batch_size; ++s) {
+    thread_pool.AddWork([this, s, &output](int tid) {
+      DecodeSample(output[s], *frames_decoders_[s]);
+    }, volume(input[s].shape()));
```
JanuszL, line 51: I'm not sure if this is a simple as that. It may very much depend on the codec and the compression level. But I don't have any better idea how account this here.

dali/operators/decoder/video/video_decoder.cc
```diff
@@ -0,0 +1,69 @@
...
+  thread_pool.RunAll();
+}
+
+DALI_SCHEMA(experimental__decoders__Video)
+    .DocStr(
+        R"code(Decodes a video file from a memory buffer (e.g. provided by external source).
+
+The video streams can be in most of the container file formats. FFmpeg is used to parse video
+ containers and returns a batch of sequences of sequence_length frames with shape (N, F, H, W, C),
+ where N is the batch size, and F is the number of frames).)code")
```
JanuszL, line 62: Do you think that writing that each sequence can have a different length adds any value here?

banasraf: Changes this doc to include that information

dali/operators/decoder/video/video_decoder.cc
banasraf: Currently, this also builds index, so I guess the amount of work might depend on number of frames

dali/operators/decoder/video/video_decoder.cc
banasraf: Currently, this also builds index, so I guess the amount of work might depend on number of frames

dali/operators/decoder/video/video_decoder.cc
JanuszL: Understood.

dali/operators/decoder/video/video_decoder.cc
JanuszL: I think we can remove that pretty soon.

dali/operators/decoder/video/video_decoder.h
```diff
@@ -0,0 +1,99 @@
...
+                 "Type of the input buffer must be uint8.");
+    DALI_ENFORCE(input.sample_dim() == 1,
+                 "Input buffer must be 1-dimensional.");
+  }
+
+  TensorListShape<4> ReadOutputShape() {
+    TensorListShape<4> shape(frames_decoders_.size());
+    for (size_t s = 0; s < frames_decoders_.size(); ++s) {
+      TensorShape<4> sample_shape;
+      sample_shape[0] = frames_decoders_[s]->NumFrames();
```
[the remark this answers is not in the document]
JanuszL: This also adds a challenge: how to allocate the buffer for each sequence if we don't know its length ahead of time.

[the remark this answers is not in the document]
awolant: After some discussions we decided to focus on scenarios that this value is obtainable here. 

In general case, we probably need to dynamically allocate memory for frames as long as they are coming and then copy them to the output buffer.

dali/test/python/decoder/test_video.py
```diff
@@ -0,0 +1,78 @@
...
+
+@pipeline_def(device_id=0)
+def video_decoder_pipeline(source):
+    data = fn.external_source(source=source, dtype=types.UINT8, ndim=1)
+    return fn.experimental.decoders.video(data)
+
+
+def video_length(filename):
+    cap = cv2.VideoCapture(filename)
+    return int(cap.get(cv2.CAP_PROP_FRAME_COUNT))
```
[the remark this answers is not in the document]
banasraf: Does dependency on cv hurt us? We use it in other tests also. And that way I would have to hardcode the list of files, so if there's a video added to dali extra it'd have to be also added to this list.

awolant (approved):

dali/operators/decoder/video/video_decoder.cc
banasraf: It's probably a good point for further investigation, but for now I guess number of frames might be an acceptable heuristic. Especially for datasets that contain similar files (formats, codecs, compression) that differ in lenght.

banasraf:
!build

dali/operators/decoder/video/video_decoder.h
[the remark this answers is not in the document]
JanuszL: > scenarios where this value is obtainable here.

I think this may constitute more cases than we think of now. But let us start with something.

JanuszL (approved):

banasraf:
!build

PR #4276, opened 2022-09-22
Change the default order of data storage objects

mzient:

## Description:
Change the default order for data storage objects to `host` - regardless if they contain any data or not.

### Tests:

- [x] Existing tests apply
The performance in RN50 and retina was validated. 
- [x] New tests added
  - [ ] Python tests
  - [x] GTests
  - [ ] Benchmark
  - [ ] Other
- [ ] N/A

### Documentation
- [ ] Documentation updated
  - [ ] Docstring
  - [ ] Doxygen
  - [ ] RST
  - [ ] Jupyter
  - [ ] Other
- [ ] N/A

#### Requirements
- [ ] N/A

**REQ IDs**: N/A

**JIRA TASK**: N/A

### DIFF
--- a/dali/pipeline/data/buffer.h
+++ b/dali/pipeline/data/buffer.h
@@ -453,7 +453,7 @@ class DLL_PUBLIC Buffer {
     free_storage();
     if (order) {
       set_order(order);
-    } else if (!order_ && pinned_) {
+    } else if (!order_) {
       set_order(AccessOrder::host());
     }
 
--- a/dali/pipeline/data/buffer.h
+++ b/dali/pipeline/data/buffer.h
@@ -368,7 +368,7 @@ class DLL_PUBLIC Buffer {
   void set_order(AccessOrder order, bool synchronize = true) {
     if (!order.has_value())
       return;
-    if (!synchronize || !order_.has_value()) {
+    if (!synchronize) {
       order_ = order;
       return;
     }
@@ -453,8 +453,6 @@ class DLL_PUBLIC Buffer {
     free_storage();
     if (order) {
       set_order(order);
-    } else if (!order_) {
-      set_order(AccessOrder::host());
     }
 
     if (device_ < 0) {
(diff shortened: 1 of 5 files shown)


klecki:
!build

dali/pipeline/data/tensor_list.cc
```diff
@@ -463,6 +463,8 @@ const TensorListShape<> &TensorList<Backend>::shape() const & {
 
 template <typename Backend>
 void TensorList<Backend>::set_order(AccessOrder order, bool synchronize) {
+  if (!order)
```
stiepan, line 466: So you cannot nullify the order of the tensor list?

stiepan: And.. what do we need `AccessOrder::null_stream()` for?

klecki:
!build

dali/pipeline/data/tensor_list.cc
klecki: I added a restriction that disallows setting the empty order to TensorList and Buffer, and allowed it to mean no-op in the APIs where it was used that way (so it can be an optional argument). For the null_stream, I think it is invalid, and if used I guess we will get an error?

klecki:
!build

dali/test/python/test_dali_cpu_only.py
```diff
@@ -60,6 +60,20 @@ def get_data():
     return out
 
 
+# The same code is used as CPU-only pipeline to test if TF plugin loads successfully
+# during its installation.
+def test_tensorflow_build_check():
```
stiepan, line 65: Is this related to the PR?

stiepan (approved):

dali/pipeline/data/buffer.h
```diff
@@ -492,7 +491,9 @@ class DLL_PUBLIC Buffer {
    *       set_order(order) separately for less synchronization.
    */
   void reset(AccessOrder order = {}) {
-    set_order(order);
+    if (order) {
+      set_order(order);
+    }
```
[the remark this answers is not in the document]
klecki: In most cases reset means that we free the allocation and you can change the properties, but we kept pinnedness, order and device_id intact.

dali/test/python/test_dali_cpu_only.py
klecki: It is a bit related, because I had to run this code to see where it is failing :P

klecki:
!build

PR #4278, opened 2022-09-23
[WIP] Add frames decoder without index

awolant:
### Tests:

- [ ] Existing tests apply
- [ ] New tests added
  - [ ] Python tests
  - [ ] GTests
  - [ ] Benchmark
  - [ ] Other
- [ ] N/A

### Documentation
- [ ] Documentation updated
  - [ ] Docstring
  - [ ] Doxygen
  - [ ] RST
  - [ ] Jupyter
  - [ ] Other
- [ ] N/A

#### Requirements
- [ ] N/A

**REQ IDs**: N/A

**JIRA TASK**: N/A

### DIFF
--- a/dali/operators/reader/loader/video/frames_decoder.cc
+++ b/dali/operators/reader/loader/video/frames_decoder.cc
@@ -157,7 +157,7 @@ FramesDecoder::FramesDecoder(const std::string &filename)
 
 
 
-FramesDecoder::FramesDecoder(const char *memory_file, int memory_file_size)
+FramesDecoder::FramesDecoder(const char *memory_file, int memory_file_size, bool build_index)
   : av_state_(std::make_unique<AvState>()),
     memory_video_file_(MemoryVideoFile(memory_file, memory_file_size)) {
   av_log_set_level(AV_LOG_ERROR);
@@ -190,6 +190,11 @@ FramesDecoder::FramesDecoder(const char *memory_file, int memory_file_size)
       av_state_->codec_->name,
       ". Supported codecs: h264, HEVC."));
   InitAvState();
+  
+  if (!build_index) {
+    return;
+  }
+  
   BuildIndex();
   DetectVfr();
 }
@@ -198,6 +203,8 @@ void FramesDecoder::BuildIndex() {
   // TODO(awolant): Optimize this function for:
   //  - CFR
   //  - index present in the header
+  
+  index_ = vector<IndexEntry>();
 
   int last_keyframe = -1;
   while (ReadRegularFrame(nullptr, false)) {
@@ -207,10 +214,10 @@ void FramesDecoder::BuildIndex() {
     entry.is_flush_frame = false;
 
     if (entry.is_keyframe) {
-      last_keyframe = index_.size();
+      last_keyframe = index_.value().size();
     }
     entry.last_keyframe_id = last_keyframe;
-    index_.push_back(entry);
+    index_.value().push_back(entry);
   }
   while (ReadFlushFrame(nullptr, false)) {
     IndexEntry entry;
@@ -218,7 +225,7 @@ void FramesDecoder::BuildIndex() {
     entry.pts = av_state_->frame_->pts;
     entry.is_flush_frame = true;
     entry.last_keyframe_id = last_keyframe;
-    index_.push_back(entry);
+    index_.value().push_back(entry);
   }
   Reset();
 }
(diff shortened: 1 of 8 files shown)


dali/operators/reader/loader/video/frames_decoder.cc
```diff
@@ -414,4 +421,8 @@ bool FramesDecoder::ReadNextFrame(uint8_t *data, bool copy_to_output) {
   }
   return ReadFlushFrame(data, copy_to_output);
 }
+
```
JanuszL, line 424: How about adding:
```
std::vector<IndexEntry> &FramesDecoder::Index() {
  return index_.value();
}

File: dali/operators/reader/loader/video/frames_decoder.h [hunk 2]
```diff
@@ -109,18 +109,19 @@ class DLL_PUBLIC FramesDecoder {
...
-  FramesDecoder(const char *memory_file, int memory_file_size);
+  FramesDecoder(const char *memory_file, int memory_file_size, bool build_index = true);
 
   /**
    * @brief Number of frames in the video
    * 
    */
   int64_t NumFrames() const {
-    return index_.size();
+    return av_state_->ctx_->streams[av_state_->stream_id_]->nb_frames;
```
JanuszL, line 124: Is this value available for all containers or only with the index available (like avi)?

JanuszL: According to the comment:
```
    int64_t nb_frames;                 ///< number of frames in this stream if known or 0
```
The code needs to be able to handle the situation where it is 0.

awolant:
!build

dali/operators/reader/loader/video/frames_decoder.h
awolant: I added a hard error to throw if there is no index and there is no `nb_frames`. In the things we support now, it should not happen. If it does, we will take it from there.

dali/operators/reader/loader/video/frames_decoder.cc
awolant: I added the other wrapper to check `has_value` and thorw, if functionality is unavailable (like `Seek`). 
I access `index_` directly only in index building, where we assume the index is there and these are all uses of the index. If the need for another wrapper arises in the future, I'll add it, ok?

JanuszL: Sure. It is just an idea.

awolant:
!build

dali/operators/reader/loader/video/frames_decoder.cc
```diff
@@ -84,6 +84,18 @@ const std::vector<AVCodecID> FramesDecoder::SupportedCodecs = {
   AVCodecID::AV_CODEC_ID_HEVC
 };
 
+int64 FramesDecoder::NumFrames() const {
```
banasraf, line 87: As this call might fail, maybe it would be useful to have some method to check if the number of frames is available (`NumFramesAvailable` or smth like that).

dali/operators/reader/loader/video/frames_decoder.cc
```diff
@@ -207,18 +226,18 @@ void FramesDecoder::BuildIndex() {
     entry.is_flush_frame = false;
 
     if (entry.is_keyframe) {
-      last_keyframe = index_.size();
+      last_keyframe = index_.value().size();
```
banasraf, line 229: I think you can write `index_->size()` (same with other method calls on `index_`)

banasraf: As far as I understand the documentation, `.value()` performs a check if the value is present, `*` and `->` don't do that so they're tiny bit more efficient in cases you're sure the value is there.

banasraf (approved):

dali/operators/reader/loader/video/frames_decoder.cc
JanuszL: Maybe it should not ENFORCE and just return the number. If it <=0  then it is not available?

dali/operators/reader/loader/video/frames_decoder.cc
awolant: Done

dali/operators/reader/loader/video/frames_decoder.cc
awolant: How about we add it when we will be building a code path that uses it? For now, we are not implementing this scenario where this is not available.

JanuszL: If we add the part that uses it, it would happen on the operator level, and the loader remains the same.
I think should move this enforcement to the operator, which doesn't support containers without the number of frames that are known ahead, and don't throw it here. Now it seems that this enforcement here seems to be artificial.

awolant: Done, I changed it to retrun 0 and added a note to the doc about it.

JanuszL (approved):

awolant:
!build

PR #4280, opened 2022-09-23
Take a stab at generating stubs

klecki:
## Description:
Generate stubs for DALI fn API and put them in appropriate .pyi files.
Signatures are generated using Signature and Parameter objects to avoid mistakes:
https://peps.python.org/pep-0362/
Stub files are described here: https://peps.python.org/pep-0484/#stub-files
They can be used by IDE to run the code completion and similar tools.

TODO: 
- [ ] Include type hints - there is a problem of real types (DataNode) vs
    the proper types that can happen in a graph.
- [ ] Need to add the stub generation as a build step.
- [ ] Format the code that was generated 
- [ ] Think about stubs for plugins
- [ ] Support enums in signatures

### Tests:

- [ ] Existing tests apply
- [ ] New tests added
  - [ ] Python tests
  - [ ] GTests
  - [ ] Benchmark
  - [ ] Other
- [ ] N/A

### Documentation
- [ ] Documentation updated
  - [ ] Docstring
  - [ ] Doxygen
  - [ ] RST
  - [ ] Jupyter
  - [ ] Other
- [ ] N/A

#### Requirements
- [ ] N/A

**REQ IDs**: N/A

**JIRA TASK**: N/A


dali/python/nvidia/dali/ops.py
```diff
@@ -773,14 +774,18 @@ def _wrap_op(op_class, submodule=[], parent_module=None):
...
 
 
-def _load_ops():
+def _get_all_registered_ops():
     global _cpu_ops
     global _gpu_ops
     global _mixed_ops
     _cpu_ops = _cpu_ops.union(set(_b.RegisteredCPUOps()))
     _gpu_ops = _gpu_ops.union(set(_b.RegisteredGPUOps()))
     _mixed_ops = _mixed_ops.union(set(_b.RegisteredMixedOps()))
```
mzient, line 783: Now that I look at it - why is it a union? Why not just replace? Can something disappear from `_b.RegisteredXXXOps`?

klecki: :man_shrugging:

klecki:
Done in DALI 1.32 and 1.33

PR #4283, opened 2022-09-26
Improve wake-up condition in ThreadPool.

mzient:

## Category:

## Description:
It seemed that in unfavorable circumstances we would only wake one thread in the threadpool even if multiple work items have been added. I'm not 100% sure it was the case, but the wake-up condition in the new form is more readable and more directly related to the real reason to notify multiple threads.

### Tests:

Everything that uses ThreadPool, really.

- [X] Existing tests apply
- [ ] New tests added
  - [ ] Python tests
  - [ ] GTests
  - [ ] Benchmark
  - [ ] Other
- [ ] N/A

### Documentation
- [ ] Documentation updated
  - [ ] Docstring
  - [ ] Doxygen
  - [ ] RST
  - [ ] Jupyter
  - [ ] Other
- [X] N/A

#### Requirements
- [X] N/A

**REQ IDs**: N/A

**JIRA TASK**: N/A

### DIFF
--- a/dali/pipeline/util/thread_pool.cc
+++ b/dali/pipeline/util/thread_pool.cc
@@ -60,20 +60,19 @@ ThreadPool::~ThreadPool() {
 }
 
 void ThreadPool::AddWork(Work work, int64_t priority, bool start_immediately) {
-  bool started_before = false;
+  bool should_start_all = false;
   {
     std::lock_guard<std::mutex> lock(mutex_);
     work_queue_.push({priority, std::move(work)});
     work_complete_ = false;
-    started_before = started_;
     started_ |= start_immediately;
+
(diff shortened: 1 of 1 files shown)


mzient:
!build

szalpal:
Since this is a bugfix, shouldn't we also introduce a test that covers this bug?

mzient:
> Since this is a bugfix, shouldn't we also introduce a test that covers this bug?

It's a _possible_ bug fix - I'm not sure the bug is really there. I just visited this code and wasn't sure that the condition is checked properly - in any case, the new code should be easier to read.
Verifying this would be quite hard - it's all about race conditions and parallelism and these are notoriously hard to test reliably (especially so if the machine we test is used by someone else).

PR #4292, opened 2022-09-27
Fix nvJPEG decoder synchronization

szkarpinski:
## Category:

## Description:
nvJPEG decoder threads were synchronizing in a wrong place and as a result the buffers were sometimes partially overwritten before being consumed, leading to decoding errors like that one, where part of a crane from another image has appeared on the dog picture:

image

In this PR I move the synchronization point so that we don't do any work until the previous decoding finishes and frees the resources.

nvJPEG

We see that synchronizing with the host after each decoded image is inefficient and @mzient is working on a different design here, which would use double-buffering to handle that. This PR is simply a bugfix to existing design, intended to make things work correctly until the double-buffering is ready.

### Tests:

- [x] Existing tests apply
- [ ] New tests added
  - [ ] Python tests
  - [ ] GTests
  - [ ] Benchmark
  - [ ] Other
- [ ] N/A

### Documentation
- [ ] Documentation updated
  - [ ] Docstring
  - [ ] Doxygen
  - [ ] RST
  - [ ] Jupyter
  - [ ] Other
- [ ] N/A

#### Requirements
- [x] N/A

**REQ IDs**: N/A

**JIRA TASK**: N/A

### DIFF
--- a/dali/imgcodec/decoders/nvjpeg/nvjpeg.cc
+++ b/dali/imgcodec/decoders/nvjpeg/nvjpeg.cc
@@ -201,6 +201,7 @@ void NvJpegDecoderInstance::DecodeJpegSample(ImageSource& in, uint8_t *out, Deco
   auto& decode_event = ctx.resources.decode_event;
   auto& device_buffer = ctx.resources.device_buffer;
 
+  CUDA_CALL(cudaEventSynchronize(decode_event));
   CUDA_CALL(nvjpegStateAttachPinnedBuffer(state, ctx.resources.pinned_buffer));
   CUDA_CALL(nvjpegJpegStreamParse(nvjpeg_handle_, in.RawData<unsigned char>(), in.Size(),
                                   false, false, ctx.resources.jpeg_stream));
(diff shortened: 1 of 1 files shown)


dali/imgcodec/decoders/nvjpeg/nvjpeg.cc
```diff
@@ -201,6 +201,7 @@ void NvJpegDecoderInstance::DecodeJpegSample(ImageSource& in, uint8_t *out, Deco
   auto& decode_event = ctx.resources.decode_event;
   auto& device_buffer = ctx.resources.device_buffer;
 
+  CUDA_CALL(cudaEventSynchronize(decode_event));
```
prak-nv, line 204: and just so I know such synchronization has to be before attaching buffer?

prak-nv (approved):

stiepan (approved):

staniewzki:
Can be closed, #4247 resolved the issue.

szkarpinski:
Closing, #4247 resolved the issue

PR #4312, opened 2022-10-05
Fix clang warnings

JanuszL:
- removes unused variables from DropDims::simplify,
  StftImplGPU::StoreComplexResult and CocoLoader::ParseJsonAnnotations
- makes `move()` call preceded by `std::` in operator.h and
  backend_impl.cc
- silents unused `o` variable warning in reduce_test.h
  as it is used in debug builds for asserts

## Category:

## Description:
- removes unused variables from DropDims::simplify,
  StftImplGPU::StoreComplexResult and CocoLoader::ParseJsonAnnotations
- makes `move()` call preceded by `std::` in operator.h and
  backend_impl.cc
- silents unused `o` variable warning in reduce_test.h
  as it is used in debug builds for asserts

- reduce_drop_dims.h
- reduce_test.h
- stft_gpu_impl.cu
- coco_loader.cc
- operator.h
- backend_impl.cc

- NA

### Tests:

- [x] Existing tests apply
  - stft_gpu_test.cu
  - reduce_drop_dims_test.cc
  - reduce_gpu_test.cc
  - operator_test.cc
  - compilation still works
- [ ] New tests added
  - [ ] Python tests
  - [ ] GTests
  - [ ] Benchmark
  - [ ] Other
- [ ] N/A

### Documentation
- [ ] Documentation updated
  - [ ] Docstring
  - [ ] Doxygen
  - [ ] RST
  - [ ] Jupyter
  - [ ] Other
- [x] N/A

#### Requirements
- [x] N/A

**REQ IDs**: N/A

**JIRA TASK**: N/A

### DIFF
--- a/dali/pipeline/operator/operator.h
+++ b/dali/pipeline/operator/operator.h
@@ -173,7 +173,7 @@ class DLL_PUBLIC OperatorBase {
     using namespace std;  // NOLINT
     static_assert(is_arithmetic_or_half<remove_reference_t<T>>::value || is_enum<T>::value,
                   "The eligible diagnostic entry types are arithmetic types or enum");
-    if (!diagnostics_.emplace(move(name), val).second) {
+    if (!diagnostics_.emplace(std::move(name), val).second) {
       DALI_FAIL("Diagnostic with given name already exists");
     }
   }
--- a/dali/python/backend_impl.cc
+++ b/dali/python/backend_impl.cc
@@ -202,7 +202,7 @@ void FillTensorFromDlPack(py::capsule capsule, SourceDataType<SrcBackend> *batch
   // empty lambda that just captures dlm_tensor_ptr unique ptr that would be destructed when
   // shared ptr is destroyed
   batch->ShareData(shared_ptr<void>(dl_tensor.data,
-                                    [dlm_tensor_ptr = move(dlm_tensor_ptr)](void*) {}),
+                                    [dlm_tensor_ptr = std::move(dlm_tensor_ptr)](void*) {}),
                                     bytes, is_pinned, typed_shape, dali_type.id(), device_id);
 
 
--- a/dali/kernels/reduce/reduce_drop_dims.h
+++ b/dali/kernels/reduce/reduce_drop_dims.h
@@ -1,4 +1,4 @@
-// Copyright (c) 2020-2021, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+// Copyright (c) 2020-2022, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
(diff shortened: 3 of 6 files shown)


JanuszL:
!build

JanuszL:
!build

klecki (approved):

dali/python/backend_impl.cc
```diff
@@ -202,7 +202,7 @@ void FillTensorFromDlPack(py::capsule capsule, SourceDataType<SrcBackend> *batch
   // empty lambda that just captures dlm_tensor_ptr unique ptr that would be destructed when
   // shared ptr is destroyed
   batch->ShareData(shared_ptr<void>(dl_tensor.data,
-                                    [dlm_tensor_ptr = move(dlm_tensor_ptr)](void*) {}),
+                                    [dlm_tensor_ptr = std::move(dlm_tensor_ptr)](void*) {}),
```
mzient, line 205: ....what was "wrong" here?

dali/pipeline/operator/operator.h
```diff
@@ -173,7 +173,7 @@ class DLL_PUBLIC OperatorBase {
     using namespace std;  // NOLINT
     static_assert(is_arithmetic_or_half<remove_reference_t<T>>::value || is_enum<T>::value,
                   "The eligible diagnostic entry types are arithmetic types or enum");
-    if (!diagnostics_.emplace(move(name), val).second) {
+    if (!diagnostics_.emplace(std::move(name), val).second) {
```
mzient, line 176: What's the warning about?

JanuszL: Clang reports usage of unqualified calls to `move`. I think it sees it as a potential threat even if `using namespace std` is used.

dali/python/backend_impl.cc
JanuszL: Clang reports usage of unqualified calls to `move`. I think it sees it as a potential threat even if `using namespace std` is used.

mzient: I don't think we need to modify our code to silence every warning that someone somewhere thought would be a good idea to have but whatever.

dali/pipeline/operator/operator.h
mzient: Fun fact: in C++20 you don't have to use `std::` to call `std::` functions on types from `std::`, because the scope of argument-dependent lookup has been increased - so, you can do:
```C++
auto foo() {
    std::vector<int> v;
    return move(v);   // argument-dependent lookup kicks in
}
```
Same goes for `copy` and basically all stuff from `algorithm` (finally!).

mzient (approved):

dali/python/backend_impl.cc
JanuszL: The full warning is:
```
test.cc:5:12: warning: unqualified call to 'std::move' [-Wunqualified-std-cast-call]
    return move(v);   // argument-dependent lookup kicks in
           ^
           std::
1 warning generated.
```
for 
```
#include <vector>

auto foo() {
    std::vector<int> v;
    return move(v);
}
```

dali/pipeline/operator/operator.h
JanuszL: `--std=c++20` still returns this warning.

dali/python/backend_impl.cc
JanuszL: The change is not that significant, so I would just let it be.

PR #4323, opened 2022-10-08
Fix executor test - add test operators

klecki:
Instead of using actual operators, use simple ones implemented only for testing and present in the test binary

### Tests:

- [ ] Existing tests apply
- [ ] New tests added
  - [ ] Python tests
  - [ ] GTests
  - [ ] Benchmark
  - [ ] Other
- [ ] N/A

### Documentation
- [ ] Documentation updated
  - [ ] Docstring
  - [ ] Doxygen
  - [ ] RST
  - [ ] Jupyter
  - [ ] Other
- [ ] N/A

#### Requirements
- [ ] N/A

**REQ IDs**: N/A

**JIRA TASK**: N/A

### DIFF
--- a/dali/pipeline/executor/executor_test.cc
+++ b/dali/pipeline/executor/executor_test.cc
@@ -631,9 +631,8 @@ TYPED_TEST(ExecutorTest, TestPinning) {
                                     .AddOutput("copy_1", "cpu")),
               "Copy_1");
 
-  graph.AddOp(this->PrepareSpec(OpSpec("Reshape")
+  graph.AddOp(this->PrepareSpec(OpSpec("PassthroughOp")
                                     .AddArg("device", "cpu")
-                                    .AddArg("layout", "")
                                     .AddInput("copy_0", "cpu")
                                     .AddOutput("pass_through_0", "cpu")),
               "PassThrough_0");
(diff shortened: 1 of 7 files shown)


klecki:
!build

dali/pipeline/executor/executor_test.cc
```diff
@@ -660,19 +659,18 @@ TYPED_TEST(ExecutorTest, TestPinning) {
...
                                     .AddInput("copy_2", "cpu")
                                     .AddOutput("pass_through_1", "cpu")),
               "PassThrough_1");
 
   // Check pinning argument inputs to operators in GPU stage
-  graph.AddOp(this->PrepareSpec(OpSpec("random__CoinFlip")
+  graph.AddOp(this->PrepareSpec(OpSpec("CopyArgumentOp")
                                     .AddArg("device", "gpu")
-                                    .AddArgumentInput("probability", "pass_through_1")
+                                    .AddArgumentInput("to_copy", "pass_through_1")
```
stiepan, line 671: I am not familair with this API so here come some basic questions:
1. How does this part correspond to what happens in `DALI_SCHEMA(CopyArgumentOp)`?
2. If they are related, we say that the argument `to_copy` from the schema can accept tensors input here?
3. What is the second name though? Is it declared somewhere else or we define the name here?
4. In particular, how does that relate to `pass_through_1` from `PassthroughOp` op (in any frameworkish way)?

jantonguirao (approved):

dali/pipeline/executor/executor_test.cc
klecki: `pass_through_1` will be a name of tensor node in graph, basically the name of the output of the `PassthroughOp`. We set it as an input for the `to_copy` argument input (we set in schema that it accepts tensor inputs). It can only be on the CPU.
This will be created in Python via:
```
fn.copy_argument_op(device="gpu", to_copy=fn.passthrough_op(...))
```

stiepan: Thanks for the example!

stiepan (approved):

PR #4327, opened 2022-10-10
Enable MPEG4 GPU decoding

JanuszL:
- enables MPEG4 support in the GPU frame decoder
   without index

## Category:

## Description:
- enables MPEG4 support in the GPU frame decoder
   without index

- frames decoder for the GPU

- NA

### Tests:

- [ ] Existing tests apply
- [x] New tests added
  - [ ] Python tests
  - [x] GTests
    - CfrFrameRateMpeg4NoIndex
    - VfrFrameRateMpeg4NoIndex
  - [ ] Benchmark
  - [ ] Other
- [ ] N/A

### Documentation
- [ ] Documentation updated
  - [ ] Docstring
  - [ ] Doxygen
  - [ ] RST
  - [ ] Jupyter
  - [ ] Other
- [x] N/A

#### Requirements
- [x] N/A

**REQ IDs**: N/A

**JIRA TASK**: DALI-3038


JanuszL:
Needs https://github.com/NVIDIA/DALI_extra/pull/108

dali/operators/reader/loader/video/frames_decoder.cc
```diff
@@ -101,41 +102,54 @@ void FramesDecoder::InitAvState() {
...
 
-void FramesDecoder::FindVideoStream() {
-  for (size_t i = 0; i < av_state_->ctx_->nb_streams; ++i) {
-    av_state_->codec_params_ = av_state_->ctx_->streams[i]->codecpar;
-    av_state_->codec_ = avcodec_find_decoder(av_state_->codec_params_->codec_id);
+void FramesDecoder::FindVideoStream(bool init_codecs) {
+  if (init_codecs) {
+    for (size_t i = 0; i < av_state_->ctx_->nb_streams; ++i) {
+      av_state_->codec_params_ = av_state_->ctx_->streams[i]->codecpar;
+      av_state_->codec_ = avcodec_find_decoder(av_state_->codec_params_->codec_id);
```
awolant, line 130: I know that this new code path with `init_codecs` is needed here, because this call fails for `AV_CODEC_ID_MPEG4`. Do you know why it does?

JanuszL: Our build configuration doesn't list anything with mpeg in the name in the `Enabled decoders:` list. I guess we would need to use  "--enable-gpl" or "--enable-nonfree" but we can't due to the license.
If we move to a solution where we can use the libaviutils the user may have installed separately we can lose that limitation.
Still, this code path should be better as we avoid unnecessary codec initialization on the CPU.

awolant: I will investigate, if we need "--enable-gpl" or "--enable-nonfree". We need it for `libxvid` but I am not 100% that we need it for native `mpeg4` decoder. Is that fine?

JanuszL: I suggest to merge this PR and enable mpeg4 support for CPU separately as the GPU support is more important for us.

awolant: https://github.com/NVIDIA/DALI/pull/4361 should cover both CPU and GPU MPEG-4 Part2 decoding.

JanuszL:
!build

awolant (approved):

szalpal (approved):

PR #4330, opened 2022-10-10
Fix checking of the return status of the bundle lib tests

JanuszL:
- when the DALI wheel is built checks is run to verify if all necessary libraries are bundled inside. To speed things up the test are run in parallel with other things in bash and the error, status is not properly propagated so the error in the test doesn't stop the execution

## Category:

## Description:
- when the DALI wheel is built checks is run to verify
  if all necessary libraries are bundled inside. To speed
  things up the test are run in parallel with other things
  in bash and the error, status is not properly propagated
  so the error in the test doesn't stop the execution

- docker/build_helper.sh

- NA

### Tests:

- [ ] Existing tests apply
- [ ] New tests added
  - [ ] Python tests
  - [ ] GTests
  - [ ] Benchmark
  - [ ] Other
- [x] N/A

### Documentation
- [ ] Documentation updated
  - [ ] Docstring
  - [ ] Doxygen
  - [ ] RST
  - [ ] Jupyter
  - [ ] Other
- [x] N/A

#### Requirements
- [x] N/A

**REQ IDs**: N/A

**JIRA TASK**: N/A

### DIFF
--- a/docker/build_helper.sh
+++ b/docker/build_helper.sh
@@ -142,8 +142,12 @@ if [ "${BUILD_PYTHON}" = "ON" ]; then
         ### This appears to be a bug in binutils
         ### (http://bugs.strategoxt.org/browse/NIXPKGS-85).
         bundle_wheel nvidia_dali[_-]*.whl NO NO ${OUT_DEBUG_WHL_NAME} "${BUNDLE_PATH_PREFIX}" &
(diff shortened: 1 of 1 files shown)


JanuszL:
!build

mzient (approved):

banasraf (approved):

PR #4358, opened 2022-10-14
Extend and relax TensorList sample APIs

klecki:
## Description:
Add ResizeSample that allows to resize individual samples. Automatically converts the batch into non-contiguous mode, same as SetSample.

SetSample no longer requires the order between the batch and new sample to match - instead it waits for the incoming order in the order of current batch, so the access in the batch order will be correct.

Fix a problem that prohibited copying from empty batch that had no samples nor type set. Now it resets the current batch to similar state.

Those features are useful for implementing the Split and Merge operators.

TensorList 

If the parameters to order::wait are in correct order.

### Tests:
- [ ] Existing tests apply
- [x] New tests added
  - [ ] Python tests
  - [ ] GTests
  - [ ] Benchmark
  - [ ] Other
- [ ] N/A

### Documentation
- [x] Documentation updated
  - [ ] Docstring
  - [x] Doxygen
  - [ ] RST
  - [ ] Jupyter
  - [ ] Other
- [ ] N/A

#### Requirements
- [ ] N/A

**REQ IDs**: N/A

**JIRA TASK**: N/A

### DIFF
--- a/dali/pipeline/data/tensor_list.cc
+++ b/dali/pipeline/data/tensor_list.cc
@@ -597,8 +597,6 @@ void TensorList<Backend>::ResizeSample(int sample_idx, const TensorShape<> &new_
   assert(sample_idx >= 0 && sample_idx < curr_num_tensors_);
   // Resizing any individual sample converts the batch to non-contiguous mode
   MakeNoncontiguous();
-  if (tensors_[sample_idx].capacity() >= volume(new_shape) * type_.size())
-    return;
   shape_.set_tensor_shape(sample_idx, new_shape);
   tensors_[sample_idx].Resize(new_shape);
 }
--- a/dali/pipeline/data/tensor_list.cc
+++ b/dali/pipeline/data/tensor_list.cc
@@ -221,7 +221,7 @@ TensorList<Backend> &TensorList<Backend>::operator=(TensorList<Backend> &&other)
 template <typename Backend>
 void TensorList<Backend>::VerifySampleShareCompatibility(DALIDataType type, int sample_dim,
                                                          TensorLayout layout, bool pinned,
-                                                         AccessOrder order, int device_id,
+                                                         int device_id,
                                                          const std::string &error_suffix) {
   // Checks in the order of class members
   DALI_ENFORCE(this->type() == type,
@@ -241,9 +241,6 @@ void TensorList<Backend>::VerifySampleShareCompatibility(DALIDataType type, int 
                make_string("Sample must have the same pinned status as target batch, current: ",
                            this->is_pinned(), ", new: ", pinned, error_suffix));
 
-  DALI_ENFORCE(this->order() == order,
-               make_string("Sample must have the same order as the target batch", error_suffix));
-
   DALI_ENFORCE(this->device_id() == device_id,
                make_string("Sample must have the same device id as target batch, current: ",
                            this->device_id(), ", new: ", device_id, error_suffix));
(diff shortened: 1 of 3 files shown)


klecki:
!build

dali/pipeline/data/tensor_list.cc
```diff
@@ -573,6 +585,25 @@ void TensorList<Backend>::Resize(const TensorListShape<> &new_shape, DALIDataTyp
...
+  DALI_ENFORCE(IsValidType(type()),
+               "Sample in TensorList cannot be resized with invalid type. Set the type first for "
+               "the whole TensorList using set_type or Resize.");
+  DALI_ENFORCE(sample_dim() == new_shape.sample_dim(),
+               "Sample in TensorList cannot be resized with non-compatible batch dimension. Use "
+               "set_sample_dim or Resize to set correct sample dimension for the whole batch.");
+  // Bounds check
+  assert(sample_idx >= 0 && sample_idx < curr_num_tensors_);
+  // Resizing any individual sample converts the batch to non-contiguous mode
+  MakeNoncontiguous();
```
mzient, line 599:
```suggestion
  if (volume[new_shape] != volume[shape_.set_tensor_shape(sample_idx)])
    MakeNoncontiguous();
```
?

klecki: As discussed on slack, I think if we want to have a variant that doesn't break the contiguity, it should be a separate call like ReshapeSample or ResizeSample with some parameter, that would always enforce that we have a valid volume. That way the bahaviour is consistent and we know the postconditions of the call always.

dali/pipeline/data/tensor_list.cc
```diff
@@ -573,6 +585,25 @@ void TensorList<Backend>::Resize(const TensorListShape<> &new_shape, DALIDataTyp
...
+  if (tensors_[sample_idx].capacity() >= volume(new_shape) * type_.size())
+    return;
+  shape_.set_tensor_shape(sample_idx, new_shape);
```
stiepan, line 602: Isn't the early exit too early? Shouldn't we remember the new shape even if the actual tensor allocation is not affected?

dali/pipeline/data/tensor_list.cc
```diff
@@ -821,6 +852,13 @@ void TensorList<Backend>::Copy(const TensorList<SrcBackend> &src, AccessOrder or
                                bool use_copy_kernel) {
   auto copy_order = copy_impl::SyncBefore(this->order(), src.order(), order);
 
+  if (!src.has_data() && !IsValidType(src.type())) {
+    Reset();
+    SetLayout(src.GetLayout());
+    // no copying to do
```
stiepan, line 858: Why not just `if(!src.has_data())`?

klecki: We can have no allocation but still have a valid type, in that case reset doesn't work that well.

dali/pipeline/data/tensor_list.cc
klecki: Yup, thanks for spotting that.

stiepan: How about adding a test for such case if that is not much of a hassle?

dali/pipeline/data/tensor_list.cc
klecki: So if the source has a valid type, I want to copy it, I leave it to the Resize(<empty_shape>, type) to handle.

stiepan (approved):

klecki:
!build

PR #4359, opened 2022-10-14
Add internal Split and Merge operators

klecki:
## Description:
(First commit):
Add samplewise passthrough information to graph - it informs the graph, that the operator may
pass through data on any-to-any basis (specifying the mapping would not be practical if we want
to enable more than two-value selection - similar to a `switch` rather than an `if`).
It also indicates that the data won't be guaranteed to be passed through without changes,
so the executor can't consider it as passing through contiguous data.

(Second commit):
Add Split and Merge operators. 
Both accept only a boolean predicate that describes the split, but can be easily extended to support 
multiple values in a switch-like fashion.

Split divides samples into two outputs based on the predicate. All the metadata regarding the input
(dimensionality, type, pinnedness, etc) is passed unchanged to both outputs.

Merge does the opposite to Split. The problems arise around inputs that can have different order
or pinnedness - the first can happen around stage boundary and with pass-through operators,
the latter might be sometimes out of our control when external source is involved.

The order problem is solved/worked around by implicit synchronization that was introduced recently
in SetSample and is now transparent for this operator.

For pinned memory (that is a problem only for CPU op), the executor tries to detect if we pinned any 
of the inputs to the Merge node, and if so, tries to pin all the inputs (including following through
the pass-through). 
The Merge op will treat the output as pinned if executor marked the outputs as pinned, or if any
of the inputs turns out to be pinned. It also needs to utilize this information to set the correct device_id.

Downgrading the status of memory from pinned to regular was considered at some point, but based
on the recent synchronization problems with Xavier, I decided on copy to a buffer of correct pinned
status, so not to get surprised by suddenly fully async copies from supposedly non-pinned memory.

In optimal case, when everything is consistently pinned and the input orders are consistent the merge
operation should not introduce additional synchronization or copies.

Executor had to be extended and now detects empty batches, resets the output and doesn't run the
operator as they were not designed to handle empty inputs.
#4404 prevents from empty batches from entering already existing pipelines, so the old behaviour
of not supporting them is preserved.

For Split and Merge the check for equal batch size in input and output is disabled.

Operators are introduced into `experimental` module with a leading underscore name and without
exposing their documentation (at least for now).

This required few fixes (commit 3 and 4), so the CamelCase to snake_case conversion actually
works with leading underscore, and a fix that correctly accesses schema when inserting the
op into fn api, so we know that the doc is hidden, an it is introduced via the `.hidden` module 
and must be reimported into module above. 

The last commit is just a Python tests.

Executor,  operator checks, schema, graph, a bit of fn API.

Do we need to introduce some new safeguards after disabling some checks for Split/Merge ops?
Can we do better with pinning and pinned memory?

### Tests:

- [ ] Existing tests apply
- [x] New tests added
  - [x] Python tests
  - [x] GTests
  - [ ] Benchmark
  - [ ] Other
- [ ] N/A

### Documentation
- [x] Documentation updated
  - [x] Docstring
  - [ ] Doxygen
  - [ ] RST
  - [ ] Jupyter
  - [ ] Other
- [ ] N/A

#### Requirements
- [ ] N/A

**REQ IDs**: N/A

**JIRA TASK**: DALI-2985


klecki:
!build

dali/pipeline/operator/builtin/merge.cc
```diff
@@ -0,0 +1,173 @@
...
+
+template <typename Backend>
+void Merge<Backend>::WriteTestsDiagnostics(const Workspace &ws) {
+  in_0_pinned_ = ws.template Input<Backend>(0).is_pinned();
+  in_1_pinned_ = ws.template Input<Backend>(1).is_pinned();
+  out_pinned_ = ws.template Output<Backend>(0).is_pinned();
+}
+
+
+DALI_SCHEMA(experimental___Merge)
```
jantonguirao, line 162: It doesn't matter much now because they are hidden, but I wonder if we should be more specific with the operator name. `BatchSplit` `BatchMerge` or similar. Split and Merge are names that we might want to use for some general purpose ops.

dali/pipeline/operator/builtin/merge.cc
```diff
@@ -0,0 +1,173 @@
...
+    .DocStr(R"code(Merge batch based on a predicate.)code")
+    .NumInput(2)
+    .NumOutput(1)
+    .AddArg("predicate", "Boolean categorization of the inputs", DALI_BOOL, true)
```
jantonguirao, line 166: I understand the usage of this op in the scope of conditional execution. However, I think it wouldn't cost us much to make this a more usable operator by changing to the following:
```
#x = [A, B, C]
#y = [D, E]
#z = [F]
out = fn.merge(x, y, z, groups=[1, 0, 0, 2, 1, 0])
# out = [D, A, B, F, E, C]
```

jantonguirao: I'd be fine to limiting to 2 inputs for now, but I think that `predicate` might not be a good name for the argument. `from`, `partitions`, `groups`... would read a bit better, I think. Just food for thought

dali/pipeline/operator/builtin/merge.cc
```diff
@@ -0,0 +1,173 @@
...
+  }
+
+  // We propagate views only, so just don't care about what is here and reset
+  output.Reset();
+  for (int input_category = 0; input_category < kMaxCategories; input_category++) {
+    const auto &input = ws.template Input<Backend>(input_category);
+    if (input.num_samples() > 0) {
+      output.set_type(input.type());
+      output.set_sample_dim(input.shape().sample_dim());
+      output.SetLayout(input.GetLayout());
```
jantonguirao, line 78: I think you should check in SetupImpl that all your inputs have the same type, dimensionality and layout.

dali/pipeline/operator/builtin/split.h
```diff
@@ -0,0 +1,58 @@
...
+
+  DISABLE_COPY_MOVE_ASSIGN(Split);
+
+ private:
+  void RegisterTestsDiagnostics();
+  void WriteTestsDiagnostics(const Workspace &ws);
+
+  USE_OPERATOR_MEMBERS();
+  static constexpr int kMaxCategories = 2;
+  std::array<int, kMaxCategories> category_counts_;
```
jantonguirao, line 50: I think that group or partition is a better name than category. It confused me quite a bit.

dali/pipeline/operator/builtin/split_merge.h
```diff
@@ -0,0 +1,56 @@
...
+  return cond_val ? 0 : 1;
+}
+
+inline bool isSplit(const OpSchema &schema) {
+  return schema.name().rfind("_Split") != std::string::npos;
+}
+
+inline bool isMerge(const OpSchema &schema) {
+  return schema.name().rfind("_Merge") != std::string::npos;
+}
```
jantonguirao, line 48: I'd sleep calmer if we checked the full name of the operator explicitly.

dali/pipeline/operator/builtin/split_merge.h
```diff
@@ -0,0 +1,56 @@
...
+ * @param is_logical
+ * @return int
+ */
+inline int get_category_index(const TensorList<CPUBackend>& condition, int condition_idx,
+                              bool is_logical = true) {
+  assert(is_logical && "Numerical conditions are not implemented");
+  bool cond_val = *condition.tensor<bool>(condition_idx);
+  //
+  return cond_val ? 0 : 1;
+}
```
jantonguirao, line 40: I think we can easily support integer indexing of partitions.

dali/pipeline/operator/builtin/split_merge_test.cc
```diff
@@ -0,0 +1,596 @@
...
+#include "dali/pipeline/workspace/workspace.h"
+#include "dali/test/dali_operator_test.h"
+
+#include "dali/pipeline/pipeline.h"
+#include "dali/test/test_tensors.h"
+
+
+namespace dali {
+
+class SplitMerge : public ::testing::Test {
```
jantonguirao, line 34:
```suggestion
namespace test {

class SplitMergeTest : public ::testing::Test {
```

dali/pipeline/operator/builtin/split_merge_test.cc
```diff
@@ -0,0 +1,596 @@
...
+      EXPECT_EQ(output[which][output_sample_idx].shape(), input[i].shape());
+      for (int elem = 0; elem < input[i].shape().num_elements(); elem++) {
+        EXPECT_EQ((output[which].template tensor<int32_t>(output_sample_idx)[elem]),
+                  (input.template tensor<int32_t>(i)[elem]));
+      }
+    }
+  }
+}
+
+class SplitMergeNegative : public SplitMerge {
```
jantonguirao, line 415:
```suggestion
class SplitMergeNegativeTest : public SplitMergeTest {
```

dali/pipeline/operator/builtin/split_merge_test.cc
```diff
@@ -0,0 +1,596 @@
...
+      EXPECT_NE(std::string(e.what()).rfind(expected), std::string::npos)
+          << expected << "\n====\nvs\n====\n"
+          << e.what();
+    } catch (...) {
+      FAIL() << "Unexpected exception.";
+    }
+  }
+}
+
+class SplitMergePinnedInputs : public SplitMerge {
```
jantonguirao, line 540:
```suggestion
class SplitMergePinnedInputsTest : public SplitMergeTest {
```

dali/pipeline/operator/builtin/merge.cc
jantonguirao: Or even better `batch.Split` `batch.Merge`

dali/pipeline/operator/builtin/split.cc
```diff
@@ -0,0 +1,109 @@
...
+}
+
+template <typename Backend>
+void Split<Backend>::WriteTestsDiagnostics(const Workspace &ws) {
+  in_pinned_ = ws.template Input<Backend>(0).is_pinned();
+  out_0_pinned_ = ws.template Output<Backend>(0).is_pinned();
+  out_1_pinned_ = ws.template Output<Backend>(1).is_pinned();
+}
+
+DALI_SCHEMA(experimental___Split)
```
jantonguirao, line 98: perhaps `experimental.batch.Split`?

dali/pipeline/operator/builtin/split.cc
```diff
@@ -0,0 +1,109 @@
...
+    .DocStr(R"code(Split batch based on a predicate.)code")
+    .NumInput(1)
+    .NumOutput(2)
+    .AddArg("predicate", "Boolean categorization of the input batch", DALI_BOOL, true)
```
jantonguirao, line 102: I'd consider something like `partition` `group` (whatever we choose that works for both Split and Merge) and make it an index rather than a bool

dali/pipeline/operator/builtin/split_merge.h
```diff
@@ -0,0 +1,56 @@
...
+
+inline bool isSplit(const OpSchema &schema) {
```
jantonguirao, line 42: `IsSplit` or `is_split`. Let's not add a new naming convention :)
Same applies for isMerge and isSplitOrMerge

dali/pipeline/operator/builtin/split_merge.h
```diff
@@ -0,0 +1,56 @@
...
+  return schema.name().rfind("_Split") != std::string::npos;
+}
+
+inline bool isMerge(const OpSchema &schema) {
+  return schema.name().rfind("_Merge") != std::string::npos;
+}
+
+inline bool isSplitOrMerge(const OpSchema &schema) {
```
jantonguirao, line 50: a bit redundant to have that I'd say.

dali/pipeline/operator/operator.cc
```diff
@@ -12,12 +12,17 @@
...
 
+#include "dali/pipeline/operator/builtin/split_merge.h"
 #include "dali/pipeline/operator/operator.h"
 
 namespace dali {
 
 template <typename Backend>
 void OperatorBase::EnforceUniformInputBatchSize(const Workspace &ws) const {
+  // Builtin operators have relaxed checks for the purpose of conditional execution
+  if (isSplitOrMerge(spec_.GetSchema())) {
```
jantonguirao, line 23: `IsSplit(schema) || IsMerge(schema)`?

dali/pipeline/operator/builtin/split.cc
awolant: I think that what you suggest is planned as an extension of this functionality. Since it makes it just a little bit easier I'd be in favor of proceeding with the boolean predicate for now.

awolant (approved):
Looks to me like it does the job. Nice!

dali/pipeline/operator/builtin/merge.cc
klecki: I can agree on the BatchSplit, but I think batch.split is a bit weird.
Still, for now they are intended to be hidden (and private).

jantonguirao: _conditional.[Split, Merge] would be nice, I think (as discussed offline)

dali/pipeline/operator/builtin/split.cc
jantonguirao: `_conditional.Split` (as discussed offline)

dali/pipeline/operator/builtin/split.cc
klecki: As discussed, we keep the boolean predicate for now. 
The integer `group` would be one of the next steps in this effort. Accepting booleans for conditions is needed, `group` would be a mutually exclusive argument with it, so we can easily extend. But handling arbitrary number of groups adds a lot of complexity on DALI side and probably requires more API extensions(to handle variadic number of outputs we probably need additional argument), and complicates testing.

dali/pipeline/operator/builtin/merge.cc
klecki: Also, `_conditional.split` was suggested.

dali/pipeline/operator/builtin/merge.cc
klecki: As discussed, this is planned for a follow-up.

dali/pipeline/operator/builtin/merge.cc
klecki: SetSample already does it, but I add more suited error messaging.

dali/pipeline/operator/builtin/split.h
klecki: Done

dali/pipeline/operator/builtin/split_merge.h
klecki: done

dali/pipeline/operator/builtin/split_merge.h
klecki: done

dali/pipeline/operator/operator.cc
klecki: I would be happy to do something like `IsThisAnOpThatDoesn'tNeedThisCheck` but it's a bit long, so I used this one, as the Split and Merge go hand in hand.

dali/pipeline/operator/builtin/split_merge_test.cc
klecki: done

dali/pipeline/operator/builtin/split_merge_test.cc
klecki: done

klecki:
!build

klecki:
!build

jantonguirao (approved):

dali/test/python/operator/test_split_merge.py
```diff
@@ -0,0 +1,146 @@
...
+
+def test_fail_conditional_split_merge():
+    base = (
+        "Divergent data found in different branches of conditional operation. All paths in "
+        "conditional operation are merged into one batch which must have consistent type, "
+        "dimensionality, layout and other metadata. Found distinct "
+    )
+    yield check_fail_conditional_split_merge, types.UINT32, None, None, base + "types*"
+    yield check_fail_conditional_split_merge, None, "HWC", None, base + "layouts*"
+    yield check_fail_conditional_split_merge, None, None, [10, -1], base + "sample dimensions*"
```
szalpal, line 146: Small hint. Now that we have `nose2`, you can write the tests without abusing the generators. I believe implementation using `@param` would be way better. Like here: https://github.com/NVIDIA/DALI/pull/4379/files?show-viewed-files=true&file-filters%5B%5D=#diff-9d4f5c833015a5d47d5c41f94b79bd93651bc9e75c3e8debb52aab2d6faa97c9R119

I've used it with `unittest.TestCase` class, but I guess it works with `test_` functions as well.

dali/test/python/operator/test_split_merge.py
```diff
@@ -0,0 +1,146 @@
+# Copyright (c) 2022, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
```
szalpal, line 5: What about nested split-merges? Are they allowed? If so, maybe it would be good to test them?

dali/test/python/operator/test_split_merge.py
```diff
@@ -0,0 +1,146 @@
...
+    with assert_raises(RuntimeError, glob=err_glob):
+        run_conditional_split_merge_reinterpret(dtype, layout, shape)
+
+
+def test_fail_conditional_split_merge():
+    base = (
+        "Divergent data found in different branches of conditional operation. All paths in "
+        "conditional operation are merged into one batch which must have consistent type, "
+        "dimensionality, layout and other metadata. Found distinct "
+    )
```
szalpal, line 143: How about also adding a test, where there is no split, but split&merge is? Like `predicate=[True]*batch_size`

szalpal (commented):
Most of suggestions are, well, suggestions, so feel free to ignore them

dali/python/nvidia/dali/ops.py
```diff
@@ -754,7 +754,7 @@ def _process_op_name(op_schema_name, make_hidden=False):
     op_full_name = op_schema_name.replace(namespace_delim, '.')
     *submodule, op_name = op_full_name.split('.')
     if make_hidden:
-        submodule = (*submodule, 'hidden')
+        submodule = [*submodule, 'hidden']
```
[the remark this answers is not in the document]
klecki:
```
>>> *x, y = "a.b.c.d".split(".")
>>> x
['a', 'b', 'c']
```
So the statement above already returns it as a list, if we are not hiding the documentation it's a list. This unifies the behaviour.

dali/pipeline/executor/executor.cc
```diff
@@ -313,6 +314,27 @@ void Executor<WorkspacePolicy, QueuePolicy>::RunHelper(OpNode &op_node, Workspac
...
   }
 
+  // Assuming that most operators don't expect empty input, and expect consistent input.
+  if (ws.NumInput() > 0) {
+    bool all_inputs_empty = true;
+    for (int i = 0; i < ws.NumInput(); i++) {
+      all_inputs_empty = all_inputs_empty && ws.GetInputBatchSize(i) == 0;
+    }
+    if (all_inputs_empty) {
+      // TODO(klecki): All ops propagate the dim, type and do correct validation with empty input.
```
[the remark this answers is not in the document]
klecki: Reworked.

dali/pipeline/operator/builtin/merge.h
```diff
@@ -0,0 +1,75 @@
...
+#include "dali/core/common.h"
+#include "dali/pipeline/operator/operator.h"
+namespace dali {
+
+template <typename Backend>
+class Merge : public Operator<Backend> {
+ public:
+  inline explicit Merge(const OpSpec &spec) : Operator<Backend>(spec) {
+    DALI_ENFORCE(spec.HasTensorArgument("predicate"),
+                 "The 'predicate' argument is required to be present as argument input.");
```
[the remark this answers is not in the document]
klecki: Done

dali/pipeline/operator/builtin/merge.h
```diff
@@ -0,0 +1,75 @@
...
+    RegisterTestsDiagnostics();
+  }
+
+  virtual inline ~Merge() = default;
```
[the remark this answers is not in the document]
klecki: done

dali/pipeline/operator/builtin/merge.h
```diff
@@ -0,0 +1,75 @@
...
+
+#include "dali/core/access_order.h"
+#include "dali/core/common.h"
+#include "dali/pipeline/operator/operator.h"
+namespace dali {
+
+template <typename Backend>
+class Merge : public Operator<Backend> {
+ public:
+  inline explicit Merge(const OpSpec &spec) : Operator<Backend>(spec) {
```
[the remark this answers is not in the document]
klecki: done

dali/pipeline/operator/builtin/split.h
```diff
@@ -0,0 +1,58 @@
...
+#include <vector>
+
+#include "dali/pipeline/operator/operator.h"
+
+namespace dali {
+
+template <typename Backend>
+class Split : public Operator<Backend> {
+ public:
+  inline explicit Split(const OpSpec &spec) : Operator<Backend>(spec) {
```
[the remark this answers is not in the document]
klecki: done

dali/pipeline/operator/builtin/split.h
```diff
@@ -0,0 +1,58 @@
...
+    DALI_ENFORCE(spec.HasTensorArgument("predicate"),
+                 "The 'predicate' argument is required to be present as argument input.");
```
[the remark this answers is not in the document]
klecki: done

dali/pipeline/operator/builtin/split.h
```diff
@@ -0,0 +1,58 @@
...
+    RegisterTestsDiagnostics();
+  }
+
+  virtual inline ~Split() = default;
```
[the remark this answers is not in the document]
klecki: done

dali/pipeline/operator/builtin/split.h
```diff
@@ -0,0 +1,58 @@
...
+  void RunImpl(Workspace &ws) override;
+
+  DISABLE_COPY_MOVE_ASSIGN(Split);
+
+ private:
+  void RegisterTestsDiagnostics();
+  void WriteTestsDiagnostics(const Workspace &ws);
+
+  USE_OPERATOR_MEMBERS();
+  static constexpr int kMaxGroups = 2;
```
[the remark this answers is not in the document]
klecki: Done

dali/pipeline/operator/builtin/split_merge.h
```diff
@@ -0,0 +1,56 @@
...
+namespace dali {
+
+/**
+ * @brief When we have logical condition, the true group is expected to go through the first (0)
+ * output and input (respectively for split and merge), so we basically do the reverse here.
+ *
+ * @param condition
+ * @param condition_idx
+ * @param is_logical
+ * @return int
```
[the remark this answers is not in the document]
klecki: done

dali/pipeline/operator/builtin/split_merge.h
```diff
@@ -0,0 +1,56 @@
...
+ */
+inline int get_group_index(const TensorList<CPUBackend>& condition, int condition_idx,
+                              bool is_logical = true) {
+  assert(is_logical && "Numerical conditions are not implemented");
+  bool cond_val = *condition.tensor<bool>(condition_idx);
+  //
```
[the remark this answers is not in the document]
klecki: done

dali/pipeline/operator/builtin/merge.cc
```diff
@@ -0,0 +1,203 @@
...
+    if (nonzero_input_sample_idx < 0 && input.num_samples() > 0) {
+      nonzero_input_sample_idx = input_group_idx;
+      continue;  // no point in comparing with ourselves
+    }
+    if (nonzero_input_sample_idx >= 0 && input.num_samples() > 0) {
+      const auto &base_input = ws.template Input<Backend>(nonzero_input_sample_idx);
+      const char *error_msg_base =
+          "Divergent data found in different branches of conditional operation. All paths in "
+          "conditional operation are merged into one batch which must have consistent type, "
+          "dimensionality, layout and other metadata. ";
```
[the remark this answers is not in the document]
klecki: done

dali/pipeline/operator/builtin/merge.cc
```diff
@@ -0,0 +1,203 @@
...
+    // We produce pinned data if the executor said so
+    pinned_ = output.is_pinned();
+    // TODO(klecki): We keep pinedness if it was passed to us:
+    for (int input_group = 0; input_group < kMaxGroups; input_group++) {
+      pinned_ = *pinned_ || ws.template Input<Backend>(input_group).is_pinned();
+    }
+  }
+
+  if (*pinned_ || std::is_same_v<Backend, GPUBackend>) {
+    cudaGetDevice(&device_id_);
```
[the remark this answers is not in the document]
klecki: done

dali/pipeline/operator/builtin/merge.cc
```diff
@@ -0,0 +1,203 @@
...
+  in_1_pinned_ = ws.template Input<Backend>(1).is_pinned();
+  out_pinned_ = ws.template Output<Backend>(0).is_pinned();
+}
+
+
+DALI_SCHEMA(_conditional__Merge)
+    .DocStr(R"code(Merge batch based on a predicate.)code")
+    .NumInput(2)
+    .NumOutput(1)
+    .AddArg("predicate", "Boolean categorization of the inputs", DALI_BOOL, true)
```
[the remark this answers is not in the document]
klecki: There is no simple usage, as this is internal now. What do you mean if the predicate can be iterable or generator? It can support whatever DALI pipeline supports for argument input, provided it has correct type and shape.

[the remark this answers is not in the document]
klecki: If we expose them at some point, we can write the doc based on how they should be used, but for now I think I need some more data about it.

dali/test/python/operator/test_split_merge.py
```diff
@@ -0,0 +1,146 @@
...
+def flip_pipe(dev):
+    input = fn.external_source(name="input", device=dev)
+    return fn.flip(input, horizontal=True)
+
+
+@pipeline_def
+def conditional_split_merge_pipe(dev):
+    input = fn.external_source(name="input", device=dev)
+    pred = fn.external_source(name="predicate")
+    true_branch, false_branch = fn._conditional.split(input, predicate=pred)
```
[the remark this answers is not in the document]
klecki: Yes, to indicate that this module is internal. I already renamed them from `experimental._split` and `experimental._merge`, do you have better idea?

dali/pipeline/operator/op_schema.h
```diff
@@ -585,20 +590,32 @@ used with DALIDataType, to avoid confusion with `AddOptionalArg<type>(name, doc,
...
-  DLL_PUBLIC bool IsPassThrough(int input_idx, int output_idx) const;
+  DLL_PUBLIC bool IsPassThrough(int input_idx, int output_idx, bool strict = true) const;
 
   /**
    * @brief Does this operator pass through any data
    */
   DLL_PUBLIC bool HasPassThrough() const;
 
+  /**
+   * @brief Does this operator pass through any data as a whole batch to batch
```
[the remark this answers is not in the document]
klecki: done

dali/pipeline/operator/builtin/merge.cc
```diff
@@ -0,0 +1,203 @@
...
+  auto &output = ws.template Output<Backend>(0);
+  const auto &predicate = ws.ArgumentInput("predicate");
+  auto sample_idx_in_input = uniform_array<kMaxGroups>(0);
+
+  WriteTestsDiagnostics(ws);
+
+  if (!pinned_) {
+    // We produce pinned data if the executor said so
+    pinned_ = output.is_pinned();
+    // TODO(klecki): We keep pinedness if it was passed to us:
```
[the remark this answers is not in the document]
klecki: Expanded.

dali/test/python/operator/test_split_merge.py
klecki: I'm waiting on the promised guideline on how to write the tests for the new framework. 
Your example only shows one parameter though, so I'm gonna stick with what I have for now.

dali/test/python/operator/test_split_merge.py
klecki: There is nothing specific that shouldn't allow nesting. I will cover the nesting use-case when we land the if-statement, that would be more interesting case, here I don't think we need it.

dali/test/python/operator/test_split_merge.py
klecki: Done*

dali/pipeline/operator/builtin/merge.cc
```diff
@@ -0,0 +1,203 @@
...
+                                           const TensorList<GPUBackend> &input,
+                                           int input_sample_idx, Workspace &ws) {
+  assert(false && "This codepath should not be executed");
+}
+
+
+template <>
+void Merge<GPUBackend>::FinalizeCopy(Workspace &ws) {
+  assert(false && "This codepath should not be executed");
+}
```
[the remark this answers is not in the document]
klecki: Hmm, FinalizeCopy is actually called, my bad. Also I can just sfinae them out, they must exist so the code can compile.

klecki:
!build

dali/test/python/operator/test_split_merge.py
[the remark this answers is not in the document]
klecki: > user might expect that it's not a part of public API, he's not supposed to use it and the API might change at any point.
This is exactly the point of private API and the intention for the split and merge operators..

szalpal (approved):

klecki:
!build

klecki:
!build

PR #4365, opened 2022-10-17
Remap kernel implementation with NPP

szalpal:
## Category:

## Description:

This PR adds the `NppRemapKernel` - a GPU `RemapKernel` implenentation using NPP.

The main part of the PR is the `NppRemapKernel` class, together with its test included in the `remap_npp_test.cc` file.

Additionally, the following has been added:
* Static and dynamic linking of `nppig` library,

## Additional information:

This PR has a prerequisite and utils PR: #4374 .

### Tests:

- [x] Existing tests apply
- [x] New tests added
  - [ ] Python tests
  - [x] GTests
  - [ ] Benchmark
  - [ ] Other
- [ ] N/A

### Documentation
- [x] Documentation updated
  - [ ] Docstring
  - [x] Doxygen
  - [ ] RST
  - [ ] Jupyter
  - [ ] Other
- [ ] N/A

#### Requirements
- [ ] N/A

**REQ IDs**: N/A

**JIRA TASK**: N/A


include/dali/core/tensor_view.h
```diff
@@ -481,6 +481,22 @@ struct TensorListView<Backend, DataType, DynamicDimensions>
...
+
+  template <int other_sample_ndim>
+  TensorListView(std::nullptr_t, const TensorListShape<other_sample_ndim> &shape)
+      : Base(shape) {}
+
+  template <int other_sample_ndim>
+  TensorListView(std::nullptr_t, TensorListShape<other_sample_ndim> &&shape)
+      : Base(std::move(shape)) {}
+
+  /** @} */
```
szalpal, line 498: These three constructors has already been added to `static_dim` flavour of `TensorListView`. Here I'm adding them also to `DynamicDimensions` flavour.

szalpal:
!build

szalpal:
!build

dali/kernels/imgproc/geom/remap.h
```diff
@@ -15,24 +15,25 @@
...
  * API for Remap operation. Remap applies a generic geometrical transformation to an image.
- * @tparam Backend
- * @tparam T
+ * @tparam Backend Storage backend for data.
+ * @tparam T Type of input and output data.
  */
 template <typename Backend, typename T>
 struct RemapKernel {
   using Border = boundary::Boundary<T>;
+  using MapType = typename std::conditional<std::is_same<T, double>::value, double, float>::type;
```
banasraf, line 36:
```suggestion
  using MapType = std::conditional_t<std::is_same<T, double>::value, double, float>;
```

banasraf, line 36: Not sure about this. The type in the map is in the coordinates domain and `T` is the data type, so those are not really related to each other in any way.

dali/kernels/imgproc/geom/remap_npp.h
```diff
@@ -0,0 +1,162 @@
...
+           TensorListView<Backend, T> output,
+           TensorListView<Backend, const T> input,
+           TensorListView<Backend, const MapType, 2> mapsx,
+           TensorListView<Backend, const MapType, 2> mapsy,
+           span<Box<2, int64_t>> output_rois = {},
+           span<Box<2, int64_t>> input_rois = {},
+           span<DALIInterpType> interpolations = {},
+           span<Border> borders = {}) override {
+    CUDA_CALL(nppSetStream(context.gpu.stream));
+    NppStreamContext npp_ctx;
```
JanuszL, line 72: Shouldn't the npp_ctx be created only once and just reused (not recreated every `Run` invocation)?
Also, shouldn't you create context on your own - like https://github.com/NVIDIA/VideoProcessingFramework/blob/master/PyNvCodec/TC/src/NppCommon.cpp ?

include/dali/core/tensor_view.h
```diff
@@ -650,6 +666,27 @@ TensorListView<StorageBackend, T, ndim> make_tensor_list(T *const *data,
...
+    const std::vector<TensorView<StorageBackend, T, ndim>> &tvs) {
+  std::vector<TensorShape<ndim>> shapes(tvs.size());
+  for (size_t i = 0; i < tvs.size(); i++) {
+    shapes[i] = tvs[i].shape;
+  }
+  TensorListView<StorageBackend, T, ndim> tlv{nullptr, shapes};
+  tlv.data.resize(tvs.size());
+  for (size_t i = 0; i < tvs.size(); i++) {
+    tlv.data[i] = tvs[i].data;
+  }
```
mzient, line 686: You don't need this intermediate vector:
```suggestion
  if (tvs.empty())
    return {};
  TensorListView<StorageBackend, T, ndim> tlv;
  tlv.resize(tvs.size(), tvs[0].dim());
  for (size_t i = 0; i < tvs.size(); i++) {
    tlv.shape.set_tensor_shape(i, tvs[i].shape);
    tlv.data[i] = tvs[i].data;
  }
```

include/dali/core/tensor_view.h
```diff
@@ -650,6 +666,27 @@ TensorListView<StorageBackend, T, ndim> make_tensor_list(T *const *data,
   return { data, std::move(shape) };
 }
 
+template <typename StorageBackend, int ndim, typename T>
+TensorListView<StorageBackend, T, ndim> make_tensor_list(
+    const TensorView<StorageBackend, T, ndim> &tv) {
+  return {tv.data, {tv.shape}};
+}
```
mzient, line 673: This overload seems a bit off. Where do you need it? Can you perhaps use:
```make_tensor_list({tv});```
?

File: dali/kernels/imgproc/geom/remap_npp.h [hunk 6]
```diff
@@ -0,0 +1,162 @@
...
+
+  void Run(KernelContext &context,
+           TensorListView<Backend, T> output,
+           TensorListView<Backend, const T> input,
+           TensorListView<Backend, const MapType, 2> mapsx,
+           TensorListView<Backend, const MapType, 2> mapsy,
+           span<Box<2, int64_t>> output_rois = {},
+           span<Box<2, int64_t>> input_rois = {},
+           span<DALIInterpType> interpolations = {},
+           span<Border> borders = {}) override {
```
banasraf, line 70: Doesn't this function lack some input validation? I guess at least to check `mapsx`, `mapsy` shapes vs. `rois`?

dali/kernels/imgproc/geom/remap_npp.h
```diff
@@ -0,0 +1,162 @@
...
+
+ private:
+  void invoke_remap_kernel(TensorView<Backend, const T> input,
+                           TensorView<Backend, T> output,
+                           TensorView<Backend, const MapType, 2> mapx,
+                           TensorView<Backend, const MapType, 2> mapy,
+                           Box<2, int64_t> input_roi,
+                           Box<2, int64_t> output_roi,
+                           int interpolation, NppStreamContext ctx) {
+    assert(input.shape == output.shape);
```
banasraf, line 119: Is it really a necessary constrain?

szalpal: I believe that yes. No matter if the ROI is specified or not, the output size should always match the input size. Unless you believe otherwise?

dali/test/test_tensors.h
```diff
@@ -55,7 +55,7 @@ class TestTensorList {
       cpumem_ = mm::alloc_raw_unique<char, mm::memory_kind::host>(size, 256);
       if (gpumem_)
         CUDA_CALL(cudaMemcpyAsync(cpumem_.get(), gpumem_.get(), size,
-                                  cudaMemcpyDeviceToHost, stream));
+                                  cudaMemcpyDefault, stream));
```
mzient, line 58: This will slow thing down - is it _really_ needed? After all, we _know_ that it's a D2H copy.

dali/kernels/kernel_traits.h
```diff
@@ -258,6 +258,33 @@ using check_kernel = detail::check_kernel_has_run<Kernel, assert_>;
...
+/**
+ * Checks, if the Tuple contains a given type T.
+ *
+ * Usage example:
+ * using Tup = std::tuple<int, double, MyType>;
+ * static_assert(has_type_v<double, Tup>);  // this will pass
+ * static_assert(has_type_v<float,  Tup>);  // this will fail
+ */
+template<typename T, typename Tuple>
+constexpr bool has_type_v = detail::has_type<T, Tuple>::type::value;
```
mzient, line 286: This doesn't belong in this file. The right file for this would be `include/core/tuple_helpers.h`
Also, it should be called something like "contains".

dali/kernels/imgproc/geom/remap_npp.h
mzient: It's not necessary and it's not the right check.
The output must have the same spatial dimensions as the _mapping_, not the input.
Ultimately, output elements are defined as:
`out[pos] = sample_at(input, ivec2(mapping_x[pos], mapping_y[pos]))`

dali/kernels/imgproc/geom/remap_npp.h
JanuszL: According to the samples - https://github.com/NVIDIA/CUDALibrarySamples/blob/master/NPP/findContour/findContour.cpp#L240, yes you may need to replace the stream (although I don't expect it changes run to run in DALI operators), but other than that the rest can be reused.

dali/kernels/imgproc/geom/remap_npp_test.cc
```diff
@@ -0,0 +1,295 @@
...
+ * @return Number of outlying pixels.
+ */
+template<typename T>
+size_t count_outlying_pixels(const cv::Mat &diff) {
+  const auto &p = reinterpret_cast<const T *>(diff.data);
+  int ret = 0;
+  for (int i = 0; i < diff.rows * diff.cols * diff.channels(); i += diff.channels()) {
+    // Effectively compare to 0, but secure from float precision.
+    auto res = std::accumulate(p, p + diff.channels(), 0.f);
+    if (-1e-7 > res || res > 1e-7) ret++;
```
mzient, line 90: I don't think we need or even _wait_ this level of similarity with OpenCV. OpenCV uses approximage linear interpolation; we don't want to replicate that.

dali/kernels/imgproc/geom/remap_npp_test.cc
```diff
@@ -0,0 +1,295 @@
...
+  }
+  return ret;
+}
```
mzient, line 93: This implementation is flawed - OpenCV matrices can have row padding.
```suggestion
size_t count_outlying_pixels(const cv::Mat &diff) {
  cv::Mat tmp = cv::abs(diff);
  cv::threshold(tmp, tmp,1e-7, 1.0, cv.THRESH_BINARY);
  return cv::countNonZero(tmp);
}
```

dali/kernels/imgproc/geom/remap.h
szalpal: Done

dali/kernels/imgproc/geom/remap_npp.h
szalpal: Done

dali/kernels/imgproc/geom/remap_npp.h
szalpal: Corrected the checks

dali/test/test_tensors.h
szalpal: I checked [cudaMemcpy docs](https://docs.nvidia.com/cuda/cuda-runtime-api/group__CUDART__MEMORY.html#group__CUDART__MEMORY_1gc263dbe6574220cc776b45438fc351e8) as well as [cudaMemcpyDefault definition](https://docs.nvidia.com/cuda/cuda-runtime-api/group__CUDART__TYPES.html#group__CUDART__TYPES_1g18fa99055ee694244a270e4d5101e95b) and there's no mention of higher resource usage, only that it uses UVA. Do you happen to have any reference for this?

Also, I believe CUDA engineers point out, that there are virtually no difference between these.

dali/kernels/kernel_traits.h
szalpal: Done

dali/operators/util/npp.h
```diff
@@ -31,6 +31,23 @@
...
+    case DALI_INTERP_NN:
+      return NPPI_INTER_NN;
+    case DALI_INTERP_LINEAR:
+      return NPPI_INTER_LINEAR;
+    case DALI_INTERP_CUBIC:
+      return NPPI_INTER_CUBIC;
+    case DALI_INTERP_LANCZOS3:
+      return NPPI_INTER_LANCZOS3_ADVANCED;
+    default:
+      DALI_FAIL(make_string("Unsupported interpolation type. Interpolation type (", interp,
```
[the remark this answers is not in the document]
szalpal: Done

include/dali/core/tensor_view.h
szalpal: My feeling is that TensorView might successfully be a 1-sample TensorListView. Therefore I've put the function. I guess using `make_tensor_list({tv});` does not hurt, but probably `make_tensor_list(tv);` is more obvious.

@banasraf , @awolant , how do you feel about this overload?

banasraf (approved):

dali/kernels/imgproc/geom/remap_npp_test.cc
szalpal: Unfortunately, `cv::countNonZero` works on 1-channel images, while here I need to operate on both 1-channel and 3-channel. As I wrote in the comment above, the cv::Mats in this file wrap the contiguous memory, so maybe this function isn't bad after all?

szalpal:
!build

dali/kernels/imgproc/geom/remap_npp_test.cc
mzient: In that case a big fat comment is in order - and perhaps also an assert
```
assert(mat.ptr<T>(1) - mat.ptr<T>(0) == mat.cols);
```
or something like that.

include/dali/core/tensor_view.h
mzient: For me it looks like it strips the outer dimension and converts it to the batch.
NCHW -> slice across N and produce N x CHW tensor.

dali/test/test_tensors.h
szalpal: Reverted

dali/kernels/imgproc/geom/remap_npp_test.cc
szalpal: Added comment and assert

dali/kernels/imgproc/geom/remap_npp_test.cc
szalpal: Good point. I've changed the way the input data is generated to avoid the halfway cases. Since this is a test for remap and not interpolation nor rounding, I believe that's legal.

include/dali/core/tensor_view.h
szalpal: Added documentation to explain what this function does.

dali/kernels/imgproc/geom/remap_npp.h
szalpal: True, I've adjusted the kernel code to reflect the usage of NppStream from the example provided.

szalpal:
!build

awolant (approved):
Reading this PR I started to think what benefit do we get from implementing this as a kernel? Maybe it could have been done in the operator without a kernel? Main thing you do here is wrapping NPP and it looks to me like the kernel just gives another layer of unnecessary abstraction and boilerplate code. 
I do not suggest changing it know, when it is ready. It was just an afterthought when I read this. Probably something for us to think about in the future.

mzient (approved):

szalpal:
!build

PR #4366, opened 2022-10-17
Add inflate operator

stiepan:
## Category:

## Description:
Adds inflate operator that allows to decompress batch of compressed samples (and batch of sequences of compressed samples). For now, only LZ4 compressin is supported. 

The decompression is done with nvCOMP library. Note, nvCOMP is not open source, the nvCOMP SDK licencse is included in the Acknwoledgment: https://github.com/NVIDIA/DALI/pull/4368. 

If the samples contain multiple compressed chunks, either chunk_sizes or chunk_offsets parameter must be provided. If both are provided, the order of the chunks in the input and output can be different.

## Additional information:
For now, we build with the nvCOMP (and thus without the operator) only for x86 due to incompatible glibcxx in the nvCOMP's aarch build.

* The new operator is added
* Python tests are added
* CI dockerfiles and wheel building scripts are affected

### Tests:

- [ ] Existing tests apply
- [x] New tests added
  - [x] Python tests
  - [ ] GTests
  - [ ] Benchmark
  - [ ] Other
- [ ] N/A

### Documentation
- [x] Documentation updated
  - [x] Docstring
  - [x] Doxygen
  - [ ] RST
  - [ ] Jupyter
  - [ ] Other
- [ ] N/A

#### Requirements
- [ ] N/A

**REQ IDs**: DECOMP.02-08.

**JIRA TASK**: DALI-3049


dali/operators/decoder/inflate/inflate.cc
```diff
@@ -0,0 +1,89 @@
...
+        layout="HWC",
+        sequence_extent="F")
+    return sequences
+
+)code")
+    .NumInput(1)
+    .NumOutput(1)
+    .AddArg(inflate::shapeArgName, "The shape of the output (inflated) chunk.", DALI_INT_VEC, true)
+    .AddOptionalTypeArg(inflate::dTypeArgName, "The output (inflated) data type.", DALI_NO_TYPE)
+    .AddOptionalArg(inflate::offsetArgName, R"code("A list of offsets within the input sample
```
mzient, line 57: It's probably a matter of taste, but I don't like these constants - they make the code only infinitesimally "harder" but much harder to read.

awolant (approved):

dali/operators/decoder/inflate/inflate.cc
```diff
@@ -0,0 +1,89 @@
...
+
+DALI_SCHEMA(experimental__Inflate)
+    .DocStr(R"code(Inflates/decompresses the input using specified decompression algorithm.
+
+The input must be a 1D tensor of bytes (uint8). Passing the ``shape`` and ``dtype`` of the
+decompressed samples is required.
+
+Each input sample can either be a single compressed chunk or consist of multiple
+compressed chunks that have the same shape and type when inflated, so that they can be
+be merged into a single tensor where the leftmost extent of the tensor corresponds
```
mzient, line 29:
```suggestion
be merged into a single tensor where the outermost extent of the tensor corresponds
```

stiepan: The real use case it targets are 3D sequences of single channel HW images. Each image is compressed separately and the results are concatenated. Using `chunks_offsets` or `chunks_sizes` you can do the reverse: uncompress the chunks into a sequence.

dali/operators/decoder/inflate/inflate_gpu.cu
```diff
@@ -0,0 +1,45 @@
...
+namespace inflate {
+
+namespace detail {
+
+__global__ void FillTheTails(void* const* chunks, size_t* actual_sizes, size_t* output_sizes) {
+  uint8_t* chunk_data = static_cast<uint8_t*>(chunks[blockIdx.z]);
+  auto output_size = output_sizes[blockIdx.z];
+  auto inflated_size = actual_sizes[blockIdx.z];
+  for (auto idx = inflated_size + threadIdx.x + blockIdx.x * blockDim.x; idx < output_size;
+       idx += blockDim.x * gridDim.x) {
```
mzient, line 29: Possible overflow.
```suggestion
       idx += static_cast<size_t>(gridDim.x) * blockDim.x) {
```

dali/operators/decoder/inflate/inflate_params.h
```diff
@@ -0,0 +1,397 @@
...
+
+namespace inflate {
+
+constexpr static const char *dTypeArgName = "dtype";
+constexpr static const char *algArgName = "algorithm";
+constexpr static const char *shapeArgName = "shape";
+constexpr static const char *offsetArgName = "chunks_offsets";
+constexpr static const char *sizeArgName = "chunks_sizes";
+constexpr static const char *layoutArgName = "layout";
+constexpr static const char *sequenceLayoutArgName = "sequence_extent";
```
mzient, line 38: As mentioned before, I consider this actually harmful. Code like this is harder to read and harder to debug. Also, there's a tendency for these names to drift away. Actually, it seems that this already happened here with
`sequenceLayoutArgName = "sequence_extent";` - is it `extent` or `layout`?

dali/operators/decoder/inflate/inflate_params.h
```diff
@@ -0,0 +1,397 @@
...
+  }
+}
+
+inline InflateAlg parse_inflate_alg(std::string str) {
+  std::transform(str.begin(), str.end(), str.begin(), [](auto c) { return std::tolower(c); });
+  if (str == "lz4") {
+    return InflateAlg::LZ4;
+  }
+  DALI_FAIL(make_string("Unknown inflate algorithm was specified for `", algArgName,
+                        "` argument: `", str, "`."));
```
mzient, line 59: I don't think there are multiple arguments where we can specify an algorithm. Also, don't use inverted quotes for a string.
```suggestion
  DALI_FAIL(make_string("Unknown inflate algorithm: \"", str, "\"."));
```

dali/operators/decoder/inflate/inflate_params.h
```diff
@@ -0,0 +1,397 @@
...
+template <typename Backend>
+class ShapeParams {
+ public:
+  explicit ShapeParams(const OpSpec &spec)
+      : spec_{spec},
+        shape_{shapeArgName, spec},
+        chunk_offsets_{offsetArgName, spec},
+        chunk_sizes_{sizeArgName, spec},
+        layout_{spec.GetArgument<TensorLayout>(layoutArgName)},
+        sequence_extent_{spec.GetArgument<TensorLayout>(sequenceLayoutArgName)} {
```
mzient, line 71: We usually call such arguments `"xxx_axis_name"`.

dali/operators/decoder/inflate/inflate_params.h
```diff
@@ -0,0 +1,397 @@
...
+    return shape;
+  }
+
+  const OpSpec &spec_;
+
+  ArgValue<int, DynamicDimensions> shape_;
+  ArgValue<int, 1> chunk_offsets_;
+  ArgValue<int, 1> chunk_sizes_;
+  TensorLayout layout_;
+  TensorLayout sequence_extent_;
```
mzient, line 383: We don't really need a tensor layout here - we need a single axis name.
```suggestion
  char sequence_axis_name_ = 0;
```

dali/operators/decoder/inflate/inflate_params.h
```diff
@@ -0,0 +1,397 @@
...
+      TensorShape<> sample_shape(data, data + sample_dim);
+      max_output_sample_vol_ = std::max(max_output_sample_vol_, volume(sample_shape));
+      shape.set_tensor_shape(sample_idx, sample_shape);
+    }
+    return shape;
+  }
+
+  const OpSpec &spec_;
+
+  ArgValue<int, DynamicDimensions> shape_;
```
mzient, line 379: Isn't it always 1D?
```suggestion
  ArgValue<int, 1> shape_;
```

dali/operators/decoder/inflate/inflate_params.h
```diff
@@ -0,0 +1,397 @@
...
+        layout_{spec.GetArgument<TensorLayout>(layoutArgName)},
+        sequence_extent_{spec.GetArgument<TensorLayout>(sequenceLayoutArgName)} {
+    DALI_ENFORCE(sequence_extent_.size() == 1,
+                 make_string("The `", sequenceLayoutArgName, "` must be a single character, got `",
+                             sequence_extent_, "`."));
+  }
+
+  void ProcessInputArgs(const Workspace &ws, int batch_size) {
+    SetupOffsetsAndSizes(ws);
+    shape_.Acquire(spec_, ws, batch_size, ArgValue_EnforceUniform, ArgShapeFromSize<1>{});
```
mzient, line 79: Can't the `shape_` be always 1D? Then you could:
```suggestion
    shape_.Acquire(spec_, ws, batch_size, ArgValue_EnforceUniform);
```

dali/operators/decoder/inflate/inflate_params.h
```diff
@@ -0,0 +1,397 @@
...
+    SetupOutputShape(shape_.get());
+    SetupOutputLayout();
+  }
+
+  auto GetMaxOutChunkVol() const {
+    return max_output_sample_vol_;
+  }
+
+  auto GetTotalChunksNum() const {
```
mzient, line 88: Nitpick (and it applies... everywhere):
item number = number of item*s*
so
```suggestion
  auto GetTotalNumChunks() const {
```
or
```suggestion
  auto GetTotalChunkNum() const {
```

dali/operators/decoder/inflate/inflate_params.h
```diff
@@ -0,0 +1,397 @@
...
+  void SetupOutputLayout() {
+    if (layout_.empty()) {
+      return;
+    }
+    auto sample_dim = GetOutputShape().sample_dim() - HasChunks();
+    DALI_ENFORCE(
+        layout_.size() == sample_dim,
+        make_string("The size of inflated chunks' layout must match the number of extents in "
+                    "the output shape. However, got the layout `",
+                    layout_, "` while the chunks output shape has ", sample_dim, " extent(s)."));
```
mzient, line 137: This message is quite convoluted. How about just:
```suggestion
        make_string("The layout \"", layout_, "\" has a different number of dimensions (", layout_.size(),
                    ") than the requested output shape (", sample_dim, ")."));
```

dali/operators/decoder/inflate/inflate_params.h
```diff
@@ -0,0 +1,397 @@
...
+
+  void AcquireInferOffsetsSizes(const Workspace &ws) {
+    assert(HasChunks());
+
+    const auto copy_to_vector = [](auto &v, const auto &tlv) {
+      v.clear();
+      v.reserve(tlv.shape.num_elements());
+      for (int sample_idx = 0; sample_idx < tlv.shape.num_samples(); sample_idx++) {
+        auto num_elements = tlv.shape[sample_idx].num_elements();
+        const auto *chunks_data = tlv.data[sample_idx];
```
mzient, line 175: ...and suddenly a function with a generic name and generic arguments gets some semantics (chunks_data).
What does this name mean, btw? Is it a single chunk or multiple chunks described somehow by one pointer?

dali/operators/decoder/inflate/inflate_params.h
```diff
@@ -0,0 +1,397 @@
...
+      offsets_.resize(batch_size, 0);
+      sizes_.clear();
+      sizes_.reserve(batch_size);
+      for (int sample_idx = 0; sample_idx < batch_size; sample_idx++) {
+        sizes_.push_back(in_shape[sample_idx].num_elements());
+      }
+    }
+  }
+
+  void AcquireInferOffsetsSizes(const Workspace &ws) {
```
mzient, line 167:
```suggestion
  void SetupOffsetsAndSizes(const Workspace &ws) {
```

dali/operators/decoder/inflate/inflate_params.h
```diff
@@ -0,0 +1,397 @@
...
+
+    const auto copy_to_vector = [](auto &v, const auto &tlv) {
+      v.clear();
+      v.reserve(tlv.shape.num_elements());
+      for (int sample_idx = 0; sample_idx < tlv.shape.num_samples(); sample_idx++) {
+        auto num_elements = tlv.shape[sample_idx].num_elements();
+        const auto *chunks_data = tlv.data[sample_idx];
+        v.insert(v.end(), chunks_data, chunks_data + num_elements);
+      }
+    };
```
mzient, line 178: Why do you need this flattening?

dali/operators/decoder/inflate/inflate_gpu.cc
```diff
@@ -0,0 +1,151 @@
...
+    auto total_chunks_num = params_.GetTotalChunksNum();
+    const auto &offsets = params_.GetInChunkOffsets();
+    input_ptrs_.clear();
+    input_ptrs_.reserve(total_chunks_num);
+    for (int64_t sample_idx = 0, chunk_flat_idx = 0; sample_idx < batch_size; sample_idx++) {
+      auto num_chunks = num_chunks_per_sample[sample_idx].num_elements();
+      const uint8_t *sample_base_ptr = in_view.data[sample_idx];
+      for (int chunk_idx = 0; chunk_idx < num_chunks; chunk_idx++, chunk_flat_idx++) {
+        input_ptrs_.push_back(sample_base_ptr + offsets[chunk_flat_idx]);
+      }
```
mzient, line 98: So, you've flattened chunk offsets and now you reconstitute the mapping between samples and chunk offsets.
```suggestion
      auto &&offsets = chunk_offsets_[sample_idx];
      for (int chunk_idx = 0; chunk_idx < num_chunks; chunk_idx++) {
        input_ptrs_.push_back(sample_base_ptr + offsets.data[chunk_idx]);
      }
```

dali/operators/decoder/inflate/inflate_params.h
```diff
@@ -0,0 +1,397 @@
...
+        DALI_ENFORCE(chunk_size > 0, is_last_chunk ? last_chunk_error_msg(sample_idx) :
+                                                     non_monotone_error(sample_idx));
+        sizes_.push_back(chunk_size);
+      }
+    }
+    assert(offsets_.size() == sizes_.size());
+  }
+
+  template <typename TLShape>
+  void InferOffsetsFromSizes(const TLShape &in_shape) {
```
mzient, line 313: Can't it be something more concrete?
```suggestion
  template <int ndim>
  void InferOffsetsFromSizes(const TensorListShape<ndim> &in_shape) {
```

dali/operators/decoder/inflate/inflate_params.h
```diff
@@ -0,0 +1,397 @@
...
+            make_string("Input chunk cannot exceed the sample size. However, for a sample of idx ",
+                        sample_idx, ", got a chunk that starts at the position ",
+                        chunk_offsets[chunk_idx], " and has size ", chunk_sizes[chunk_idx],
+                        " while the sample contains only ", sample_num_bytes, " bytes."));
+      }
+    }
+  }
+
+  template <typename TLShape>
+  void InferSizesFromOffsets(const TLShape &in_shape) {
```
mzient, line 278:
```suggestion
  template <int ndim>
  void InferSizesFromOffsets(const TensorListShape<ndim> &in_shape) {
```

dali/operators/decoder/inflate/inflate_params.h
```diff
@@ -0,0 +1,397 @@
...
+      }
+    };
+
+    const auto validate_offset = [](auto sample_idx, auto sample_num_bytes, auto chunk_offset) {
+      DALI_ENFORCE(chunk_offset >= 0,
+                   make_string("Input chunks offsets must be non-negative. Got ", chunk_offset,
+                               " offset for sample of idx ", sample_idx, "."));
+      DALI_ENFORCE(
+          chunk_offset < sample_num_bytes,
+          make_string("Input chunks offsets cannot exceed the sample size. Got chunk offset ",
```
mzient, line 186: Strictly speaking, the offset == sample_num_bytes doesn't _exceed_ the sample size.
```suggestion
          make_string("Input chunks offsets cannot point past the last byte. Got chunk offset ",
```

dali/operators/decoder/inflate/inflate_params.h
```diff
@@ -0,0 +1,397 @@
...
+    if (HasChunks()) {
+      AcquireInferOffsetsSizes(ws);
+    } else {
+      chunks_per_sample_ = uniform_list_shape(batch_size, TensorShape<1>{1});
+      offsets_.resize(batch_size, 0);
+      sizes_.clear();
+      sizes_.reserve(batch_size);
+      for (int sample_idx = 0; sample_idx < batch_size; sample_idx++) {
+        sizes_.push_back(in_shape[sample_idx].num_elements());
+      }
```
mzient, line 163: How about going the other way round and creating a TensorListView from this instead of flattening the data obtained from ArgValue?

dali/operators/decoder/inflate/inflate_params.h
```diff
@@ -0,0 +1,397 @@
...
+      if (num_chunks == 0) {
+        continue;
+      }
+      auto in_chunk_size = in_shape[sample_idx].num_elements();
+      for (int chunk_idx = 0; chunk_idx < num_chunks; chunk_idx++, flat_chunk_idx++) {
+        bool is_last_chunk = chunk_idx == num_chunks - 1;
+        int64_t next_chunk_offset = is_last_chunk ? in_chunk_size : offsets_[flat_chunk_idx + 1];
+        int64_t chunk_size = next_chunk_offset - offsets_[flat_chunk_idx];
+        DALI_ENFORCE(chunk_size > 0, is_last_chunk ? last_chunk_error_msg(sample_idx) :
+                                                     non_monotone_error(sample_idx));
```
mzient, line 305: This cannot happen for the last chunk - offset validation takes care of that.
```suggestion
        assert(chunk_size > 0 || !is_last_chunk);
        DALI_ENFORCE(chunk_size > 0, non_monotone_error(sample_idx));
```

dali/operators/decoder/inflate/inflate_params.h
```diff
@@ -0,0 +1,397 @@
...
+  }
+
+  template <typename TLShape>
+  void InferSizesFromOffsets(const TLShape &in_shape) {
+    assert(offsets_.size() == chunks_per_sample_.num_elements());
+    auto last_chunk_error_msg = [](auto sample_idx) {
+      return make_string(
+          "The last chunk's offset exceeds the size of the sample for sample of idx ", sample_idx,
+          ".");
+    };
```
mzient, line 284: This will never happen - offset validation makes it impossible.

dali/operators/decoder/inflate/inflate_params.h
```diff
@@ -0,0 +1,397 @@
...
+constexpr static const char *offsetArgName = "chunks_offsets";
+constexpr static const char *sizeArgName = "chunks_sizes";
+constexpr static const char *layoutArgName = "layout";
+constexpr static const char *sequenceLayoutArgName = "sequence_extent";
+
+enum class InflateAlg {
+  LZ4
+};
+
+inline std::string to_string(const InflateAlg &alg) {
```
mzient, line 44: Nitpick
```suggestion
inline std::string to_string(InflateAlg alg) {
```

dali/operators/decoder/inflate/inflate_gpu.cc
```diff
@@ -0,0 +1,151 @@
...
+    TYPE_SWITCH(output.type(), type2id, Out, INFLATE_SUPPORTED_TYPES, (
+      BOOL_SWITCH(params_.HasChunks(), HasChunks, (
+        SetupOutChunksTyped<HasChunks, Out>(output);
+      ));  //NOLINT
+    ), DALI_FAIL(  //NOLINT
+      make_string("Unsupported output type was specified for GPU inflate LZ4 operator: `",
+                  output.type(), "`.")));
+  }
+
+  template <int sequence_dim, typename Out, typename TL>
```
mzient, line 113:
```suggestion
  template <bool has_chunks, typename Out, typename TL>
```

dali/operators/decoder/inflate/inflate_gpu.cc
```diff
@@ -0,0 +1,151 @@
...
+  void SetupOutChunksTyped(TL &output) {
+    auto out_view = view<Out>(output);
+    auto batch_size = out_view.shape.num_samples();
+    auto total_chunks_num = params_.GetTotalChunksNum();
+    inflated_ptrs_.clear();
+    inflated_ptrs_.reserve(total_chunks_num);
+    inflated_sizes_.clear();
+    inflated_sizes_.reserve(total_chunks_num);
+    for (int64_t sample_idx = 0; sample_idx < batch_size; sample_idx++) {
+      auto sample_range = sequence_utils::unfolded_view_range<sequence_dim>(out_view[sample_idx]);
```
mzient, line 123:
```suggestion
      const int dims_to_unfold = has_chunks ? 1 : 0;
      auto sample_range = sequence_utils::unfolded_view_range<dims_to_unfold>(out_view[sample_idx]);
```

dali/operators/decoder/inflate/inflate_gpu.cu
```diff
@@ -0,0 +1,45 @@
...
+    chunk_data[idx] = 0;
+  }
+}
+
+}  // namespace detail
+
+void FillTheTails(DALIDataType output_type, int batch_size, void* const* chunks_dev,
+                  size_t* actual_sizes_dev, size_t* output_sizes_dev, cudaStream_t stream) {
+  dim3 grid(1, 1, batch_size);
```
mzient, line 38: This grid is very small. How likely it is that this will happen at all and that there's by a substantial space to fill?

CMakeLists.txt
```diff
@@ -109,6 +109,9 @@ cmake_dependent_option(BUILD_NVOF "Build with NVIDIA OPTICAL FLOW SDK support" O
                        "NOT BUILD_DALI_NODEPS" OFF)
 cmake_dependent_option(BUILD_NVML "Build with NVIDIA Management Library (NVML) support" ON
                        "NOT BUILD_DALI_NODEPS" OFF)
+
+cmake_dependent_option(BUILD_NVCOMP "Build with nvCOMP support" OFF
+                       "NOT BUILD_DALI_NODEPS" OFF)
```
mzient, line 114: Shouldn't it be `ON` by default?

dali/operators/decoder/inflate/inflate.cc
```diff
@@ -0,0 +1,89 @@
...
+                    std::vector<int>{}, true)
+    .AddOptionalArg(inflate::algArgName, R"code(Algorithm to be used to decode the data.
+
+Currently only ``LZ4`` is supported.)code",
+                    "LZ4")
+    .AddOptionalArg(inflate::layoutArgName, R"code(Layout of the output (inflated) chunk.
+
+If the samples consist of multiple chunks, additionally, the ``sequence_extent`` extent
+will be added to the beginning of the specified layout.)code",
+                    TensorLayout(""))
```
mzient, line 77:
```suggestion
    .AddOptionalArg<TensorLayout>(inflate::layoutArgName, R"code(Layout of the output (inflated) chunk.

If the samples consist of multiple chunks, additionally, the ``sequence_extent`` extent
will be added to the beginning of the specified layout.)code",
                    nullptr)
```

dali/operators/decoder/inflate/inflate.cc
```diff
@@ -0,0 +1,89 @@
...
+    .AddOptionalArg(inflate::sequenceLayoutArgName, R"code(The name for the sequence extent.
```
mzient, line 78:
```suggestion
    .AddOptionalArg(inflate::sequenceLayoutArgName, R"code(The name for the sequence axis.
```

dali/operators/decoder/inflate/inflate.cc
```diff
@@ -0,0 +1,89 @@
...
+
+If the samples consist of multiple chunks, the extra sequence extent will be added as the leftmost
+extent to the output tensor. By default it is assumed to be video frames,
+hence the default ``F`` extent.
```
mzient, line 82:
```suggestion
If the samples consist of multiple chunks, an extra outer dimension will be added to the output tensor.
By default it is assumed to be video frames, hence the default label 'F'
```

dali/operators/decoder/inflate/inflate_params.h
mzient: After reading the PR, this argument should be called `sequence_axis` (or perhaps even `sequence_axis_name`) - that's the nomenclature we use in other operators that deal with axis labels.

dali/operators/decoder/inflate/inflate.cc
```diff
@@ -0,0 +1,89 @@
...
+        shape=uncompressed_hwc_shape,
+        layout="HWC",
+        sequence_extent="F")
+    return sequences
+
+)code")
+    .NumInput(1)
+    .NumOutput(1)
+    .AddArg(inflate::shapeArgName, "The shape of the output (inflated) chunk.", DALI_INT_VEC, true)
+    .AddOptionalTypeArg(inflate::dTypeArgName, "The output (inflated) data type.", DALI_NO_TYPE)
```
mzient, line 56: Don't add a default where you don't need one.
```suggestion
    .AddOptionalTypeArg(inflate::dTypeArgName, "The output (inflated) data type.")
```

dali/operators/decoder/inflate/inflate.cc
```diff
@@ -0,0 +1,89 @@
...
+    .AddOptionalArg(inflate::offsetArgName, R"code("A list of offsets within the input sample
+describing where the consecutive chunks begin.
+
+If the ``chunks_sizes`` is not specified, it is assumed that the chunks are densely packed
+in the input tensor and the last chunk ends with the sample's end.)code",
+                    std::vector<int>{}, true)
```
mzient, line 62:
```suggestion
    .AddOptionalArg<std::vector<int>>(inflate::offsetArgName, R"code("A list of offsets within the input sample
describing where the consecutive chunks begin.

If the ``chunks_sizes`` is not specified, it is assumed that the chunks are densely packed
in the input tensor and the last chunk ends with the sample's end.)code",
                    nullptr, true)
```

dali/operators/decoder/inflate/inflate.cc
```diff
@@ -0,0 +1,89 @@
...
+    .AddOptionalArg(inflate::sizeArgName,
```
mzient, line 63: Likewise - don't use a default value when it's invalid.

dali/operators/decoder/inflate/inflate.h
```diff
@@ -0,0 +1,105 @@
...
+        IsFloatingPoint(dtype_) || IsIntegral(dtype_),
+        make_string("The inflate output type must have floating point or integral type, got `",
+                    dtype_, "` instead."));
+  }
+
+  bool SetupImpl(std::vector<OutputDesc> &output_desc, const Workspace &ws) override {
+    auto input_type = ws.GetInputDataType(0);
+    auto input_shape = ws.GetInputShape(0);
+    DALI_ENFORCE(
+        input_type == type2id<uint8_t>::value,
```
mzient, line 48:
```suggestion
        input_type == DALI_UINT8,
```
?

dali/operators/decoder/inflate/inflate_gpu.cc
```diff
@@ -0,0 +1,151 @@
...
+    auto stream = ws.stream();
+
+    kernels::DynamicScratchpad scratchpad({}, stream);
+    size_t *actual_out_sizes = scratchpad.AllocateGPU<size_t>(total_chunks_num);
+
+    size_t *in_sizes;
+    void const **in;
+    size_t *out_sizes;
+    void **out;
+    std::tie(in_sizes, in, out_sizes, out) = scratchpad.ToContiguousGPU(
```
mzient, line 64: Hmmm.... would that work?
```suggestion
    auto [in_sizes, in, out_sizes, out] = scratchpad.ToContiguousGPU(
```

dali/operators/decoder/inflate/inflate_params.h
stiepan: I follwed requirements specified by @klecki that forsee supporting more compression algorithms. It seems reasonable.

dali/operators/decoder/inflate/inflate_params.h
stiepan: I though about it, but I do need flattened size_t sizes in the end, so decided on not going over the top with abstraction here and just preprocess the argument to the format that is really needed by the operator.

dali/operators/decoder/inflate/inflate_gpu.cu
stiepan: User is expected to pass the actual output size through shape and dtype, this is just to prevent leaking uninited memory. Maybe we could make that a feature at some point (padding samples in a sequence), but for now it is only for the incorrect input case.

dali/operators/decoder/inflate/inflate_gpu.cc
stiepan: I wasn't sure if it end up as .cu or .cc file, but now, yeah, sure.

dali/operators/decoder/inflate/inflate.h
stiepan: done

dali/operators/decoder/inflate/inflate.cc
stiepan: done

dali/operators/decoder/inflate/inflate.cc
stiepan: done

dali/operators/decoder/inflate/inflate.cc
stiepan: done

dali/operators/decoder/inflate/inflate.cc
stiepan: done

dali/operators/decoder/inflate/inflate.cc
stiepan: done

dali/operators/decoder/inflate/inflate_gpu.cc
stiepan: done

dali/operators/decoder/inflate/inflate_gpu.cc
stiepan: done

dali/operators/decoder/inflate/inflate_params.h
stiepan: done

dali/operators/decoder/inflate/inflate_params.h
stiepan: Nice catch.

dali/operators/decoder/inflate/inflate_params.h
stiepan: removed it.

dali/operators/decoder/inflate/inflate_params.h
stiepan: done

dali/operators/decoder/inflate/inflate_params.h
stiepan: Sure, thanks.

dali/operators/decoder/inflate/inflate_params.h
stiepan: done

dali/operators/decoder/inflate/inflate_params.h
stiepan: I want to support passing scalars here too.

dali/operators/decoder/inflate/inflate_params.h
stiepan: done

dali/operators/decoder/inflate/inflate_params.h
stiepan: done

dali/operators/decoder/inflate/inflate_params.h
stiepan: Thanks, I renamed the arg.

dali/operators/decoder/inflate/inflate_gpu.cu
stiepan: I launch them super small, but maybe that is more futureproof.

dali/operators/decoder/inflate/inflate_params.h
```diff
@@ -0,0 +1,389 @@
...
+
+namespace dali {
+
+namespace inflate {
+
+constexpr static const char *dTypeArgName = "dtype";
+constexpr static const char *algArgName = "algorithm";
+constexpr static const char *shapeArgName = "shape";
+constexpr static const char *offsetArgName = "chunks_offsets";
+constexpr static const char *sizeArgName = "chunks_sizes";
```
mzient, line 36: Grammar: Offsets of chunks = chunk offsets
```suggestion
constexpr static const char *offsetArgName = "chunk_offsets";
constexpr static const char *sizeArgName = "chunk_sizes";
```

stiepan: done

dali/test/python/test_optical_flow.py
```diff
@@ -28,19 +28,6 @@
 is_of_supported_var = None
 
 
-def get_arch(device_id=0):
```
stiepan, line 31: Moved it to test_utils

JanuszL:
Please adjust https://github.com/NVIDIA/DALI/blob/main/conda/recipe/build.sh as well.

docker/Dockerfile.cuda117.x86_64.deps
```diff
@@ -22,3 +22,13 @@ RUN NVJPEG2K_VERSION=0.5.0.25-1 && \
     cp /usr/include/nvjpeg2k* /usr/local/cuda/include/ && \
     cp /usr/lib/x86_64-linux-gnu/libnvjpeg2k* /usr/local/cuda/lib64/ && \
     rm -rf /var/lib/apt/lists/*
+
+RUN NVCOMP_VERSION=2.4 && \
```
JanuszL, line 26: Can you merge it with L14 (where we install nvJPEG2k and libcufile)?

CMakeLists.txt
mzient: Well, in case of NVJPEG2k you can set it to on and it will disable itself if the library isn't found.

dali/operators/decoder/inflate/inflate_params.h
mzient: More algorithms - yes. What I meant is multiple _arguments_. Besides, the message doesn't look good:

> Unknown inflate algorithm was specified for 'algorithm' argument: 'wtf'

I think the shorter one conveys just as much information:

> Unknown inflate algorithm 'wtf'

dali/operators/decoder/inflate/inflate_params.h
mzient: > the user wouldn't be able to pass a single interger as a shape.
Are you sure? I think it would only restrict the use of 0D tensors (if at all), while plain scalars would work just fine.

dali/operators/decoder/inflate/inflate.cc
mzient: `outermost` still applies - we never use `leftmost` in this context (the word appears 6 times in our code base and refers to position within image/array, not to dimensions). There are 67 occurrences of `outermost`.

dali/operators/decoder/inflate/inflate_params.h
stiepan: done

dali/operators/decoder/inflate/inflate_params.h
stiepan: done

CMakeLists.txt
stiepan: For example for cufile we have OFF by default. I am not sure if overriding the option silently rather than explicitly failing if we cannot build requested conf is a good way to go.

dali/operators/decoder/inflate/inflate_params.h
mzient: OK.

dali/test/python/operator/test_inflate.py
```diff
@@ -0,0 +1,368 @@
...
+    pipe.build()
+    for iter_size in iteration_sizes:
+        inflated, baseline = pipe.run()
+        check_batch(inflated, baseline, iter_size)
+
+
+@has_operator("experimental.inflate")
+def test_sample_inflate():
+    seed = 42
+    for batch_size in [1, 8, 64, 256, 348]:
```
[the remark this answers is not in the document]
stiepan: Not really tbh, 1 is "special case", 8 is small, 348 big and the remaining 2 medium ones. I though there may be some issues that would show up for small or big batch sizes specifically. But maybe there's too much cases.

CMakeLists.txt
mzient: OK, it's not a blocker anyway.

dali/operators/decoder/inflate/inflate_params.h
stiepan: Oh, sure, now I get your point. I've changed the error message.

docker/Dockerfile.cuda117.x86_64.deps
stiepan: done

dali/operators/decoder/inflate/inflate_params.h
```diff
@@ -0,0 +1,388 @@
...
+      }
+    }
+  }
+
+  TensorListShape<> ParseOutputShape(const TensorListView<StorageCPU, const int> &provided_shape) {
+    auto num_samples = provided_shape.num_samples();
+    if (num_samples == 0) {
+      return {};
+    }
+    int sample_dim = provided_shape[0].num_elements();
```
mzient, line 350: Shouldn't you verify that there's at least one element?

dali/operators/decoder/inflate/inflate_params.h
stiepan: To document slack discussion: this is a way to handle dynamic scalars, i.e. those passed as tensor input for example through external source.

dali/operators/decoder/inflate/inflate_params.h
stiepan: Just for completeness (because it is silly example to use compression), allowing for num_elements = 0 covers compressed scalars.

dali/operators/decoder/inflate/inflate.cc
stiepan: I've resolved that disccusion by misclick and couldn't find this anywhere. :p
Anyway, renamed it.

stiepan:
> Please adjust https://github.com/NVIDIA/DALI/blob/main/conda/recipe/build.sh as well.

I think it should be covered.

Btw. Sorry, I edited your comment instead of replying to it at first. I didn't know I could do that. :D

dali/operators/decoder/inflate/inflate_params.h
mzient: OK.

mzient (approved):

docker/Dockerfile.cuda110.x86_64.deps
```diff
@@ -19,4 +20,12 @@ RUN NVJPEG2K_VERSION=0.1.0 && \
...
-    rm -rf /var/lib/apt/lists/*
+    rm -rf /var/lib/apt/lists/* && \
+    mkdir nvcomp && \
+    cd nvcomp && \
+    wget https://developer.download.nvidia.com/compute/nvcomp/${NVCOMP_VERSION}/local_installers/nvcomp_install_CUDA_11.x.tgz && \
+    tar -xvf nvcomp*.tgz && \
+    cp -rv include/nvcomp* /usr/local/cuda/include/ && \
+    cp -v lib/*.so /usr/local/cuda/lib64/ && \
+    cd .. && \
+    rm -rf nvcomp
```
JanuszL, line 31:
```suggestion
    rm -rf nvcomp*
```
To remove nvcomp*.tgz as well.

stiepan: I wget `nvcomp*.tgz` into nvcomp dir that is deleted here.

szalpal (approved):

stiepan:
Rebasing on the CI fixes

PR #4378, opened 2022-10-21
Fix namespace resolution for CUDA and STL math functions

szalpal:

## Category:

## Description:

This PR fixes a couple of problems that come from the similar reasons:

When a situation happens, that a `vec.h` file gets included together with a different file, that defines some math-like functions inside `dali` namespace, the CUDA or `std` functions used in this file aren't picked, because of namespace precedence. For example:

`vec.h` defines `vec<N, T> floor(const vec<N, T> &a)` function. Let's consider a following test file:
```c++
#include "vec.h"
#include "math_overloads.h"

namespace dali {

double a = 3.4;
auto b = floor(a);

}
```
The `floor(a)` function will fire a compilation error, because there's no definition for `floor(double)` - the function that actually gets picked is `dali::floor`, which has only `vec<N,T>` arguments.

Similarly the other problem - when `vec.h` is included in directory tree inside `dali` namespace, using `floor(float)` without `std` or `::` namespaces is a compilation error, since the `dali::floor` gets picked - and there's no such function.

### Tests:

- [x] Existing tests apply
- [ ] New tests added
  - [ ] Python tests
  - [ ] GTests
  - [ ] Benchmark
  - [ ] Other
- [ ] N/A

### Documentation
- [ ] Documentation updated
  - [ ] Docstring
  - [ ] Doxygen
  - [ ] RST
  - [ ] Jupyter
  - [ ] Other
- [ ] N/A

#### Requirements
- [ ] N/A

**REQ IDs**: N/A

**JIRA TASK**: N/A

### DIFF
--- a/dali/kernels/mask/grid_mask_cpu.h
+++ b/dali/kernels/mask/grid_mask_cpu.h
@@ -49,7 +49,7 @@ class GridMaskCpu {
       for (int x = 0; x < in.shape[1]; x++) {
         float fx = fxy + x * ca;
         float fy = fyy + x * sa;
-        auto m = (fx - floor(fx) >= ratio) || (fy - floor(fy) >= ratio);
+        auto m = (fx - std::floor(fx) >= ratio) || (fy - std::floor(fy) >= ratio);
         for (int c = 0; c < in.shape[2]; c++)
           *out_ptr++ = *in_ptr++ * m;
       }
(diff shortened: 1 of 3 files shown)


szalpal:
!build

banasraf (approved):

stiepan (approved):

PR #4406, opened 2022-10-31
Serialize wrapped global functions by value

stiepan:

## Category:

## Description:
Extends the idea of serialization by value from local functions to global ones that were manually overwritten.

Fixes an issue with a new ipython notebook env, where `open` (which should be just `io.open`) is replaced with a custom wrapper (https://github.com/ipython/ipython/commit/93992a7ecd086bb24840ba03cd69960daf41575d). This makes serailization by value of any callback that use `open` call fail with the pickle check that the `open is io.open`.

## Additional information:

Plain python pickler serializes functions by name, just remembering "a path" to the function. However, if the function is not statically available (because it is dynamically created closure or it is global function but in a notebook), this approach fails.
DALI uses sligtly cutomized pickler to handle serialization of local functions and global functions in a notebook mode. The pickler serializes local functions by value (marshalling their bytecode) and serializing recursivelly the dependencies. The same procudure can be applied to a global function if it is marked with `dali.pickling.pikcle_by_value`.

Now, ipython recently added a wrapper around global `open`, which makes it non-serializable with pickle. As a result, serialization by value fails on callbacks that use `open`, because it needs to be serialzied as a dependency. The change in this PR just tries to catch any such cases in advance and apply "by value" mode to them too.

### Tests:

- [x] Existing tests apply
- [ ] New tests added
  - [ ] Python tests
  - [ ] GTests
  - [ ] Benchmark
  - [ ] Other
- [ ] N/A

### Documentation
- [ ] Documentation updated
  - [ ] Docstring
  - [ ] Doxygen
  - [ ] RST
  - [ ] Jupyter
  - [ ] Other
- [ ] N/A

#### Requirements
- [x] N/A

**REQ IDs**: N/A

**JIRA TASK**: N/A

### DIFF
--- a/dali/python/nvidia/dali/reducers.py
+++ b/dali/python/nvidia/dali/reducers.py
@@ -1,4 +1,4 @@
-# Copyright (c) 2020-2021, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# Copyright (c) 2020-2022, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


stiepan:
!build

jantonguirao (approved):

mzient (approved):

PR #4416, opened 2022-11-04
Optimize TensorList and Tensor creation in the Python backend

JanuszL:
- properly propagate information about pinnes from the python to
  TensorListCPU
- removes unnecessary copy to the continuous memory when TensorList
   is created from a list of Tensors

## Category:

## Description:
- properly propagate information about pinnes from the python to
  TensorListCPU
- removes unnecessary copy to the continuous memory when TensorList
   is created from a list of Tensors

- dali/python/backend_impl.cc

### Tests:

- [x] Existing tests apply
   -test_backend_impl:test_create_tensorlist_list
  - test_backend_impl_gpu:test_cuda_array_interface_tensor_list_gpu_direct_creation_list
  - test_backend_impl_gputest_dlpack_tensor_list_gpu_direct_creation_list
  - test_backend_impl_torch_dlpack:test_dlpack_tensor_list_cpu_direct_creation_list
  - TensorListSuite.IsContiguous 
- [ ] New tests added
  - [ ] Python tests
  - [ ] GTests
  - [ ] Benchmark
  - [ ] Other
- [ ] N/A

### Documentation
- [ ] Documentation updated
  - [ ] Docstring
  - [ ] Doxygen
  - [ ] RST
  - [ ] Jupyter
  - [ ] Other
- [x] N/A

#### Requirements
- [x] N/A

**REQ IDs**: N/A

**JIRA TASK**: N/A

### DIFF
--- a/dali/pipeline/data/tensor_list.cc
+++ b/dali/pipeline/data/tensor_list.cc
@@ -791,11 +791,16 @@ void TensorList<Backend>::SetContiguity(BatchContiguity state) {
 
 
 template <typename Backend>
-void TensorList<Backend>::MakeContiguous(std::weak_ptr<void> owner) {
+void TensorList<Backend>::MakeContiguous(AccessOrder copy_order) {
   if (state_.IsContiguous()) {
     return;
   }
-  DALI_FAIL("Coalescing the buffer to Contiguous state is not yet implemented.");
+  auto contiguous_tmp = TensorList<Backend>(shape().num_samples());
+  contiguous_tmp.SetContiguity(BatchContiguity::Contiguous);
+  contiguous_tmp.SetupLike(*this);
+  contiguous_tmp.Copy(*this, order());
+  copy_order.wait(order());
+  *this = std::move(contiguous_tmp);
 }
 
 
@@ -1014,6 +1019,7 @@ Tensor<Backend> TensorList<Backend>::AsTensor() {
                  "To convert empty batch to a Tensor, valid dimensionality must be set");
     return AsReshapedTensor(TensorShape<>::empty_shape(sample_dim()));
   }
+  MakeContiguous();
   return AsReshapedTensor(shape_cat(shape().num_samples(), shape()[0]));
 }
 
--- a/dali/python/backend_impl.cc
+++ b/dali/python/backend_impl.cc
@@ -1197,16 +1197,6 @@ void ExposeTensorList(py::module &m) {
     .def("copy_to_external",
         [](TensorList<GPUBackend> &tl, py::object p, py::object cuda_stream,
            bool non_blocking, bool use_copy_kernel) {
-          CUDAStreamLease lease;
-          AccessOrder copy_order;
-          int device = tl.device_id();
-          if (!cuda_stream.is_none()) {
-            cudaStream_t stream = static_cast<cudaStream_t>(ctypes_void_ptr(cuda_stream));
-            copy_order = AccessOrder(stream, device);
-          } else {
-            lease = CUDAStreamPool::instance().Get(device);
-            copy_order = AccessOrder(lease, device);
-          }
           CopyToExternalImplGPU(tl, p, cuda_stream, non_blocking, use_copy_kernel);
         },
       "ptr"_a,
--- a/dali/python/backend_impl.cc
+++ b/dali/python/backend_impl.cc
@@ -341,7 +341,7 @@ auto FromPythonTrampoline(const char *python_module, const char *attr_name) {
  * associated access order with the provided stream; otherwise, the function will wait until the
  * copy completes.
  *
- * @tparam SourceObject  a data store on GPUBackend (Tensor, TensorList, TensorList)
+ * @tparam SourceObject  a data store on GPUBackend (Tensor, TensorList)
  * @param src             Source batch
  * @param dst_ptr         Destination pointer, wrapped in a C void_ptr Python type
  * @param cuda_stream     CUDA stream, wrapped in a C void_ptr type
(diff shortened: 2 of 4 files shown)


JanuszL:
!build

klecki:
Hmm, removing the contiguity from Python is actually a breaking change. Now, some of the TensorLists that the user may create won't be contiguous and as_tensor() would fail.
We enforce that DALI pipeline returns only contiguous data, as the framework iterators rely on that.

stiepan:
> Hmm, removing the contiguity from Python is actually a breaking change. Now, some of the TensorLists that the user may create won't be contiguous and as_tensor() would fail. We enforce that DALI pipeline returns only contiguous data, as the framework iterators rely on that.

Can we defer the copy to contigious memory up until the `as_tensor` call?

JanuszL:
> Hmm, removing the contiguity from Python is a breaking change. Now, some of the TensorLists that the user may create won't be contiguous and as_tensor() would fail. We enforce that the DALI pipeline returns only contiguous data, as the framework iterators rely on that.
Do we guarantee that the user-created TensorList is continuous? I don't think so. The TensorList returned by DALI is still continuous and `as_tensor` should work.

> Can we defer the copy to contigious memory up until the as_tensor call?
Why not, I see that some test has already failed because of that. Will do.

klecki:
> Do we guarantee that the user-created TensorList is continuous? I don't think so

I don't think we allowed users to create non-contiguous TensorLists (in Python).

JanuszL:
!build

JanuszL:
> I don't think we allowed users to create non-contiguous TensorLists (in Python).

I added `MakeContiguous` calls in the Python backend when we need the contiguous data as the user access it directly.

JanuszL:
!build

dali/pipeline/data/tensor_list.cc
```diff
@@ -791,11 +791,16 @@ void TensorList<Backend>::SetContiguity(BatchContiguity state) {
...
-void TensorList<Backend>::MakeContiguous(std::weak_ptr<void> owner) {
+void TensorList<Backend>::MakeContiguous(AccessOrder copy_order) {
   if (state_.IsContiguous()) {
     return;
   }
-  DALI_FAIL("Coalescing the buffer to Contiguous state is not yet implemented.");
+  auto contiguous_tmp = TensorList<Backend>(shape().num_samples());
+  contiguous_tmp.SetContiguity(BatchContiguity::Contiguous);
+  contiguous_tmp.SetupLike(*this);
+  contiguous_tmp.Copy(*this, order());
```
klecki, line 801: Won't this already run on the `order()`. Both tmp and this already have it set.

dali/pipeline/data/tensor_list.cc
```diff
@@ -791,11 +791,16 @@ void TensorList<Backend>::SetContiguity(BatchContiguity state) {
...
+  copy_order.wait(order());
```
klecki, line 802: I don't remember how it works exactly, but probably this copy_order can be used as parameter for Copy.

dali/pipeline/data/tensor_list.cc
```diff
@@ -791,11 +791,16 @@ void TensorList<Backend>::SetContiguity(BatchContiguity state) {
...
+  *this = std::move(contiguous_tmp);
```
klecki, line 803: I am not sure how do I feel about the swap of the allocation, at least to happen automatically for some of the calls. 

If we ever want to allow modification of TensorList memory from Python side, this will prevent things like:
```
tl = TensorListGPU(some_list_of_samples)
t0 = tl[0]  # this will take a reference to memory
t = tl.as_tensor()  # this will make the internal representation contiguous, detaching t0 from tl.
modify(tl)  # the changes are not propagated to t0, as it's separate allocation.
```
With current design `tl` and `t0` would be still aliasing. 

Also, I just realized, that the backend_impl's  "at" is probably broken (and it would even be more broken), as it just takes a raw pointer and wraps it into Python buffer without any reference count subscription. We already had such issues with `__get__`.

Should we get this to open floor to decide what is fine and what is not? Should we just allow non-contiguous TensorList and make the conversion explicit? Should we pursue the current approach?

Also, how much do we want to synchronize the data with host or user provided stream in such cases? 
This mostly impacts the copy_to_external.

dali/python/backend_impl.cc
```diff
@@ -1201,9 +1195,19 @@ void ExposeTensorList(py::module &m) {
...
+          AccessOrder copy_order;
+          int device = tl.device_id();
+          if (!cuda_stream.is_none()) {
+            cudaStream_t stream = static_cast<cudaStream_t>(ctypes_void_ptr(cuda_stream));
+            copy_order = AccessOrder(stream, device);
+          } else {
+            lease = CUDAStreamPool::instance().Get(device);
+            copy_order = AccessOrder(lease, device);
+          }
+          CopyToExternalImplGPU(tl, p, cuda_stream, non_blocking, use_copy_kernel);
```
klecki, line 1210: Doesn't CopyToExternalImplGPU already have similar code obtaining some stream? Why is it repeated here? `copy_order` seems unused.

dali/pipeline/data/tensor_list.cc
JanuszL: Sure, let us discuss this further.

> This mostly impacts the copy_to_external.

I think it is not affected. We have a copy there that works no matter if the underlying is continuous or not.

dali/pipeline/data/tensor_list.cc
JanuszL: I don't think so:
```
  template <typename SrcBackend>
  void Copy(const TensorList<SrcBackend> &in_tv, AccessOrder order = {},
            bool use_copy_kernel = false);
```
so you need to provide an order, in this case I provide the order from `this`.

dali/pipeline/data/tensor_list.cc
JanuszL: I think both ways would work, I don't know which is better.

dali/python/backend_impl.cc
JanuszL: Mhh, leftover. Removed

dali/pipeline/data/tensor_list.cc
klecki: Copy already does a three way merge between the dst order, source order and the order parameter I think.

PR #4434, opened 2022-11-10
Do not call nvcomp with no input

stiepan:

## Category:
**Bug fix**/**Other**

## Description:
If the user specifies the chunk_sizes/offsets, but they are empty lists for all the samples, there are no chunks for the nvcomp to decompress. Such calls to nvcomp error out, while we can simply return a batch of zero-volume samples according to the logic of the operator.

Alternatively we could forbid specyfing empty list of chunks for any sample. I just prefer having explicit control over this behaviour rather than slightly obsucre error from the lib call.

### Tests:

- [ ] Existing tests apply
- [x] New tests added
  - [x] Python tests
  - [ ] GTests
  - [ ] Benchmark
  - [ ] Other
- [ ] N/A

### Documentation
- [ ] Documentation updated
  - [ ] Docstring
  - [ ] Doxygen
  - [ ] RST
  - [ ] Jupyter
  - [ ] Other
- [x] N/A

#### Requirements
- [x] N/A

**REQ IDs**: N/A

**JIRA TASK**: N/A

### DIFF
--- a/dali/test/python/operator/test_inflate.py
+++ b/dali/test/python/operator/test_inflate.py
@@ -256,14 +256,15 @@ def test_total_no_chunks(ex_kwargs):
     @pipeline_def
     def pipeline():
         inflate = fn.external_source(source=lambda _: deflated, batch=False)
-        return fn.experimental.inflate(inflate.gpu(), shape=(128, 128, 3), **ex_kwargs)
+        return fn.experimental.inflate(inflate.gpu(), shape=(128, 128, 3), layout="HWC",
+                                       **ex_kwargs)
 
     batch_size = 8
     pipe = pipeline(batch_size=batch_size, num_threads=4, device_id=0)
     pipe.build()
     for _ in range(2):
         (inflated, ) = pipe.run()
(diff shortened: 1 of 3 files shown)


stiepan:
!build

klecki (approved):

dali/test/python/operator/test_inflate.py
```diff
@@ -241,6 +243,30 @@ def test_chunks():
                 seed += 1
 
 
+@has_operator("experimental.inflate")
+@restrict_platform(min_compute_cap=6.0, platforms=["x86_64"])
+@params({"chunk_offsets": []}, {"chunk_sizes": []}, {"chunk_offsets": np.array([], dtype=np.int32)},
+        {"chunk_sizes": np.array([], dtype=np.int32)})
+def test_total_no_chunks(ex_kwargs):
```
mzient, line 250: Can't we have proper parameter names with `@params`?

mzient (approved):

stiepan:
!build

PR #4453, opened 2022-11-17
Fix TensorList copy ordering issues

mzient:

## Category:
**Breaking change** (C++ API only)

## Description:
When a buffer has a stream ordering set, it could happen that it was allocated in stream order, but then copied _actually_ in host order (even though device order was specified). This PR changes several things:
- it makes it an error to specify a device order for H2H copy (breaking change in C++ API)
- it uses host order when copying between CPUBackend Tensors and TensorLists
- it changes the management of orders in the executor, so that the CPUBackend outputs are visible as having host order (it's only changed to the stage's stream after the operator.Run completes)

### Tests:

- [X] Existing tests apply
- [ ] New tests added
  - [ ] Python tests
  - [ ] GTests
  - [ ] Benchmark
  - [ ] Other
- [ ] N/A

### Documentation
- [ ] Documentation updated
  - [ ] Docstring
  - [ ] Doxygen
  - [ ] RST
  - [ ] Jupyter
  - [ ] Other
- [X] N/A

#### Requirements
- [X] N/A

**REQ IDs**: N/A

**JIRA TASK**: N/A

### DIFF
--- a/dali/pipeline/executor/executor.cc
+++ b/dali/pipeline/executor/executor.cc
@@ -407,6 +407,7 @@ void Executor<WorkspacePolicy, QueuePolicy>::RunHelper(OpNode &op_node, Workspac
     op.Run(ws);
   }
 
+  /* TODO(michalz): Find a way to make this valid in presence of passthrough between stages
   // Set the output order to the stage's stream
   for (int i = 0; i < ws.NumOutput(); i++) {
     if (ws.OutputIsType<CPUBackend>(i)) {
@@ -415,6 +416,7 @@ void Executor<WorkspacePolicy, QueuePolicy>::RunHelper(OpNode &op_node, Workspac
       ws.Output<GPUBackend>(i).set_order(ws_order);
     }
   }
+  */
 
   for (int i : empty_layout_in_idxs) {
     if (ws.InputIsType<CPUBackend>(i)) {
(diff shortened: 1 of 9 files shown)


jantonguirao (approved):

dali/pipeline/executor/executor.cc
```diff
@@ -296,21 +296,33 @@ void Executor<WorkspacePolicy, QueuePolicy>::RunHelper(OpNode &op_node, Workspac
...
+    else if (order == mixed_op_stream_)
+      return 1;
+    else if (order == gpu_op_stream_)
+      return 2;
+    else
+      return 3;  // foreign order, comes last
+  };
+
+  auto set_order = [&](auto &output, AccessOrder order) {
+    // Check if this stage (identified by ws_order) is already synchronized with the output's
```
klecki, line 315: Apart from foreign order, can this situation actually appear with our internal orders? That would require, for example, for the GPU order to appear in the Mixed stage, which I don't think is easily achievable.

klecki (approved):

dali/pipeline/executor/executor.cc
mzient: No, it's not, but I think it should be - that is, setting the order in TensorLists that share storage should be observable in the others (but it's not that easy).

PR #4462, opened 2022-11-22
Execute whole Keras code in the expected device scope

klecki:
## Description:
TL;DR: move `model.fit` under `with tf.device`.

In newer TF version, when the model.fit is executed without explicit scope (which gives it default placement) we may encounter transfers between the device on which we defined the model and inputs to the device which executes the model - which is not supported. Adjust those tests to both be defined and execute on the desired device - so all the tensors/variables placements are uniform.

Additionally some cleanup code is added for Keras tests so the graph placement doesn't collide with other tests - @with_setup(tf.keras.backend.clear_session) - after the previous change, on my machine running all graph tests in sequence caused subsequent tests to fail with allocation error.

TF Tests

I am not sure why the previous one worked (as the offending variables are defined in the internal TF code) and if we might want to keep the old behavior of running model.fit without device scope.

### Tests:

- [x] Existing tests apply
This is a fix for test_dali_tf_dataset_mnist_eager.py 
- [ ] New tests added
  - [ ] Python tests
  - [ ] GTests
  - [ ] Benchmark
  - [ ] Other
- [ ] N/A

### Documentation
- [ ] Documentation updated
  - [ ] Docstring
  - [ ] Doxygen
  - [ ] RST
  - [ ] Jupyter
  - [ ] Other
- [ ] N/A

#### Requirements
- [ ] N/A

**REQ IDs**: N/A

**JIRA TASK**: N/A

### DIFF
--- a/dali/test/python/test_dali_tf_dataset_mnist.py
+++ b/dali/test/python/test_dali_tf_dataset_mnist.py
@@ -110,14 +110,14 @@ def run_keras_single_device(device='cpu', device_id=0):
         model = keras_model()
         train_dataset = get_dataset(device, device_id)
 
-    model.fit(
-        train_dataset,
-        epochs=EPOCHS,
-        steps_per_epoch=ITERATIONS)
+        model.fit(
+            train_dataset,
+            epochs=EPOCHS,
+            steps_per_epoch=ITERATIONS)
 
-    assert model.evaluate(
-        train_dataset,
-        steps=ITERATIONS)[1] > TARGET
+        assert model.evaluate(
+            train_dataset,
+            steps=ITERATIONS)[1] > TARGET
 
 
 def graph_model(images, reuse, is_training):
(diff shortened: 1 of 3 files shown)


klecki:
!build

JanuszL (approved):
As I understand `@with_setup(tf.keras.backend.clear_session)` is the fix, the rest is formating.

klecki:
> As I understand @with_setup(tf.keras.backend.clear_session) is the fix, the rest is formating.

Nope, precisely the opposite.
The formatting is meaningful, it brings the model.fit under the `with tf.device` changing where we are running.

JanuszL:
> > As I understand @with_setup(tf.keras.backend.clear_session) is the fix, the rest is formating.
> 
> Nope, precisely the opposite. The formatting is meaningful, it brings the `model.fit` under the `with tf.device` changing where we are running.

Oh, now I get it.

klecki:
!build

PR #4466, opened 2022-11-24
Fix numpy reader crash

mzient:
## Category:

## Description:
The new libcufile doesn't exit cleanly (for some reasons) and has problems with library use count tracking.
This PR does the following:
- it improves the stub generator to support functions that are redefined with C macros
- it adds support for cuFileDriverClose_v2
- it uses library reference counting when the new library is detected - it happens in GDSMem and in NumpyReaderGPU
- it deletes the pipelines in the test so that the pipeline doesn't linger when the temporary directory with the test data goes out of scope and is deleted

NumpyReaderGPU
dynlink stubs

### Tests:
GDSMemTest (GTest)
NumpyReader tests (reader/test_numpy.py)

- [X] Existing tests apply
- [ ] New tests added
  - [ ] Python tests
  - [ ] GTests
  - [ ] Benchmark
  - [ ] Other
- [ ] N/A

### Documentation
- [ ] Documentation updated
  - [ ] Docstring
  - [ ] Doxygen
  - [ ] RST
  - [ ] Jupyter
  - [ ] Other
- [X] N/A

#### Requirements
- [X] N/A

**REQ IDs**: N/A

**JIRA TASK**: N/A

### DIFF
--- a/dali/imgcodec/decoders/memory_pool.h
+++ b/dali/imgcodec/decoders/memory_pool.h
@@ -16,6 +16,7 @@
 #define DALI_IMGCODEC_DECODERS_MEMORY_POOL_H_
 
 #include <atomic>
+#include <array>
 #include <shared_mutex>
 #include <thread>
 #include <map>
--- a/dali/operators/reader/numpy_reader_gpu_op.cc
+++ b/dali/operators/reader/numpy_reader_gpu_op.cc
@@ -28,6 +28,7 @@ NumpyReaderGPU::NumpyReaderGPU(const OpSpec& spec)
       thread_pool_(num_threads_, spec.GetArgument<int>("device_id"), false, "NumpyReaderGPU"),
       sg_(1 << 18),
       header_cache_(spec.GetArgument<bool>("cache_header_information")) {
+  InitDriverScope();
   prefetched_batch_tensors_.resize(prefetch_queue_depth_);
   // make the device current
   DeviceGuard g(device_id_);
@@ -41,6 +42,12 @@ NumpyReaderGPU::NumpyReaderGPU(const OpSpec& spec)
   loader_ = InitLoader<NumpyLoaderGPU>(spec, std::vector<string>(), shuffle_after_epoch);
 
   kmgr_transpose_.Resize<TransposeKernel>(1);
+}
+
+NumpyReaderGPU::~NumpyReaderGPU() {
+  // Stop the prefetch thread as it uses the thread pool from this class. So before we can
+  // destroy the thread pool make sure no one is using it anymore.
+  this->StopPrefetchThread();
 }
 
 void NumpyReaderGPU::Prefetch() {
@@ -134,6 +141,7 @@ void NumpyReaderGPU::ScheduleChunkedRead(SampleView<GPUBackend> &out_sample,
     ssize_t copy_end = file_offset + chunk_read_length;
     ssize_t chunk_copy_length = copy_end - copy_start;
     thread_pool_.AddWork([=, &load_target](int tid) {
+      assert(chunk_read_length <= static_cast<ssize_t>(staging_.chunk_size()));
       auto buffer = staging_.get_staging_buffer();
       load_target.ReadRawChunk(buffer.at(0), chunk_read_length, 0, file_offset);
       assert(dst_ptr >= base_ptr && dst_ptr + chunk_copy_length <= base_ptr + data_bytes);
--- a/dali/util/cufile_helper.h
+++ b/dali/util/cufile_helper.h
@@ -18,6 +18,7 @@
 // general stuff
 #include <cstdio>
 #include <string>
+#include <utility>
 
 #if !defined(__AARCH64_QNX__) && !defined(__AARCH64_GNU__) && !defined(__aarch64__)
 #include <linux/sysctl.h>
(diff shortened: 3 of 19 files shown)


dali/imgcodec/decoders/memory_pool.h
```diff
@@ -16,6 +16,7 @@
 #define DALI_IMGCODEC_DECODERS_MEMORY_POOL_H_
 
 #include <atomic>
+#include <array>
```
mzient, line 19: This failed to compile with libc++ 12

dali/operators/reader/numpy_reader_gpu_op.cc
```diff
@@ -43,6 +44,12 @@ NumpyReaderGPU::NumpyReaderGPU(const OpSpec& spec)
   kmgr_transpose_.Resize<TransposeKernel>(1);
 }
 
+NumpyReaderGPU::~NumpyReaderGPU() {
```
mzient, line 47: Moved as-is from the header.

dali/util/cufile_helper.h
```diff
@@ -28,35 +29,60 @@
...
+    std::swap(fd, other.fd);
+    std::swap(fdd, other.fdd);
+    std::swap(cufh, other.cufh);
+    other.Close();
+    return *this;
+  }
 
   void Close() {
-    if ((fd != -1) && (fdd != -1))
+    if (cufh) {
```
mzient, line 73: Use the real value we're interested in instead of relying on side effects.

dali/util/mmaped_file.cc
```diff
@@ -147,6 +147,11 @@ MmapedFileStream::MmapedFileStream(const std::string& path, bool read_ahead) :
   DALI_ENFORCE(p_ != nullptr, "Could not open file " + path + ": " + std::strerror(errno));
 }
 
+MmapedFileStream::~MmapedFileStream() {
```
mzient, line 150: Moved from the header.

dali/util/std_cufile.cc
```diff
@@ -79,11 +80,15 @@ namespace dali {
 StdCUFileStream::StdCUFileStream(const std::string& path) : CUFileStream(path) {
   // open file
   cufile_open(f_, length_, path.c_str());
-
   // set the path to current path
   path_ = path;
 }
 
+StdCUFileStream::~StdCUFileStream() {
```
mzient, line 87: Moved from the header.

dali/util/std_file.cc
```diff
@@ -29,6 +29,11 @@ StdFileStream::StdFileStream(const std::string& path) : FileStream(path) {
   DALI_ENFORCE(fp_ != nullptr, "Could not open file " + path + ": " + std::strerror(errno));
 }
 
+StdFileStream::~StdFileStream() {
```
mzient, line 32: Moved from the header.

tools/stub_generator/stub_codegen.py
```diff
@@ -89,28 +95,29 @@ def main():
...
+           function_name not in all_declaration:
             continue
 
         # make sure that we deal with every function only once
-        all_declaration.remove(cursor.spelling)
-
-        with open(cursor.location.file.name, 'r', encoding='latin-1') as file:
-            start = cursor.extent.start.offset
-            end = cursor.extent.end.offset
-            declaration = file.read()[start:end]
```
mzient, line 103: This copied the declaration from the original file - it didn't work correctly, because there might have been some macros - and they could be used in different order, yielding different results:
```C++
#define cuFileDriverClose cuFileDriverClose_v2
CUFileError_t cuFileDriverClose();
```
or
```C++
CUFileError_t cuFileDriverClose_v2();
#define cuFileDriverClose cuFileDriverClose_v2
```

dali/util/cufile_helper.h
```diff
@@ -28,35 +29,60 @@
...
+  CUFileHandle() = default;
+  CUFileHandle(CUFileHandle &&other) {
+    *this = std::move(other);
+  }
+
+  CUFileHandle &operator=(CUFileHandle &&other) {
+    std::swap(fd, other.fd);
+    std::swap(fdd, other.fdd);
+    std::swap(cufh, other.cufh);
+    other.Close();
```
banasraf, line 68: Do you have to explicitly call `Close`? Wouldn't it just be called when destructing `other`

mzient: I read somewhere that implementing move assignment as a pure swap isn't a good practice, because it prolongs the life of the object being overwritten in a surprising way. This is especially important in case of resources like files, which have global side effects. Consider:
```C++
File a = File::open("a.dat", "rw");
File b = File::open("b.dat", "rw");

a = std::move(b);  // the user would expect that "a.dat" is closed, but this assignment actually did a swap
File c = File::open("a.dat", "rw");  // file busy, because the moved-out object `b` is keeping the file open.
```

banasraf: Huh. That's interesting. I didn't know how the move works, then. I thought `b` would be destroyed just after evaluation of this assignment.

banasraf (approved):

dali/util/cufile_helper.h
mzient: We learn something new every day ;) The object still "lives", even though it shouldn't be used for anything other than destruction. I'm not sure if the compiler is allowed to destroy a moved-out local variable before the scope end, but there are certainly cases where you can move-out objects with more permanent storage - so they cannot be destroyed (calling a destructor multiple times is an error). For example, you can loop over an array and move every element from it to some other container and then destroy the array. The destructors will be called when the array is destroyed, not when each object in the array is moved out.

awolant (approved):

PR #4480, opened 2022-11-30
Extend decoding support

awolant:
## Category:

**New feature** 

## Description:

Adds ability for `FramesDecoderGpu` to decode raw H264 i H265 streams from memory.

`FramesDecoderGpu` and its tests. Some changes to its base class `FramesDecoder` were also necessary.

### Tests:

- [ ] Existing tests apply
- [x] New tests added
  - [ ] Python tests
  - [ ] GTests
  - [ ] Benchmark
  - [ ] Other
- [ ] N/A

### Documentation
- [ ] Documentation updated
  - [ ] Docstring
  - [ ] Doxygen
  - [ ] RST
  - [ ] Jupyter
  - [ ] Other
- [ ] N/A

#### Requirements
- [ ] N/A

**REQ IDs**: N/A

**JIRA TASK**: DALI-3147


dali/operators/reader/loader/video/frames_decoder.cc
```diff
@@ -236,20 +243,84 @@ FramesDecoder::FramesDecoder(const char *memory_file, int memory_file_size, bool
...
+      curr_num_frames++;
     }
-    curr_num_frames++;
+
+    num_frames_ = curr_num_frames;
+    Reset();
+  } else {
+    // Failover for unseekable video
+    auto current_position = memory_video_file_->position_;
+    memory_video_file_->Seek(0, SEEK_SET);
```
JanuszL, line 299: Maybe we can copy `memory_video_file_` and pass this copy to `CreateAvState` ?

dali/operators/reader/loader/video/frames_decoder.cc
```diff
@@ -236,20 +243,84 @@ FramesDecoder::FramesDecoder(const char *memory_file, int memory_file_size, bool
   DetectVfr();
 }
 
+void FramesDecoder::CreateAvState(std::unique_ptr<AvState> &av_state, bool init_codecs) {
```
JanuszL, line 246: Do you know what is the overhead of this function?

szalpal (approved):

dali/operators/reader/loader/video/frames_decoder.h
```diff
@@ -58,8 +58,13 @@ struct AvState {
       av_frame_free(&frame_);
     }
     avcodec_free_context(&codec_ctx_);
-    avformat_close_input(&ctx_);
-    avformat_free_context(ctx_);
+    if (ctx_ != nullptr) {
```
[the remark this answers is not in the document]
awolant: I think it was leaking this context before.

dali/operators/reader/loader/video/frames_decoder.cc
awolant: Can we stay like this for now? I would like to decouple this even more in the next step. I just wanted to have minimal changes and the working decoding.

JanuszL: Sure

dali/operators/reader/loader/video/frames_decoder.h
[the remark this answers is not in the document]
JanuszL: Understood

dali/operators/reader/loader/video/frames_decoder.cc
```diff
@@ -236,20 +243,97 @@ FramesDecoder::FramesDecoder(const char *memory_file, int memory_file_size, bool
...
+    Reset();
+  } else {
+    // Failover for unseekable video
+    auto current_position = memory_video_file_->position_;
+    memory_video_file_->Seek(0, SEEK_SET);
+    std::unique_ptr<AvState> tmp_av_state = std::make_unique<AvState>();
+    CreateAvState(tmp_av_state, false);
+
+    while (av_read_frame(tmp_av_state->ctx_, tmp_av_state->packet_) >= 0) {
+      // We want to make sure that we call av_packet_unref in every iteration
```
JanuszL, line 311: Do you think we can extract this to a function as the same patter in applied in L291.

dali/operators/reader/loader/video/frames_decoder.cc
```diff
@@ -236,20 +243,97 @@ FramesDecoder::FramesDecoder(const char *memory_file, int memory_file_size, bool
...
-  num_frames_ = curr_num_frames;
-  Reset();
+bool FramesDecoder::IsFormatSeekable() {
+  if (
+    av_state_->ctx_->iformat->read_seek == nullptr &&
+    av_state_->ctx_->iformat->read_seek2 == nullptr) {
+    return false;
+  }
+
+  return av_state_->ctx_->pb->read_seek != nullptr;
```
JanuszL, line 336: So still if one of `iformat->read_seek` and `iformat->read_seek2` is not null we still need to check `pb->read_seek`?

dali/operators/reader/loader/video/frames_decoder.h
```diff
@@ -58,8 +58,13 @@ struct AvState {
...
+      if (ctx_->pb != nullptr) {
+        avio_context_free(&ctx_->pb);
+      }
+      avformat_close_input(&ctx_);
```
JanuszL, line 65: Based on https://ffmpeg.org/doxygen/trunk/demux_8c_source.html#l00369, 'avformat_close_input' seems to call `avformat_free_context` and `avio_close` which calls `avio_context_free`. Still please double check this in the version of FFmpeg we use (asan doesn't complain so I think we are not leaking anything).

dali/operators/reader/loader/video/frames_decoder_gpu.cc
```diff
@@ -529,7 +545,12 @@ bool FramesDecoderGpu::SendFrameToParser() {
 
   // Store pts from current packet to indicate,
   // that this frame is in the decoder
-  piped_pts_.push(av_state_->packet_->pts);
+  if (av_state_->packet_->pts != AV_NOPTS_VALUE) {
+    piped_pts_.push(av_state_->packet_->pts);
+  } else {
+    piped_pts_.push(frame_index_if_no_pts_);
+    frame_index_if_no_pts_++;
+  }
```
JanuszL, line 553: Is it possible that some frames in the video have PTS and some not?

dali/operators/reader/loader/video/frames_decoder_gpu.cc
```diff
@@ -557,35 +578,76 @@ bool FramesDecoderGpu::ReadNextFrameWithoutIndex(uint8_t *data, bool copy_to_out
...
   current_frame_output_ = data;
 
+  int frame_to_return_index = -1;
+
+  // Handle the case, when packet has more frames that we have empty spots
+  // in the buffer.
+  // If so, we need to return frame from the buffer before sending last packet.
+  if (frame_index_if_no_pts_ != 0) {
+    if (NumEmptySpots() < (piped_pts_.size())) {
+      for (size_t i = 0; i < frame_buffer_.size(); ++i) {
```
JanuszL, line 588: Is it possible that we won't find desired frame in the `frame_buffer_` as it has not been decoded yet?

dali/operators/reader/loader/video/frames_decoder.cc
awolant: Done, called `CountFrames`.

dali/operators/reader/loader/video/frames_decoder.h
awolant: I did that based on the documentation that said that allocation of this context should be freed. Since in this scenario that we use it the wrapping object destructor takes care of it, I removed the call in our code.

dali/operators/reader/loader/video/frames_decoder.cc
awolant: Yes, we need both (one of the read functions to be not null and `pb->read_seek` to be not null) to be able to seek. I was not able to 100% confirm that in all possible scenarios one of these implies the other, so I decided to just check for both.

dali/operators/reader/loader/video/frames_decoder_gpu.cc
awolant: Based on my understanding this should not happen in well formed files. If there is no PTS value, this assumes that the whole video is constant frame rate and this frame rate is present in the metadata. 
If we have PTS and no other information we assume that this file might be variable frame rate and we need to have PTS of every frame to be able to decode it and show it correctly.

dali/operators/reader/loader/video/frames_decoder_gpu.cc
awolant: I think that theoretically, this can happen. The ultimately resilient solution for this issue I have found is to allocate spaces in the buffer dynamically as much as needed.
In practice, I have never encountered this issue yet, we have buffer big enough for the files that we have seen so far.  

I would like to leave it as it is now. When we do decoupling mentioned above, we can revisit it and allow for buffer dynamic extension or something like that. Is that ok?

JanuszL: Maybe we can error in such case instead of silently returning the wrong result so the user can provide us with a failing sample and we can adjust?

dali/operators/reader/loader/video/frames_decoder.cc
```diff
@@ -236,20 +243,83 @@ FramesDecoder::FramesDecoder(const char *memory_file, int memory_file_size, bool
...
+    av_state->codec_ctx_ = avcodec_alloc_context3(av_state->codec_);
+    DALI_ENFORCE(av_state->codec_ctx_, "Could not alloc av codec context");
+
+    ret = avcodec_parameters_to_context(av_state->codec_ctx_, av_state->codec_params_);
+    DALI_ENFORCE(
+      ret >= 0,
+      make_string("Could not fill the codec based on parameters: ", detail::av_error_string(ret)));
+
+    av_state->packet_ = av_packet_alloc();
+    DALI_ENFORCE(av_state->packet_, "Could not allocate av packet");
```
JanuszL, line 284: I think you can replace it with InitAvState().

dali/operators/reader/loader/video/frames_decoder.h
```diff
@@ -207,6 +209,8 @@ class DLL_PUBLIC FramesDecoder {
 
   bool is_full_range_ = false;
 
+  std::optional<bool> zero_latency_ = {};
```
JanuszL, line 212: I think it is no longer used.

dali/operators/reader/loader/video/frames_decoder_gpu.cc
```diff
@@ -399,20 +406,29 @@ FramesDecoderGpu::FramesDecoderGpu(
...
+}
 
-  // Process decoded frame for output
+int FramesDecoderGpu::HandlePictureDisplay(CUVIDPARSERDISPINFO *picture_display_info) {
   CUVIDPROCPARAMS videoProcessingParameters = {};
-  videoProcessingParameters.progressive_frame = !picture_params->field_pic_flag;
+  videoProcessingParameters.progressive_frame = !picture_display_info->progressive_frame;
   videoProcessingParameters.second_field = 1;
-  videoProcessingParameters.top_field_first = picture_params->bottom_field_flag ^ 1;
+  videoProcessingParameters.top_field_first = picture_display_info->top_field_first;
```
JanuszL, line 431: What is the difference now?

JanuszL (approved):

dali/operators/reader/loader/video/frames_decoder_gpu.cc
awolant: I think right now it should error out with `Could not find empty slot in the frame buffer`. It should try to find the next frame, fail to do it, than call for another frame, try to place it in the buffer and error out due to lack of empty spots.

dali/operators/reader/loader/video/frames_decoder.cc
awolant: `CreateAvState` covers some parts of the `InitAvState` and some parts of other functions. I would like to decouple the av state manipulation from this code altogether,  so I will refactor it during that, if that is fine?

awolant:
!build

PR #4491, opened 2022-12-01
Remove TF-specific codepaths from AutoGraph 

klecki:
## Description:
Remove the `tensorflow.python` prefix from pure AutoGraph imports.
Remove all imports that start with `from tensorflow...` and lead to
non-AutoGraph functionality.
Adjust some of the imports (mainly inspect and unittest) and add 
some placeholder utilites.

Replace all TensorFlow Operator overloads with calls to OperatorBase 
class that currently is a no-op. 
The new code is in third_party/autograph/autograph/utils/hooks.py
which keeps a _DISPATCH global symbol. We replace it with the instance 
inheriting from OperatorBase. You can see it's replaced in api.py:202 
for the `ag_internal`. `ag_internal` translates to the `ag__`  that is inserted 
as a local object for every transformed function.

Introduce `initialize_autograph` call that allows to set the name and 
provide custom (AutoGraph) operator overloads. User is supposed 
to inherit from OperatorBase.

The library can be initialized once, I don't expect to need several different
instances of AutoGraph in one scope, to handle this in a NumPy random
generator style would need to much refactoring to be worth it.

Remove tests that utilize the TF Tensors, Datasets, operators, etc.
Replace TF-specific test class with unittest.TestCase.
For particular subdirectory:
* `converters/` - replace tf.constant with regular value and test the
    Python fallbacks (as the OperatorBase always indicates that we 
    don't need to select the overload)
* `core/`, `impl/` - keep the test that are not TF-specific, for example
    not use TF decorators, defuns, etc.
* `operators/` - keep mostly the tests that have `python` in name,
    the others indicate that it's TF specific. Those tests explicitly check
    the Python fallback implementation for transformed statements.
* `pyct` - this is the transpiler part, most tests are pure AutoGraph,
    most of the time only imports and test class used need to be adjusted.

Some functionality around decorators and stack trace remapping was removed
or commented out. 
The former was tightly coupled with TF's decorator system.
The latter requires some compiled utilities which I didn't
port as it would require plugging it into the build system. This can be added
after the integration with DALI.

After this PR, the next one should be able to move `autograph` into `nvidia.dali`
under some `_internal` name and move the test into our test directory
so they can be executed.

AutoGraph

N/A

### Tests:
This PR does the work necessary to be able to use the tests.
- [ ] Existing tests apply
- [ ] New tests added
  - [ ] Python tests
  - [ ] GTests
  - [ ] Benchmark
  - [ ] Other
- [ ] N/A

### Documentation
- [ ] Documentation updated
  - [ ] Docstring
  - [ ] Doxygen
  - [ ] RST
  - [ ] Jupyter
  - [ ] Other
- [x] N/A

#### Requirements
- [ ] N/A

**REQ IDs**: N/A

**JIRA TASK**: N/A


klecki:
!build

third_party/autograph/autograph/core/converter_testing.py
```diff
@@ -59,11 +59,11 @@ def is_inside_generated_code():
     del frame
 
 
-class TestingTranspiler(api.PyToTF):
+class TestingTranspiler(api.PyToLib):
   """Testing version that only applies given transformations."""
 
-  def __init__(self, converters, ag_overrides):
-    super(TestingTranspiler, self).__init__()
+  def __init__(self, converters, ag_overrides, hooks_impl=hooks.OperatorBase()):
```
mzient, line 65: As usual, I'll complain about misused plural:
```suggestion
  def __init__(self, converters, ag_overrides, hook_impl=hooks.OperatorBase()):
```

third_party/autograph/autograph/impl/api.py
```diff
@@ -142,56 +128,58 @@ def _attach_error_metadata(e, f):
                                        __file__)
 
 
-class StackTraceMapper(tf_stack.StackTraceMapper):
-  """Remaps generated code to code it originated from."""
+# class StackTraceMapper(tf_stack.StackTraceMapper):
```
mzient, line 131: Do you intend to revive this piece?

klecki: This is the piece mentioned in the PR description about stack remapping. I would need to plug it into DALI build, so I will wait with it till I move it into DALI, integrate the tests etc.

third_party/autograph/autograph/core/converter_testing.py
klecki: I changed it, still I prefer the plural name here, we have multiple hooks or operators to overload. 
I decided to use `operator overload` in the name rather than `hook impl` which looks like something internal.

mzient (approved):
Remarks:
- some nitpicks about formatting
- some comments in OperatorBase would be nice
- rename tf_logging.py to get rid of "tf_" in the name.
-

third_party/autograph/autograph/utils/tf_logging.py
```diff
@@ -0,0 +1,310 @@
...
+
+  def _logger_find_caller(stack_info=False, stacklevel=1):  # pylint: disable=g-wrong-blank-lines
+    code, frame = _get_caller(4)
+    sinfo = None
+    if stack_info:
+      sinfo = '\n'.join(_traceback.format_stack())
+    if code:
+      return (code.co_filename, frame.f_lineno, code.co_name, sinfo)
+    else:
+      return '(unknown file)', 0, '(unknown function)', sinfo
```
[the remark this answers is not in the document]
klecki: Looks like I no longer need this file.

jantonguirao (approved):

third_party/autograph/test/impl/api_test.py
```diff
@@ -65,7 +60,20 @@ def __init__(self):
...
+  def evaluate(self, x):
+    return x
+
+
+  def assertAllEqual(self, a, b):
+    if isinstance(a, np.ndarray) or isinstance(b, np.ndarray):
+      same = np.array(a) == np.array(b)
+      self.assertTrue(np.all(same))
+    else:
+      self.assertEqual(a, b)
```
stiepan, line 76: The blank lines need some moving around here. :D

stiepan (approved):

third_party/autograph/test/impl/api_test.py
klecki: You are right, I am working on a followup to enable linting and tidy it up. I will do it in a separate PR.

klecki:
!build

PR #4507, opened 2022-12-12
Don't look for duplicate arguments in parent schemas.

mzient:

## Category:

## Description:
Prior to this change, AddParent had to be the last command, so that AddArgument didn't look at the parents. Failure to do so could result in:
- failure to look up the parent because of undefined order of initialization of globals in different translation units
- order-dependent behaviour
- if AddParent was first, otherwise legal argument shadowing became illegal.

Additionally, a bug in `GetDefaultValueForArgument` was fixed. Prior to this change, `GetDefaultValueForArgument` tried to dynamic-cast a Value to `Value<T>` instead of `Value<argument_storage_t<T>>`, which always failed for types as simple as `int` (which is stored as `int64_t`).

A dependence of dali_test on libdali_operators has been removed - it's not necessary and prolongs the build. This has already been done once, but the dependence was somehow reintroduced.

* Schema
* dali_test

### Tests:
Modified a test for AddParent, complete with shadowing

- [ ] Existing tests apply
- [X] New tests added
  - [ ] Python tests
  - [X] GTests
  - [ ] Benchmark
  - [ ] Other
- [ ] N/A

### Documentation
- [X] Documentation updated
  - [ ] Docstring
  - [X] Doxygen
  - [ ] RST
  - [ ] Jupyter
  - [ ] Other
- [ ] N/A

#### Requirements
- [X] N/A

**REQ IDs**: N/A

**JIRA TASK**: N/A

### DIFF
--- a/dali/CMakeLists.txt
+++ b/dali/CMakeLists.txt
@@ -100,7 +100,7 @@ if (BUILD_DALI_PIPELINE AND BUILD_TEST)
   adjust_source_file_language_property("${DALI_TEST_SRCS}")
   add_executable(dali_test "${DALI_TEST_SRCS}")
 
-  target_link_libraries(dali_test PUBLIC dali dali_core dali_kernels dali_operators ${DALI_LIBS} gtest)
+  target_link_libraries(dali_test PUBLIC dali dali_core dali_kernels ${DALI_LIBS} gtest)
   target_link_libraries(dali_test PRIVATE dynlink_cuda ${CUDART_LIB})
   if (BUILD_NVML)
     target_link_libraries(dali_test PRIVATE dynlink_nvml)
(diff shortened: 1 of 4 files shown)


awolant (approved):

szalpal:
For the record, I believe that this change is not actually tested here. The same situation has been in my `InputOperatorBase`-as-parent problem. When I put both Schemas in the same compilation unit, everything was working fine. Only putting those in separate compilation units resulted in the problem.

I might be wrong, but did you happen to run the test before the fix? To make sure it's actually tested?

mzient:
> For the record, I believe that this change is not actually tested here. The same situation has been in my `InputOperatorBase`-as-parent problem. When I put both Schemas in the same compilation unit, everything was working fine. Only putting those in separate compilation units resulted in the problem.
> 
> I might be wrong, but did you happen to run the test before the fix? To make sure it's actually tested?

I did and it indeed worked - but the test doesn't cover exactly this case (i.e. reverse initialization order). I've added a test that does.

szalpal (approved):

PR #4513, opened 2022-12-15
Utilities and refactoring pre-`VideoInput` operator

szalpal:

## Category:

## Description:
This PR is a pre-`VideoInput` operator PR. It contains ironed-out part of the VideoInput operator PR to make the latter shorter and easier to review. For more context on these changes, please refer to the CI-only PR of `VideoInput` operator: #4512 

* Moving `caching_list.h` to builtin operators directory.
* Adding `InputOperator` documentation.
* Relaxing enforcing uniform batch size for InputOperators.
* Allowing feed_input to work with all InputOperators (instead of ExternalSource only).
* Introducing `DecodeFrames` in `VideoDecoderBase`.
* Reverting template method pattern in InputOperator's `SetupImpl`.

During the development of `VideoInput` it tuned out, that the template method pattern for InputOperator `SetupImpl` function does not suffice the behaviour requirements. The main reason was that `HandleDataAvailability` function has been called every Operator iteration, while `VideoInput` mustn't check it that often. On the contrary - `VideoInput` may work on the data fed a few iterations ago. Since `HandleDataAvailability` checks, if the data is enqueued - it might be a false result. Therefore, `HandleDataAvailability` shall be called by every subclass of `InputOperator`, what is stated in the documentation.

The `VideoDecoderBase` needed an additional method to decode parts of the video injected into `FramesDecoder`. This function received also a `padding` feature, required by `VideoInput` operator.

Lastly, the uniform-batch-size-requirement for the `InputOperators` had to be relaxed, since these operators actually create a batch. Therefore, they can decide not to create a uniform batch across iteration and it is perfectly fine behaviour.

### Tests:

- [x] Existing tests apply
- [ ] New tests added
  - [ ] Python tests
  - [ ] GTests
  - [ ] Benchmark
  - [ ] Other
- [ ] N/A

### Documentation
- [ ] Documentation updated
  - [ ] Docstring
  - [x] Doxygen
  - [ ] RST
  - [ ] Jupyter
  - [ ] Other
- [ ] N/A

#### Requirements
- [ ] N/A

**REQ IDs**: N/A

**JIRA TASK**: N/A


szalpal:
!build

szalpal:
!build

dali/operators/decoder/video/video_decoder_base.h
```diff
@@ -60,19 +61,59 @@ class DLL_PUBLIC VideoDecoderBase {
...
+    //    (a) There are no more frames, or
+    //    (b) Sufficient number of frames has been decoded.
+    for (; f < num_frames && frames_decoder.NextFrameIdx() != -1; f++) {
       frames_decoder.ReadNextFrame(output_data + f * frame_size);
     }
+    assert(f <= num_frames);
+    bool full_sequence_decoded = f == num_frames;
+    // If there's an insufficient number of frames, pad if requested.
+    for (; f < num_frames && pad_value.has_value(); f++) {
+      CUDA_CALL(cudaMemcpy(output_data + f * frame_size, pad_value->raw_data(), frame_size,
```
JanuszL, line 101: Do you want `pad_value` to house the whole frame?

dali/operators/decoder/video/video_decoder_base.h
```diff
@@ -60,19 +61,59 @@ class DLL_PUBLIC VideoDecoderBase {
...
-    for (int f = 0; f < num_frames; ++f) {
+
+    int64_t f = 0;
+    // Work until:
+    //    (a) There are no more frames, or
+    //    (b) Sufficient number of frames has been decoded.
+    for (; f < num_frames && frames_decoder.NextFrameIdx() != -1; f++) {
       frames_decoder.ReadNextFrame(output_data + f * frame_size);
     }
+    assert(f <= num_frames);
```
JanuszL, line 97: Is that possible at all?

szalpal: I intended it to be a sanity check, that `f` is not be higher than `num_frames`. Because it shouldn't be.

dali/operators/decoder/video/video_decoder_base.h
szalpal: That's correct. I'll put it in the documentation, since it's probably not clear.

szalpal:
!build

dali/operators/decoder/video/video_decoder_base.h
mzient: Better keep it. Hopefully we _will_ have a CI with asserts at some point.

dali/operators/decoder/video/video_decoder_base.h
```diff
@@ -60,19 +62,64 @@ class DLL_PUBLIC VideoDecoderBase {
...
+   * @param num_frames How many frames shall be decoded.
+   * @param pad_value What to (optionally) pad the partial sequence with.
+   * @return False, if less than `num_frames` have been decoded.
    */
-  void DecodeSample(SampleView<OutBackend> output, int64_t idx) {
-    auto &frames_decoder = *frames_decoders_[idx];
-    int64_t num_frames = output.shape()[0];
+  bool DecodeFrames(SampleView<OutBackend> output, int64_t sample_idx, int64_t num_frames,
+                    std::optional<SampleView<OutBackend>> pad_value = std::nullopt,
+                    cudaStream_t stream = 0) {
```
mzient, line 86: I'd recommend swapping the two arguments and leaving the `stream` without a default value - this will help us avoid accidentally decoding stuff on stream 0.

dali/pipeline/operator/builtin/input_operator.h
```diff
@@ -437,6 +441,17 @@ class InputOperator : public Operator<Backend>, virtual public BatchSizeProvider
   AccessOrder internal_copy_order_ = AccessOrder::host();
 };
 
+
+/**
+ * Checks, if the Operator defined by provided Schema is an InputOperator
+ */
+inline bool IsInputOperator(const OpSchema &schema) {
+  auto parents = schema.GetParents();
```
mzient, line 449: Unnecessary copy
```suggestion
  const auto &parents = schema.GetParents();
```

dali/operators/decoder/video/video_decoder_base.h
mzient: Another take: use `std::optional<cudaStream_t>` and keep the default as `nullopt` - the CPU variant will try to obtain the stream and fail fast in case where the value wasn't supplied.

dali/pipeline/operator/operator.cc
```diff
@@ -23,6 +24,10 @@ void OperatorBase::EnforceUniformInputBatchSize(const Workspace &ws) const {
   if (IsSplitOrMerge(spec_.GetSchema())) {
     return;
   }
+  // InputOperators have relaxed checks, since they actually create a batch.
+  if (IsInputOperator(spec_.GetSchema())) {
+    return;
+  }
```
mzient, line 30: I don't think it's OK - ExternalSource worked before, so why wouldn't Input Operators? I don't think we can relax it for them.

szalpal: Good point, it shouldn't be relaxed for input ops, but for VideoInput. Changing.

dali/pipeline/pipeline.h
```diff
@@ -134,13 +134,13 @@ class DLL_PUBLIC Pipeline {
   template<typename T, typename OperatorBackend>
   void
   SetDataSourceHelper(const string &name, const T &tl, OperatorBase *op_ptr, AccessOrder order = {},
-                      InputOperatorSettingMode ext_src_setting_mode = {}) {
+                      InputOperatorSettingMode in_op_setting_mode = {}) {
```
mzient, line 137: What's a "setting mode"? This is not a good name - verbose, but still ambiguous.

szalpal: This has previously been called `ExtSrcSettingMode`.
https://github.com/NVIDIA/DALI/blob/1fee392318c202b82c7ccf4d0bfd60736fd2fd4f/dali/pipeline/operator/builtin/external_source.h#L68

So I just followed the pattern. While this PR might not be a best place to conduct the renaming, if you feel there's a better name I can refactor it in the next PR.

dali/pipeline/operator/builtin/input_operator.h
szalpal: Done

dali/operators/decoder/video/video_decoder_base.h
szalpal: Changed to `std::optional<cudaStream_t>`

mzient (approved):

stiepan (approved):

szalpal:
!build

PR #4543, opened 2022-12-30
Split DALI operator tests into two jobs

JanuszL:
- as DALI operator tests last longer and longer
  split it into two jobs to avoid timeouts in CI

## Category:

## Description:
- as DALI operator tests last longer and longer
  split it into two jobs to avoid timeouts in CI
 

- tests

- NA

### Tests:

- [x] Existing tests apply
  - all tests from dali/test/python/operator and dali/test/python/operator_b
- [ ] New tests added
  - [ ] Python tests
  - [ ] GTests
  - [ ] Benchmark
  - [ ] Other
- [ ] N/A

### Documentation
- [ ] Documentation updated
  - [ ] Docstring
  - [ ] Doxygen
  - [ ] RST
  - [ ] Jupyter
  - [ ] Other
- [x] N/A

#### Requirements
- [x] N/A

**REQ IDs**: N/A

**JIRA TASK**: N/A


qa/TL0_python-self-test/test.sh
```diff
@@ -11,3 +11,7 @@ popd
 pushd ../TL0_python-self-test-operators
 bash -e ./test.sh
 popd
+
+pushd ../TL0_python-self-test-operators_b
+bash -e ./test.sh
+popd
```
mzient, line 17:
```suggestion
pushd ../TL0_python-self-test-operators_1
bash -e ./test.sh
popd

pushd ../TL0_python-self-test-operators_2
bash -e ./test.sh
popd
```
or
```suggestion
pushd ../TL0_python-self-test-operators_a
bash -e ./test.sh
popd

pushd ../TL0_python-self-test-operators_b
bash -e ./test.sh
popd
```

JanuszL: Changed to `operators_1` and `operators_2`.

mzient (approved):

JanuszL:
> I don't quite like the change to the directory structure, as I believe it's going to be hard to maintain an order there. I would have prefer a directory structure reflecting the nature of the tests:
> 
> ```
> - operators
> --- audio
> --- random
> --- image
> --- ...
> ```
> 
> And later the two jobs picking a subset of the directories.
> 
> However, let's go with this approach and we'll adjust if we see the need

Adding new dirs to the python test could be troublesome, now we just have a bag and that is all. Still, we can do a follow-up if the initial approach works fine.

PR #4564, opened 2023-01-13
Operator trace mechanism

szalpal:
## Category:

## Description:
This PR adds an Operator Trace mechanism. Until now, user had no inside to what happens in the particular operators - he has only seen whatever makes it to the output of the pipeline. The Operator Trace mechanism allows user to communicate with specific operators within the pipeline.

The operator traces are mainly required for the VideoInput operator. When using the VideoInput from the C API, user needs to know, when the VideoInput has consumed the data provided by external source. Using operator traces, user will be able to provide the `data_id` (this feature is the next step of this PR) and use it to identify the status of the input data in the VideoInput operator.

Operator traces are currently exposed only in C API. The entry points to the feature are the following functions:
```c++
int daliHasOperatorTrace(daliPipelineHandle *pipe_handle, const char *operator_name, const char *trace_name);
const char * daliGetOperatorTrace(daliPipelineHandle *pipe_handle, const char *operator_name, const char *trace_name);
```
Please refer to their documentation for more details.

## Additional information:
Additional things, that happened in this PR:
* Introducing `daliCreatePipeline2` function in C API.
* Introducing functions to use Operator Traces feature in the C API.
* Introduced a notion of an `iteration_id`.
* Adding an Operator Name to the Workspace.
* Introduced an `IterationData`.
* Added `PassthoughOperator` also to `dali_operator_test` binary.
* Added tests for operator traces.
* Extending test-wise `Passthrough` operator with Operator Trace feature.

Operator traces are kept in a new data structure called `IterationData`. The point of the operator traces is to keep the data for the whole iteration - maintaining it across the stages. Therefore this new data structure has been introduced. The executor keeps a queue of these `IterationData`, to properly handle the Prefetching.

While working on the operator traces, it became apparent, that we need an identifier of the iteration, that couples the workspaces across the stages. From now on, the Executor tracks the iteration ID. The iteration ID is used to access the `IterationData` corresponding to a given iteration. Also, to properly handle the OperatorTraceMap, the Workspace needs to know, which operator does it belong to. Therefore, an information about the Operator instance has been added to the Workspace.

To make the tests easier, PassthroughOperator has been added to the `dali_operator_test` binary. Previously it has only been enabled in dali_test binary.

Existing `daliCreatePipeline` function in the C API did not allow to create a Pipeline, that executes within a `SimpleExecutor` (i.e. the one that has `exec_async=False` and `exec_pipelined=False`). Since such Pipeline has been required for the test of the Operator Traces, `daliCreatePipeline2` function has been created.

### Tests:

- [ ] Existing tests apply
- [x] New tests added
  - [ ] Python tests
  - [x] GTests
  - [ ] Benchmark
  - [ ] Other
- [ ] N/A

### Documentation
- [ ] Documentation updated
  - [ ] Docstring
  - [x] Doxygen
  - [ ] RST
  - [ ] Jupyter
  - [ ] Other
- [ ] N/A

#### Requirements
- [ ] N/A

**REQ IDs**: N/A

**JIRA TASK**: N/A


szalpal:
!build

dali/pipeline/executor/executor.cc
```diff
@@ -436,6 +441,67 @@ void Executor<WorkspacePolicy, QueuePolicy>::RunHelper(OpNode &op_node, Workspac
 }
 
 
+template <typename WorkspacePolicy, typename QueuePolicy>
+void Executor<WorkspacePolicy, QueuePolicy>::ShareOutputsImpl(Workspace *ws, size_t iteration_id) {
+  DALI_ENFORCE(ws != nullptr, "Workspace is nullptr");
```
szalpal, line 446: This whole function has been copy-pasted from `executor.h`, with only the lines `470:471` added.

dali/c_api/c_api.cc
```diff
@@ -547,6 +563,23 @@ device_type_t daliGetOutputDevice(daliPipelineHandle *pipe_handle, int id) {
...
+  auto traces = ws->GetOperatorTraces(operator_name);
+  return traces.find(trace_name) != traces.end() ? 0 : 1;
+}
+
+
+const char *
+daliGetOperatorTrace(daliPipelineHandle *pipe_handle, const char *operator_name,
+                     const char *event_name) {
+  auto *ws = reinterpret_cast<dali::Workspace *>(pipe_handle->ws);
+  return ws->GetOperatorTraces(operator_name).at(event_name).c_str();
```
JanuszL, line 579: What if there is not `event_name`? Will we get an exception?

dali/pipeline/executor/executor.cc
```diff
@@ -461,6 +527,34 @@ int Executor<WorkspacePolicy, QueuePolicy>::InferBatchSize(
...
+template<typename WorkspacePolicy, typename QueuePolicy>
+IterationData &
+Executor<WorkspacePolicy, QueuePolicy>::GetCurrentIterationData(size_t iteration_id) {
+  return iteration_data_[iteration_id % iteration_data_.size()];
+}
+
+
+template<typename WorkspacePolicy, typename QueuePolicy>
+size_t Executor<WorkspacePolicy, QueuePolicy>::CalcIterationDataSize() const {
+  return 2;
```
JanuszL, line 554: Why 2?

dali/pipeline/workspace/workspace.h
```diff
@@ -526,20 +527,92 @@ class WorkspaceBase : public ArgumentWorkspace {
...
+    return operator_id_;
+  }
+  /// @}
+
+
+  ///@{
+  /**
+   * Set the whole operator trace map in this workspace.
+   *
+   * Sets the operator trace map that corresponds to all operators in the current iteration.
```
JanuszL, line 567: `to all operators`? Isn't this maps specific to the particular operator?

include/dali/c_api.h
```diff
@@ -120,6 +120,34 @@ DLL_PUBLIC void daliCreatePipeline(daliPipelineHandle *pipe_handle, const char *
                                    int cpu_prefetch_queue_depth, int gpu_prefetch_queue_depth,
                                    int enable_memory_stats);
 
+/**
```
JanuszL, line 123: I would align the docs string with the `daliCreatePipeline`

include/dali/c_api.h
```diff
@@ -483,6 +511,43 @@ DLL_PUBLIC const char *daliGetOutputName(daliPipelineHandle *pipe_handle, int id
...
+ *
+ * Operator Traces is a communication mechanism with particular operators in the pipeline. For
+ * more information @see operator_trace_map_t.
+ *
+ * User does not own the returned value. In a situation, when changing of this value is necessary,
+ * user shall copy it to his own memory. The lifetime of this value ends, when the
+ * daliOutputRelease() is called.
+ *
+ * User shall check, if the trace with given name exists (@see daliHasOperatorTrace). If it does
+ * not, the result of this function is undefined. Specifying the name of non-existing operator is
```
JanuszL, line 538: `undefined`? I'm afraid it would just crash. Undefined means usually just garbage.

dali/pipeline/executor/executor.cc
```diff
@@ -461,6 +527,34 @@ int Executor<WorkspacePolicy, QueuePolicy>::InferBatchSize(
...
+template<typename WorkspacePolicy, typename QueuePolicy>
+void Executor<WorkspacePolicy, QueuePolicy>::InitIterationData() {
+  cpu_iteration_id_ = 0;
+  mixed_iteration_id_ = 0;
+  gpu_iteration_id_ = 0;
+  output_iteration_id_ = 0;
+  size_t iteration_data_size = CalcIterationDataSize();
+  iteration_data_.resize(iteration_data_size);
+  for (auto& id : iteration_data_) {
+    id.operator_traces = std::make_shared<operator_trace_map_t>();
```
JanuszL, line 540: Do we need to make it a shared_ptr?

dali/c_api/operator_trace_test.cc
```diff
@@ -0,0 +1,154 @@
...
+  bool exec_pipelined_, exec_async_, exec_separated_;
+  int cpu_queue_depth_, gpu_queue_depth_;
+  std::string serialized_pipeline_;
+};
+
+
+TEST_P(OperatorTraceTest, OperatorTraceTest) {
+  daliPipelineHandle h;
+  daliDeserializeDefault(&h, serialized_pipeline_.c_str(), serialized_pipeline_.size());
+  daliCreatePipeline2(&h, serialized_pipeline_.c_str(), serialized_pipeline_.length(), batch_size_,
```
JanuszL, line 117: Do we need to call both of them?

include/dali/c_api.h
szalpal: I believe they are aligned, it's just GitHub's weird rendering:
image

include/dali/c_api.h
szalpal: True. I've added some exception handling, but I've left the `undefined` specification. Although the behaviour right now is pretty defined (returning `nullptr`), I feel that checking the presence of a given trace is a proper usage of the API. That's because we don't want the user to abuse the exception mechanism to check the trace - he should use `daliHasOperatorTrace` function for this.

dali/c_api/operator_trace_test.cc
szalpal: Good catch. Removed the former.

dali/pipeline/workspace/workspace.h
szalpal: `operator_trace_map_t` corresponds to all operators in a given iteration:
```c++
using operator_trace_map_t = std::unordered_map<
        std::string /* op_name */,
        std::unordered_map<std::string /* trace_name */, std::string /* trace_value */>
>;
```
All Workspaces in a given iteration must know all the traces in that iteration. This is stipulated by the output mechanism - output is just another workspace. Therefore, to access any particular op in the iteration, the knowledge of every operator traces is required.

dali/c_api/c_api.cc
szalpal: Yes, the exception would be thrown. Added the exception handling here, since I believe we don't want to throw an exception from the C API.

dali/pipeline/executor/executor.cc
szalpal: Yes. This is because all the workspaces in a given iteration have to know all the operator traces. I've explained it also in [this comment](https://github.com/NVIDIA/DALI/pull/4564#discussion_r1070287929). Therefore, the easiest way would be if they share the same `operator_trace_map_t`.

dali/pipeline/executor/executor.cc
```diff
@@ -436,6 +441,67 @@ void Executor<WorkspacePolicy, QueuePolicy>::RunHelper(OpNode &op_node, Workspac
...
+    auto op_type = graph_->Node(out_tensor.producer.node).op_type;
+    auto storage_dev = out_tensor.producer.storage_device;
+    VALUE_SWITCH(storage_dev, storage_dev_static, (StorageDevice::GPU, StorageDevice::CPU), (
+      VALUE_SWITCH(op_type, op_type_static, (OpType::CPU, OpType::MIXED, OpType::GPU), (
+        auto &queue =
+               get_queue<op_type_static, storage_dev_static>(tensor_to_store_queue_[out_tensor_id]);
+        auto stage_output_idx = output_idx[op_type_static];
+        ws->AddOutput(queue[stage_output_idx]);
+        ws->InjectOperatorTraces(std::make_shared<operator_trace_map_t>(
+                *GetCurrentIterationData(iteration_id).operator_traces));
```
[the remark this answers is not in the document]
szalpal: Yes, good catch, thanks.

dali/pipeline/executor/executor.cc
szalpal: It'd be best to explain it on the sketch
```
iteration ID
 CPU 0       1
 MXD   0       1
 GPU       0       1
time 0 1 2 3 4 5 6 7 8 9
           ^^^ Here the iteration 0 is being read by the user (therefore we need to keep it)
               while iteration 1 is being conducted, thus Ops write to OpTraceMap.
```

dali/pipeline/workspace/workspace.h
```diff
@@ -526,20 +527,92 @@ class WorkspaceBase : public ArgumentWorkspace {
...
+
+  /**
+   * Set the trace value for the current operator.
+   *
+   * Typically, this function shall be called by an operator in RunImpl or SetupImpl.
+   *
+   * @see operator_trace_map_t
+   */
+  DLL_PUBLIC void SetOperatorTrace(const std::string &trace_key, std::string trace_value) {
+    (*operator_traces_)[GetOperatorId()].insert_or_assign(trace_key, std::move(trace_value));
```
klecki, line 585: Can this potentially modify the trace map from multiple threads? We discussed some scenario close to this problem, assuming that subsequent stages do not run in parallel, it should not happen.

klecki (commented):
Some comments, I will finish the rest tomorrow.

dali/pipeline/executor/executor.cc
szalpal: Is such action possible with the `SimpleExecutor`. If it is, than probably this is wrong.

dali/pipeline/workspace/workspace.h
szalpal: Yes, I believe it can (when Operator does it in using ThreadPool). Added a safe lock.

dali/pipeline/workspace/workspace.h
```diff
@@ -526,20 +527,92 @@ class WorkspaceBase : public ArgumentWorkspace {
...
+  }
+
+
+
+  /**
+   * Get the trace map for a given operator.
+   *
+   * Returns a map, that maps an trace key to a trace value: `ret_value[trace_key] = trace_value`.
```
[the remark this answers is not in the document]
szalpal: Done

dali/pipeline/workspace/workspace.h
```diff
@@ -526,20 +527,92 @@ class WorkspaceBase : public ArgumentWorkspace {
...
+   * @brief Returns the index of the thread that will process this data.
+   */
   DLL_PUBLIC virtual inline int thread_idx() const {
     return 0;
   }
 
+
+  /// @{
+  /**
+   * Sets the operator ID that this Workspace belongs to.
```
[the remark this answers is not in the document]
szalpal: Done

dali/pipeline/workspace/workspace.h
```diff
@@ -684,10 +757,17 @@ class WorkspaceBase : public ArgumentWorkspace {
     return index_map[idx];
   }
 
+
   AccessOrder output_order_ = AccessOrder::host();
   ThreadPool *thread_pool_ = nullptr;
   cudaEvent_t event_ = nullptr;
   SmallVector<cudaEvent_t, 4> parent_events_;
+
+  /// Name of the instance of the operator, to which this Workspace belongs.
```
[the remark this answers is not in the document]
szalpal: Done

dali/pipeline/workspace/workspace.h
```diff
@@ -526,20 +527,92 @@ class WorkspaceBase : public ArgumentWorkspace {
...
+  /**
+   * Sets the operator ID that this Workspace belongs to.
+   */
+  void SetOperatorId(std::string operator_id) {
+    operator_id_ = std::move(operator_id);
+  }
+
+
+  /**
+   * Returns the operator ID that this Workspace belongs to.
```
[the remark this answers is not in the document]
szalpal: Fixed the wording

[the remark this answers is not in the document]
szalpal: Added a remark about the operator id in error messages.

Do you believe we should figure out better mechanism, which will be accessible also when Workspace is not available? Essentially, I believe the Workspace is always available, right?

dali/pipeline/workspace/workspace.h
szalpal: On the second thought, I won't.

We decided, that if the operator writes the operator trace in multiple threads - he should handle the thread safety. Furthermore, it's impossible for multiple operators of one iteration to run in parallel, so this also is safe. Lastly, `std::mutex` has no move-ctor, thus implementing the safe lock would require some hacking.

awolant (approved):

dali/test/operators/passthrough.h
```diff
@@ -35,9 +35,15 @@ class PassthroughOp : public Operator<Backend> {
     return false;
   }
 
+
   void RunImpl(Workspace &ws) override {
     ws.Output<Backend>(0).ShareData(ws.Input<Backend>(0));
+    ws.SetOperatorTrace("test_trace", make_string("test_value", iteration_id++));
```
klecki, line 41: One thing I am not very fond of is the fact, that this is operator intended to be used in some test and exhibit specific behaviour. I won't mind creating another one that is used to test traces, but I am not sure if we should pile all functionality that we want to test into it. It's not a PassthroughAndSetTraceOp, and I don't want it to become PassthroughAndSetTraceAndDoSomeOtherThingOrTwoOp.

dali/c_api/c_api.cc
```diff
@@ -186,16 +186,31 @@ void daliInitialize() {
...
+
+DLL_PUBLIC void
+daliCreatePipeline2(daliPipelineHandle *pipe_handle, const char *serialized_pipeline, int length,
+                    int max_batch_size, int num_threads, int device_id, int pipelined_execution,
+                    int async_execution, int separated_execution, int prefetch_queue_depth,
+                    int cpu_prefetch_queue_depth, int gpu_prefetch_queue_depth,
+                    int enable_memory_stats) {
+  bool se = separated_execution == 0;
+  bool pe = pipelined_execution == 0;
+  bool ae = async_execution == 0;
```
klecki, line 209: OMG, why do you treat 0 as true?
```suggestion
  bool se = separated_execution;
  bool pe = pipelined_execution;
  bool ae = async_execution;
```

dali/pipeline/workspace/workspace.h
```diff
@@ -526,20 +527,95 @@ class WorkspaceBase : public ArgumentWorkspace {
...
   DLL_PUBLIC virtual inline int thread_idx() const {
     return 0;
   }
 
+
+  /// @{
+  /**
+   * Sets the operator ID that this Workspace in associated with.
+   */
+  void SetOperatorId(std::string operator_id) {
```
klecki, line 549: One more thing, why do we call it ID if it's instance_name? Should we keep this naming?

include/dali/c_api.h
```diff
@@ -120,6 +120,34 @@ DLL_PUBLIC void daliCreatePipeline(daliPipelineHandle *pipe_handle, const char *
...
+ * @param pipe_handle Pipeline handle.
+ * @param serialized_pipeline Serialized pipeline.
+ * @param length Length of the serialized pipeline string.
+ * @param max_batch_size Maximum batch size.
+ * @param num_threads Number of CPU threads which this pipeline uses.
+ * @param device_id ID of the GPU device which this pipeline uses.
+ * @param pipelined_execution If 0, this pipeline will execute in Pipeline mode.
+ * @param async_execution If 0, this pipeline will execute asynchronously.
+ * @param separated_execution If 0, this pipeline will have different depths
+ *                            of the CPU and GPU prefetching queues.
```
klecki, line 135: Please, treat the 1 and 0 as true and false, not the other way round.

dali/c_api/c_api.cc
```diff
@@ -547,6 +563,31 @@ device_type_t daliGetOutputDevice(daliPipelineHandle *pipe_handle, int id) {
   return device_type_t::CPU;
 }
 
+
+int daliHasOperatorTrace(daliPipelineHandle *pipe_handle, const char *operator_name,
+                         const char *trace_name) {
+  auto *ws = reinterpret_cast<dali::Workspace *>(pipe_handle->ws);
+  try {
+    auto traces = ws->GetOperatorTraces(operator_name);
```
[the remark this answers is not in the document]
szalpal: Actually, no idea. `GetOperatorTraces` returns `const&`, so I guess it shouldn't copy. To be sure added the `&`

dali/c_api/c_api.cc
```diff
@@ -547,6 +563,31 @@ device_type_t daliGetOutputDevice(daliPipelineHandle *pipe_handle, int id) {
...
+    return traces.find(trace_name) != traces.end() ? 0 : 1;
```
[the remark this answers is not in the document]
szalpal: Done

dali/c_api/c_api.cc
szalpal: Done

dali/pipeline/executor/executor.cc
```diff
@@ -436,6 +439,66 @@ void Executor<WorkspacePolicy, QueuePolicy>::RunHelper(OpNode &op_node, Workspac
...
+        auto &queue =
+               get_queue<op_type_static, storage_dev_static>(tensor_to_store_queue_[out_tensor_id]);
+        auto stage_output_idx = output_idx[op_type_static];
+        ws->AddOutput(queue[stage_output_idx]);
+        ws->InjectOperatorTraces(GetCurrentIterationData(iteration_id).operator_traces);
+      ), DALI_FAIL("Invalid op type"));  // NOLINT(whitespace/parens)
+    ), DALI_FAIL("Invalid storage device"));  // NOLINT(whitespace/parens)
+  }
+
+
```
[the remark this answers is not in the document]
szalpal: Done

dali/pipeline/executor/executor.cc
```diff
@@ -436,6 +439,66 @@ void Executor<WorkspacePolicy, QueuePolicy>::RunHelper(OpNode &op_node, Workspac
...
+    auto &out_tensor = graph_->Tensor(out_tensor_id);
+    auto op_type = graph_->Node(out_tensor.producer.node).op_type;
+    auto storage_dev = out_tensor.producer.storage_device;
+    VALUE_SWITCH(storage_dev, storage_dev_static, (StorageDevice::GPU, StorageDevice::CPU), (
+      VALUE_SWITCH(op_type, op_type_static, (OpType::CPU, OpType::MIXED, OpType::GPU), (
+        auto &queue =
+               get_queue<op_type_static, storage_dev_static>(tensor_to_store_queue_[out_tensor_id]);
+        auto stage_output_idx = output_idx[op_type_static];
+        ws->AddOutput(queue[stage_output_idx]);
+        ws->InjectOperatorTraces(GetCurrentIterationData(iteration_id).operator_traces);
```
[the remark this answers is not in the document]
szalpal: Done

dali/c_api/c_api.cc
```diff
@@ -547,6 +563,31 @@ device_type_t daliGetOutputDevice(daliPipelineHandle *pipe_handle, int id) {
...
+  }
+}
+
+
+const char *
+daliGetOperatorTrace(daliPipelineHandle *pipe_handle, const char *operator_name,
+                     const char *trace_name) {
+  auto *ws = reinterpret_cast<dali::Workspace *>(pipe_handle->ws);
+  try {
+    return ws->GetOperatorTraces(operator_name).at(trace_name).c_str();
```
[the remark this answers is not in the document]
szalpal: Done

dali/pipeline/workspace/workspace.h
szalpal: I'd say, that `id` and `instance_name` can be two different concepts and it's just a coincidence that they have the same value for now. ID would uniquely identify every operator, while I don't know if it's a good idea to guarantee that for `instance_name`. Furthermore, `instance_name` is what Executor knows, while ID is something that Workspace is aware about. As I said, from my perspective these are two separate things.

If you really insist I'll change to `SetOperatorInstanceName`, but I'd prefer to keep the ID. It's shorter...

dali/test/operators/passthrough.h
szalpal: Added another operator. 100 extra lines of code :D

dali/pipeline/workspace/workspace.h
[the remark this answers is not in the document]
klecki: Workspace is not available in the constructor, only in the Setup and Run.

dali/pipeline/workspace/workspace.h
klecki: What I mean here, we set it to instance name. Executor also has operator id, but it's just some integer identifying the operator node in the graph.

klecki (approved):

szalpal:
!build

PR #4588, opened 2023-01-19
Add a requirement for CUDA toolkit for CUDA 12 builds

klecki:
## Description:
Improve the documentation by linking to the CUDA toolkit installation instruction while mentioning it is required for CUDA 12-based builds.

docs/installation

Better wording?

### Tests:

- [ ] Existing tests apply
- [ ] New tests added
  - [ ] Python tests
  - [ ] GTests
  - [ ] Benchmark
  - [ ] Other
- [x] N/A

### Documentation
- [ ] Documentation updated
  - [ ] Docstring
  - [ ] Doxygen
  - [x] RST
  - [ ] Jupyter
  - [ ] Other
- [ ] N/A

#### Requirements
- [ ] N/A

**REQ IDs**: N/A

**JIRA TASK**: N/A

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -138,7 +138,7 @@ To install the latest DALI release for the latest CUDA version (12.x)::
     pip install --extra-index-url https://developer.download.nvidia.com/compute/redist --upgrade nvidia-dali-cuda120
 
 DALI requires `NVIDIA driver <https://www.nvidia.com/drivers>`_ supporting the appropriate CUDA version.
-In the case of DALI based on CUDA 12 it requires `CUDA Toolkit <https://docs.nvidia.com/cuda/cuda-installation-guide-linux/index.html>`_
+In case of DALI based on CUDA 12, it requires `CUDA Toolkit <https://docs.nvidia.com/cuda/cuda-installation-guide-linux/index.html>`_
 to be installed.
 
 DALI comes preinstalled in the `TensorFlow <https://catalog.ngc.nvidia.com/orgs/nvidia/containers/tensorflow>`_,
--- a/README.rst
+++ b/README.rst
@@ -1,5 +1,4 @@
 |License|  |Documentation|
-|LGTM Python badge|
 
 NVIDIA DALI
 ===========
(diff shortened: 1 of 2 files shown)


klecki:
!build

jantonguirao (approved):

README.rst
```diff
@@ -127,16 +127,16 @@ We strongly encourage you to comment on our roadmap and provide us feedback on t
...
 ----
 
 Installing DALI
 ---------------
 
-To install the latest DALI release for the latest CUDA version (11.x)::
+To install the latest DALI release for the latest CUDA version (12.x)::
 
-    pip install --extra-index-url https://developer.download.nvidia.com/compute/redist --upgrade nvidia-dali-cuda110
+    pip install --extra-index-url https://developer.download.nvidia.com/compute/redist --upgrade nvidia-dali-cuda120
```
szalpal, line 139: How about putting the information about necessary CUDA install also here?
image
So that when somebody goes to the README, pastes these instructions in "installing DALI", it works?

klecki: Do you have some wording to suggest? I think the problem is, that the CUDA toolkit may be installed in several different ways depending on the system and you shouldn't mix those if you already used one of installation paths already.
Especially if you have toolkit, you don't want to repeat the install by copy-pasting it.

szalpal: Generally, right now user needs to know, that he needs to install CUDA first and then DALI. Anything that would suffice it works.

How about something along these lines:
```
A first step to installing DALI is to install CUDA toolkit. There are many possible ways of installing it, e.g.:
1. [Using `pip install`](https://nvidia.github.io/cuda-python/install.html#installing-from-pypi): `pip install cuda-python`
1. [Using runfile](https://developer.nvidia.com/cuda-downloads?target_os=Linux&target_arch=x86_64&Distribution=Ubuntu&target_version=22.04&target_type=runfile_local) 
1. [Installing from conda](https://nvidia.github.io/cuda-python/install.html#installing-from-conda)

If the CUDA has been installed previously, it is not necessary to reinstall using any of these methods.

To install the latest DALI release [...]
```

klecki: It is already in requirements, linked to the CUDA Toolkit installation guide https://docs.nvidia.com/cuda/cuda-installation-guide-linux/index.html
which covers the topic extensively. I don't think it is good idea to copy a small subset of this page here.
Typically users working on DL will already have the CUDA installed for the purposes of the framework.
We don't specify how to install any of the frameworks or the driver either.

klecki:
!build

szalpal (approved):

PR #4591, opened 2023-01-20
Add experimental FITS reader for CPU backend

aderylo:
### Tests:

- [ ] Existing tests apply
- [x] New tests added
  - [x] Python tests
  - [ ] GTests
  - [ ] Benchmark
  - [ ] Other
- [ ] N/A

### Documentation
- [ ] Documentation updated
  - [ ] Docstring
  - [ ] Doxygen
  - [ ] RST
  - [ ] Jupyter
  - [ ] Other
- [ ] N/A

#### Requirements
- [ ] N/A

**REQ IDs**: N/A

**JIRA TASK**: N/A


dali/test/python/reader/test_fits.py
```diff
@@ -0,0 +1,161 @@
...
+from numpy.testing import assert_array_equal
+
+rng = np.random.RandomState(12345)
+
+gds_data_root = '/scratch/'
+if not os.path.isdir(gds_data_root):
+    gds_data_root = os.getcwd() + "/scratch/"
+    if not os.path.isdir(gds_data_root):
+        os.mkdir(gds_data_root)
+        assert os.path.isdir(gds_data_root)
```
JanuszL, line 32: You don't need that for fits. It is used for the GPU Direct Storage solution that works only on the real file system, as our tests run inside the docker environment and its tmp file system doesn't work with GDS so we need to mount the external file system and use it for tests.

dali/operators/reader/fits_reader_op.cc
```diff
@@ -0,0 +1,152 @@
...
+    .DocStr("Legacy alias for :meth:`readers.fits`.")
+    .NumInput(0)
+    .NumOutput(1)  // (Arrays)
+    .AddParent("readers__Fits")
+    .MakeDocPartiallyHidden()
+    .Deprecate(
+        "readers__Fits",
+        R"code(In DALI 1.0 all readers were moved into a dedicated :mod:`~nvidia.dali.fn.readers`
+submodule and renamed to follow a common pattern. This is a placeholder operator with identical
+functionality to allow for backward compatibility.)code");  // Deprecated in 1.0;
```
klecki, line 129: Please remove.

dali/operators/reader/fits_reader_op.cc
```diff
@@ -0,0 +1,152 @@
...
+      int64_t b_end = prev_b_start = nbytes * (b + 1) / req_nblocks;
+      int64_t b_size = b_end - b_start;
+      thread_pool.AddWork([=](int tid) {
+        std::memcpy(out_ptr + b_start, in_ptr + b_start, b_size); 
+      }, b_size);
+    }
+  }
+}
+
+DALI_REGISTER_OPERATOR(readers__Fits, FitsReaderCPU, CPU);
```
klecki, line 50: I think, we should start by placing this in experimental module, at least untill we get some more functionality:
```suggestion
DALI_REGISTER_OPERATOR(experimental__readers__Fits, FitsReaderCPU, CPU);
```
This will mean that you can access the operator as `experimental.readers.fits`, so you would need to adjust the tests.

dali/operators/reader/fits_reader_op.cc
```diff
@@ -0,0 +1,152 @@
...
+
+DALI_SCHEMA(readers__Fits)
```
klecki, line 52: And adjust here as well:
```suggestion
DALI_SCHEMA(experimental__readers__Fits)
```

dali/operators/reader/fits_reader_op.cc
```diff
@@ -0,0 +1,152 @@
...
+                    R"code(Applies **only** to the ``gpu`` backend type.
+
+.. warning::
+    This argument is temporarily disabled and left for backward compatibility.
+    It will be reenabled in the future releases.
+
+If true, the device I/O buffers will be registered with cuFile. It is not recommended if the sample
+sizes vary a lot.)code",
+                    true)
+    .AddParent("LoaderBase");
```
klecki, line 113: As discussed, I would ask for following adjustments to the API:
* allow the user to specify which HDU indices they want to read. This would ensure easy mapping of 1 sample (potentially consisting of several images) to 1 file as well as allowing us to extend the reader with ability to load other types of HDUs (like some arrays with metadata). This can be something like `hdu_indices=[1, 4]` argument

Each requested HDU should map to one output. You can handle this by specifying a lambda via `AdditionalOutputsFn` that would calculate how many outputs are there.

Maybe we can have some kind of default right now, for example if the user doesn't provide any required outpus, they want the first image HDU so we always have at least one output.

* add the ability to specify the `dtype` and `ndim` of each output - this should be mandatory, it should accept one or more values (the same number as requested HDUs to return). We should check if the user provided correct values and raise a meaningful error if not. 

* Optionally, we can add some argument that allows the user to ensure if they are reading HDU with a correct extension.

dali/operators/reader/loader/fits_loader.cc
```diff
@@ -0,0 +1,92 @@
...
+  Index pastIndex = current_index_;
+
+  // hanlde moving through HDUs, assuming that primary hdu is for meta data
+  auto path = filesystem::join_path(file_root_, filename);
+  fits_open_file(&current_file, path.c_str(), READONLY, &status);
+  fits_get_num_hdus(current_file, &hdunum, &status);
+  fits_movabs_hdu(current_file, hdu_index_, NULL, &status);
+
+  if (hdu_index_ < hdunum) {
+    hdu_index_++;
```
klecki, line 45: The suggested approach would mean, that we potentially read one sample consisting of multiple HDUs.

klecki (commented):
The comments related to the operator API and some code trimming suggestions.

dali/operators/reader/loader/fits_loader.cc
```diff
@@ -0,0 +1,92 @@
...
+    fits_report_error(stderr, status);
+  }
+
+  // read the header
+  fits::HeaderData header;
+  try {
+    fits::ParseHeader(header, current_file);
+  } catch (const std::runtime_error& e) {
+    DALI_FAIL(e.what() + ". File: " + filename);
+  }
```
klecki, line 72: It should be enough to run it directly and let the exception be handled further. I think only the DALI_ENFORCE in this function may throw.
```suggestion 
  fits::ParseHeader(header, current_file); 
```

dali/util/fits.cc
```diff
@@ -0,0 +1,82 @@
...
+  long dims[MAX_DIMS] = {0};
+
+  fits_get_hdu_type(src, &hdu_type, &status);
+  bool is_image = (hdu_type == IMAGE_HDU);
+  DALI_ENFORCE(is_image, "Only IMAGE_HDUs are supported!");
+  parsed_header.hdu_type = hdu_type;
+
+  fits_get_img_param(src, MAX_DIMS, &img_type, &n_dims, dims, &status);
+  parsed_header.type_info = &TypeFromFitsImageType(img_type);
+  parsed_header.compressed = (fits_is_compressed_image(src, &status) == 1);
```
klecki, line 58: Should we error out on the compressed files until we merge the option to decompress them?

dali/test/python/reader/test_fits.py
aderylo: Sure, addressed by removing this fragment.

dali/operators/reader/fits_reader_op.cc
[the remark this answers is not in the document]
aderylo: Issue addressed.

dali/operators/reader/fits_reader_op.cc
```diff
@@ -0,0 +1,152 @@
...
+
+.. note::
+  The ``gpu`` backend requires cuFile/GDS support (418.x driver family or newer). which is
+  shipped with the CUDA toolkit starting from CUDA 11.4. Please check the GDS documentation
+  for more details.
+
+  The ``gpu`` reader reads the files in chunks. The size of the chunk can be controlled
+  process-wide with an environment variable ``DALI_GDS_CHUNK_SIZE``. Valid values are powers of 2
+  between 4096 and 16M, with the default being 2M. For convenience, the value can be specified
+  with a k or M suffix, applying a multiplier of 1024 and 2^20, respectively.
```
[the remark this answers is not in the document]
aderylo: Addressed.

dali/operators/reader/fits_reader_op.cc
```diff
@@ -0,0 +1,152 @@
...
+If not using ``file_list`` or ``files``, this directory is traversed to discover the files.
+``file_root`` is required in this mode of operation.)",
+                            nullptr)
+    .AddOptionalArg(
+        "file_filter",
+        R"(If a value is specified, the string is interpreted as glob string to filter the
+list of files in the sub-directories of the ``file_root``.
+
+This argument is ignored when file paths are taken from ``file_list`` or ``files``.)",
+        "*.npy")
```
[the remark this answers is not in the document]
aderylo: Addressed.

dali/operators/reader/fits_reader_op.cc
```diff
@@ -0,0 +1,152 @@
...
+                                    nullptr)
+    .AddOptionalArg("register_buffers",
+                    R"code(Applies **only** to the ``gpu`` backend type.
+
+.. warning::
+    This argument is temporarily disabled and left for backward compatibility.
+    It will be reenabled in the future releases.
+
+If true, the device I/O buffers will be registered with cuFile. It is not recommended if the sample
+sizes vary a lot.)code",
```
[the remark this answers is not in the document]
aderylo: Addressed.

dali/operators/reader/fits_reader_op.cc
aderylo: Addressed, followed through with the suggestion.

dali/operators/reader/fits_reader_op.cc
aderylo: Addressed.

dali/operators/reader/fits_reader_op.h
```diff
@@ -0,0 +1,81 @@
...
+    DataReader<Backend, Target>::SetupImpl(output_desc, ws);
+
+    int batch_size = GetCurrBatchSize();
+    const auto& file_0 = GetSample(0);
+    DALIDataType output_type = file_0.get_type();
+    int ndim = file_0.get_shape().sample_dim();
+    TensorListShape<> sh(batch_size, ndim);
+
+    // TODO: implement checking that all images have same dimensions
+    // also all general calculations for all images such as roi
```
klecki, line 55: In the initial case it might be ok to assume uniform batch but maybe there shouldn't be such constraint in general and we should support image of different sizes.

Also, you need to peek the shape of the sample and set it for all element of `sh` so the executor can correctly resize the buffers. Now it will probably have some random contents (we just resized the internal vector to `batch_size * ndim` elements), so this is probably a bug.

dali/operators/reader/fits_reader_op.cc
aderylo: Okay, I have a few questions then: Given that reader can take as an argument path to a directory with potentially multiple files, should the `hdu_indices` by a vector<vector<int>> for each file?

klecki: I think we should stick with the same vector of indices for all files that we are reading.

dali/operators/reader/fits_reader_op.h
```diff
@@ -0,0 +1,85 @@
...
+    return true;
+  }
+
+  USE_READER_OPERATOR_MEMBERS(Backend, Target);
+  using DataReader<Backend, Target>::GetCurrBatchSize;
+  using DataReader<Backend, Target>::GetSample;
+  using Operator<Backend>::spec_;
+
+  // rewrite this part given new approach with multiple outputs
+  // do i
```
klecki, line 44: There was a question here :)

dali/operators/reader/fits_reader_op.cc
```diff
@@ -0,0 +1,116 @@
...
+inline int FitsReaderOutputFn(const OpSpec &spec) {
+  return spec.HasArgument("hdu_indices") ?
+             spec.GetRepeatedArgument<std::string>("hdu_indices").size() : 1;
+}
+
+}  // namespace detail
+
+DALI_REGISTER_OPERATOR(experimental_readers__Fits, FitsReaderCPU, CPU);
+
+DALI_SCHEMA(experimental_readers__Fits)
```
klecki, line 36:
```suggestion
DALI_REGISTER_OPERATOR(experimental__readers__Fits, FitsReaderCPU, CPU);

DALI_SCHEMA(experimental__readers__Fits)
```
`__` - double underscore is needed.

dali/test/python/reader/test_fits.py
```diff
@@ -0,0 +1,160 @@
...
+    return pipe
+
+
+all_numpy_types = set([
+    np.bool_, np.byte, np.ubyte, np.short, np.ushort, np.intc, np.uintc, np.int_, np.uint,
+    np.longlong, np.ulonglong, np.half, np.float16, np.single, np.double, np.longdouble, np.csingle,
+    np.cdouble, np.clongdouble, np.int8, np.int16, np.int32, np.int64, np.uint8, np.uint16,
+    np.uint32, np.uint64, np.intp, np.uintp, np.float32, np.float64, np.float_, np.complex64,
+    np.complex128, np.complex_
+])
```
klecki, line 64: Let's limit ourselves to some subset of the image types supported by FITS:
u8, i16, i32, i64, u32, u64.

dali/test/python/reader/test_fits.py
```diff
@@ -0,0 +1,160 @@
...
+    pipe = Pipeline(batch_size=batch_size, num_threads=num_threads, device_id=device_id)
+    data = fn.readers.fits(device=device,
+                           file_list=file_list,
+                           files=files,
+                           file_root=path,
+                           file_filter=file_filter,
+                           shard_id=0,
+                           num_shards=1)
+    pipe.set_outputs(data)
+    return pipe
```
klecki, line 55: Two nitpicks, I recommend using `@pipeline_def` decorator and you need to add the `experimental` module:
```suggestion
@pipeline_def
def FitsReaderPipeline(path, device="cpu", file_list=None, files=None,
                       file_filter="*.fits", hdu_indices=None, dtype=None):
    data = fn.experimental.readers.fits(device=device,
                           file_list=file_list,
                           files=files,
                           file_root=path,
                           file_filter=file_filter,
                           shard_id=0,
                           num_shards=1)
   return data
```

This function would accept arguments for Pipeline constructor as kwargs, so you can still pass the `batch_size`, `num_threads` and `device_id`.

dali/test/python/reader/test_fits.py
```diff
@@ -0,0 +1,160 @@
...
+    hdu = fits.ImageHDU(imageData, imageHeader)
+    if compressed:
+        hdu = fits.CompImageHDU(imageData, imageHeader)
+
+    hdu.writeto(filename)
+
+
+def delete_fits_file(filename):
+    if os.path.isfile(filename):
+        os.remove(filename)
```
klecki, line 40: Unused, you could probably take a look into setting up a Unittest setUp and tearDown functions for this purpose.

klecki (commented):
Bunch of comments, but I think we are close to finishing with this.

dali/operators/reader/fits_reader_op.cc
```diff
@@ -0,0 +1,116 @@
...
+
+If ``file_root`` is provided, the paths are treated as being relative to it.
+
+This argument is mutually exclusive with ``file_list``.)",
+                                    nullptr)
+    .AddOptionalArg("hdu_indices",
+                                 R"(HDU indexes to read. If not provided, first HDU after primary 
+  will be yielded (i.e. will default to  hdu_indices=[2]).)",
+                                 std::vector<int>{2})
+    .AddOptionalArg("dtypes", R"code(Data types of the respective outputs.
```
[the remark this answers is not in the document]
klecki: Ok, we have previous example in webdataset where we use `dtypes` already, so the name should stay.
Maybe we can have similar behaviour: https://docs.nvidia.com/deeplearning/dali/user-guide/docs/operations/nvidia.dali.fn.readers.webdataset.html
Either one value applied to all HDUs that we read (by default uint8) otherwise a list of values.

dali/operators/reader/fits_reader_op.cc
aderylo: On it!

dali/operators/reader/loader/fits_loader.cc
```diff
@@ -0,0 +1,89 @@
...
+    }
+
+    int anynul = 0, nulval = 0;
+    Index nbytes = header.nbytes();
+
+    // reset, resize target
+    if (target.data[output_idx].shares_data()) {
+      target.data[output_idx].Reset();
+    }
+    target.data[output_idx].Resize(header.shape, header.type());
```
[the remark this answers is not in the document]
aderylo: Addressed

dali/util/fits.cc
```diff
@@ -0,0 +1,82 @@
...
+    return TypeTable::GetTypeInfo<uint64_t>();
+  if (imgtype == ULONG_IMG)
+    return TypeTable::GetTypeInfo<uint64_t>();
+  if (imgtype == ULONGLONG_IMG)
+    return TypeTable::GetTypeInfo<uint64_t>();
+  if (imgtype == FLOAT_IMG)
+    return TypeTable::GetTypeInfo<float>();
+  if (imgtype == DOUBLE_IMG)
+    return TypeTable::GetTypeInfo<double>();
+  DALI_FAIL("Unknown FITS image type type string");
```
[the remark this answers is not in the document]
aderylo: My bad, thanks for point out! Addressed in [8f512bc]

dali/util/fits.cc
```diff
@@ -0,0 +1,82 @@
...
+// limitations under the License.
+
+#include "dali/util/fits.h"
+#include <fitsio.h>
+#include <string>
+#include <vector>
+#include "dali/pipeline/data/types.h"
+#include "dali/pipeline/data/views.h"
+
+#define MAX_DIMS 999
```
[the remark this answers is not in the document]
aderylo: I looked up what actually goes on under the hood in `fits_get_img_param` and rewrote header parsing so it doesn't use any constants.

dali/util/fits.cc
```diff
@@ -46,24 +54,27 @@ const TypeInfo &TypeFromFitsImageType(int imgtype) {
...
 
   fits_get_hdu_type(src, &hdu_type, &status);
   bool is_image = (hdu_type == IMAGE_HDU);
   DALI_ENFORCE(is_image, "Only IMAGE_HDUs are supported!");
   parsed_header.hdu_type = hdu_type;
 
-  fits_get_img_param(src, MAX_DIMS, &img_type, &n_dims, dims, &status);
+  fits_get_img_type(src, &img_type, &status);    /* get BITPIX value */
+  fits_get_img_dim(src, &n_dims, &status);       /* get NAXIS value */
+  long dims[n_dims] = {0};                       /* create array for storing img dims*/
```
klecki, line 65: Isn't this a VLA? I don't think it will work in C++

dali/operators/reader/fits_reader_op.cc
```diff
@@ -0,0 +1,114 @@
...
+    .AddOptionalArg<vector<string>>("files", R"(A list of file paths to read the data from.
+
+If ``file_root`` is provided, the paths are treated as being relative to it.
+
+This argument is mutually exclusive with ``file_list``.)",
+                                    nullptr)
+    .AddOptionalArg("hdu_indices",
+                                 R"(HDU indexes to read. If not provided, first HDU after primary 
+  will be yielded (i.e. will default to  hdu_indices=[2]).)",
+                                 std::vector<int>{2})
```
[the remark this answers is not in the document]
klecki: Try the `GetRepeatedArgument` method.

klecki:
!build

klecki:
!build

dali/operators/reader/fits_reader_op.cc
```diff
@@ -0,0 +1,116 @@
...
+
+If ``file_root`` is provided, the paths are treated as being relative to it.
+
+This argument is mutually exclusive with ``file_list``.)",
+                                    nullptr)
+    .AddOptionalArg("hdu_indices",
+                    R"(HDU indexes to read. If not provided, first HDU after primary 
+                                 R"(HDU indexes to read. If not provided, first HDU after primary 
+                    R"(HDU indexes to read. If not provided, first HDU after primary 
+  will be yielded (i.e. will default to  hdu_indices=[2]).)",
```
klecki, line 81:
```suggestion
                    R"(HDU indexes to read. If not provided, first HDU after primary 
  will be yielded.)",
```
I think that the default is generated automatically in the docs. Looks like the line got duplicated.
You can maybe mention that the indexing starts from 1.

klecki:
The problems in CI:
```
ERROR: test_fits (nose2.loader.ModuleImportFailure)
ImportError: Failed to import test module: test_fits
  File "/usr/local/lib/python3.8/dist-packages/nose2/plugins/loader/discovery.py", line 204, in _find_tests_in_file
    module = util.module_from_name(module_name)
  File "/usr/local/lib/python3.8/dist-packages/nose2/util.py", line 76, in module_from_name
    __import__(name)
  File "/opt/dali/dali/test/python/reader/test_fits.py", line 18, in <module>
    from astropy.io import fits
ModuleNotFoundError: No module named 'astropy'
```
`qa/TL0_python-self-test-readers-decoders/test_nofw.sh` <- needs astropy added to the list of `pip_packages`.


```
FAIL: test_dali_variable_batch_size.test_coverage
AssertionError: Test doesn't cover:
 {'experimental.readers.fits'}
```
The readers are not working with variable batch, add the fits reader to `excluded_methods` in `dali/test/python/test_dali_variable_batch_size.py`


```
FAIL: test_dali_cpu_only.test_coverage
AssertionError: Test doesn't cover:
 {'experimental.readers.fits'}
 ```

Use `check_no_input` to implement a simple test with the reader in `dali/test/python/test_dali_cpu_only.py` and add it to the `tested_methods`.

File: dali/test/python/test_dali_cpu_only.py [hunk 25]
```diff
@@ -1379,6 +1379,7 @@ def explicit_conditional_ops_pipeline():
     "experimental.filter",  # not supported for CPU
     "experimental.inflate",  # not supported for CPU
     "experimental.remap",  # operator is GPU-only
+    "experimental.readers.fits", # lacking test files in DALI_EXTRA
```
klecki, line 1382:
```suggestion
    "experimental.readers.fits",  # lacking test files in DALI_EXTRA 
```

klecki:
!build

dali/operators/reader/fits_reader_op.cc
```diff
@@ -0,0 +1,114 @@
...
+    .AddOptionalArg(
+        "file_filter",
+        R"(If a value is specified, the string is interpreted as glob string to filter the
+list of files in the sub-directories of the ``file_root``.
+
+This argument is ignored when file paths are taken from ``file_list`` or ``files``.)",
+        "*.fits")
+    .AddOptionalArg<string>("file_list",
+                            R"(Path to a text file that contains filenames (one per line)
+where the filenames are relative to the location of that file or to ``file_root``, if specified.
```
stiepan, line 62: nitpick. I'd split that sentence in two for clarity.

```suggestion
                            R"(Path to a text file that contains filenames (one per line).
The filenames are relative to the location of that file or to ``file_root``, if specified.
```

dali/operators/reader/fits_reader_op.cc
```diff
@@ -0,0 +1,114 @@
...
+                    false)
+    .AddOptionalArg<vector<string>>("files", R"(A list of file paths to read the data from.
+
+If ``file_root`` is provided, the paths are treated as being relative to it.
+
+This argument is mutually exclusive with ``file_list``.)",
+                                    nullptr)
+    .AddOptionalArg("hdu_indices",
+                    R"(HDU indices to read. If not provided, first HDU after primary 
+  will be yielded. Since, HDUs are indexed starting from 1, the default value is as follows: hdu_indices = [2].)",
```
stiepan, line 79:
```suggestion
                    R"(HDU indices to read. If not provided, the first HDU after the primary 
  will be yielded. Since HDUs are indexed starting from 1, the default value is as follows: hdu_indices = [2].)",
```

dali/operators/reader/fits_reader_op.cc
```diff
@@ -0,0 +1,114 @@
...
+
+``stick_to_shard`` and ``random_shuffle`` cannot be used when this argument is set to True.)",
+                    false)
+    .AddOptionalArg<vector<string>>("files", R"(A list of file paths to read the data from.
+
+If ``file_root`` is provided, the paths are treated as being relative to it.
+
+This argument is mutually exclusive with ``file_list``.)",
+                                    nullptr)
+    .AddOptionalArg("hdu_indices",
```
stiepan, line 77: Did I miss it or the fact that the number of outputs is `spec.GetRepeatedArgument<int>("hdu_indices").size();` is not documented in the spec?

@klecki, I guess we could mention that both in the top section of the docs and here.

dali/operators/reader/fits_reader_op.cc
```diff
@@ -0,0 +1,114 @@
...
+                    R"(HDU indices to read. If not provided, first HDU after primary 
+  will be yielded. Since, HDUs are indexed starting from 1, the default value is as follows: hdu_indices = [2].)",
+                    std::vector<int>{2})
+    .AddOptionalArg("dtypes", R"code(Data types of the respective outputs.
+
+The default output data types are UINT8. However, if set, each output data type should be specified.")code",
+                    DALI_DATA_TYPE_VEC,
```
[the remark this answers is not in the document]
stiepan: @klecki, do you know if we can easily broadcast the input here, like `dtypes=type.UINT8`?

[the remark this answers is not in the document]
klecki: I am not sure if it allows to have a scalar default if we specify it, one element vector would work for certain. From the Python API I think it will now accept both scalars and vectors.

dali/test/python/reader/test_fits.py
```diff
@@ -0,0 +1,150 @@
...
+        # setup file
+        filenames = ["test_{:02d}.fits".format(i) for i in range(nsamples)]
+        full_paths = [os.path.join(test_data_root, fname) for fname in filenames]
+        for i in range(nsamples):
+            compressed = compressed_arg
+            if compressed is None:
+                compressed = random.choice([False, True])
+            create_fits_file(full_paths[i], shapes[i], type, compressed)
+
+        # laod manually, we skip primary hdu since they don't have any data
```
stiepan, line 85:
```suggestion
        # load manually, we skip primary hdu since they don't have any data
```

dali/operators/reader/fits_reader_op.cc
klecki: Yup, I think that we can also expand the docs, but that can be next step.

dali/util/fits.cc
```diff
@@ -0,0 +1,127 @@
...
+#include <vector>
+#include "dali/pipeline/data/types.h"
+#include "dali/pipeline/data/views.h"
+
+namespace dali {
+namespace fits {
+
+void HandleFitsError(int status) {
+  std::string status_str;
+  status_str.reserve(FLEN_STATUS);
```
stiepan, line 27: I think formally it should be `resize`, as `reserve` does not change the `size` of the string (see https://en.cppreference.com/w/cpp/string/basic_string/operator_at does not guarantee anything with respect to capacity).

stiepan: Also you could do the resize inside the `if`.

dali/util/fits.cc
```diff
@@ -0,0 +1,127 @@
...
+  parsed_header.datatype_code = ImgTypeToDatatypeCode(img_type);
+  parsed_header.type_info = &TypeFromFitsDatatypeCode(parsed_header.datatype_code);
+  parsed_header.compressed = (fits_is_compressed_image(src, &status) == 1);
+
+  for (auto it = dims.rbegin(); it != dims.rend(); ++it) {
+    parsed_header.shape.shape.push_back(static_cast<int64_t>(*it));
+  }
+
+  if (status)
+    HandleFitsError(status);
```
stiepan, line 111: nitpick: I guess you may just call the `HandleFitsError` and rely on the if inside `HandleFitsError`.

dali/operators/reader/loader/fits_loader.cc
```diff
@@ -0,0 +1,93 @@
...
+
+    // copy the image
+    fits_read_img(current_file, header.datatype_code, 1, nelem, &nulval,
+                  static_cast<uint8_t*>(target.data[output_idx].raw_mutable_data()), &anynul,
+                  &status);
+    DALI_ENFORCE(status == 0, make_string("Failed to read a file: ", filename));
+
+
+    // close the file handle
+    fits_close_file(current_file, &status);
```
stiepan, line 83: +1 to Krzysztof's comment about the `UniqueHandle`. This looks like a handle's leak on unsuccessful `fits_read_img`.

dali/operators/reader/loader/fits_loader.cc
```diff
@@ -0,0 +1,93 @@
...
+    if (target.data[output_idx].shares_data()) {
+      target.data[output_idx].Reset();
+    }
+    target.data[output_idx].Resize(header.shape, header.type());
+
+    // copy the image
+    fits_read_img(current_file, header.datatype_code, 1, nelem, &nulval,
+                  static_cast<uint8_t*>(target.data[output_idx].raw_mutable_data()), &anynul,
+                  &status);
+    DALI_ENFORCE(status == 0, make_string("Failed to read a file: ", filename));
```
stiepan, line 79: For consideration:

I wonder if it's not more user-friendly if we forward a message from the fits itself here. 
```
make_string("Failed to read a file: ", filename, ". ", GetFitsErrorMessage(status))
```

where `GetFitsErrorMessage` is like `HandleFitsError` but simply returns the string. (You could then use it inside `HandleFitsError` too.)

dali/operators/reader/loader/fits_loader.cc
```diff
@@ -0,0 +1,93 @@
...
+    fits_movabs_hdu(current_file, hdu_indices_[output_idx], NULL, &status);
+    DALI_ENFORCE(status == 0, make_string("Issue moving to hdu with an index: ",
+                                          hdu_indices_[output_idx], ", in a file: ", filename));
+
+    // read the header
+    fits::HeaderData header;
+    try {
+      fits::ParseHeader(header, current_file);
+    } catch (const std::runtime_error& e) {
+      DALI_FAIL(e.what() + ". File: " + filename);
```
stiepan, line 63: 👍

dali/operators/reader/fits_reader_op.h
```diff
@@ -0,0 +1,86 @@
...
+    for (int output_idx = 0; output_idx < num_outputs; output_idx++) {
+      int ndim = sample_0.data[output_idx].shape().sample_dim();
+      output_desc[output_idx].shape = TensorListShape<>(num_samples, ndim);
+    }
+
+    for (int sample_idx = 0; sample_idx < num_samples; sample_idx++) {
+      auto& sample = GetSample(sample_idx);
+      for (int output_idx = 0; output_idx < num_outputs; output_idx++) {
+        output_desc[output_idx].shape.set_tensor_shape(sample_idx, sample.data[output_idx].shape());
+        output_desc[output_idx].type = sample.data[output_idx].type();
```
stiepan, line 61: Do you (or one of the base classes?) validate the `dtypes`  against what we actually get in the header?

dali/operators/reader/fits_reader_op.h
```diff
@@ -0,0 +1,86 @@
...
+    // If necessary start prefetching thread and wait for a consumable batch
+    DataReader<Backend, Target>::SetupImpl(output_desc, ws);
+    int num_outputs = ws.NumOutput();
+    int num_samples = GetCurrBatchSize();  // samples here are synonymous with files
+
+    output_desc.resize(num_outputs);
+    const auto& sample_0 = GetSample(0);
+
+    for (int output_idx = 0; output_idx < num_outputs; output_idx++) {
+      int ndim = sample_0.data[output_idx].shape().sample_dim();
```
stiepan, line 53: What happens if the ndim's in the actual headers differ between files?

dali/test/python/reader/test_fits.py
stiepan, line 1: I wonder how the reader acts if
1. some dtypes in the header don't match the declarated dtypes
2. ndims differ between files
3*. there are no files to read (due to empty dir, empty list file, restritive glob

I belive the third point should be handled by the base class, so it is nice to have but for the first two I think we should have tests that assert we fail in orderly manner (take a look at `assert_raises`).

dali/operators/reader/fits_reader_op.cc
aderylo: Ok, all suggestions regarding spec were applied.

dali/util/fits.cc
```diff
@@ -0,0 +1,135 @@
...
+  int32_t hdu_type, img_type, n_dims, status = 0;
+
+  fits_get_hdu_type(src, &hdu_type, &status);
+  bool is_image = (hdu_type == IMAGE_HDU);
+  DALI_ENFORCE(is_image, "Only IMAGE_HDUs are supported!");
+
+  fits_get_img_equivtype(src, &img_type, &status); /* get IMG_TYPE code value */
+  fits_get_img_dim(src, &n_dims, &status);         /* get NAXIS value */
+
+  DALI_ENFORCE(n_dims > 0, "NAXIS (image dimensions) value for each HDU has to be greater than 0!");
```
stiepan, line 105: I think we still need an extra `HandleFitsError` check here (or `n_dims=0`) in the definition. Otherwise, if `fits_get_img_dim` fails, we DALI_ENFORCE with something unitilizied. It could be positive but still garabge.

aderylo: Yup, although I went ahead and simply wrapped everything in FITS_CALL.

stiepan: Cool!

dali/operators/reader/fits_reader_op.h
aderylo: No idea. In any case, I've added check for consistency of dimensions and types  between files in corresponding outputs.

dali/operators/reader/loader/fits_loader.cc
aderylo: Done!

klecki:
!build

dali/operators/reader/fits_reader_op.cc
```diff
@@ -0,0 +1,117 @@
...
+DALI_SCHEMA(experimental__readers__Fits)
+    .DocStr(R"(Reads Fits image HDUs from a directory.
+
+This operator can be used in the following modes:
+
+1. Read all files from a directory indicated by ``file_root`` that match given ``file_filter``.
+2. Read file names from a text file indicated in ``file_list`` argument.
+3. Read files listed in ``files`` argument.
+4. Number of outputs per sample corresponds to the length of ``hdu_indices`` argument. By default,
+first HDU with data is read from each file, so the number of outputs defaults to 1.
```
stiepan, line 45: Nitpick, can be done as a follow-up

```suggestion

The number of outputs per sample corresponds to the length of ``hdu_indices`` argument. By default,
first HDU with data is read from each file, so the number of outputs defaults to 1. 
```

dali/util/fits.cc
```diff
@@ -0,0 +1,136 @@
...
+  std::vector<int64_t> dims(n_dims, 0); /* create vector for storing img dims*/
+
+  FITS_CALL(fits_get_img_size(src, n_dims, &dims[0], &status)); /* get NAXISn values */
+
+  parsed_header.hdu_type = hdu_type;
+  parsed_header.datatype_code = ImgTypeToDatatypeCode(img_type);
+  parsed_header.type_info = &TypeFromFitsDatatypeCode(parsed_header.datatype_code);
+  parsed_header.compressed = (fits_is_compressed_image(src, &status) == 1);
```
stiepan, line 116: We don't validate this particular status ever, do we?

dali/test/python/reader/test_fits.py
stiepan: Nice! Nontheless, it would be good to have a simple test cases that validate it is really the case. For now and future changes.

stiepan:
It looks good, it seems that it works, great job!

As a follow-up, please keep in mind: 1. a few test cases on incorrect inputs would be helpful, 2. enforcing the dtypes if we accept them 3. there is one status of fits call that is not validated (it seems like a special case when the status is not returned as an output; if the call cannot fail according to fits docs <I did not check that> then please ignore that remark).

stiepan (approved):

PR #4598, opened 2023-01-20
Enable Python 3.10 tests in CI

JanuszL:
- updates test dependencies to enable Python 3.10 tests in CI
- adds wrapper for nose to overcome problem with dependency on
  collections vs collections.abc
- the inclusion of the code that fixes the problem inside the test itself
  doesn't help as when the test attributes are used the collections
  are accessed before the test file is loaded

## Category:

## Description:
- updates test dependencies to enable Python 3.10 tests in CI
- adds wrapper for nose to overcome problem with dependency on
  collections vs collections.abc
- the inclusion of the code that fixes the problem inside the test itself
  doesn't help as when the test attributes are used the collections
  are accessed before the test file is loaded

- qa/setup_packages.py
- tests

- NA

### Tests:

- [x] Existing tests apply
  - all python tests
- [ ] New tests added
  - [ ] Python tests
  - [ ] GTests
  - [ ] Benchmark
  - [ ] Other
- [ ] N/A

### Documentation
- [ ] Documentation updated
  - [ ] Docstring
  - [ ] Doxygen
  - [ ] RST
  - [ ] Jupyter
  - [ ] Other
- [x] N/A

#### Requirements
- [x] N/A

**REQ IDs**: N/A

**JIRA TASK**: N/A


docs/examples/use_cases/video_superres/model/model.py
```diff
@@ -177,7 +177,7 @@ def forward(self, inputs, iteration, writer, im_out=False):
             y[:, :, fr, :, :] = self.upsample(downsampled_input)
 
             if fr == self.mi:
-                conv1_out.append(self.conv1(y[:, :, self.mi, :, :]))
+                conv1_out.append(self.conv1(y[:, :, self.mi, :, :].clone()))
```
JanuszL, line 180: Newer PyTorch version complains that this is in place operation and it cannot run autograd.

qa/TL1_ssd_training/test.sh
```diff
@@ -10,7 +10,7 @@ test_body() {
     git clone https://github.com/nvidia/apex
     pushd apex
     # get the lastest stable and tested APEX version
-    git checkout 4ef930c1c884fdca5f472ab2ce7cb9b505d26c1a
+    git checkout ad68a2fd8519d2e2fd565966a14a4f68f78ebc71
```
JanuszL, line 13: Older apex doesn't support newer PyTorch.

qa/TL1_ssd_training/test.sh
```diff
@@ -21,8 +21,10 @@ test_body() {
     pip wheel -v --no-cache-dir --global-option="--cpp_ext" --global-option="--cuda_ext" .
     # for some reason the `pip install -v --no-cache-dir --global-option="--cpp_ext" --global-option="--cuda_ext" .` doesn't install
     # wheel but only builds the binaries with the setuptools>=50.0, so now we builds wheel explicitly and then install it
-    pip install apex*.whl
     unset CUDA_HOME
+    # escape frome APEX dir or it pip will tell that is already installed
```
szalpal, line 25:
```suggestion
    # We need to escape from APEX dir. Otherwise pip will tell that it is already installed.
```

qa/TL1_jupyter_conda/test_cupy.sh
```diff
@@ -1,4 +1,4 @@
-#!/bin/bash -e
+[#!/bin/bash -e
```
szalpal, line 1: Out of curiosity, why the `[]` brackets around this test?

JanuszL: Not intended

qa/TL1_ssd_training/test.sh
JanuszL: Fixed

qa/TL0_python-self-test_xavier/test_nofw.sh
```diff
@@ -1,7 +1,6 @@
 #!/bin/bash -e
 # used pip packages
-# due to https://github.com/numpy/numpy/issues/18131 we cannot use 1.19.5
-pip_packages='${python_test_runner_package} dataclasses numpy>=1.17,<=1.19.4 opencv-python pillow psutil'
+pip_packages='${python_test_runner_package} dataclasses numpy>=1.17 opencv-python pillow psutil'
```
stiepan, line 3: Regarding the comment and previous limitations, shouldn't it be `numpy>=1.20`?

JanuszL: Good questions. For sure 1.19.5 didn't work, but I don't know if 1.20 does. All I know it works now whatever version is available there. So I would prefer not to add `>1.20` if we are not sure if that is the case.

qa/setup_packages.py
```diff
@@ -439,50 +463,46 @@ def get_pyvers_name(self, url, cuda_version):
...
+                            {"113": [PckgVer("9.6.0", python_max_ver="3.6",
+                                             dependencies=["numpy<1.24"]),
+                                     PckgVer("10.0.0", python_min_ver="3.7",
+                                             dependencies=["numpy<1.24"])]},
                             "cupy-cuda{cuda_v}"),
                 CudaPackage("mxnet",
-                            {"100": [PckgVer("1.9.0", dependencies=["numpy<1.24"])],
-                             "110": [PckgVer("1.9.1", dependencies=["numpy<1.24"])],
-                             "111": [PckgVer("1.9.1", dependencies=["numpy<1.24"])]},
+                            {"113": [PckgVer("1.9.1", dependencies=["numpy<1.24"])]},
```
stiepan, line 474: What is special about 113? Is this some special build with that cuda version being the highest supported?

qa/setup_packages.py
```diff
@@ -572,6 +595,14 @@ def get_extra_indices(idx, packages, cuda_version):
     return " ".join(ret)
 
 
+def get_links_indices(idx, packages, cuda_version):
```
stiepan, line 598: The `idx` arg is not used, is it?

qa/setup_packages.py
JanuszL: In case of PyTorch it seems to be the only version (with that cuda) that supports Python 3.6 and 3.10 at the same time, so we can have only one CUDA version in the test image. For other CUDA version there is either no variant with 3.6 or 3.10.

qa/TL0_python-self-test_xavier/test_nofw.sh
stiepan: Up to you. My bet was for 1.20 because

https://github.com/numpy/numpy/issues/18131 is marked as fixed in https://github.com/numpy/numpy/pull/18196. The latter commit is merged in [1.20.0](https://github.com/numpy/numpy/releases/tag/v1.20.0) and there are no releases between 1.20.0 and 1.19.5.

qa/setup_packages.py
stiepan: Oh I see, thanks.

stiepan (approved):

qa/TL0_python-self-test_xavier/test_nofw.sh
JanuszL: Done

szalpal (approved):

PR #4599, opened 2023-01-23
Make daliPipelineHandle a pointer to an opaque C++ structure.

mzient:

## Category:

## Description:
DALIPipelineHandle used to be a structure that contained some type-erased pointers. This leaked implementation details (the structures accompanyting the pipeline) to user code and also caused binary incompatibility whenever we changed this structure.
Also, confusiingly, the "handle" wasn't a handle (opaque pointer or a number) but a structure.
This change makes daliPipelineHandle a pointer to a forward-declared struct.
The structure is now defined in C++ and doesn't need to employ type erasure. This improves the separation of the API and the implementation and makes the implementation cleaner and safer.

For backward compatibility, the APIs still take the handle by pointer (effectively a double pointer) - this will be changed in the future, but is out of scope of this PR.

C API

### Tests:

C API tests
TF plugin tests
- [X] Existing tests apply
- [ ] New tests added
  - [ ] Python tests
  - [ ] GTests
  - [ ] Benchmark
  - [ ] Other
- [ ] N/A

### Documentation
- [ ] Documentation updated
  - [ ] Docstring
  - [ ] Doxygen
  - [ ] RST
  - [ ] Jupyter
  - [ ] Other
- [X] N/A

#### Requirements
- [X] N/A

**REQ IDs**: N/A

**JIRA TASK**: N/A

### DIFF
--- a/dali/c_api/c_api.cc
+++ b/dali/c_api/c_api.cc
@@ -37,7 +37,23 @@ using dali::AccessOrder;
 using dali::CPUBackend;
 using dali::GPUBackend;
 
+/**
+ * Maps operator name to the batch size set prior to daliSetExternal... call.
+ *
+ * Typically, this operator will be BatchSizeProvider.
+ * Negative values denote max batch size (default state).
+ * Typical usage:
+ * auto *batch_size_map = reinterpret_cast<batch_size_map_t *>(handle->batch_size_map);
+ */
 using batch_size_map_t = std::unordered_map<std::string /* op_name */, int /* batch_size */>;
+
+/**
+ * Maps operator name to the data_id set prior to daliSetExternal... call.
+ *
+ * Usually the operator with given `op_name` will be an InputOperator.
+ * Typical usage:
+ * auto *data_id_map = reinterpret_cast<data_id_map_t *>(handle->data_id_map);
+ */
 using data_id_map_t = std::unordered_map<std::string /* op_name */, std::string /* data_id */>;
 
 /**
@@ -183,6 +199,17 @@ inline dali::mm::memory_kind_id GetMemKind(device_type_t device_type, bool is_pi
         : (is_pinned ? dali::mm::memory_kind_id::pinned : dali::mm::memory_kind_id::host);
 }
 
+inline std::unique_ptr<DALIPipeline> WrapPipeline(std::unique_ptr<dali::Pipeline> pipeline) {
+  auto pipe_wrap = std::make_unique<DALIPipeline>();
+
+  if (pipeline->device_id() >= 0) {
+    pipe_wrap->copy_stream = dali::CUDAStreamPool::instance().Get(pipeline->device_id());
+  }
+
+  pipe_wrap->pipeline = std::move(pipeline);
+  return pipe_wrap;
+}
+
 }  // namespace
 
 
@@ -206,17 +233,6 @@ void daliCreatePipeline(daliPipelineHandle *pipe_handle, const char *serialized_
   daliCreatePipeline2(pipe_handle, serialized_pipeline, length, max_batch_size, num_threads,
                       device_id, 1, 1, separated_execution, prefetch_queue_depth,
                       cpu_prefetch_queue_depth, gpu_prefetch_queue_depth, enable_memory_stats);
-}
-
-inline std::unique_ptr<DALIPipeline> WrapPipeline(std::unique_ptr<dali::Pipeline> pipeline) {
-  auto pipe_wrap = std::make_unique<DALIPipeline>();
-
-  if (pipeline->device_id() >= 0) {
-    pipe_wrap->copy_stream = dali::CUDAStreamPool::instance().Get(pipeline->device_id());
-  }
-
-  pipe_wrap->pipeline = std::move(pipeline);
-  return pipe_wrap;
 }
 
 DLL_PUBLIC void
@@ -246,11 +262,6 @@ void daliDeserializeDefault(daliPipelineHandle *pipe_handle, const char *seriali
                             int length) {
   auto pipeline = std::make_unique<dali::Pipeline>(std::string(serialized_pipeline, length));
   pipeline->Build();
-  dali::CUDAStreamLease stream;
-  if (pipeline->device_id() >= 0) {
-    stream = dali::CUDAStreamPool::instance().Get(pipeline->device_id());
-  }
-
   *pipe_handle = WrapPipeline(std::move(pipeline)).release();
 }
 
--- a/dali/c_api/c_api.cc
+++ b/dali/c_api/c_api.cc
@@ -67,6 +67,13 @@ struct DALIPipeline {
   dali::CUDAStreamLease copy_stream;
 };
 
+/** Temporary workaround for backward compatibility
+ *
+ * In the long run, daliPipeline_t is going to be change to daliPipeline and is going to be
+ * passed by value.
+ */
+typedef daliPipelineHandle *daliPipelineHandle_t;
+
 namespace {
 
 bool dali_initialized = false;
--- a/dali/c_api/c_api.cc
+++ b/dali/c_api/c_api.cc
@@ -201,7 +201,7 @@ void daliInitialize() {
 }
 
 
-void daliCreatePipeline(daliPipelineHandle_t *pipe_handle, const char *serialized_pipeline,
+void daliCreatePipeline(daliPipelineHandle *pipe_handle, const char *serialized_pipeline,
                         int length, int max_batch_size, int num_threads, int device_id,
                         int separated_execution, int prefetch_queue_depth,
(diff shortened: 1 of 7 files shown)


dali/c_api/c_api.cc
```diff
@@ -37,27 +37,23 @@ using dali::AccessOrder;
...
-
-/**
- * Maps operator name to the batch size set prior to daliSetExternal... call.
- * Typically, this operator will be BatchSizeProvider.
- * Negative values denote max batch size (default state).
- * Typical usage:
- * auto *batch_size_map = reinterpret_cast<batch_size_map_t *>(handle->batch_size_map);
- */
 using batch_size_map_t = std::unordered_map<std::string /* op_name */, int /* batch_size */>;
+using data_id_map_t = std::unordered_map<std::string /* op_name */, std::string /* data_id */>;
```
szalpal, line 41: Why removing documentation of these structures?

dali/c_api/c_api.cc
```diff
@@ -37,27 +37,23 @@ using dali::AccessOrder;
...
- * Maps operator name to the data_id set prior to daliSetExternal... call.
- * Usually the operator with given `op_name` will be an InputOperator.
- * Typical usage:
- * auto *data_id_map = reinterpret_cast<data_id_map_t *>(handle->data_id_map);
+ * @brief Aggregates a DALI pipeline and auxiliary objects
  */
-using data_id_map_t = std::unordered_map<std::string /* op_name */, std::string /* data_id */>;
+struct DALIPipeline {
+  std::unique_ptr<dali::Pipeline> pipeline;
+  dali::Workspace workspace;
```
szalpal, line 48: Please follow our naming pattern `DaliPipeline`:
https://google.github.io/styleguide/cppguide.html#General_Naming_Rules

> For the purposes of the naming rules below, a "word" is anything that you would write in English without internal spaces. This includes abbreviations, such as acronyms and initialisms. For names written in mixed case (also sometimes referred to as "[camel case](https://en.wikipedia.org/wiki/Camel_case)" or "[Pascal case](https://en.wiktionary.org/wiki/Pascal_case)"), in which the first letter of each word is capitalized, prefer to capitalize abbreviations as single words, e.g., StartRpc() rather than StartRPC().


Additionally, how about not naming it `Pipeline`? Inside this structure there's more than the Pipeline - Workspace, weird maps, something that wraps CUDA stream... Maybe `DaliInstance` would be more suitable?

dali/c_api/c_api.cc
```diff
@@ -212,6 +208,16 @@ void daliCreatePipeline(daliPipelineHandle *pipe_handle, const char *serialized_
                       cpu_prefetch_queue_depth, gpu_prefetch_queue_depth, enable_memory_stats);
 }
 
+inline std::unique_ptr<DALIPipeline> WrapPipeline(std::unique_ptr<dali::Pipeline> pipeline) {
+  auto pipe_wrap = std::make_unique<DALIPipeline>();
```
szalpal, line 212: How about making this `static`? Or a part of anonymous namespace above?

dali/c_api/c_api.cc
```diff
@@ -255,14 +250,8 @@ void daliDeserializeDefault(daliPipelineHandle *pipe_handle, const char *seriali
...
-  auto ws = std::make_unique<dali::Workspace>();
-  auto bs_map = std::make_unique<batch_size_map_t>();
-  auto di_map = std::make_unique<data_id_map_t>();
-  pipe_handle->ws = ws.release();
-  pipe_handle->copy_stream = stream.release().release();
-  pipe_handle->pipe = pipeline.release();
-  pipe_handle->batch_size_map = bs_map.release();
-  pipe_handle->data_id_map = di_map.release();
+
+  *pipe_handle = WrapPipeline(std::move(pipeline)).release();
```
szalpal, line 254: Shouldn't the 3 lines above be deleted?

include/dali/c_api.h
```diff
@@ -30,13 +30,11 @@ extern "C" {
...
-  void *ws;
-  void *batch_size_map;      /// @see batch_size_map_t
-  cudaStream_t copy_stream;  /// Stream to perform copy operations on
-  void *data_id_map;         /// @see data_id_map_t
-} daliPipelineHandle;
+typedef struct DALIPipeline *daliPipelineHandle;
+
+// Temporary workaround for backward compatibility
+// - in the long run, daliPipeline_t is going to be change to daliPipeline and passed by value
+typedef daliPipelineHandle *daliPipelineHandle_t;
```
szalpal, line 37: Could you add explicitly in the comment, which structure should be used? `daliPipelineHandle` or `daliPipelineHandle_t`? It's not clear for me here.

dali/c_api/c_api.cc
mzient: Isn't it all equivalent?

dali/c_api/c_api.cc
mzient: Perhaps it was added recently and I removed it during rebasing?

dali/c_api/c_api.cc
mzient: In general: local conventions always prevail over global ones.
`[^A-Za-z0-9]Dali[A-Za-z0-9]+` - 149 occurrences
`[^A-Za-z0-9]DALI[A-Za-z0-9]+` - 1566 occurences

dali/c_api/c_api.cc
mzient: Even 4.

dali/c_api/c_api.cc
mzient: I forgot the most important thing - this name is actually public and appears in the C API header as:
`typedef struct DALIPipeline *daliPipelineHandle_t`.
That's why it starts with DALI - like other structs/enums/... we use in C API. And that's why it's called a Pipeline - because that's what we call the handle.
In the implementation I could use a local typedef to make _our_ code cleaner, but wouldn't that be an overkill?

dali/c_api/c_api.cc
mzient: Restored.

include/dali/c_api.h
mzient: I'll just remove it or make it private (in the CC file). It will only confuse the users and won't make the transition to the future API any easier.

dali/c_api/c_api.cc
mzient: Moved.

stiepan:
Nitpick to the description:

`whereas all other APIs take the pointer by value` - they don't, right? It is a pointer to a pointer everywhere, so that the code that uses the API still compiles?

dali/c_api/c_api.cc
```diff
@@ -52,12 +49,34 @@ using batch_size_map_t = std::unordered_map<std::string /* op_name */, int /* ba
...
+  std::unique_ptr<dali::Pipeline> pipeline;
+  dali::Workspace workspace;
+  batch_size_map_t batch_size_map;
+  data_id_map_t data_id_map;
+  dali::CUDAStreamLease copy_stream;
+};
+
+/** Temporary workaround for backward compatibility
+ *
+ * In the long run, daliPipeline_t is going to be change to daliPipeline and is going to be
```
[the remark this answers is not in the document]
stiepan: or replaced with?

stiepan (approved):
Please adjust the description, to spare confusion for future readers.

dali/c_api/c_api.cc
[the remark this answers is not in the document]
mzient: Fixed.

mzient:
> Please adjust the description, to spare confusion for future readers.

Done.

dali/c_api/c_api.cc
szalpal: Still, there is a clear and concise rule in our style guide, how we should format acronyms. I'm aware, that it's not consistent in our codebase, but since the name is public, I'd vote for keeping the rule - we're publicly stating that it's our rule after all.

Should you believe that we should break this rule, please elevate this issue to the team meeting, like we tend to do in such cases.

JanuszL: I agree that `DALI*` appears way more often in the search query. Still, this is not something we use in this file, and introducing it here breaks this file's consistency as well.
If we really want to follow the convention we keep in this file it should be rather `daliPipeline`.

JanuszL: It is not about bikeshedding but about understanding the other's point of view and ideas that don't deserve to be bashed.
In most cases, people are fine accepting the suggestions from the reviewers if it is about my opinion against yours (and that is the case I believe). This is approach is encouraging the reviewers to add any comment they feel can improve the code. The other way around is very discouraging and long term affects the code quality.

szalpal (approved):

PR #4617, opened 2023-01-26
Support inferring batch size from tensor argument inputs

stiepan:

## Category:s

## Description:
This PR updates the way executor sets requested batch size. If there are any inputs available, executor assumes the operator expected uniform batch sizes across inputs and outputs and sets requested batch size to any tensor input it can find: either positional or named (argument input). Otherwise, if there are none avialable, the stage queue batch size is used.

## Additional information:

The common patern introduced with the end2end variable batch size was to look at the regular inputs or, if there were none, resort to `GetRequestedBatchSize`. This PR moves that behaviour to the executor and includes named arguments in the search of batch before resorting to stage batch size. @klecki  updated the autograph layer to reflect that by appropriate "hoisting" of the splits here https://github.com/NVIDIA/DALI/pull/4618

The regular usages of the operators should not see any difference.

### Tests:

- [ ] Existing tests apply
- [x] New tests added
  - [x] Python tests
  - [ ] GTests
  - [ ] Benchmark
  - [ ] Other
- [ ] N/A

### Documentation
- [ ] Documentation updated
  - [ ] Docstring
  - [ ] Doxygen
  - [ ] RST
  - [ ] Jupyter
  - [ ] Other
- [ ] N/A

#### Requirements
- [x] N/A

**REQ IDs**: N/A

**JIRA TASK**: N/A

### DIFF
--- a/dali/test/python/conditionals/test_pipeline_conditionals.py
+++ b/dali/test/python/conditionals/test_pipeline_conditionals.py
@@ -617,3 +617,135 @@ def test_uninitialized():
                                 " (both `if` branches). The `else` branch must also have a return"
                                 " statement.")):
         one_return()
+
+
+def _tensor_arg_permute_batch_params():
+    batch_sizes = [1, 5, 8]
+    inp0 = [[np.full((2, 2), i, dtype=np.float32) for i in range(batch_size)]
+            for batch_size in batch_sizes]
+    mask_batches = [
+        np.array([i % 2 for i in range(batch_size)], dtype=bool) for batch_size in batch_sizes
+    ]
+    kwarg_batches = [np.array([pred for pred in mask], dtype=np.int32) for mask in mask_batches]
+    return (inp0, ), mask_batches, {'indices': kwarg_batches}
+
+
+def _tensor_arg_transform_per_dim_params(arg_name):
+
+    def inner():
+        batch_sizes = [5, 1, 2, 8]
+        mask_batches = [
+            np.array([i % 2 for i in range(batch_size)], dtype=bool) for batch_size in batch_sizes
+        ]
+        kwarg_batches = [
+            np.array([[pred, pred] for pred in mask], dtype=np.float32) for mask in mask_batches
+        ]
+        return tuple(), mask_batches, {arg_name: kwarg_batches}
+
+    return inner
+
+
+def _tensor_arg_rotate_params():
+    batch_sizes = [3, 1, 2, 4]
+    mask_batches = [
+        np.array([i % 2 for i in range(batch_size)], dtype=bool) for batch_size in batch_sizes
+    ]
+    kwarg_batches = [
+        np.array([10 + 45 * pred for pred in mask], dtype=np.float32) for mask in mask_batches
+    ]
+    return tuple(), mask_batches, {'angle': kwarg_batches}
+
+
+def _tensor_arg_roi_random_crop_params():
+    batch_sizes = [1, 2, 7, 3]
+    crop_shape = [[
+        np.array([100 * i + 50, 200 * i + 50, 3], dtype=np.int32) for i in range(batch_size)
+    ] for batch_size in batch_sizes]
+    roi_start = [[
+        np.array([sample[0] // 2, sample[1] // 2, sample[2]], dtype=np.int32) for sample in batch
+    ] for batch in crop_shape]
+    mask_batches = [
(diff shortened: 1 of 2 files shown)


dali/operators/python_function/dltensor_function.h
```diff
@@ -245,30 +245,6 @@ class DLTensorPythonFunctionImpl : public Operator<Backend> {
   bool synchronize_stream_;
   bool batch_processing;
   std::vector<TensorLayout> output_layouts_;
-
- private:
-  int GetCurrBatchSize(Workspace &ws) {
```
stiepan, line 250: Getting the batch size can be done with the `InferBatchSizeFromInput`. The uniformity check is done at the operator class level, with the `EnforceUniformInput/OutputBatchSize`. And as for the check of requested batch size, it is set uniformly `batch_sizes_.resize(NumOutput(), batch_size);`.

dali/operators/generic/constant.h
```diff
@@ -83,7 +83,7 @@ class Constant : public Operator<Backend> {
       output_.Reset();
     }
     output_shape_ = max_output_shape_;
-    output_shape_.resize(ws.GetRequestedBatchSize(0));
+    output_shape_.resize(InferBatchSizeFromInput(ws));
```
stiepan, line 86: It does not have a tensor inputs.

dali/operators/generic/permute_batch.h
```diff
@@ -45,9 +45,10 @@ class PermuteBatchBase : public Operator<Backend> {
       this->spec_.TryGetRepeatedArgument(indices_, "indices");
     }
     // TODO(michalz): Remove when fully variable batch size is supported
-    DALI_ENFORCE(static_cast<int>(indices_.size()) == ws.GetRequestedBatchSize(0), make_string(
+    int cur_batch_size = InferBatchSizeFromInput(ws);
```
stiepan, line 48: It accepts a regular input, so the check will enforce the same input size as it did in non `split_` case (although the same thing is checked at the op level), but enables using it within split_ branches,

dali/operators/geometry/affine_transforms/transform_base_op.h
```diff
@@ -72,7 +72,7 @@ class TransformBaseOp : public SequenceOperator<Backend, true> {
 
   bool SetupImpl(std::vector<OutputDesc> &output_descs, const Workspace &ws) override {
     has_input_ = ws.NumInput() > 0;
-    auto curr_batch_size = has_input_ ? ws.GetInputBatchSize(0) : ws.GetRequestedBatchSize(0);
+    auto curr_batch_size = InferBatchSizeFromInput(ws);
```
mzient, line 75: This defeats the purpose of RequestedBatchSize. Instead, move `InferBatchSize` to the executor and invoke it prior to SetBatchSize. It will be a workaround, but at least we'll have _one_ place where we can change/improve batch size inference instead of spreading it across operators. It also allows for some smart(er) solutions or non-local batch source.

mzient (changes requested):
Move the batch size calculation to the executor and keep RequestedBatchSize.

dali/test/python/conditionals/test_pipeline_conditionals.py
```diff
@@ -617,3 +617,135 @@ def one_return():
...
+
+
+# Test operators that infer their batch sizes from the tensor argument inputs
+@params(fn.permute_batch, fn.roi_random_crop, fn.transforms.crop, fn.transforms.scale,
+        fn.transforms.shear, fn.transforms.translation, fn.transforms.rotation,
+        fn.random.uniform, fn.random.normal, fn.random.coin_flip)
+def test_named_tensor_arguments(op):
+
+    ops2params = {
+        fn.permute_batch: _tensor_arg_permute_batch_params,
```
mzient, line 694: This one is a nasty little bugger - in principle, the output batch_size comes from the `indices` argument and not the input.

mzient, line 694: `permute_batch` does have an input - although by original design the output batch size was taken from `indices`. I don't think there's any special case for this op - or any other means to infer the batch size - which would mean that it takes the batch_size just from the input, not arg. input.

stiepan: I put it on purpose here as it calls GetRequestedBatchSize manually as an extra check that GetRequestedBatchSize returns the "split" batch size.

stiepan:
!build

klecki (approved):

mzient (approved):

PR #4618, opened 2023-01-26
Fix classification of argument input-only operators in AutoGraph

klecki:
## Description:
Follow up for the #4617. 
Currently, an operator without inputs is considered a generator and is registered in the global scope - it will always produce full batch, and each use will insert appropriate split nodes - as there is no way of signaling the requested batch size from the pipeline definition level and some generators do not support variable batch size at all.
Some operators can have 0 positional inputs while having some argument inputs. Now they would be classified as generator and relegated to the global scope.
Include detection of argument inputs when registering the outputs of the operator - if the operator has any input or *argument input* it can detect the batch size of the input and infer the size of the output. This makes it usable in an `if` scope and doesn't require placing such operators in global scope. Placing them in the global scope results in unnecessary splits to be inserted, even though the produced batch size is correct for given scope.

Conditionals/AutoGraph

### Tests:

- [ ] Existing tests apply
- [x] New tests added
  - [x] Python tests
  - [ ] GTests
  - [ ] Benchmark
  - [ ] Other
- [ ] N/A

### Documentation
- [ ] Documentation updated
  - [ ] Docstring
  - [ ] Doxygen
  - [ ] RST
  - [ ] Jupyter
  - [ ] Other
- [ ] N/A

#### Requirements
- [x] N/A

**REQ IDs**: N/A

**JIRA TASK**: N/A

### DIFF
--- a/dali/test/python/conditionals/test_pipeline_conditionals.py
+++ b/dali/test/python/conditionals/test_pipeline_conditionals.py
@@ -535,6 +535,81 @@ def test_dot_gpu():
     compare_pipelines(*pipes, bs, iters)
 
 
+# Test if operators without positional inputs but with argument inputs are correctly handled
+# in the split/merge - so they are tracked in the local scope.
+
+
+def test_arg_inputs_scoped_tracking():
+    test_data_root = get_dali_extra_path()
+    caffe_db_folder = os.path.join(test_data_root, 'db', 'lmdb')
+
+    bs = 10
+    iters = 5
+    kwargs = {"batch_size": bs, "num_threads": 4, "device_id": 0, "seed": 42}
+
+    @experimental.pipeline_def(enable_conditionals=True, **kwargs)
+    def global_transform_pipe():
+        encoded, _ = fn.readers.caffe(path=caffe_db_folder)
+        decoded = fn.decoders.image(encoded, device="mixed")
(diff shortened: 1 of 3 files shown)


szalpal (approved):

dali/test/python/conditionals/test_pipeline_conditionals.py
```diff
@@ -535,6 +535,81 @@ def conditional_pipe():
...
+        else:
+            output = decoded
+        return output
+
+    @experimental.pipeline_def(enable_conditionals=True, **kwargs)
+    def scoped_transform_pipe():
+        encoded, _ = fn.readers.caffe(path=caffe_db_folder)
+        decoded = fn.decoders.image(encoded, device="mixed")
+        pred = fn.random.coin_flip(dtype=types.DALIDataType.BOOL)
+        angle = fn.random.uniform(values=[10, 20, 30])
```
stiepan, line 568: Can we have a test when you put the angle inside the if?

stiepan (approved):

klecki:
!build

dali/test/python/conditionals/test_pipeline_conditionals.py
klecki: There is a test that checks if pure generators are registered in the top scope (test_generators).

klecki:
!build

PR #4628, opened 2023-02-01
Fix @autoserialize problem with unknown module

szalpal:

## Category:

## Description:
When `autograph` has been introduced to DALI, we've faced the problem with autoserialization. Turns out, that `autograph` in runtime swaps the modules using `six` - probably to maintain compatibility between Python2 and Python3. Therefore, when traversing the module tree in search for `@autoserialize` decorator, inspect has not been able to find some modules and was prompting an error.

The solution is that if given module is unavailable, it is not traversed any more. DALI won't be able to retrieve `@autoserialize` from there anyhow, so no harm done.

### Tests:

- [x] Existing tests apply
- [ ] New tests added
  - [ ] Python tests
  - [ ] GTests
  - [ ] Benchmark
  - [ ] Other
- [ ] N/A

### Documentation
- [ ] Documentation updated
  - [ ] Docstring
  - [ ] Doxygen
  - [ ] RST
  - [ ] Jupyter
  - [ ] Other
- [ ] N/A

#### Requirements
- [ ] N/A

**REQ IDs**: N/A

**JIRA TASK**: N/A

### DIFF
--- a/dali/python/nvidia/dali/_utils/autoserialize.py
+++ b/dali/python/nvidia/dali/_utils/autoserialize.py
@@ -32,7 +32,7 @@ def _discover_autoserialize(module, visited):
     ret = []
     try:
         module_members = inspect.getmembers(module)
-    except ModuleNotFoundError as e:
+    except ModuleNotFoundError:
         # If any module can't be inspected, DALI will not be able to find the @autoserialize
         # anyway. We can just skip this module.
         return ret
--- a/dali/python/nvidia/dali/_utils/autoserialize.py
+++ b/dali/python/nvidia/dali/_utils/autoserialize.py
@@ -30,7 +30,12 @@ def _discover_autoserialize(module, visited):
     """
     assert module is not None
     ret = []
-    module_members = inspect.getmembers(module)
+    try:
+        module_members = inspect.getmembers(module)
+    except ModuleNotFoundError as e:
+        # If any module can't be inspected, DALI will not be able to find the @autoserialize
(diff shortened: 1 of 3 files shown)


szalpal:
!build

JanuszL (approved):

dali/python/nvidia/dali/_utils/autoserialize.py
```diff
@@ -30,7 +30,12 @@ def _discover_autoserialize(module, visited):
     """
     assert module is not None
     ret = []
-    module_members = inspect.getmembers(module)
+    try:
```
JanuszL, line 33: Can you add a simple test with `import xyz` inside?

szalpal:
!build

dali/python/nvidia/dali/_utils/autoserialize.py
szalpal: Done

dali/test/python/autoserialize_test/custom_module_inside.py
```diff
@@ -0,0 +1,9 @@
+from nvidia.dali.plugin.triton import autoserialize
+from nvidia.dali import pipeline_def
+
+
+@autoserialize
+@pipeline_def(batch_size=1, num_threads=1, device_id=0)
+def func_under_test():
+    import numpy as np  # noqa: F401
```
JanuszL, line 8: It is likely that the test env may have numpy installed. How about `import doesnt_exist`?

PR #4629, opened 2023-02-01
Add lazy `and` and `or`, and not lazy `not` support 

klecki:
## Description:
Add support for logical expressions when `conditionals_enabled=True`.
AutoGraph allows for the detection and overloading of logical expressions.

The `not` expression is executed eagerly as an operator - it's directly inserted in the AutoGraph overload.
For `and` and `or` AutoGraph captures the right argument of the logical expression as lambdas. We can utilize the same mechanism as in the `if` statement, and evaluate it lazily in accordance with proper Python semantics. 
The implementation is based on simplified form of the `if` statement. As we don't have an actual `and` and `or` operators, a special operator that validates both sides of the input is inserted in the binary logical expressions: `fn._conditional.validate_logical`.

All the logical expressions and `if` conditions are restricted to 0-d boolean inputs placed on CPU. The `boolean` part is important, as Python semantics for logical expression: https://docs.python.org/3/reference/expressions.html#boolean-operations require to evaluate the truthness (or call `__bool__`) of the inputs, and in the case of `and` and `or`, return the unchanged inputs. 

In cases like batches of: `[1, 0]` or `[0.0, 1.0]` it would mean that samplewise `or` results in `[1, 1.0]` - a batch of mixed types. If we restrict the inputs to just `booleans` we don't hit that problem.

The same error checking API as in `_conditional.validate_logical` is additionally added to check the inputs of the `if` statement condition and `not` operator to provide better and more consistent error messages between conditional use-cases. 
In `if` it is implemented by notifying the `split` operator that it is used as `if` statement implementation and it should apply the stricter checks. There is no need to do it in `merge`, as `split` is executed first.

All those operator land as builtin in a dedicated `conditional` directory, where the `split` and `merge` were moved for consistency.

The followup would be to include those expressions as a part of the tutorial/documentation.
There is another followup planned to lift some restrictions of boolean input type where it is possible without introducing the mentioned problems (for `if` condition and `not`).

Conditionals

### Tests:

- [ ] Existing tests apply
- [x] New tests added
  - [x] Python tests
  - [ ] GTests
  - [ ] Benchmark
  - [ ] Other
- [ ] N/A

### Documentation
- [ ] Documentation updated
  - [ ] Docstring
  - [ ] Doxygen
  - [ ] RST
  - [ ] Jupyter
  - [ ] Other
- [ ] N/A

#### Requirements
- [ ] N/A

**REQ IDs**: N/A

**JIRA TASK**: N/A


klecki:
!build

awolant (approved):
Looks like it works, well tested, well documented.

dali/pipeline/operator/builtin/conditional/validate_logical_expr.cc
```diff
@@ -0,0 +1,72 @@
...
+This operator is inserted when Python ``and`` and ``or`` expressions of ``if`` statements are used
+with ``enabled_conditionals=True``.
+It provides better error messages about the restrictions of inputs to particular expression
+rather than falling back to the error reported by the underlying implementation (which is based
+on ``if`` statements).
+The inputs are restricted to scalar values of boolean type - this makes them consistent
+with Python semantics and allows for unambiguous ``if`` evaluation.
+You can use mathematical operators `&`, `|`, or ``==`` and compare with 0 to emulate elementwise
+equivalent of logical operations on inputs of other types, shapes, and placements.
+Note that such operations would be eager.
```
stiepan, line 56: I am a bit wary of using word eager here as we have eager execution which is different thing. Having logical expressions evaluated lazily (in the skipping exection sense) and bitwise eagerly (in the same sense) is just the standard way of doing it, so maybe we are good without mentioning it?

stiepan (approved):

klecki:
!build

PR #4676, opened 2023-02-24
Fix the logical expression tests to avoid short-cutting them

klecki:
## Description:
Make one of the inputs a batch that always contains one True and one False value, so the right hand side is always evaluated.

### Tests:

- [x] Existing tests apply
The particular test was adjusted by changing the input to little bit less random.
Otherwise there was a chance to skip the validation of RHS expression and not
raise an error.
- [ ] New tests added
  - [ ] Python tests
  - [ ] GTests
  - [ ] Benchmark
  - [ ] Other
- [ ] N/A

### Documentation
- [ ] Documentation updated
  - [ ] Docstring
  - [ ] Doxygen
  - [ ] RST
  - [ ] Jupyter
  - [ ] Other
- [ ] N/A

#### Requirements
- [ ] N/A

**REQ IDs**: N/A

**JIRA TASK**: N/A

### DIFF
--- a/dali/test/python/conditionals/test_logical_expressions.py
+++ b/dali/test/python/conditionals/test_logical_expressions.py
@@ -181,12 +181,30 @@ def test_lazy_eval_with_oob():
     compare_pipelines(*pipes, bs, iters)
 
 
+def logical_true_false_random():
+    """ Return an External Source that returns batch of [True, False, <random booleans>]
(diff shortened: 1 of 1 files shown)


klecki:
!build

JanuszL (approved):

banasraf (approved):

PR #4694, opened 2023-03-07
[AA] Add auto augmentation wrapper

stiepan:
## Category:

## Description:
This is a prerequisite for automatic augmentation library (https://github.com/NVIDIA/DALI/pull/4648).

The automatic augmentations in this context are just different schemes of random selection of augmentation per sample. 

The common idea is that, apart from sampling the augmentation, the schemes provide parameters to the randomly selected augmentations. In order to do that in abstract way, they use the notion of magnitude: each operation has adimissible magnitude range, which is divided into a number of bins. The automatic augmentation schemere selects the magnitude bin and the augmentation's role is to understand how it maps to exact paramter. For example, for `fn.rotate` magnitude is an angle.

This PR adds `@augmentation` decorator that lets user wrap a function that uses DALI processing so that it can adhere to automatic augmentation scheme "protocol".

The user who is interested in only running a standard suite for [Auto/Rand/Trivial]Augment should not need to use any of the utils here directly. 

The simplest interaction needed is changeing augmenation setup (such as magnitudes range), which is supposed to look as follows (to change the range of angles to (0, 60), randomly negated for each sample:

```
rotate = a.rotate.augmentation((0, 60), randomly_negate=True)
```

For more advanced cases, the `@augmentation` is the user-facing tool for wrapping custom code as an augmentation.

```
@augmentation(mag_range=(0, 30), randomly_negate=True)
def rotate(sample, angle, fill_value=0, interp_type=None):
    return fn.rotate(sample, angle=angle, fill_value=fill_value, interp_type=interp_type)
```

```
def warp_x_param(magnitude):
    return [magnitude, 0]

@augmentation(mag_range=(0, 0.3), randomly_negate=True, as_param=warp_x_param)
def shear_x(sample, shear, fill_value=0, interp_type=None):
    mt = fn.transforms.shear(shear=shear)
    return fn.warp_affine(sample, matrix=mt, fill_value=fill_value, interp_type=interp_type,
                          inverse_map=False)

```

### Tests:

- [ ] Existing tests apply
- [x] New tests added
  - [x] Python tests
  - [ ] GTests
  - [ ] Benchmark
  - [ ] Other
- [ ] N/A

### Documentation
- [x] Documentation updated
  - [x] Docstring
  - [ ] Doxygen
  - [ ] RST
  - [ ] Jupyter
  - [ ] Other
- [ ] N/A

#### Requirements
- [x] N/A

**REQ IDs**: N/A

**JIRA TASK**: DALI-3230

### DIFF
--- a/dali/python/nvidia/dali/auto_aug/core/_args.py
+++ b/dali/python/nvidia/dali/auto_aug/core/_args.py
@@ -1,0 +1,87 @@
+# Copyright (c) 2023, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import inspect
+
+
+class MissingArgException(Exception):
+
+    def __init__(self, message, augmentation, missing_args):
+        super().__init__(message)
+        self.augmentation = augmentation
+        self.missing_args = missing_args
+
+
+class UnusedArgException(Exception):
+
+    def __init__(self, message, unused_args):
+        super().__init__(message)
+        self.unused_args = unused_args
+
+
+def filter_extra_accepted_kwargs(fun, kwargs, skip_positional=0):
+    sig = inspect.signature(fun)
+    # the params from signature with up to skip_positional filtered out
+    # (less only if there is not enough of positional args)
+    params = [(name, param) for i, (name, param) in enumerate(sig.parameters.items())
+              if i >= skip_positional or param.kind not in
+              [inspect.Parameter.POSITIONAL_OR_KEYWORD, inspect.Parameter.POSITIONAL_ONLY]]
+    extra = [
+        name for (name, param) in params
+        if param.kind in [inspect.Parameter.POSITIONAL_OR_KEYWORD, inspect.Parameter.KEYWORD_ONLY]
+    ]
+    return {name: value for name, value in kwargs.items() if name in extra}
+
+
+def get_required_kwargs(fun, skip_positional=0):
+    sig = inspect.signature(fun)
+    # the params from signature with up to skip_positional filtered out
+    # (less only if there is not enough of positional args)
+    params = [(name, param) for i, (name, param) in enumerate(sig.parameters.items())
+              if i >= skip_positional or param.kind not in
+              [inspect.Parameter.POSITIONAL_OR_KEYWORD, inspect.Parameter.POSITIONAL_ONLY]]
+    return [
+        name for name, param in params if param.default is inspect.Parameter.empty
+        and param.kind in [inspect.Parameter.POSITIONAL_OR_KEYWORD, inspect.Parameter.KEYWORD_ONLY]
+    ]
+
+
+def get_num_positional_args(fun):
+    sig = inspect.signature(fun)
+    return len([
+        name for name, param in sig.parameters.items() if param.kind in
+        [inspect.Parameter.POSITIONAL_OR_KEYWORD, inspect.Parameter.POSITIONAL_ONLY]
+    ])
+
+
+def get_missing_kwargs(fun, kwargs, skip_positional=0):
+    required = get_required_kwargs(fun, skip_positional=skip_positional)
+    return [name for name in required if name not in kwargs]
+
+
+def filter_unused_args(augmentations, kwargs):
+    used_kwargs = set(kwarg_name for augment in augmentations
+                      for kwarg_name in filter_extra_accepted_kwargs(augment.op, kwargs, 2))
+    return [kwarg_name for kwarg_name in kwargs if kwarg_name not in used_kwargs]
+
+
+def forbid_unused_kwargs(augmentations, kwargs, call_name):
+    unused_args = filter_unused_args(augmentations, kwargs)
+    if unused_args:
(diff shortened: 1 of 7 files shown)


stiepan:
!build

stiepan:
!build

dali/python/nvidia/dali/auto_aug/core/decorator.py
```diff
@@ -0,0 +1,100 @@
...
+        A callback that transforms the magnitude into a parameter. The parameter will be passed to
+        the decorated operation instead of the plain magnitude. This way, the parameters for the
+        range of magnitudes can be computed once in advance and stored as a Constant node.
+        Note, the callback must return numpy arrays or data directly convertible to numpy arrays
+        (in particular, no pipelined DALI operators can be used in the callback).
+        The output type and dimensionality must be consistent and not depend on the magnitude.
+    param_device: str
+        A "cpu" or "gpu", describes where to store the precomputed parameters.
+    name: str
+        Name of the augmentation. By default, the name of the decorated function is used.
```
klecki, line 72: I don't see the name of the function accessed anywhere, or maybe I missed it somehow.

dali/python/nvidia/dali/auto_aug/core/_args.py
```diff
@@ -0,0 +1,87 @@
...
+
+
+class UnusedArgException(Exception):
+
+    def __init__(self, message, unused_args):
+        super().__init__(message)
+        self.unused_args = unused_args
+
+
+def filter_extra_accepted_kwargs(fun, kwargs, skip_positional=0):
```
klecki, line 33: I am asking for docs as the name can be understood as removing the extra kwargs that are accepted rather than obtaining them.

dali/test/python/auto_aug/test_augmentation_decorator.py
```diff
@@ -0,0 +1,392 @@
...
+        with assert_warns(Warning, glob=warn_glob):
+            p = pipeline(num_threads=4, device_id=0, batch_size=batch_size)
+    p.build()
+    magnitudes, = p.run()
+    magnitudes = [np.array(el) for el in magnitudes]
+    if use_implicit_sign:
+        # the implicit sign is random, just sanity check if
+        # there are some positive and negative magnitudes
+        assert any(el < 0 for el in magnitudes)
+        assert any(el > 0 for el in magnitudes)
```
klecki, line 142: I guess you will get a fully positive or negative batch at some point, unless the seed will be fixed.

klecki (approved):
Some nitpicks, some suggestions (I think we can take care of potential improvements in a follow up PR), and asking for a bit more docs.

Also, at some point we should dedicate a page in the docs for this and list the contents of the modules.

dali/python/nvidia/dali/auto_aug/core/_augmentation.py
```diff
@@ -0,0 +1,296 @@
...
+                 num_magnitude_bins: Optional[int] = None, random_sign: Optional[_DataNode] = None,
+                 **kwargs) -> _DataNode:
+        """
+        Applies the decorated transformation to the `sample` as if by calling
+        `self.op(sample, param, **kwargs)` where
+        `param = as_param(magnitudes[magnitude_bin] * ((-1) ** random_sign))`.
+
+        Parameter
+        ---------
+        sample : DataNode
```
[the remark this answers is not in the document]
stiepan: The conditionals are written in this sample-like style, so I stuck to it.

dali/python/nvidia/dali/auto_aug/core/_augmentation.py
```diff
@@ -0,0 +1,296 @@
...
+            raise Exception(
+                f"The augmentation `{self.name}` has `mag_range` specified, so when called, "
+                f"it requires `magnitude_bin` parameter to select the magnitude from the "
+                f"`mag_range`.\nError in augmentation: {self}.")
+        # if the bin is runtime defined, create Constant node with the array of params
+        # for any possible bin and use subscript to take the right one at each iteration
+        if isinstance(magnitude_bin, _DataNode):
+            if self.randomly_negate:
+                magnitudes = _remap_bins_to_signed_magnitudes(magnitudes)
+                magnitude_bin = 2 * magnitude_bin + random_sign
```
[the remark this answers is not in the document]
stiepan: Good point, I replaced the explicit random_sign passing with `signed_bin()` call, you invoke it like this `signed_bin(magnitude_bin)` and the necessary node is created at the place of the call. Then, when wrapper sees it as `magnitude_bin` param, it can use that node. It is meant as an optimisation - if user does not make the call but the augmentation has randomly_negate set to True, it will wrap the magnitude_bin on its own.

dali/test/python/auto_aug/test_augmentation_decorator.py
stiepan: Spot-on, I've added the seeds

stiepan:
!build

dali/python/nvidia/dali/auto_aug/core/decorator.py
stiepan:
```
@property
    def name(self):
        return self._name or self.op.__name__
```

awolant (approved):
Nice, well tested, lovely job :+1: 

From the user perspective I struggled with one small thing. From this code sample:
```
@augmentation(mag_range=(0, 30), randomly_negate=True)
def rotate(sample, angle, fill_value=0, interp_type=None):
    return fn.rotate(sample, angle=angle, fill_value=fill_value, interp_type=interp_type)
```
it was not immediately clear to me to which param `mag_range` applies. After going through the doc it is clear that the second argument is being worked on here. 

One thing I have seen in similar situations was to explicitly point the argument by name or by index.
Additionally, when I first saw `as_param` I thought that it does exactly that - tells to the decorator which argument of the function decorated is the param to be modified. 

Maybe something to consider would be to rename `as_param` to something like `map_param` or `transform_param` and use `as_param` or `param_id` name to point to the argument that `mag_range` applies to.

Regardless, I think it would be ok to leave it as it is.

PR #4697, opened 2023-03-07
Reduce memory footprint of conditional execution

klecki:
## Description:
Keep only the currently required memory for operators producing outputs within a conditional scope.

The executor detects if there were conditional statements used within the graph. In that case processing a partial batch is assumed to be an execution of operator within a conditional scope.
For such operators their outputs are cleared of reserved memory before run (and new allocation for that iteration) - this means that we keep only the currently processed batch size instead of the biggest one we saw.

Conditional execution.

Should we have some leeway in freeing the reserved memory? For example some margin in how big the batch size change would be?

### Tests:

- [x] Existing tests apply
This should not change any outside functionality but it might impact perf of conditional execution.
The executor statistics track the maximum of the seen allocation so we can't even detect it from Python.
- [ ] New tests added
  - [ ] Python tests
  - [ ] GTests
  - [ ] Benchmark
  - [ ] Other
- [ ] N/A

### Documentation
- [ ] Documentation updated
  - [ ] Docstring
  - [ ] Doxygen
  - [ ] RST
  - [ ] Jupyter
  - [ ] Other
- [ ] N/A

#### Requirements
- [x] N/A

**REQ IDs**: N/A

**JIRA TASK**: 3282

### DIFF
--- a/dali/pipeline/executor/executor.cc
+++ b/dali/pipeline/executor/executor.cc
@@ -23,10 +23,11 @@
 #include "dali/pipeline/data/backend.h"
 #include "dali/pipeline/executor/executor.h"
 #include "dali/pipeline/executor/queue_metadata.h"
+#include "dali/pipeline/executor/source_info_propagation.h"
 #include "dali/pipeline/graph/op_graph_storage.h"
+#include "dali/pipeline/operator/builtin/conditional/split_merge.h"
 #include "dali/pipeline/operator/common.h"
 #include "dali/pipeline/workspace/workspace_data_factory.h"
-#include "dali/pipeline/executor/source_info_propagation.h"
 
 namespace dali {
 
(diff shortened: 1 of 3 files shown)


klecki:
!build

dali/pipeline/executor/executor.cc
```diff
@@ -398,6 +399,38 @@ void Executor<WorkspacePolicy, QueuePolicy>::RunHelper(OpNode &op_node, Workspac
...
   }
 
+  // If we are using conditional statements try to keep smaller memory footprint by not keeping
+  // the largest batch produced in conditional block as reserved memory.
+  // If the conditionals were detected in this graph (split operator with `_if_stmt` argument)
+  // see if the operator is executed in conditional scope - seeing a partial batch processed
+  // would indicate it in most cases (unless someone mixes partial batches and conditionals).
+  // In such case reset the outputs before adjusting their size so we free the reserved memory
+  // and allocated only batch of currently needed size.
+  if (has_conditionals_) {
```
JanuszL, line 409: So if the condition is applied to any part of the graph we reset buffer for all workspaces, right?

JanuszL: `has_conditionals_` is global for the whole graph.
Here we check `ws.GetInputBatchSize(i) < max_batch_size_` for all ops in the `has_conditionals_` is set, not only for the ops inside the scope.
Am I right?

klecki: >   // If the conditionals were detected in this graph (split operator with `_if_stmt` argument)
  // see if the operator is executed in conditional scope - seeing a partial batch processed

If there are any conditionals in the graph I treat an op processing a batch smaller than max_batch_size as running in a conditional scope - as this is typically a good indication, apart for some edge cases.

We can extend it by adding proper scoping information to every operator, but that is more involved change and I am not sure how to make it in a good, forward compatible way.
My current idea is to add a hidden parameter with the predicate to every operator in conditional scope, but that just adds a lot of edges in graph. Another idea is to number every scope and pass those ids around so we can look it up. This seems a bit better, but would be a bigger change.

JanuszL: I think that long term it would be to move to a fully dynamic allocation of the memory for each workspace, each iteration, and limit the liveness of the buffers. For now, I would stay with the proposed approach and don't try harder with it.

klecki: I am not sure what we are discussing here any more. This basically does dynamic allocation when there are conditionals in the graph for anything that looks like running in a conditional scope. If there are no conditionals or something is processing a max_batch_size, we keep the current behaviour.

JanuszL: I'm just trying to understand how it works and if it is aligned with the comment.

dali/pipeline/executor/executor.h
```diff
@@ -373,6 +374,16 @@ class DLL_PUBLIC Executor : public ExecutorBase, public QueuePolicy {
...
 
+  /**
+   * @brief Traverses the Graph to check if there is a _conditional.split with _if_stmt argument
+   * specified, indicating that the `if` statement was used and the graph is operating in
+   * conditional mode (enable_conditionals=True).
+   *
+   * TODO(klecki): Consider adding a specific callback to Pipeline so the frontend can indicate
+   * it directly.
+   */
+  void DetectConditionals();
```
awolant, line 385: Maybe it would be good to add some super small test for it just to cover it and have it running in the tests for future reference.

dali/pipeline/executor/executor.cc
```diff
@@ -398,6 +399,38 @@ void Executor<WorkspacePolicy, QueuePolicy>::RunHelper(OpNode &op_node, Workspac
...
+      for (int i = 0; i < spec.NumOutput(); i++) {
+        if (ws.template OutputIsType<CPUBackend>(i)) {
+          ws.template Output<CPUBackend>(i).Reset();
+        } else {
+          ws.template Output<GPUBackend>(i).Reset();
+        }
+      }
+    }
+  }
+
```
awolant, line 433: Maybe this whole block would be more readable as a separate method?

awolant (approved):

dali/pipeline/executor/executor.h
klecki: Sounds like a splendid idea.

klecki:
!build

dali/pipeline/executor/executor.cc
klecki: done

klecki:
!build

stiepan (approved):

PR #4712, opened 2023-03-14
DALI Pipeline inputs as named arguments to `Pipeline.run()`

szalpal:
## Category:

## Description:

Currently, providing inputs to DALI Pipeline is highly cumbersome. We have a couple of ways:
1. Using Readers,
2. Using `external_source` operator and its argument `source`,
3. Using an input operator (e.g. `external_source`) and `feed_input` function.

All of these are quite problematic. This PR attempts to tackle this with allowing yet another way of providing inputs.

My proposal is to augment `pipe.run()` function with `**kwargs` in Python. Every such argument, provided that it has a corresponding input operator in DALI pipeline, can accept any data assigned to the name of such input operator. For example:

```python
@pipeline_def
def pipe():
    return fn.external_source(name="my_inp")

p = pipe(...)
p.build()
p.run(my_inp=np.random(...))
```
The change is added run `.run()` only, since I figured it's always possible to add it to the `schedule_run` too, but removing it will be problematic due to backwards compatibility issues.

### Tests:

- [x] Existing tests apply
- [x] New tests added
  - [ ] Python tests
  - [ ] GTests
  - [ ] Benchmark
  - [ ] Other
- [ ] N/A

### Documentation
- [x] Documentation updated
  - [x] Docstring
  - [ ] Doxygen
  - [ ] RST
  - [ ] Jupyter
  - [ ] Other
- [ ] N/A

#### Requirements
- [ ] N/A

**REQ IDs**: N/A

**JIRA TASK**: N/A

### DIFF
--- a/dali/python/nvidia/dali/pipeline.py
+++ b/dali/python/nvidia/dali/pipeline.py
@@ -1036,7 +1036,6 @@ Parameters
             raise RuntimeError("Pipeline must be built first.")
         return self._pipe.Outputs()
 
-
     def _are_pipeline_inputs_possible(self):
         """
         Returns True if using pipeline_inputs argument in .run() function is possible.
@@ -1046,7 +1045,6 @@ Parameters
         if self.exec_separated:
             return self._cpu_queue_size <= 1 and self._gpu_queue_size <= 1
         return self.prefetch_queue_depth <= 1
-
 
     def run(self, **pipeline_inputs):
         """


szalpal:
!build

szalpal:
!build

szalpal:
!build

dali/python/nvidia/dali/pipeline.py
```diff
@@ -965,9 +965,9 @@ def schedule_run(self):
         Should not be mixed with :meth:`run` in the same pipeline"""
         with self._check_api_type_scope(types.PipelineAPIType.SCHEDULED):
             if self._first_iter and self._exec_pipelined:
-                self._prefetch()
+                self._prefetch(**pipeline_inputs)
```
mzient, line 968: That's bad.... It would most likely cause very weird behavior of the pipeline and explaining it to the user would be extremely hard. Instead, rather enforce that queue depth is 1 when using pipeline inputs.

szalpal: Unfortunately, with how currently the running logic is organised, it has to be there. `if self._first_iter and self._exec_pipelined:` means, that when we legitimately set `exec_pipelined=True` and `prefetch_queue_depth=1`, the `prefetch` function will be executed. And we want to have `pipeline_inputs` in this scenario.

I've added the enforce and extensive explanation in the `run` function. Hopefully this will solve the issue.

mzient: (to put our call in writing):
We don't have to propagate it down to `_prefetch` and `_run_once` - we could just call `feed_input` in `schedule_run` or even restrict the new API to run, so we avoid _yet another_ way to feed the (prefetching) pipeline.

dali/python/nvidia/dali/pipeline.py
```diff
@@ -1046,14 +1047,57 @@ def run(self):
...
+            Please note, that using this feature requires setting either ``prefetch_queue_depth=1``
+            or ``exec_pipelined=False`` in DALI Pipeline constructor.
+
+            This feature can be considered as a syntactic sugar over :meth:`feed_input` function.
+
+        Returns
+        -------
             A list of `TensorList` objects for respective pipeline outputs
         """
+        if len(pipeline_inputs) > 0 and (self.prefetch_queue_depth > 1 and self.exec_pipelined):
```
mzient, line 1085: Note: I wonder what's the value of `prefetch_queue_depth` in case of separated queues - and, specifically, what `prefetch_queue_depth > 1` would evaluate to. At least I'd try to specify separated queues and see the error below pops up.

szalpal:
!build

szalpal:
!build

szalpal:
!build

szalpal:
!build

mzient (approved):

szalpal:
!build

awolant (approved):

PR #4715, opened 2023-03-15
Support numerical types in `if` predicate and `not` expression

klecki:
## Description:
Evaluate scalar numerical types treating them as truthy and falsy values.
Utilize regular promotion from numeric types to boolean. We can do it safely as `not` produces a new value and the predicate in `if` is not related to the output.
`and` and `or` expression stay limited to `bool` inputs to not risk creating scenarios where batches of mixed types can arise.

Conditional expressions, split/merge.

### Tests:

- [x] Existing tests apply
Existing test was modified to not expect exception when non-boolean type is passed as `if` predicate or into `not`.
- [x] New tests added
  - [x] Python tests
  - [ ] GTests
  - [ ] Benchmark
  - [ ] Other
- [ ] N/A

### Documentation
- [x] Documentation updated
  - [ ] Docstring
  - [ ] Doxygen
  - [ ] RST
  - [x] Jupyter
  - [ ] Other
- [ ] N/A

#### Requirements
- [ ] N/A

**REQ IDs**: N/A

**JIRA TASK**: N/A

### DIFF
--- a/dali/pipeline/operator/builtin/conditional/split_merge.h
+++ b/dali/pipeline/operator/builtin/conditional/split_merge.h
@@ -1,4 +1,4 @@
-// Copyright (c) 2022, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+// Copyright (c) 2022-2023, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -16,8 +16,11 @@
 #define DALI_PIPELINE_OPERATOR_BUILTIN_CONDITIONAL_SPLIT_MERGE_H_
 
 #include <string>
+#include "dali/core/static_switch.h"
 #include "dali/pipeline/data/backend.h"
 #include "dali/pipeline/data/tensor_list.h"
+#include "dali/pipeline/data/types.h"
+#include "dali/pipeline/operator/builtin/conditional/validation.h"
 #include "dali/pipeline/operator/op_schema.h"
 
 namespace dali {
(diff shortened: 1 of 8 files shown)


klecki:
!build

dali/pipeline/operator/builtin/conditional/split_merge.h
```diff
@@ -30,10 +33,16 @@ namespace dali {
  * @param condition_idx Index of the sample that this condition applies to (input for split, output
  * for merge)
  */
-inline int get_group_index(const TensorList<CPUBackend>& condition, int condition_idx,
-                              bool is_logical = true) {
+inline int get_group_index(const TensorList<CPUBackend> &condition, int condition_idx,
```
stiepan, line 36: Just a food for thought, could it affect perf if we accepted TensorView here and did the switch for the while batch instead as we do most of the time (when samples are far from scalars of course).

docs/examples/general/conditionals.ipynb
```diff
@@ -153,7 +153,7 @@
     "\n",
     "### Requirements\n",
     "There are several requirements that needs to be met when using the conditionals within DALI pipeline:\n",
-    "1. The `if` condition must be a DALI DataNode and the underlying data must be of type Bool.\n",
+    "1. The `if` condition must be a DALI DataNode and the underlying data must be of type `BOOL` or other numerical type that can be evaluated to True or False in accordance with Python semantics.\n",
```
mzient, line 156: I think that at this point there's no reason to mention bool.
```suggestion
    "1. The `if` condition must be a DALI DataNode and the underlying data must be of a numerical type that can be evaluated to True or False in accordance with Python semantics.\n",
```
This is a nitpick / thing to consider, nothing serious.

mzient (approved):
+1 for updating stale comments

docs/examples/general/conditionals.ipynb
klecki: CI is super lazy, I will update it when I revisit the docs about the logical expressions `and`, `or` and `not`.

PR #4733, opened 2023-03-21
Add timestamps to QA test output.

mzient:
## Category:

## Description:
It redirects test script output to a perl incantation that adds logs to stdout and stderr.

Tests

### Tests:

- [ ] Existing tests apply
- [ ] New tests added
  - [ ] Python tests
  - [ ] GTests
  - [ ] Benchmark
  - [ ] Other
- [X] N/A

### Documentation
- [ ] Documentation updated
  - [ ] Docstring
  - [ ] Doxygen
  - [ ] RST
  - [ ] Jupyter
  - [ ] Other
- [X] N/A

#### Requirements
- [X] N/A

**REQ IDs**: N/A

**JIRA TASK**: N/A


qa/test_template_impl.sh
```diff
@@ -0,0 +1,206 @@
+#!/bin/bash
```
mzient, line 1: Just renamed

qa/test_template_impl.sh
mzient, line 1: Just renamed

JanuszL (approved):

szalpal (approved):
My only concern would be if the `perl` dependency is sufficed in all the configurations we have in the CI

mzient:
> My only concern would be if the `perl` dependency is sufficed in all the configurations we have in the CI. Maybe running L1 manually would be a good idea?

Good point. I don't think L1 would change anything, but I've triggered Conda and AARCH64 builds - the environments there are quite different, so if anything goes sideways, I'd expect it to happen there.

JanuszL:
> > My only concern would be if the `perl` dependency is sufficed in all the configurations we have in the CI. Maybe running L1 manually would be a good idea?
> 
> Good point. I don't think L1 would change anything, but I've triggered Conda and AARCH64 builds - the environments there are quite different, so if anything goes sideways, I'd expect it to happen there.

The conda build doesn't work right now due to internal conda things. I would skip that for now.

PR #4747, opened 2023-03-27
Move enable_conditionals option to regular @pipeline_def

klecki:
## Description:
Refactor the `@pipeline_def` and `@experimental.pipeline_def` code to group up the common steps in separate functions.
Move the `enable_conditionals` to an actual keyword-only argument of `@pipeline_def`. 

Expand and rearrange the documentation.
Adjust some of the tests to use the regular decorator instead of the experimental one.

pipeline_def, conditionals

### Tests:

- [x] Existing tests apply
The conditionals tests use mixed decorators as some of them will be used with debug_mode tests
the notebook example for conditionals was adjusted to use regular decorator
automatic augments were adjusted to use regular decorator.
- [ ] New tests added
  - [ ] Python tests
  - [ ] GTests
  - [ ] Benchmark
  - [ ] Other
- [ ] N/A

### Documentation
- [x] Documentation updated
  - [x] Docstring
  - [ ] Doxygen
  - [x] RST
  - [x] Jupyter
  - [ ] Other
- [ ] N/A

#### Requirements
- [ ] N/A

**REQ IDs**: N/A

**JIRA TASK**: N/A

### DIFF
--- a/dali/python/nvidia/dali/pipeline.py
+++ b/dali/python/nvidia/dali/pipeline.py
@@ -1530,7 +1530,7 @@ def _generate_graph(pipe, func, fn_args, fn_kwargs):
         pipe.set_outputs(*po)
 
 
-def pipeline_def(fn=None, **pipeline_kwargs):
+def pipeline_def(fn=None, *, enable_conditionals=False, **pipeline_kwargs):
     """
     Decorator that converts a graph definition function into a DALI pipeline factory.
 
@@ -1603,21 +1603,27 @@ def pipeline_def(fn=None, **pipeline_kwargs):
 
         pipe = my_pipe(batch_size=42, num_threads=3)
         ...
+
+    Keyword args
+    ------------
+
+    enable_conditionals : bool, optional
+        Enable support for conditional execution of DALI operators using ``if`` statements
+        in the pipeline definition, by default False.
     """
-    pipeline_conditionals = pipeline_kwargs.pop('enable_conditionals', False)
 
     def actual_decorator(func):
 
         @functools.wraps(func)
         def create_pipeline(*args, **kwargs):
-            conditionals_on = kwargs.get('enable_conditionals', pipeline_conditionals)
+            conditionals_on = kwargs.get('enable_conditionals', enable_conditionals)
 
-            func = _preprocess_pipe_func(func, conditionals_on)
-            pipeline_args, fn_kwargs = _regroup_args(func, pipeline_kwargs, kwargs)
+            pipe_func = _preprocess_pipe_func(func, conditionals_on)
+            pipeline_args, fn_kwargs = _regroup_args(pipe_func, pipeline_kwargs, kwargs)
             pipe = Pipeline(**pipeline_args)
             _preprocess_pipe_object(pipe, conditionals_on, args, fn_kwargs)
 
-            _generate_graph(pipe, func, args, fn_kwargs)
+            _generate_graph(pipe, pipe_func, args, fn_kwargs)
             return pipe
 
         # Add `is_pipeline_def` attribute to the function marked as `@pipeline_def`
@@ -1681,7 +1687,7 @@ def _collect_ops(output_nodes):
     return ops
 
 
-def _pipeline_def_experimental(fn=None, **pipeline_kwargs):
+def _pipeline_def_experimental(fn=None, *, enable_conditionals=False, **pipeline_kwargs):
     """Variant of :meth:`@pipeline_def <nvidia.dali.pipeline_def>` decorator that enables additional
     experimental features. It has the same API as its non-experimental variant with the addition of
     the keyword arguments listed below.
@@ -1696,11 +1702,6 @@ def _pipeline_def_experimental(fn=None, **pipeline_kwargs):
             This mode is intended only for debugging purposes - the pipeline performance will be
             significantly worse than the non-debug mode.
 
-    enable_conditionals : bool, optional
-        Enable support for conditional execution of DALI operators using ``if`` statements
-        in pipeline definition, by default False.
-
-
     .. note::
 
         The features enabled by this decorator are experimental. The API may change and the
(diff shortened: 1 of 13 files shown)


klecki:
!build

dali/python/nvidia/dali/pipeline.py
```diff
@@ -1455,7 +1455,82 @@ def _discriminate_args(func, **func_kwargs):
     return ctor_args, fn_args
 
 
-def pipeline_def(fn=None, **pipeline_kwargs):
+def _regroup_args(func, pipeline_def_kwargs, fn_call_kwargs):
+    """Regroup arguments that are directed into Pipeline object construction and those that
+    are passed into pipeline definition function.
+
+    Parameters
```
szalpal, line 1462: I know this doc implies it, but please map explicitly return value with its definition. Like:
```
Regroup arguments that are directed into Pipeline object construction (Pipeline kwargs) and those that are passed into pipeline definition function (Function kwargs).
```

dali/python/nvidia/dali/pipeline.py
```diff
@@ -1528,23 +1603,27 @@ def my_pipe():
 
         pipe = my_pipe(batch_size=42, num_threads=3)
         ...
+
+    Keyword args
+    ------------
+
+    enable_conditionals : bool, optional
+        Enable support for conditional execution of DALI operators using ``if`` statements
+        in the pipeline definition, by default False.
```
szalpal, line 1612: I understand, that `enable_conditionals` flag has been added to the `pipeline_def`, because it is supported only in decorator mode. Have you considered adding it to the `Pipeline` constructor and handle it there?

The purpose from my perspective would be to accumulate all pipeline parameters in one place (`Pipeline` ctor) and make the decorator pass-through only.

dali/python/nvidia/dali/pipeline.py
```diff
@@ -1528,23 +1603,27 @@ def my_pipe():
 
         pipe = my_pipe(batch_size=42, num_threads=3)
         ...
+
+    Keyword args
+    ------------
+
+    enable_conditionals : bool, optional
```
szalpal, line 1610: How about `enable_cond`? I know it's not that well descriptive than `enable_conditionals`, but my thinking here is that I never know, whether I should use singular or plural... `enable_conditional` or `enable_conditionals`?

dali/python/nvidia/dali/pipeline.py
```diff
@@ -1455,7 +1455,82 @@ def _discriminate_args(func, **func_kwargs):
...
+        pipe._conditionals_enabled = True
+        pipe._condition_stack = _conditionals._ConditionStack()
+        # Add all parameters to the pipeline as "know" nodes in the top scope.
+        for arg in args:
+            if isinstance(arg, DataNode):
+                _conditionals.register_data_nodes(arg)
+        for _, arg in fn_kwargs.items():
+            if isinstance(arg, DataNode):
+                _conditionals.register_data_nodes(arg)
+        Pipeline.pop_current()
```
stiepan, line 1505: Shouldn't that be a context manager?

klecki: Context manager doesn't work with debug mode, that's why I push and pop by hand.

stiepan: Oh. In what way it does not work? Is this because the `pipe` is somehow spetial in that case? How about using `contexlib` or just `try: finially:` to somehow make sure we won't accidentaly leak pipelines.

klecki: It is explicitly disabled in debug mode (it raises an error), so people don't define pipelines via `with pipe` for debug mode - we need a function that we capture and are able to run.
I can add the `try: finally:` or similar to always release it.

dali/python/nvidia/dali/pipeline.py
klecki: The problem is, the using `Pipeline` may direct people to use the stateful pipeline definition via `define_graph`, where one can define helper methods that utilize the `self` as a state to keep some DataNodes. And it is not recommended to use, as it may either error out during tracing or produce a wrong trace.

dali/python/nvidia/dali/pipeline.py
awolant: Wouldn't this issue be better solved by IntelliSense/syntax support in Python instead? When possbile, I am up for making things as descriptive as possible. But this is just a thought, I am fine either way

awolant (approved):

dali/python/nvidia/dali/pipeline.py
klecki: Done

dali/python/nvidia/dali/pipeline.py
klecki: Done

dali/python/nvidia/dali/pipeline.py
```diff
@@ -1455,7 +1455,82 @@ def _discriminate_args(func, **func_kwargs):
...
+        if isinstance(pipe_outputs, tuple):
+            po = pipe_outputs
+        elif pipe_outputs is None:
+            po = ()
+        else:
+            po = (pipe_outputs, )
+        pipe.set_outputs(*po)
+
+
+def pipeline_def(fn=None, *, enable_conditionals=False, **pipeline_kwargs):
```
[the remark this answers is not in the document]
klecki: As far as I am aware, it should just work based on how keyword only arguments and `**kwargs` work.
https://peps.python.org/pep-3102/
```
For each keyword argument:
    If there is a parameter with the same name as the keyword, then the argument value is assigned to that parameter slot. However, if the parameter slot is already filled, then that is an error.
    Otherwise, if there is a ‘keyword dictionary’ argument, the argument is added to the dictionary using the keyword name as the dictionary key, unless there is already an entry with that key, in which case it is an error.
    Otherwise, if there is no keyword dictionary, and no matching named parameter, then it is an error.
```

I can still double check if you think it is needed.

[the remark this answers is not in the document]
klecki: I added a test for no-parenthesis option of decorator.

klecki:
!build

Issue #4740, opened 2023-03-23
Dali leaks memory on invalid images


joey-trigo:
Hi!

Due to an issue earlier in my pipeline dali received invalid JPEGs. When that happened, dali ended up leaking memory on each request until my process was killed by OOM. I don't necessarily expect dali to handle invalid JPEGs elegantly but it definitely shouldn't be leaking memory

Here is a minimal test case to reproduce this behaviour:

```
from nvidia.dali import pipeline_def
import nvidia.dali.fn as fn
import nvidia.dali.types as types

image_dir = "./"

while True:
    @pipeline_def
    def simple_pipeline():
        jpegs, labels = fn.readers.file(file_root=image_dir)
        images = fn.decoders.image(jpegs, device='mixed', output_type=types.BGR)
        return images, labels

    pipe = simple_pipeline(batch_size=1, num_threads=1, device_id=0)
    pipe.build()

    pipe_out = pipe.run()
```

Running this with a folder `foo/` containing a regular jpg (such as [good.jpg](https://user-images.githubusercontent.com/125357480/227272721-f7945ccb-1835-4a90-8103-fb74910977b9.jpg) works fine. The script has no output and the memory is stable. Running with a corrupted jpg such as 
[0.jpg](https://user-images.githubusercontent.com/125357480/227272969-ba005433-e869-471a-b09e-10c85d39f921.jpg) causes a rapid memory increase until the process is killed

joey-trigo:
Hi,
The leak is in CPU memory, not GPU memory. I ran the exact commands you did and can see the memory climbing with, with, for example:

`while true; do awk '{ print $24 }' /proc/$(pgrep python3)/stat; sleep 1; done`


The bug you referenced does sound very similar, is it possible that only the GPU memory is freed there and not the CPU memory?

JanuszL:
Thanks, @joey-trigo, the repro is very helpful. I believe I managed to fix the issue in https://github.com/NVIDIA/DALI/pull/4748.
Please check the nightly build that follows the merge of the mentioned commit.

joey-trigo:
Will do, thanks! That fix looks good, I'll test it as soon as it merges

joey-trigo:
Hi, tested the change and it looks like it's working. Thanks again!

PR #4748, opened 2023-03-27
Fix memory leak in host JPEG decoder

JanuszL:
- the JPEG decoder is based on TensorFlow implementation and
  uses setjmp/longjmp for exception handling. At the same time
  to avoid leaks of dynamically allocated memory the allocations
  are wrapped by unique_ptrs. However, when the pointer is created
  on stack after the context was preserved by setjmp and the
  exceptional situation occurs, the context is restored and the
  pointer is forgotten but the destructor is not called (so the
  managed memory is not freed). This PR moves the instantiation
  of the unique pointer to place before the context capture so when
  the function exits its destructor is properly called.

## Category:

## Description:
- the JPEG decoder is based on TensorFlow implementation and
  uses setjmp/longjmp for exception handling. At the same time
  to avoid leaks of dynamically allocated memory the allocations
  are wrapped by unique_ptrs. However, when the pointer is created
  on stack after the context was preserved by setjmp and the
  exceptional situation occurs, the context is restored and the
  pointer is forgotten but the destructor is not called (so the
  managed memory is not freed). This PR moves the instantiation
  of the unique pointer to place before the context capture so when
  the function exits its destructor is properly called.
- relates to https://github.com/NVIDIA/DALI/issues/4740

- dali/imgcodec/decoders/jpeg/jpeg_mem.cc

- NA

### Tests:

- [x] Existing tests apply
  - test_imgcodec.py
  - test_image.py
- [ ] New tests added
  - [ ] Python tests
  - [ ] GTests
  - [ ] Benchmark
  - [ ] Other
- [ ] N/A

### Documentation
- [ ] Documentation updated
  - [ ] Docstring
  - [ ] Doxygen
  - [ ] RST
  - [ ] Jupyter
  - [ ] Other
- [x] N/A

#### Requirements
- [x] N/A

**REQ IDs**: N/A

**JIRA TASK**: N/A

### DIFF
--- a/dali/imgcodec/decoders/jpeg/jpeg_mem.cc
+++ b/dali/imgcodec/decoders/jpeg/jpeg_mem.cc
@@ -96,8 +96,9 @@ std::unique_ptr<uint8[]> UncompressLow(const void* srcdata, FewerArgsForCompiler
   // if empty image, return
   if (datasize == 0 || srcdata == nullptr) return nullptr;
 
-  // Declare temporary buffer pointer here so that we can free on error paths
+  // Declare buffers here so that we can free on error paths
   std::unique_ptr<JSAMPLE[]> temp;
+  std::unique_ptr<uint8[]> dstdata;
   JSAMPLE *tempdata = nullptr;
 
   // Initialize libjpeg structures to have a memory source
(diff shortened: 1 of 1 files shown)


JanuszL:
!build

stiepan (approved):

mzient (approved):

PR #4752, opened 2023-03-29
Fits reader gpu

aderylo:
## Category:

## Description:
Adds simplistic implementation of a fits reader that reads to the GPU backend. 

## Additional information:
Set to draft, since there are segfaults when running tests. 

### Tests:

- [ ] Existing tests apply
- [ ] New tests added
  - [x] Python tests
  - [ ] GTests
  - [ ] Benchmark
  - [ ] Other
- [ ] N/A

### Documentation
- [ ] Documentation updated
  - [ ] Docstring
  - [ ] Doxygen
  - [ ] RST
  - [ ] Jupyter
  - [ ] Other
- [ ] N/A

#### Requirements
- [ ] N/A

**REQ IDs**: N/A

**JIRA TASK**: N/A


dali/operators/reader/loader/fits_loader_gpu.cc
```diff
@@ -0,0 +1,91 @@
...
+
+    // copy the image
+
+
+    fits::FITS_CALL(fits_read_img(current_file, header.datatype_code, 1, nelem, &nulval,
+                                  static_cast<uint8_t*>(buffer.raw_mutable_data()), &anynul,
+                                  &status));
+
+    cudaMemcpy(target.data[output_idx].raw_mutable_data(), buffer.raw_mutable_data(),
+               buffer.nbytes(), cudaMemcpyHostToDevice);
```
klecki, line 81: It is a ok starting point, but we can't put the decoding kernel here, as we have access to just one sample. We need to delay the usage of the decoding kernel till the RunImpl of the FitsReaderGpu, so we can decode whole batch at once.

I assume, we need to extend the FitsFileWrapperGPU with the shape and dtype information, and just put raw, encoded bytes in each Tensor.

Than we can decode them in RunImpl.
The question would be if we should be doing CUDA memcopy here or in the RunImpl of the reader - we should probably see the impact on the performance - again, in the RunImpl we can coalesce the copy into one.

dali/operators/reader/loader/fits_loader_gpu.h
```diff
@@ -0,0 +1,80 @@
...
+#include "dali/core/common.h"
+#include "dali/operators/reader/loader/file_loader.h"
+#include "dali/pipeline/data/types.h"
+#include "dali/util/file.h"
+#include "dali/util/fits.h"
+
+namespace dali {
+
+struct FitsFileWrapperGPU {
+  std::vector<Tensor<GPUBackend>> data;
```
klecki, line 42:
```suggestion
  std::vector<Tensor<CPUBackend>> data;
  TensorShape shape;
  DALIDataType dtype;
```

dali/operators/reader/fits_reader_gpu_op.cc
```diff
@@ -0,0 +1,38 @@
...
+  int num_outputs = ws.NumOutput();
+  int batch_size = GetCurrBatchSize();
+
+  for (int output_idx = 0; output_idx < num_outputs; output_idx++) {
+    auto &output = ws.Output<GPUBackend>(output_idx);
+    for (int sample_id = 0; sample_id < batch_size; ++sample_id) {
+      auto &sample = GetSample(sample_id);
+
+      cudaMemcpy(output.raw_mutable_tensor(sample_id), sample.data[output_idx].raw_data(),
+                 sample.data[output_idx].nbytes(), cudaMemcpyDeviceToDevice);
```
klecki, line 31: 1. Collect all samples
2. cudaMemcpy them to the GPU (use the ToContiguousGpu for example).
3. call the kernel.

dali/operators/reader/fits_reader_gpu_op.cu
```diff
@@ -0,0 +1,384 @@
...
+            tile_size_cuda, sample.header[output_idx].bytepix, sample.header[output_idx].blocksize,
+            tiles, maxtilelen);
+
+        cudaFree(undecoded_data_cuda);
+        cudaFree(tile_size_cuda);
+        cudaFree(tile_offset_cuda);
+
+      } else {
+        cudaMemcpyAsync(output.raw_mutable_tensor(sample_id), sample.data[output_idx].raw_data(),
+                        sample.data[output_idx].nbytes(), cudaMemcpyHostToDevice);
```
JanuszL, line 376: You should use a proper operator stream.

dali/operators/reader/fits_reader_gpu_op.cu
```diff
@@ -0,0 +1,384 @@
...
+        int blockSize = 256;
+        int numBlocks = (int)(tiles + blockSize - 1) / blockSize;
+
+        DALI_ENFORCE(false, "got here!");
+        rice_decompress<<<numBlocks, blockSize>>>(
+            undecoded_data_cuda, output.raw_mutable_tensor(sample_id), tile_offset_cuda,
+            tile_size_cuda, sample.header[output_idx].bytepix, sample.header[output_idx].blocksize,
+            tiles, maxtilelen);
+
+        cudaFree(undecoded_data_cuda);
```
JanuszL, line 370: For tmp memory please use scratchpad - https://github.com/NVIDIA/DALI/blob/main/dali/operators/image/remap/remap.cu#L69 (declaration), https://github.com/NVIDIA/DALI/blob/main/dali/operators/image/remap/remap.cuh#L74 (allocates memory on the GPU  and copies data from the CPU to it using given stream).
Scratchpad should live until to the last call touching given memory. So:
```
dali::kernels::DynamicScratchpad ds;
auto stream = ws.stream();
auto tile_offset_cuda = std::get<0>(ds.ToContiguousGPU(stream, sample.tile_offset[output_idx]));
auto tile_offset_cuda = std::get<0>(ds.ToContiguousGPU(stream, sample.tile_size[output_idx]));
auto undecoded_data_cuda= std::get<0>(ds.ToContiguousGPU(stream, sample.data[output_idx]));
```

dali/operators/reader/fits_reader_gpu_op.cu
```diff
@@ -0,0 +1,384 @@
...
+                maxtilelen = sample.header[output_idx].maxtilelen,
+                zbitpix = sample.header[output_idx].zbitpix;
+
+        if (zbitpix == 8) {
+          cudaMalloc(&decoded_data_cuda, tiles * maxtilelen * sizeof(char));
+        } else if (zbitpix == 16) {
+          cudaMalloc(&decoded_data_cuda, tiles * maxtilelen * sizeof(short));
+        } else {
+          cudaMalloc(&decoded_data_cuda, tiles * maxtilelen * sizeof(int));
+        }
```
JanuszL, line 345: Is it used anywhere?

dali/operators/reader/fits_reader_gpu_op.cu
```diff
@@ -0,0 +1,384 @@
...
+      5, 5, 5, 6, 6, 6, 6, 6, 6, 6, 6, 6, 6, 6, 6, 6, 6, 6, 6, 6, 6, 6, 6, 6, 6, 6, 6, 6, 6,
+      6, 6, 6, 6, 6, 6, 7, 7, 7, 7, 7, 7, 7, 7, 7, 7, 7, 7, 7, 7, 7, 7, 7, 7, 7, 7, 7, 7, 7,
+      7, 7, 7, 7, 7, 7, 7, 7, 7, 7, 7, 7, 7, 7, 7, 7, 7, 7, 7, 7, 7, 7, 7, 7, 7, 7, 7, 7, 7,
+      7, 7, 7, 7, 7, 7, 7, 7, 7, 7, 7, 7, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8,
+      8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8,
+      8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8,
+      8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8,
+      8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8};
+
+  if (bytepix == 1) {
```
JanuszL, line 38: What is the difference in the flow for `bytepix==1`, `bytepix==2` and `else`? Can the common part be extracted?

dali/util/fits.cc
```diff
@@ -115,11 +128,131 @@ void ParseHeader(HeaderData &parsed_header, fitsfile *src) {
...
+                            &charnull, raw_data[size].data(), nullptr, status);
+
+              ++size;
+              sum_nelemll += nelemll;
+            }
+          }
+        }
+      }
+    }
+  }
```
JanuszL, line 244: Can this be converted to two nested loop and ask the compiler to unroll it (or just add a comment why it is unrolled by hand)?

dali/operators/reader/fits_reader_gpu_op.cu
klecki: I see that it differs mostly in the types used and in one block of code, I suggest extracting a template to handle it, and maybe extract a common function to handle the lastpix & compressed_tile calculation at the beginning.

dali/util/fits.cc
mskwr: I added explanation comment. I do not think it can be converted to two nested loop and still support decompression up to six dimensions of FITS image. Does it make sense to use here #pragma GCC unroll N?

dali/operators/reader/fits_reader_gpu_op.cu
mskwr: So adding bytepix as a template argument still did not resolve excessive casting which occurred in a kernel, so for now I split it into three kernels - each representing different type. Now type casting occurs only during scaling.

mskwr:
Excuse my force-pushing, it was just to edit the message of some commit which I forgot to sign.

dali/util/fits.cc
JanuszL: @klecki 
Do you think that something like could work here or the compiler wouldn't inline it?
```
template <unsigned level>
inline void foo(tfpixel, tilesize, minvalue, rowdim, naxis, max_level=level) {
    for (i = ftile[level]; i <= ltile[level]; ++i) {
        tfpixel[level] = (i - 1) * tilesize[level] + 1;
        tlpixel[level] = minvalue(tfpixel[level] + tilesize[level] - 1, naxis[level]);
        if (level == max_level) {
            thistilesize[level] = tlpixel[level] - tfpixel[level] + 1;
            offset[level] = (i - 1) * rowdim[level];
        } else {
            thistilesize[level] = thistilesize[level + 1] * (tlpixel[level] - tfpixel[level] + 1);
            offset[level] = (i - 1) * rowdim[level] + offset[level+1];
        }
        foo<level-1>(tfpixel, tilesize, minvalue, rowdim, naxis, max_level);
    }    
}

template <>
inline void foo<0>(tfpixel, tilesize, minvalue, rowdim, naxis, max_level=0) {
    for (i = ftile[level]; i <= ltile[level]; ++i) {
        irow = i + offset[1];

        LONGLONG nelemll = 0, noffset = 0;
        ffgdesll(fptr, (fptr->Fptr)->cn_compressed, irow, &nelemll, &noffset, status);

        tile_offset[size] = sum_nelemll;
        tile_size[size] = (long)thistilesize[0];

        unsigned char charnull = 0;
        raw_data[size].resize(nelemll / sizeof(unsigned char));
        fits_read_col(fptr, TBYTE, (fptr->Fptr)->cn_compressed, irow, 1, (long)nelemll,
                    &charnull, raw_data[size].data(), nullptr, status);

        ++size;
        sum_nelemll += nelemll;
    }
}
foo<5>(tfpixel, tilesize, minvalue, rowdim, naxis);
```

dali/operators/reader/fits_reader_gpu_op.cu
```diff
@@ -0,0 +1,417 @@
...
+
+    if (bscale != 1. || bzero != 0.) {
+      for (int j = 0; j < tile_size[tile]; ++j) {
+        uncompressed_data[beg + j] = (char)((char)uncompressed_data[beg + j] * bscale + bzero);
+      }
+    }
+  }
+}
+
+__global__ void rice_decompress_short(unsigned char *compressed_data,
```
klecki, line 136: Replace the other kernels by the invocation of the template defined above.

dali/util/fits.cc
klecki: Sure, that should work.

klecki:
!build

klecki:
Please ensure that there are no linting issues.
To run the linter call `make lint` in the build directory.

dali/operators/reader/fits_reader_gpu_op.cu
```diff
@@ -0,0 +1,417 @@
...
+    7, 7, 7, 7, 7, 7, 7, 7, 7, 7, 7, 7, 7, 7, 7, 7, 7, 7, 7, 7, 7, 7, 7, 7, 7, 7, 7, 7, 7, 7, 7, 7,
+    8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8,
+    8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8,
+    8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8,
+    8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8};
+
+__global__ void rice_decompress_byte(unsigned char *compressed_data,
+                                     unsigned char *uncompressed_data, const long *tile_offset,
+                                     const long *tile_size, int blocksize, long tiles,
+                                     long maxtilelen, double bscale, double bzero) {
```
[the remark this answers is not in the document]
mskwr: It looks like the function argument cannot be of type unsigned T (where T is a typename) and it leads to incorrect syntax. I changed it so that T is already unsigned.

dali/operators/reader/fits_reader_gpu_op.cu
```diff
@@ -0,0 +1,417 @@
...
+    int nbits, nzero, fs;
+    unsigned int b, diff, lastpix;
+    int fsmax, fsbits, bbits;
+    long beg;
+
+    fsbits = 3;
+    fsmax = 6;
+    bbits = 1 << fsbits;
+
+    lastpix = compressed_tile[0];
```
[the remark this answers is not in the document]
mskwr: Done with a little bit of modification
```template <typename T>
__device__ unsigned int read_lastpix(unsigned char *compressed_tile) {
  unsigned int lastpix = 0;
  for (int i = 0; i < sizeof(T); ++i) {
    unsigned char bytevalue = compressed_tile[i];
    lastpix = lastpix | (bytevalue << ((sizeof(T) - i - 1) << 3));
  }
  return lastpix;
}

File: dali/operators/reader/fits_reader_gpu_op.cu [hunk 12]
```diff
@@ -0,0 +1,417 @@
...
+          }
+          uncompressed_data[beg + i] = diff + lastpix;
+          lastpix = uncompressed_data[beg + i];
+        }
+      }
+    }
+
+    if (bscale != 1. || bzero != 0.) {
+      for (int j = 0; j < tile_size[tile]; ++j) {
+        uncompressed_data[beg + j] = (char)((char)uncompressed_data[beg + j] * bscale + bzero);
```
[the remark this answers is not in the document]
mskwr: As T is already unsigned now, I used std::make_signed_t here.
```
if (bscale != 1. || bzero != 0.) {
    for (int j = 0; j < tile_size[tile]; ++j) {
        uncompressed_data[beg + j] = static_cast<std::make_signed_t<T>>(
            static_cast<std::make_signed_t<T>>(uncompressed_data[beg + j]) * bscale + bzero);
    }
}

**Comment by klecki** [comment]
!build

File: dali/operators/reader/fits_reader_gpu_op.cu [hunk 13]
```diff
@@ -0,0 +1,417 @@
...
+    unsigned char *compressed_tile = compressed_data + tile_offset[tile];
+    int i, imax;
+    int k;
+    int nbits, nzero, fs;
+    unsigned int b, diff, lastpix;
+    int fsmax, fsbits, bbits;
+    long beg;
+
+    fsbits = 3;
+    fsmax = 6;
```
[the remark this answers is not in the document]
mskwr: Making them dependent would lead to some unpleasant calculations (especially for fsmax) so I moved them into arrays.
`const int fsbits[3] = {3, 4, 5};`
`const int fsmax[3] = {6, 14, 25};`

dali/operators/reader/fits_reader_gpu_op.cu
```diff
@@ -0,0 +1,216 @@
...
+
+    for (int sample_id = 0; sample_id < batch_size; ++sample_id) {
+      auto &sample = GetSample(sample_id);
+      sample_list_cpu.SetSample(sample_id, sample.data[output_idx]);
+    }
+
+
+    if (compressed) {
+      TensorList<GPUBackend> sample_list_gpu;
+      sample_list_gpu.Copy(sample_list_cpu, ws.stream());
```
klecki, line 177: Better to use scratchpad here.

dali/operators/reader/fits_reader_gpu_op.cu
```diff
@@ -0,0 +1,216 @@
...
+
+      for (int sample_id = 0; sample_id < batch_size; ++sample_id) {
+        int64_t *tile_offset_cuda, *tile_size_cuda;
+        auto &sample = GetSample(sample_id);
+        auto header = sample.header[output_idx];
+        int blockSize = 256, numBlocks = static_cast<int>(header.rows + blockSize - 1) / blockSize;
+
+        std::tie(tile_offset_cuda, tile_size_cuda) = s.ToContiguousGPU(
+            ws.stream(), sample.tile_offset[output_idx], sample.tile_size[output_idx]);
```
klecki, line 186: Let's try to unify this copy into one.

dali/operators/reader/loader/fits_loader_gpu.cc
```diff
@@ -0,0 +1,104 @@
...
+      memcpy(static_cast<uint8_t*>(target.data[output_idx].raw_mutable_data()), raw_data.data(),
+             raw_data.size());
+    } else {
+      target.data[output_idx].Resize(header.shape, header.type());
+
+      // copy the image to host memory
+      fits::FITS_CALL(fits_read_img(
+          current_file, header.datatype_code, 1, nelem, &nulval,
+          static_cast<uint8_t*>(target.data[output_idx].raw_mutable_data()), &anynul, &status));
+    }
```
klecki, line 93: Maybe let's try to extract this as a helper function and unify the implementation of parsing the header, etc with the LoaderCPU.

dali/operators/reader/fits_reader_gpu_op.cu
```diff
@@ -0,0 +1,216 @@
...
+    sample_list_cpu.SetSize(batch_size);
+
+    for (int sample_id = 0; sample_id < batch_size; ++sample_id) {
+      auto &sample = GetSample(sample_id);
+      sample_list_cpu.SetSample(sample_id, sample.data[output_idx]);
+    }
+
+
+    if (compressed) {
+      TensorList<GPUBackend> sample_list_gpu;
```
klecki, line 176:
```suggestion
      TensorList<GPUBackend> sample_list_gpu;
      sample_list_gpu.set_order(ws.stream());
```
I think with that change we can stick with the implementation using the Copy from one TL to another.
It is possible to use scratchpad, but the copy would still be quite manual, I think we can live with this version for now, unless this poses a performance problem, we can take a look into first coalescing it into one buffer.
Let's stick with this version (+ the setting of the order for the temporary buffer - it will ensure that we release the memory on the correct stream).

dali/operators/reader/fits_reader_gpu_op.cu
aderylo: Ok, ok, definitively I will try to implement that although I also think that having benchmarks first might be a good idea.

klecki:
!build

klecki:
!build

klecki:
!build

dali/operators/reader/loader/fits_loader_gpu.cc
aderylo: Done!

dali/operators/reader/fits_reader_gpu_op.cu
klecki: tbd in follow-up if time allows.

klecki:
!build

klecki (commented):
Few minor comments, otherwise looks ok.

dali/util/fits.cc
```diff
@@ -115,11 +130,134 @@ void ParseHeader(HeaderData &parsed_header, fitsfile *src) {
...
+    FITS_CALL(fits_get_num_rows(src, &parsed_header.rows, &status)); /*get NROW value */
+    parsed_header.bscale = (src->Fptr)->cn_bscale;
+    parsed_header.bzero = (src->Fptr)->cn_bzero;
+    parsed_header.bytepix = (src->Fptr)->rice_bytepix;
+    parsed_header.zbitpix = (src->Fptr)->zbitpix;
+    parsed_header.blocksize = (src->Fptr)->rice_blocksize;
+    parsed_header.maxtilelen = (src->Fptr)->maxtilelen;
+    parsed_header.tile_sizes = GetTileSizes(src, n_dims);
+    parsed_header.tiles = std::accumulate(parsed_header.tile_sizes.begin(),
+                                          parsed_header.tile_sizes.end(), 1, std::multiplies<>());
```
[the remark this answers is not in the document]
mskwr: Generally if the keywords are not present, the following values are specified by CFITSIO. Some of them get some default values (usually 0 or 1) and some of them are undefined.

dali/operators/reader/fits_reader_gpu_op.cu
```diff
@@ -0,0 +1,217 @@
...
```
JanuszL, line 176: @mzient - do you think it will work or as soon as `sample_list_gpu` goes out of scope we can get in troubles?

JanuszL: Or setting the order is sufficient?

dali/operators/reader/fits_reader_op.h
```diff
@@ -53,34 +53,37 @@ class FitsReader : public DataReader<Backend, Target> {
...
-      ndims[output_idx] = sample_0.data[output_idx].shape().sample_dim();
-      output_dtypes[output_idx] = sample_0.data[output_idx].type();
+      ndims[output_idx] = sample_0.header[output_idx].shape.sample_dim();
+      output_dtypes[output_idx] = sample_0.header[output_idx].type();
       output_desc[output_idx].shape = TensorListShape<>(num_samples, ndims[output_idx]);
     }
 
     for (int sample_idx = 0; sample_idx < num_samples; sample_idx++) {
       const auto& sample = GetSample(sample_idx);
+      // here we don't validate current data dims and type, only the final dimentions and type
```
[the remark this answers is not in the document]
mskwr: Done.

dali/util/fits.cc
```diff
@@ -115,11 +130,134 @@ void ParseHeader(HeaderData &parsed_header, fitsfile *src) {
...
 
+template <unsigned level>
+inline void extract_data(fitsfile* fptr, std::vector<std::vector<uint8_t>>& raw_data,
+                         std::vector<int64_t>& tile_offset, std::vector<int64_t>& tile_size,
+                         int64* ftile, int64* ltile, int64* tilesize, int64* thistilesize,
+                         int64* rowdim, int64* naxis, int64* offset, int* size, int64* sum_nelemll,
+                         int* status, unsigned max_level = level) {
+  for (int i = ftile[level]; i <= ltile[level]; ++i) {
+    int64 tfpixel = (i - 1) * tilesize[level] + 1;
+    int64 tlpixel = minvalue(tfpixel + tilesize[level] - 1, naxis[level]);
```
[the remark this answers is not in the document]
mskwr: Done.

dali/util/fits.cc
```diff
@@ -115,11 +130,134 @@ void ParseHeader(HeaderData &parsed_header, fitsfile *src) {
...
+
+int extract_undecoded_data(fitsfile* fptr, std::vector<uint8_t>& data,
+                           std::vector<int64_t>& tile_offset, std::vector<int64_t>& tile_size,
+                           int64 rows, int* status) {
+  std::vector<std::vector<uint8_t>> raw_data;
+  raw_data.resize(rows);
+  tile_offset.resize(rows + 1);
+  tile_size.resize(rows);
+
+  LONGLONG fpixel[MAX_COMPRESS_DIM], lpixel[MAX_COMPRESS_DIM];
```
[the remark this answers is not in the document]
mskwr: "f" stands for "first" and "l" stands for "last". Done.

dali/util/fits.cc
```diff
@@ -115,11 +130,134 @@ void ParseHeader(HeaderData &parsed_header, fitsfile *src) {
...
+  }
+
+  int64 naxis[MAX_COMPRESS_DIM] = {1, 1, 1, 1, 1, 1};
+  int64 tiledim[MAX_COMPRESS_DIM] = {1, 1, 1, 1, 1, 1};
+  int64 tilesize[MAX_COMPRESS_DIM] = {1, 1, 1, 1, 1, 1};
+  int64 ftile[MAX_COMPRESS_DIM] = {1, 1, 1, 1, 1, 1};
+  int64 ltile[MAX_COMPRESS_DIM] = {1, 1, 1, 1, 1, 1};
+  int64 rowdim[MAX_COMPRESS_DIM] = {1, 1, 1, 1, 1, 1};
+  int64 offset[MAX_COMPRESS_DIM], thistilesize[MAX_COMPRESS_DIM];
+  int64 mfpixel[MAX_COMPRESS_DIM], mlpixel[MAX_COMPRESS_DIM];
```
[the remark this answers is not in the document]
mskwr: Mostly yes. Done.

dali/operators/reader/loader/fits_loader.h
```diff
@@ -64,17 +61,86 @@ class FitsLoader : public FileLoader<CPUBackend, FitsFileWrapper> {
...
+
+    for (size_t output_idx = 0; output_idx < hdu_indices_.size(); output_idx++) {
+      // move to appropiate hdu
+      fits::FITS_CALL(fits_movabs_hdu(current_file, hdu_indices_[output_idx], NULL, &status));
+
+      // read the header
+      fits::HeaderData header;
+      try {
+        fits::ParseHeader(header, current_file);
+        target.header[output_idx] = header;
```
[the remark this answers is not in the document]
mskwr: Some tests report an error after making this change. I think the reason may be a case in which `target.header[output_idx]` is not initialized before the reference is created - then we could end up with undefined behavior. The first solution to this problem would be to keep the currently working code. The second, initialization, for example:
```suggestion
      target.header[output_idx] = {};
      fits::HeaderData& header = target.header[output_idx];
      try {
        fits::ParseHeader(header, current_file);
```

klecki (approved):

klecki:
!build

dali/operators/reader/fits_reader_gpu_op.cu
```diff
@@ -0,0 +1,217 @@
...
+        auto header = sample.header[output_idx];
+        int blockSize = 256, numBlocks = static_cast<int>(header.rows + blockSize - 1) / blockSize;
+
+        std::tie(tile_offset_cuda, tile_size_cuda) = s.ToContiguousGPU(
+            ws.stream(), sample.tile_offset[output_idx], sample.tile_size[output_idx]);
+
+        if (sample.header[output_idx].bytepix == 1) {
+          rice_decompress<uint8_t><<<numBlocks, blockSize, 0, ws.stream()>>>(
+              static_cast<uint8_t *>(sample_list_gpu.raw_mutable_tensor(sample_id)),
+              static_cast<uint8_t *>(output.raw_mutable_tensor(sample_id)), tile_offset_cuda,
```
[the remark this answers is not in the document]
mskwr: Some tests also report an error after making this change, although I am still during investigation why is it happening.
```
ERROR: test_fits.test_reading_compressed:2
1, <class 'numpy.int8'>, 1, 'gpu'
  File "/home/mskwarek/DALI/dali/test/python/reader/test_fits.py", line 170, in test_reading_compressed
    _testimpl_types_and_shapes(np_rng, device, shapes, num_outputs, dtype, batch_size, num_threads,
  File "/home/mskwarek/DALI/dali/test/python/reader/test_fits.py", line 132, in _testimpl_types_and_shapes
    pipe_out = pipe.run()
  File "/home/mskwarek/.local/lib/python3.10/site-packages/nvidia/dali/pipeline.py", line 1110, in run
    return self.outputs()
  File "/home/mskwarek/.local/lib/python3.10/site-packages/nvidia/dali/pipeline.py", line 953, in outputs
    return self._outputs()
  File "/home/mskwarek/.local/lib/python3.10/site-packages/nvidia/dali/pipeline.py", line 1037, in _outputs
    return self._pipe.Outputs()
RuntimeError: Critical error in pipeline:
Error when executing GPU operator experimental__readers__Fits encountered:
[/home/mskwarek/DALI/dali/pipeline/data/buffer.h:196] Assert on "type_.id() == TypeTable::GetTypeId<T>()" failed: Calling type does not match buffer data type, requested type: uint8 current buffer type: int8. To set type for the Buffer use 'set_type<T>()' or Resize(shape, type) first.
Stacktrace (24 entries):
[frame 0]: /home/mskwarek/DALI/build/dali/python/nvidia/dali/libdali_operators.so(+0x149227c) [0x7f79ac09227c]
[frame 1]: /home/mskwarek/DALI/build/dali/python/nvidia/dali/libdali_operators.so(unsigned char* dali::Buffer<dali::GPUBackend>::mutable_data<unsigned char>()+0x18a) [0x7f79ac1c58b6]
[frame 2]: /home/mskwarek/DALI/build/dali/python/nvidia/dali/libdali_operators.so(unsigned char* dali::TensorList<dali::GPUBackend>::mutable_tensor<unsigned char>(int)+0x36) [0x7f79ac1c56f2]
[frame 3]: /home/mskwarek/DALI/build/dali/python/nvidia/dali/libdali_operators.so(+0x68f074d) [0x7f79b14f074d]
[frame 4]: /home/mskwarek/DALI/build/dali/python/nvidia/dali/libdali_operators.so(+0x14c342c) [0x7f79ac0c342c]
[frame 5]: /home/mskwarek/DALI/build/dali/python/nvidia/dali/libdali_operators.so(+0x68faaaf) [0x7f79b14faaaf]
[frame 6]: /home/mskwarek/DALI/build/dali/python/nvidia/dali/libdali_operators.so(+0x68f9ba5) [0x7f79b14f9ba5]
[frame 7]: /home/mskwarek/DALI/build/dali/python/nvidia/dali/libdali.so(dali::Executor<dali::AOT_WS_Policy<dali::UniformQueuePolicy>, dali::UniformQueuePolicy>::RunHelper(dali::OpNode&, dali::Workspace&, unsigned long)+0x124e) [0x7f79c78afc1a]
[frame 8]: /home/mskwarek/DALI/build/dali/python/nvidia/dali/libdali.so(dali::Executor<dali::AOT_WS_Policy<dali::UniformQueuePolicy>, dali::UniformQueuePolicy>::RunGPUImpl(unsigned long)+0x38d) [0x7f79c78ac5af]
[frame 9]: /home/mskwarek/DALI/build/dali/python/nvidia/dali/libdali.so(dali::Executor<dali::AOT_WS_Policy<dali::UniformQueuePolicy>, dali::UniformQueuePolicy>::RunGPU()+0x49) [0x7f79c78aa197]
[frame 10]: /home/mskwarek/DALI/build/dali/python/nvidia/dali/libdali.so(+0x265ea1) [0x7f79c7865ea1]
[frame 11]: /home/mskwarek/DALI/build/dali/python/nvidia/dali/libdali.so(+0x2669c8) [0x7f79c78669c8]
[frame 12]: /home/mskwarek/DALI/build/dali/python/nvidia/dali/libdali.so(+0x266762) [0x7f79c7866762]
[frame 13]: /home/mskwarek/DALI/build/dali/python/nvidia/dali/libdali.so(+0x2663a8) [0x7f79c78663a8]
[frame 14]: /home/mskwarek/DALI/build/dali/python/nvidia/dali/libdali.so(std::function<void ()>::operator()() const+0x36) [0x7f79c792666e]
[frame 15]: /home/mskwarek/DALI/build/dali/python/nvidia/dali/libdali.so(+0x32597a) [0x7f79c792597a]
[frame 16]: /home/mskwarek/DALI/build/dali/python/nvidia/dali/libdali.so(+0x335101) [0x7f79c7935101]
[frame 17]: /home/mskwarek/DALI/build/dali/python/nvidia/dali/libdali.so(+0x334b71) [0x7f79c7934b71]
[frame 18]: /home/mskwarek/DALI/build/dali/python/nvidia/dali/libdali.so(void std::thread::_Invoker<std::tuple<void (dali::WorkerThread::*)(int, bool, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&), dali::WorkerThread*, int, bool, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > > >::_M_invoke<0ul, 1ul, 2ul, 3ul, 4ul>(std::_Index_tuple<0ul, 1ul, 2ul, 3ul, 4ul>)+0x9b) [0x7f79c7933e71]
[frame 19]: /home/mskwarek/DALI/build/dali/python/nvidia/dali/libdali.so(std::thread::_Invoker<std::tuple<void (dali::WorkerThread::*)(int, bool, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&), dali::WorkerThread*, int, bool, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > > >::operator()()+0x1c) [0x7f79c792fc1a]
[frame 20]: /home/mskwarek/DALI/build/dali/python/nvidia/dali/libdali.so(std::thread::_State_impl<std::thread::_Invoker<std::tuple<void (dali::WorkerThread::*)(int, bool, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&), dali::WorkerThread*, int, bool, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > > > >::_M_run()+0x20) [0x7f79c792f4d2]
[frame 21]: /lib/x86_64-linux-gnu/libstdc++.so.6(+0xdc2b3) [0x7f79c72dc2b3]
[frame 22]: /lib/x86_64-linux-gnu/libc.so.6(+0x94b43) [0x7f79c9094b43]
[frame 23]: /lib/x86_64-linux-gnu/libc.so.6(+0x126a00) [0x7f79c9126a00]

Current pipeline object is no longer valid.
```

dali/operators/reader/fits_reader_gpu_op.cu
mzient: As I understand, it's a termporary buffer. Setting the order should be enough from the correctness perspective. Performancewise, using a full-blown TensorList as a temporary storage for raw data may seem excessive, but at least the code, as it is, has the benefit of simplicity.

dali/operators/reader/fits_reader_gpu_op.cu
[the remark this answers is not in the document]
mzient: Is the output always uint8? The proposed code would require that.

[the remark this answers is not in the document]
mskwr: Oh yes, this is the reason. Output is not always uint8, uint16 or uint32 due to scaling at the end of kernel.

dali/operators/reader/fits_reader_gpu_op.cu
mzient: There's a problem with the source (host) TensorList, however. If it's pinned _and_ it uses host order, then its contents may be clobbered before this H2D copy finishes.

JanuszL: Maybe we can:
```
auto out = s.ToContiguousGPU(make_span(sample_list_cpu));
TensorListView<StorageGPU, uint8_t> sample_list_gpu(out , sample_list_cpu.shape());
``` ?

Comment (klecki) [reply dali/operators/reader/fits_reader_gpu_op.cu:175]: Does ToContiguousGPU accept span as an argument? I don't see any overload that would work like that.

The simplest fix here, would be to add yet another copy, this is what ToContiguousGPU does internally, so

```
TensorList<CPU> samples_tmp;
samples_tmp.SetContiguity(BatchContiguity::Contiguous);
samples_tmp.Copy(sample_list_cpu);
sample_list_gpu.Copy(sample_list_cpu);
```
Honestly, I don't know why we don't have any better API for that, unless I'm missing something, but using scratchpad or TLV would require you to use manual copy.

File: dali/operators/reader/fits_reader_gpu_op.cu (hunk 25 above)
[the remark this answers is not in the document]
Comment (klecki) [reply dali/operators/reader/fits_reader_gpu_op.cu:191]: The buffer at the operator output should have the correct type provided, does this mean we are using incorrect one?

File: dali/util/fits.cc [hunk 26]
```diff
@@ -115,11 +131,142 @@ void ParseHeader(HeaderData &parsed_header, fitsfile *src) {
...
+                  &charnull, raw_data[*size].data(), nullptr, status);
+
+    ++(*size);
+    *sum_nelemll += nelemll;
+  }
+}
+
+int extract_undecoded_data(fitsfile* fptr, std::vector<uint8_t>& data,
+                           std::vector<int64_t>& tile_offset, std::vector<int64_t>& tile_size,
+                           int64 rows, int* status) {
```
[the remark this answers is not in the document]
aderylo: Sure! There might be a bit of a problem with getting grand truths for the tests, since as far as I am concerned, cfitsio doesn't support reading raw uncompressed data without doing compression and scaling first.

dali/operators/reader/CMakeLists.txt
```diff
@@ -25,6 +25,7 @@ list(APPEND DALI_OPERATOR_SRCS "${CMAKE_CURRENT_SOURCE_DIR}/numpy_reader_op.cc")
 
 if(BUILD_CFITSIO)
   list(APPEND DALI_OPERATOR_SRCS "${CMAKE_CURRENT_SOURCE_DIR}/fits_reader_op.cc")
+  list(APPEND DALI_OPERATOR_SRCS "${CMAKE_CURRENT_SOURCE_DIR}/fits_reader_gpu_op.cu")
```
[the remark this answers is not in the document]
aderylo: With regards to naming, we followed the same convention as numpy_reader.  Although, I agree that op suffix seems redundant.

dali/operators/reader/fits_reader_gpu_op.cu
```diff
@@ -0,0 +1,216 @@
+// Copyright (c) 2022, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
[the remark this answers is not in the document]
aderylo: done

dali/operators/reader/fits_reader_gpu_op.h
```diff
@@ -0,0 +1,43 @@
+// Copyright (c) 2020-2022, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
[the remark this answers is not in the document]
aderylo: done

dali/operators/reader/loader/fits_loader.h
```diff
@@ -64,16 +60,83 @@ class FitsLoader : public FileLoader<CPUBackend, FitsFileWrapper> {
...
+      } catch (const std::runtime_error& e) {
+        DALI_FAIL(e.what() + ". File: " + filename);
```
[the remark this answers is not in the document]
aderylo: done

klecki:
!build

dali/operators/reader/fits_reader_gpu_op.cu
```diff
@@ -0,0 +1,216 @@
...
+    unsigned char bytevalue = compressed_tile[i];
+    lastpix = lastpix | (bytevalue << ((sizeof(T) - i - 1) << 3));
+  }
+  return lastpix;
+}
+
+template <typename T>
+__global__ void rice_decompress(unsigned char *compressed_data, T *uncompressed_data,
+                                const int64 *tile_offset, const int64 *tile_size, int blocksize,
+                                int64 tiles, int64 maxtilelen, double bscale, double bzero) {
```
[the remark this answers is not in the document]
klecki: I think with the test that checks the extraction, it will be enough to test the decoding on the operator level.

dali/util/fits_test.cc
```diff
@@ -0,0 +1,110 @@
...
+namespace fits {
+
+namespace {
+
+template <typename T>
+vector<T> ReadVector(InputStream *src) {
+  vector<T> data;
+  data.resize(src->Size() / sizeof(T));
+  auto ret = src->Read(reinterpret_cast<uint8_t *>(data.data()), src->Size());
+  // static_cast<uint8_t *>(data.data());
```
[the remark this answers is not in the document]
aderylo: No, I've cleaned it up in the latest commit.

klecki:
!build

PR #4765, opened 2023-04-03
Adjust docs config for newer Sphinx version

klecki:
## Description:
Force MathJax to render SVG by default, as the version 3 that is used by newer Sphinx, renders vertical and horizontal lines incorrectly in Firefox.

The setting is controled by import path - the import uses the same cdn that is used in default configuration.

Docs

N/A

### Tests:

- [ ] Existing tests apply
- [ ] New tests added
  - [ ] Python tests
  - [ ] GTests
  - [ ] Benchmark
  - [ ] Other
- [x] N/A

### Documentation
- [x] Documentation updated
  - [ ] Docstring
  - [ ] Doxygen
  - [ ] RST
  - [ ] Jupyter
  - [ ] Other
- [ ] N/A

#### Requirements
- [ ] N/A

**REQ IDs**: N/A

**JIRA TASK**: N/A

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -150,7 +150,7 @@ autodoc_mock_imports = ['paddle', 'torch', 'torchvision']
 
 # -- Options for MathJax -----------------------------------------------------
 
-# Configure the MathJax to use SVG output as a default instead of the
+# Configure the MathJax to use SVG rendering as a default instead of the CHTML one.
 # Apparently, this is how MathJax is supposed to be configured based on their converter
 # https://mathjax.github.io/MathJax-demos-web/convert-configuration/convert-configuration.html
 # The import is crucial, in version two it was apparently enough to set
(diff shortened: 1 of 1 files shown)


docs/conf.py
```diff
@@ -147,6 +147,29 @@
 # so we do not mock tensorflow so we do not need to extend the logic there.
 autodoc_mock_imports = ['paddle', 'torch', 'torchvision']
 
+
+# -- Options for MathJax -----------------------------------------------------
+
+# Configure the MathJax to use SVG output as a default instead of the
```
mzient, line 153: Instead of what?

klecki: hmm, removed part of the comment by mistake when editing.
```suggestion
# Configure the MathJax to use SVG rendering as a default instead of the CHTML one.
```

JanuszL:
Can you provide before/after screenshot?

PR #4800, opened 2023-04-21
Make DALI array_interface memory writable

JanuszL:
- marking DALI memory shared by array_interface read-only prevents it
  from sharing directly with Torch and possibly other libraries.
  This PR makes it writable.
- memory can be shared anyway but it requires an intermediate step
  through cupy which makes the memory writable

## Category:
*Other**/**Breaking change**(I hope not)

## Description:
- marking DALI memory shared by array_interface read-only prevents it
  from sharing directly with Torch and possibly other libraries.
  This PR makes it writable.
- memory can be shared anyway but it requires an intermediate step
  through cupy which makes the memory writable

- tensor cuda_array_interface and array_interface

- NA

### Tests:

- [x] Existing tests apply
  - adjusted test_array_interface_tensor_cpu  and test_cuda_array_interface_tensor_gpu
- [ ] New tests added
  - [ ] Python tests
  - [ ] GTests
  - [ ] Benchmark
  - [ ] Other
- [ ] N/A

### Documentation
- [ ] Documentation updated
  - [ ] Docstring
  - [ ] Doxygen
  - [ ] RST
  - [ ] Jupyter
  - [ ] Other
- [ ] N/A

#### Requirements
- [x] N/A

**REQ IDs**: N/A

**JIRA TASK**: N/A

### DIFF
--- a/dali/test/python/test_backend_impl_torch_dlpack.py
+++ b/dali/test/python/test_backend_impl_torch_dlpack.py
@@ -63,7 +63,7 @@ def test_dlpack_tensor_list_gpu_to_cpu():
 
 
 def check_dlpack_types_gpu(t):
-    arr = torch.tensor([[-0.39, 1.5], [-1.5, 0.33]], device="cuda", dtype=t)
+    arr = torch.tensor([[0.39, 1.5], [1.5, 0.33]], device="cuda", dtype=t)
     tensor = TensorGPU(to_dlpack(arr), "NHWC")
     dali_torch_tensor = convert_to_torch(tensor, device=arr.device, dtype=arr.dtype,
                                          size=tensor.shape())
@@ -149,7 +149,7 @@ def test_tensor_list_gpu_from_dlpack():
 
 
 def check_dlpack_types_cpu(t):
-    arr = torch.tensor([[-0.39, 1.5], [-1.5, 0.33]], device="cpu", dtype=t)
+    arr = torch.tensor([[0.39, 1.5], [1.5, 0.33]], device="cpu", dtype=t)
     tensor = TensorCPU(to_dlpack(arr), "NHWC")
     dali_torch_tensor = convert_to_torch(tensor, device=arr.device, dtype=arr.dtype,
                                          size=tensor.shape())
--- a/dali/python/backend_impl.cc
+++ b/dali/python/backend_impl.cc
@@ -121,7 +121,8 @@ py::dict ArrayInterfaceRepr(Tensor<Backend> &t) {
   d["shape"] = py::tuple(py_shape<Backend>(t));
   // tuple of (raw_data_pointer, if_data_is_read_only)
   tup[0] = py::reinterpret_borrow<py::object>(PyLong_FromVoidPtr(t.raw_mutable_data()));
-  tup[1] = true;
+  // if we make it readonly, it prevents us from sharing memory with PyTorch tensor
+  tup[1] = false;
   d["data"] = tup;
   if (std::is_same<Backend, GPUBackend>::value) {
     // see https://numba.pydata.org/numba-doc/dev/cuda/cuda_array_interface.html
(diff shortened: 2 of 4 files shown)


dali/test/python/test_backend_impl_torch_dlpack.py
```diff
@@ -63,7 +63,7 @@ def test_dlpack_tensor_list_gpu_to_cpu():
 
 
 def check_dlpack_types_gpu(t):
-    arr = torch.tensor([[-0.39, 1.5], [-1.5, 0.33]], device="cuda", dtype=t)
+    arr = torch.tensor([[0.39, 1.5], [1.5, 0.33]], device="cuda", dtype=t)
```
JanuszL, line 66: The latest torch complains `value cannot be converted to type uint8 without overflow`.

dali/test/python/test_backend_impl_torch_dlpack.py
```diff
@@ -149,7 +149,7 @@ def create_tmp(idx):
 
 
 def check_dlpack_types_cpu(t):
-    arr = torch.tensor([[-0.39, 1.5], [-1.5, 0.33]], device="cpu", dtype=t)
+    arr = torch.tensor([[0.39, 1.5], [1.5, 0.33]], device="cpu", dtype=t)
```
JanuszL, line 152: The latest torch complains `value cannot be converted to type uint8 without overflow`.

mzient:
Shouldn't it be "writable" in the title?

mzient (approved):

JanuszL:
> Shouldn't it be "writable" in the title?

Fixed

klecki:
Just a question: when converting form one Tensor type to another, like when we do TensorList -> Tensor, we typically pack the shared_ptr to the memory somewhere. Here, I don't think there is even Python object reference to the original tensor.  Wouldn't it crash in an iteration or two in most cases?

JanuszL:
> Just a question: when converting form one Tensor type to another, like when we do TensorList -> Tensor, we typically pack the shared_ptr to the memory somewhere. Here, I don't think there is even Python object reference to the original tensor. Wouldn't it crash in an iteration or two in most cases?

I don't think the pipeline exposes a non-continuous tensor as the output. In the general case, it may not work, in most cases, we just expose the pointer to the underlying, continuous memory, so the tensor itself doesn't have to live long.
The problem you describes is rather about the conversion from TensorList to Tensor itself, and possible problems of the array_interface are just a derivate of it.

klecki:
The problem I'm referring to is specifically how long the allocation is valid. But thinking about it, the array interface is just a property of the object, and accessing it doesn't impact the lifetime of the original object. So the user of array interface must take lifetimes into account and keep the original object alive as long as they are using the raw pointer that they obtain. I think we should be fine, until someone keeps this pointer for too long thinking it contents won't change. I don't know what Pytorch may do with such memory.

JanuszL:
> The problem I'm referring to is specifically how long the allocation is valid. But thinking about it, the array interface is just a property of the object, and accessing it doesn't impact the lifetime of the original object. So the user of array interface must take lifetimes into account and keep the original object alive as long as they are using the raw pointer that they obtain. I think we should be fine, until someone keeps this pointer for too long thinking it contents won't change. I don't know what Pytorch may do with such memory.

Yes, that is what I had in mind. I think PyTorch just wraps the memory up and does nothing with it. After all, it was possible already but with cupy in the middle.

klecki:
Also, reading the docs of array_interface:
> A reference to the object exposing the array interface must be stored by the new object if the memory area is to be secured.

I think it solves my issues.
BTW @JanuszL, maybe you want to follow up with CUDA array interface v3? https://numba.readthedocs.io/en/stable/cuda/cuda_array_interface.html

klecki (approved):

PR #4826, opened 2023-05-05
Improve VA reservation robustness

mzient:
## Category:

## Description:
On some systems, the amount of virtual address space is limited. This PR improves VA reservation rebustness by two means:
* retrying VA reservation with smaller size, if the default fails.
* trying to free unused VA if reducing the requested size doesn't work.

### Tests:

- [ ] Existing tests apply
- [X] New tests added
  - [ ] Python tests
  - [X] GTests
  - [ ] Benchmark
  - [ ] Other
- [ ] N/A

### Documentation
- [ ] Documentation updated
  - [ ] Docstring
  - [ ] Doxygen
  - [ ] RST
  - [ ] Jupyter
  - [ ] Other
- [X] N/A

#### Requirements
- [X] N/A

**REQ IDs**: N/A

**JIRA TASK**: N/A

### DIFF
--- a/dali/core/mm/default_resource_test.cc
+++ b/dali/core/mm/default_resource_test.cc
@@ -379,7 +379,9 @@ static void ReleaseUnusedTestImpl(ssize_t max_alloc_size = std::numeric_limits<s
 }
 
 TEST(MMDefaultResource, ReleaseUnusedBasic) {
-  ReleaseUnusedTestImpl(256 << 20);
+  size_t free, total;
+  CUDA_CALL(cudaMemGetInfo(&free, &total));
+  ReleaseUnusedTestImpl(align_down(free / 2, 64 << 20));
 }
 
 TEST(MMDefaultResource, ReleaseUnusedMaxMem) {
--- a/dali/core/mm/default_resource_test.cc
+++ b/dali/core/mm/default_resource_test.cc
@@ -379,14 +379,22 @@ static void ReleaseUnusedTestImpl(ssize_t max_alloc_size = std::numeric_limits<s
 }
 
 TEST(MMDefaultResource, ReleaseUnusedBasic) {
-  size_t free, total;
-  CUDA_CALL(cudaMemGetInfo(&free, &total));
-  ReleaseUnusedTestImpl(align_down(free / 2, 64 << 20));
+  cudaDeviceProp device_prop;
+  CUDA_CALL(cudaGetDeviceProperties(&device_prop, 0));
+  if (device_prop.integrated)
+    GTEST_SKIP() << "The memory usage on integrated GPUs cannot be reliably tracked.";
+
+  ReleaseUnusedTestImpl(256 << 20);
 }
 
 TEST(MMDefaultResource, ReleaseUnusedMaxMem) {
   if (!UseVMM())
     GTEST_SKIP() << "Cannot reliably test ReleaseUnused with max mem usage without VMM support";
+
+  cudaDeviceProp device_prop;
+  CUDA_CALL(cudaGetDeviceProperties(&device_prop, 0));
+  if (device_prop.integrated)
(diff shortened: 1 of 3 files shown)


include/dali/core/mm/cuda_vm_resource.h
```diff
@@ -191,13 +205,124 @@ class cuda_vm_resource : public memory_resource<memory_kind::device>,
...
+              for (int b = avail; b < end_avail; b++) {
+                blocks_to_free.push_back(region.unmap_block(b));
+                stat_.total_unmaps++;
+              }
+              auto *start_ptr = region.block_ptr<char>(avail);
+              auto *end_ptr = region.block_ptr<char>(end_avail);
+              free_mapped_.get_specific_block(start_ptr, end_ptr);
+
+              if (end_avail == end_block)
+                break;
```
JanuszL, line 270: Wouldn't this be covered by L259?

include/dali/core/mm/cuda_vm_resource.h
```diff
@@ -191,13 +205,124 @@ class cuda_vm_resource : public memory_resource<memory_kind::device>,
...
+
+        auto *start_ptr = region.block_ptr<char>(rd.start_block);
+        auto *end_ptr = region.block_ptr<char>(rd.end_block);
+        ptrdiff_t range_size = (end_ptr - start_ptr);
+
+        if (region.empty()) {
+          if (tail.empty())
+            va_regions_.erase(va_regions_.begin() + rd.region_idx);
+          else
+            region = std::move(tail);
```
JanuszL, line 304: So in this case the `start_block` is the start of the old region?

include/dali/core/mm/cuda_vm_resource.h
```diff
@@ -191,13 +205,124 @@ class cuda_vm_resource : public memory_resource<memory_kind::device>,
...
+        va_region tail = region.split(rd.end_block);
+        region.resize(rd.start_block);
+
+        auto *start_ptr = region.block_ptr<char>(rd.start_block);
+        auto *end_ptr = region.block_ptr<char>(rd.end_block);
+        ptrdiff_t range_size = (end_ptr - start_ptr);
+
+        if (region.empty()) {
+          if (tail.empty())
+            va_regions_.erase(va_regions_.begin() + rd.region_idx);
```
JanuszL, line 302: I don't fully understand that case - if both region and tail are empty does it mean that the old region was empty as well?

mzient: It means that the range covered the entire region: the tail is empty and the head is shrunk to zero blocks.

include/dali/core/mm/cuda_vm_resource.h
mzient: Yes.

include/dali/core/mm/cuda_vm_resource.h
mzient: If we already reach `end_block` there's no point in running the search in L271.

dali/core/mm/default_resource_test.cc
```diff
@@ -379,7 +379,9 @@ static void ReleaseUnusedTestImpl(ssize_t max_alloc_size = std::numeric_limits<s
 }
 
 TEST(MMDefaultResource, ReleaseUnusedBasic) {
-  ReleaseUnusedTestImpl(256 << 20);
+  size_t free, total;
+  CUDA_CALL(cudaMemGetInfo(&free, &total));
+  ReleaseUnusedTestImpl(align_down(free / 2, 64 << 20));
```
mzient, line 384: This change was necessary because some systems (Orin) didn't even notice the 256M allocation.

JanuszL (approved):

banasraf (approved):

PR #4835, opened 2023-05-08
Fix FITS reader test with Python3.6

stiepan:
## Category:

## Description:
* Exclude int8 from tested dtypes when running tests with Python3.6 - the astropy used as a baseline has older version of astropy available which does not support int8.
* Remove yileds for better compatibility with nose2
* Avoid non-deterministic tests: replace calls that rely on global random state with scoped random generators.

### Tests:

- [ ] Existing tests apply
- [ ] New tests added
  - [ ] Python tests
  - [ ] GTests
  - [ ] Benchmark
  - [ ] Other
- [ ] N/A

### Documentation
- [ ] Documentation updated
  - [ ] Docstring
  - [ ] Doxygen
  - [ ] RST
  - [ ] Jupyter
  - [ ] Other
- [ ] N/A

#### Requirements
- [ ] N/A

**REQ IDs**: N/A

**JIRA TASK**: N/A

### DIFF
--- a/dali/test/python/reader/test_fits.py
+++ b/dali/test/python/reader/test_fits.py
@@ -1,4 +1,4 @@
-# Copyright (c) 2020-2022, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# Copyright (c) 2020-2023, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


dali/test/python/reader/test_fits.py
```diff
@@ -12,21 +12,27 @@
...
 import numpy as np
 import tempfile
 import random
+from nose2.tools import params
 from test_utils import to_array
 from numpy.testing import assert_array_equal
 
-rng = np.random.RandomState(12345)
 
+def enum_product(*params):
```
JanuszL, line 30: Please don't. Use `cartesian_params`.

stiepan: But then I don't have the enumeration.

JanuszL: Good point. Do we need enumeration after all?

JanuszL (approved):

dali/test/python/reader/test_fits.py
stiepan: The tests were written so that they randomly select some paramters (`random.choice` and np's random generator declared globally used for generating ndarray). Depending on the order or subset of tests called, the random choices would be different. To avoid that, I create the random generators locally in each test case. The enumeration is used as a seed.

stiepan:
!build

awolant (approved):

PR #4836, opened 2023-05-08
Add a test that triggers a failure in Python

klecki:
## Description:
This is part of sanity test for CI configuration.
Using this test we can detect if CI correctly handles failing tests.

CI/Tests

### Tests:

- [ ] Existing tests apply
- [ ] New tests added
  - [ ] Python tests
  - [ ] GTests
  - [ ] Benchmark
  - [x] Other
- [ ] N/A

### Documentation
- [ ] Documentation updated
  - [ ] Docstring
  - [ ] Doxygen
  - [ ] RST
  - [ ] Jupyter
  - [ ] Other
- [x] N/A

#### Requirements
- [x] N/A

**REQ IDs**: N/A

**JIRA TASK**: N/A

### DIFF
--- a/dali/test/python/unittest_failure.cfg
+++ b/dali/test/python/unittest_failure.cfg
@@ -1,0 +1,15 @@
+[unittest]
+plugins = nose2.plugins.attrib
+          nose2.plugins.collect
+          nose2.plugins.printhooks
+
+test-method-prefix = failure_
+
+[collect-only]
+always-on = False
+
+[parameters]
+always-on = True
+
+[print-hooks]
+always-on = False
--- a/qa/TL1_python-self-test-failure/test.sh
+++ b/qa/TL1_python-self-test-failure/test.sh
@@ -1,0 +1,12 @@
+#!/bin/bash -e
+# used pip packages
+pip_packages='${python_test_runner_package}'
+target_dir=./dali/test/python
+
+test_body() {
+    ${python_new_invoke_test} --config unittest_failure.cfg test_trigger_failure.py
+}
+
+pushd ../..
+source ./qa/test_template.sh
+popd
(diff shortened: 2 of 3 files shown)


dali/test/python/unittest_failure.cfg
```diff
@@ -0,0 +1,15 @@
+[unittest]
```
JanuszL, line 1: Do we need this config?

klecki:
!build

qa/TL1_python-self-test-failure/test.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash -e
+# used pip packages
+pip_packages='${python_test_runner_package}'
+target_dir=./dali/test/python
+
+test_body() {
+    ${python_new_invoke_test} --config unittest_failure.cfg test_trigger_failure.py
```
JanuszL, line 7: How about just `python -c "raise TypeError()"`?

dali/test/python/unittest_failure.cfg
klecki: So I can have a test case starting with `failure_` that won't be picked up by anyone that wants to run all the tests.
I also want to test the typical unit-testing scenario, that's why I want a test file and nose invocation.

qa/TL1_python-self-test-failure/test.sh
klecki: I can add another one, but I think it's better to try to test something that is as close as possible to the proper testing behaviour.

klecki:
The other part is the MR 520 in CI.

qa/TL1_python-self-test-failure/test.sh
JanuszL: OK

mzient: Mixed feelings here. I'd perhaps go as far the other way as possible and made the `test_body` execute something trivially failing, such as `rm /tmp/this_file_does_not_exist`.

PR #4851, opened 2023-05-12
Adjust numpy reader tests for nose2

stiepan:
## Category:

## Description:
1. Get rid of `yield` style tests and replace them with yielding generators for `@params`. Some of the existing ones used the context managers incorrectly.

```
with context_manager():
   yield _test_case, params
```

doesn't work with nose2 as intended. The whole generator is iterated first, the `_test_case, params` are all collected and then the cases are run. So the context managers exits before the test case is run.

2. Separate test runs into cases nose2 can understand, i.e. avoid just looping over different params and running them all into a single case. In case of error it is difficult to learn what failed if the whole suite is a single case.

3. Make the cases reworked for the 1. or 2. reason deterministic - use explicit random generators with known seeds.
4. Add missing `index` increments in `roi` tests (otherwise all samples in the batch used the same filename).

### Tests:

- [x] Existing tests apply
- [ ] New tests added
  - [ ] Python tests
  - [ ] GTests
  - [ ] Benchmark
  - [ ] Other
- [ ] N/A

### Documentation
- [ ] Documentation updated
  - [ ] Docstring
  - [ ] Doxygen
  - [ ] RST
  - [ ] Jupyter
  - [ ] Other
- [x] N/A

#### Requirements
- [x] N/A

**REQ IDs**: N/A

**JIRA TASK**: N/A

### DIFF
--- a/dali/test/python/reader/test_numpy.py
+++ b/dali/test/python/reader/test_numpy.py
@@ -231,8 +231,9 @@ def test_unsupported_types(device, dtype):
     num_threads = 3
     batch_size = 3
     with assert_raises(RuntimeError, glob="Unknown Numpy type string"):
-        _testimpl_types_and_shapes(device, shapes, dtype, batch_size, num_threads,
-                                    fortran_order, file_arg_type, cache_header_information)
+        _testimpl_types_and_shapes(
+            device, shapes, dtype, batch_size, num_threads,
+            fortran_order, file_arg_type, cache_header_information)
 
 
 @params(*(["cpu", "gpu"] if is_gds_supported() else ["cpu"]))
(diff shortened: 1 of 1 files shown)


JanuszL (approved):

dali/test/python/reader/test_numpy.py
```diff
@@ -510,66 +521,85 @@ def pipe():
...
     dtype = np.uint8
     batch_size = 8
     file_filter = "*.npy"
+    rng = np.random.default_rng(4242 + i)
 
+    with tempfile.TemporaryDirectory(prefix=gds_data_root) as test_data_root:
+        index = 0
+        for sh in shapes:
+            filename = os.path.join(test_data_root, "test_{:02d}.npy".format(index))
+            index += 1
```
stiepan, line 551: The increment was missing.

banasraf (approved):

PR #4880, opened 2023-05-29
Support nesting data structures in conditionals

klecki:
## Description:
It was assumed, that the output of a helper function in case of conditional code is plain tuple of DataNode.

Now the return values can be built out of any complex nestings of tuples and dictionaries (they must still match between branches).

The Tree library is used to apply the splitting (and constant promotion) for branch outputs and merging the if/else paths for arbitrarily nested DataNodes instead of working only with tuples or single elements.

### Tests:

- [ ] Existing tests apply
- [x] New tests added
  - [x] Python tests
  - [ ] GTests
  - [ ] Benchmark
  - [ ] Other
- [ ] N/A

### Documentation
- [x] Documentation updated
  - [ ] Docstring
  - [ ] Doxygen
  - [ ] RST
  - [x] Jupyter
  - [ ] Other
- [ ] N/A

#### Requirements
- [ ] N/A

**REQ IDs**: N/A

**JIRA TASK**: N/A


klecki:
!build

klecki:
!build

banasraf (approved):

dali/python/nvidia/dali/_conditionals.py
```diff
@@ -281,15 +283,15 @@ def preprocess_input(self, data_node):
...
         """Register the data nodes as produced in current scope, otherwise if `global_scope` is True
         put them in the outermost scope.
         """
         if not self._is_registration_allowed:
             return
-        logging.log(8, (f"{self._indent()}[IF/Register] {data_node} at {self.stack_depth() -1}"))
+        logging.log(8, (f"{self._indent()}[IF/Register] {data_nodes} at {self.stack_depth() -1}"))
         scope = self._stack[0] if global_scope else self.top()
-        scope.add_produced(data_node)
+        tree.map_structure(lambda node: scope.add_produced(node), data_nodes)
```
stiepan, line 294: Btw. I can see that `add_produced` does on its own some kind of list flattening (up to list of the lists). Is that needed?

dali/python/nvidia/dali/_conditionals.py
```diff
@@ -516,14 +522,34 @@ def if_stmt(self, cond, body, orelse, get_state, set_state, symbol_names, nouts)
...
+                           " contents of the tuples may be different. If we define a variable as"
+                           " a dictionary, the other branch must define it as a dictionary with the"
+                           " same set of keys, the values may be different.\n")
+
+                try:
+                    tree.assert_same_structure(body_outputs, orelse_outputs, check_types=True)
+                except ValueError as e:
+                    # Suppress the original exception, add DALI explanation at the beginning,
+                    # raise the full error message.
+                    raise ValueError(err_msg + str(e)) from None
```
stiepan, line 540: why not just from e?

klecki: This gives us just one exception message, instead of something like:
```ValueError <message from Tree>

while handling this exception, another exception occurred:

ValueError <message from DALI>```

File: dali/python/nvidia/dali/_conditionals.py (hunk 1 above)
Comment (klecki) [reply dali/python/nvidia/dali/_conditionals.py:294]: I can check as this might be used in one another place, but maybe untangling it, or moving this map_structure one level higher would also work.

File: dali/python/nvidia/dali/_conditionals.py [hunk 3]
```diff
@@ -516,14 +522,34 @@ def if_stmt(self, cond, body, orelse, get_state, set_state, symbol_names, nouts)
...
                                     f" at {this_condition_stack().stack_depth() -1}:"
                                     f" merge({new_body_val}, {new_orelse_val}, predicate="
                                     f"{split_predicate}."))
-                    merged = fn._conditional.merge(new_body_val, new_orelse_val,
-                                                   predicate=split_predicate)
-                    output_values.append(merged)
+                    return fn._conditional.merge(new_body_val, new_orelse_val,
+                                                 predicate=split_predicate)
+
+                output_values = tree.map_structure(merge_branches, body_outputs, orelse_outputs)
```
stiepan, line 552: 1. Do we need `output_values = []` at L:515 then? 
2. The `body_outputs` and `orelse_outputs` are always some sort of tuple, right? Just sometimes with elements that may themselves be structures, but we'll never get here dictionary or plan DataNode? I am asking in the context of `output_values += init_state[nouts:]` in L:558

dali/python/nvidia/dali/_conditionals.py
klecki: Yeah, but it still adds the previous one, so you even get the `e` displayed twice, once as a preamble of `During handling of the above exception, another exception occurred:` and once for `The above exception was the direct cause of the following exception:` due to the `from e`. It's massive and confusing.

dali/python/nvidia/dali/_conditionals.py
klecki: Yes, we can probably remove the initialization, as we just overwrite it with something that is created as a list.

Yes, the state is a tuple/list that is assigned to list of state that are symbolic variables.

stiepan:
Cool to see the "generic" select op working!

I think the unused initilization value and potentially unused nested path in `add_produced` can wait for a follow-up.

stiepan (approved):

PR #4885, opened 2023-05-31
Assure same operator initialization order in the AA graph

stiepan:
## Category:

## Description:

If the operators in the graph are instantiated in different order, then, if they don't have seeds set, they can end up with different one as the default seed for an operator depends on the order in the pipeline.

For AutoAugment to be deterministic one either has to provide seeds manually to all random operators or make sure the operators are instantiated in the same order. 

The following PR does the second thing by removing relience on `set()` order.

It also provides (fixes) the way of doing the first thing - i.e. one of the random ops inside the AA used to not use the optional seed argument.

### Tests:

- [ ] Existing tests apply
- [x] New tests added
  - [x] Python tests
  - [ ] GTests
  - [ ] Benchmark
  - [ ] Other
- [ ] N/A

### Documentation
- [ ] Documentation updated
  - [ ] Docstring
  - [ ] Doxygen
  - [ ] RST
  - [ ] Jupyter
  - [ ] Other
- [ ] N/A

#### Requirements
- [x] N/A

**REQ IDs**: N/A

**JIRA TASK**: N/A

### DIFF
--- a/dali/python/nvidia/dali/auto_aug/auto_augment.py
+++ b/dali/python/nvidia/dali/auto_aug/auto_augment.py
@@ -32,8 +32,8 @@ except ImportError:
         "Please install numpy to use the examples.")
 
 
-def auto_augment(data: _DataNode, policy_name: str = 'image_net',
-                 shape: Optional[_DataNode] = None, fill_value: Optional[int] = 128,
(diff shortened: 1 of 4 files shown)


stiepan:
!build

awolant (approved):

JanuszL (approved):

PR #4886, opened 2023-05-31
Use memory pool for large host allocs

mzient:
## Category:

## Description:
DALI uses dynamic allocation rather aggressively. Large allocations with `malloc` are handled with `mmap`, which causes slowdowns. This PR adds a binning resource and redirects large allocations to a host memory pool (as opposed to passing them directly to malloc). It also introduces an internal knob `DALI_MALLOC_POOL_THRESHOLD` allowing the user to set the maximum size of allocations that go to malloc.

### Tests:

- [ ] Existing tests apply
- [X] New tests added
  - [ ] Python tests
  - [X] GTests
  - [ ] Benchmark
  - [ ] Other
- [ ] N/A

### Documentation
- [ ] Documentation updated
  - [ ] Docstring
  - [ ] Doxygen
  - [ ] RST
  - [ ] Jupyter
  - [ ] Other
- [X] N/A

#### Requirements
- [X] N/A

**REQ IDs**: N/A

**JIRA TASK**: N/A

### DIFF
--- a/dali/core/mm/default_resources.cc
+++ b/dali/core/mm/default_resources.cc
@@ -406,32 +406,28 @@ device_async_resource *GetDefaultDeviceResource(int device_id) {
 }
 
 template <typename Kind>
-mm::pool_resource_base<Kind> *GetPoolInterface(mm::memory_resource<Kind> *mr) {
-  while (mr) {
-    if (auto *pool = dynamic_cast<mm::pool_resource_base<Kind>*>(mr))
-      return pool;
-    if (auto *up = dynamic_cast<mm::with_upstream<Kind>*>(mr)) {
-      mr = up->upstream();
-    } else {
-      break;
+void ReleaseUnusedMemory(mm::memory_resource<Kind> *mr) {
+  if (auto *pool = dynamic_cast<mm::pool_resource_base<Kind>*>(mr)) {
+    pool->release_unused();
+  } else if (auto *up_rsrc = dynamic_cast<mm::with_upstream<Kind>*>(mr)) {
+    ReleaseUnusedMemory(up_rsrc->upstream());
+  } else if (auto *bin_rsrc = dynamic_cast<mm::binning_resource_base<Kind>*>(mr)) {
+    for (int bin = 0; bin < bin_rsrc->num_bins(); bin++) {
+      ReleaseUnusedMemory(bin_rsrc->resource(bin));
     }
   }
-  return nullptr;
 }
 
 DLL_PUBLIC
 void ReleaseUnusedMemory() {
   if (auto *devs = g_resources.device.get()) {
     for (int i = 0, n = g_resources.num_devices; i < n; i++) {
-      if (auto *pool = GetPoolInterface(devs[i].get())) {
-        pool->release_unused();
-      }
+      ReleaseUnusedMemory(devs[i].get());
     }
   }
 
-  if (auto *pool = GetPoolInterface(g_resources.pinned_async.get())) {
-    pool->release_unused();
-  }
+  ReleaseUnusedMemory(g_resources.pinned_async.get());
+  ReleaseUnusedMemory(g_resources.host.get());
 }
 
 DLL_PUBLIC
--- a/dali/core/mm/default_resources.cc
+++ b/dali/core/mm/default_resources.cc
@@ -18,6 +18,7 @@
 #include "dali/core/error_handling.h"
 #include "dali/core/format.h"
 #include "dali/core/mm/malloc_resource.h"
+#include "dali/core/mm/binning_resource.h"
 #include "dali/core/device_guard.h"
 #include "dali/core/mm/async_pool.h"
 #include "dali/core/mm/composite_resource.h"
@@ -126,11 +127,6 @@ struct DefaultResources {
 
 #define g_resources (DefaultResources::instance())
 
-inline std::shared_ptr<host_memory_resource> CreateDefaultHostResource() {
-  static auto rsrc = std::make_shared<malloc_memory_resource>();
-  return rsrc;
-}
-
 struct CUDARTLoader {
   CUDARTLoader() {
     int device_id = 0;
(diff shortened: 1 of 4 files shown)


dali/core/mm/default_resources.cc
```diff
@@ -163,6 +159,50 @@ bool UseVMM() {
...
+    if (thresh)
+      return atoll(thresh);
+    return (32 << 20);  // max for 64-bit Linux systems
+  }
+}
+
+inline std::shared_ptr<host_memory_resource> CreateDefaultHostResource() {
+  static auto rsrc = std::make_shared<malloc_memory_resource>();
+  size_t threshold = MallocThreshold();
+  if (threshold > 0) {
```
JanuszL, line 194: Maybe we should have a default value for that?

dali/core/mm/default_resources.cc
```diff
@@ -163,6 +159,50 @@ bool UseVMM() {
   return value;
 }
 
+size_t MallocThreshold() {
+  char *env = getenv("DALI_MALLOC_POOL_THRESHOLD");
```
JanuszL, line 163: Can you describe it in the `Performance Tuning` section of the docs?

include/dali/core/mm/binning_resource.h
```diff
@@ -0,0 +1,162 @@
...
+   * @tparam SplitPointCollection   a collection of split points; it must have element type
+   *                                convertible to size_t and support iterationa and obtaining size
+   *                                with std::size
+   *
+   * @tparam ResourceCollection     a collection of pointers or iterators to memory_resource<Kind>
+   *                                objects; the elements must be convertible to a plain pointer
+   *                                by applying operators &* in succession
+   *
+   * @param split_points            a collection of split points; it must have one fewer element
+   *                                than split_points
```
JanuszL, line 64:
```suggestion
   *                                than resources
```

dali/core/mm/default_resources.cc
mzient: We do - it's in MallocThreshold.

JanuszL: Right, missed that.

dali/core/mm/default_resources.cc
```diff
@@ -366,32 +406,28 @@ device_async_resource *GetDefaultDeviceResource(int device_id) {
...
-      break;
+void ReleaseUnusedMemory(mm::memory_resource<Kind> *mr) {
+  if (auto *pool = dynamic_cast<mm::pool_resource_base<Kind>*>(mr)) {
+    pool->release_unused();
+  } else if (auto *up_rsrc = dynamic_cast<mm::with_upstream<Kind>*>(mr)) {
+    ReleaseUnusedMemory(up_rsrc->upstream());
+  } else if (auto *bin_rsrc = dynamic_cast<mm::binning_resource_base<Kind>*>(mr)) {
+    for (int bin = 0; bin < bin_rsrc->num_bins(); bin++) {
+      ReleaseUnusedMemory(bin_rsrc->resource(bin));
     }
```
JanuszL, line 417: Should we fail in all other cases?

docs/advanced_topics_performance_tuning.rst
```diff
@@ -83,6 +83,37 @@ For convenience, the DALI_BUFFER_GROWTH_FACTOR environment variable and the
 `nvidia.dali.backend.SetBufferGrowthFactor` Python function can be used to set the same
 growth factor for the host and the GPU buffers.
 
+
+Allocator Configuration
+-----------------------
+
+DALI uses several types of memory resources for various kinds of memory.
```
JanuszL, line 90:
```suggestion
DALI uses several types of memory allocators for various kinds of memory.
```
`resources` sounds more like a jargon.

docs/advanced_topics_performance_tuning.rst
```diff
@@ -83,6 +83,37 @@ For convenience, the DALI_BUFFER_GROWTH_FACTOR environment variable and the
...
+
+For regular host memory, DALI uses (aligned) ``malloc`` for small allocations and a custom memory
+pool for large allocations (to prevent costly calls to ``mmap`` by ``malloc``).
```
JanuszL, line 93: You can link https://www.gnu.org/software/libc/manual/html_node/Malloc-Tunable-Parameters.html here for the reference.

dali/core/mm/default_resources.cc
```diff
@@ -163,6 +159,50 @@ bool UseVMM() {
...
+      s <<= 20;
+    return s;
+  } else {
+    char *mmap = getenv("MALLOC_MMAP_MAX_");
+    if (mmap && !atoi(mmap))
+      return 0;
+    char *thresh = getenv("MALLOC_MMAP_THRESHOLD_");
+    if (thresh)
+      return atoll(thresh);
+    return (32 << 20);  // max for 64-bit Linux systems
```
JanuszL, line 187: How about `128 KiB` ?

docs/advanced_topics_performance_tuning.rst
```diff
@@ -83,6 +83,37 @@ For convenience, the DALI_BUFFER_GROWTH_FACTOR environment variable and the
...
+The maximum size of a host allocation that is allocated directly with ``malloc`` can be customized
+by setting ``DALI_MALLOC_POOL_THRESHOLD`` environment variable. If not specified, the value is
+either derived from environment variables controlling ``malloc`` or, if not found, a default value
+of 32M is used.
+
+For host pinned memory, DALI uses a stream-aware memory pool on top of ``cudaMallocHost``.
```
JanuszL, line 99: Maybe you can link `https://developer.nvidia.com/blog/using-cuda-stream-ordered-memory-allocator-part-1/` for the reference.

dali/core/mm/default_resources.cc
mzient: We don't support 32-bit systems.

dali/core/mm/default_resources.cc
```diff
@@ -163,6 +159,50 @@ bool UseVMM() {
...
+    size_t s = atoll(env);
+    if (env[len-1] == 'k')
+      s <<= 10;
+    else if (env[len-1] == 'M')
+      s <<= 20;
+    return s;
+  } else {
+    char *mmap = getenv("MALLOC_MMAP_MAX_");
+    if (mmap && !atoi(mmap))
+      return 0;
```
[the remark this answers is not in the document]
mzient: The logic is: when MALLOC_MMAP_MAX_ is 0, then the use of mmap by malloc is totally disabled, so switching to internal pool makes no sense.

dali/core/mm/default_resources.cc
JanuszL: I mean to use the default value of `M_MMAP_THRESHOLD` (according to the docs) if `MALLOC_MMAP_THRESHOLD_` is not set,

dali/core/mm/default_resources.cc
mzient: No - just do nothing. Originally, we looked for a pool resource and if not found, ignored it. Now I fused the search and freeing of the memory because, theoretically, a binning resource could have multiple pools, all of which might need freeing.

dali/core/mm/default_resources.cc
mzient: M_MMAP_THRESHOLD by default is adaptive - it starts with 128k and then grows to the possible max as the user keeps freeing memory. In the end we would likely end up slowing the allocations down.

JanuszL (approved):

dali/core/mm/default_resources.cc
stiepan: Is there a valid use of this function in our code-base when it can fail to do anything? I am worried it may be error prone that we call a catch-all release function that can accidently not mention/forsee one of the resources and leak it.

include/dali/core/mm/binning_resource.h
```diff
@@ -0,0 +1,195 @@
...
+  }
+
+ private:
+  void *do_allocate(size_t bytes, size_t alignment) override {
+    int bin = find_bin(bytes);
+    return resource(bin)->allocate(bytes, alignment);
+  }
+
+  void do_deallocate(void *ptr, size_t bytes, size_t alignment) override {
+    int bin = find_bin(bytes);
```
stiepan, line 116: We can rely on the fact that the deallocate will be called with the same number of bytes as the allocation? There are no allocation shrinking tricks possible here that would make us forward the call to wrong resource?

dali/core/mm/default_resources.cc
```diff
@@ -163,6 +159,50 @@ bool UseVMM() {
...
 
+size_t MallocThreshold() {
+  char *env = getenv("DALI_MALLOC_POOL_THRESHOLD");
+  int len = 0;
+  if (env && (len = strlen(env))) {
+    for (int i = 0; i < len; i++) {
+      bool valid = std::isdigit(env[i]) || (i == len - 1 && (env[i] == 'k' || env[i] == 'M'));
+      if (!valid) {
+        DALI_FAIL(make_string(
+          "DALI_MALLOC_POOL_THRESHOLD must be a number, optionally followed by 'k' or 'M', got: ",
```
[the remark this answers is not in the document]
mzient: kilo and Mega? Like in science?

include/dali/core/mm/binning_resource.h
mzient: No, partial deallocation is illegal. The size and alignment must match exactly - so that resources dealing with host-inaccessible memory don't need to maintain (as much) auxiliary information, including maps, etc.

dali/core/mm/default_resources.cc
mzient: Well, yes. If we don't use a pool (but a resource that directly allocates) or when a user externally provides an allocator via C++ API (which is also possible), then not doing anything is quite expected. Until recently this function wasn't even called for host memory (which wasn't really intentional) but if it had, it wouldn't do anything anyway, because we simply used malloc.

dali/core/mm/default_resources.cc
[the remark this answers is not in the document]
stiepan: Oh, ok, it's just metric prefix. I am used to the upper-case in case of powers of two for both.

dali/core/mm/default_resources.cc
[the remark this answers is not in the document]
mzient: We could - but burning another CI cycle for that seems a bit excessive...

stiepan (approved):

PR #4751, opened 2023-03-28
Simplify AutoAugment graph

stiepan:
## Category:
**Refactoring** (*Redesign of existing code that doesn't affect functionality*

## Description:
This PR combines a couple of improvements to AutoAugment processing graph.

1. The list of augmentations to split processing into is created independently for each stage - this can reduces the size of split-merge tree generated by the `select` operation.

For example, the default v0 policy uses 25 sub-policies, each sub-policy is a sequence of (at most) 2 augmentations. Even though there are 25 sub-policies, there are only 11 unique augmentations present in those sub-polcies. DALI already splits the computation only in 11 parts. However, it can be further improved if the operators used in given stages of sub-policies form even smaller groups. In case of the default v0 policy, it is 8 vs 11.

In other words, when the comutation is split to apply the "i-th" augmentation in a sequence according to selected sub-polciy, there is no need to take into account all augmentations present in all sub-polcies, but only the augmentations at the i-th position in all sub-policies.

2. The `signed_bin` helper that handles random negation of magnitudes can now accept additional shape parameter. With that, if you have a multiple random augmentations to apply in a sequence, you can generate the random signs of magnitudes once for all stages, rather than do it for every stage separately.

### Affected modules and functionalities:
AutoAugment (points 1., 2.), RandAugment (point 2.).

### Tests:

- [ ] Existing tests apply
- [x] New tests added
  - [x] Python tests
  - [ ] GTests
  - [ ] Benchmark
  - [ ] Other
- [ ] N/A

### Documentation
- [ ] Existing documentation applies
- [ ] Documentation updated
  - [ ] Docstring
  - [ ] Doxygen
  - [ ] RST
  - [ ] Jupyter
  - [ ] Other
- [x] N/A

#### Requirements
- [ ] Implements new requirements
- [ ] Affects existing requirements
- [x] N/A

**REQ IDs**: N/A

**JIRA TASK**: DALI-3357

PR #4887, opened 2023-05-31
Fix the identity operator id in AA

stiepan:
## Category:

## Description:
Tl;dr: for the end user, the AA works corectly, but under the hood there is a bug in handling sub-policies of different length that is hidden by a happy coincidence.

Sub-policies in AA can have different length. When preparing the matrix that describes mapping from sub-policy and stage to augmentation to run, the shorter sub-polciies are padded with the identity operator. However, some time ago, https://github.com/NVIDIA/DALI/pull/4751 intrduced a bug hat uses index out of range instead of the index of the identity operator.

Fortunately, it has no implications for correctness so far because: the `select` helper runs the last augmentation in the list if the index is out of range and the last augmentation in the list is always the identity.

This PR fixes the out-of-range index and adds unit test to the function that prepares the mapping matrix.

### Tests:

- [ ] Existing tests apply
- [ ] New tests added
  - [ ] Python tests
  - [ ] GTests
  - [ ] Benchmark
  - [ ] Other
- [ ] N/A

### Documentation
- [ ] Documentation updated
  - [ ] Docstring
  - [ ] Doxygen
  - [ ] RST
  - [ ] Jupyter
  - [ ] Other
- [ ] N/A

#### Requirements
- [ ] N/A

**REQ IDs**: N/A

**JIRA TASK**: N/A


stiepan:
!build

dali/python/nvidia/dali/auto_aug/auto_augment.py
```diff
@@ -477,7 +478,7 @@ def _sub_policy_to_augmentation_map(policy: Policy) -> Tuple[_DataNode, List[Lis
                     stage_augments.add(aug)
                     stage_augments_list.append(aug)
         augmentations.append(stage_augments_list + [a.identity])
-    identity_id = [len(stage_augments) for stage_augments in augmentations]
+    identity_id = [len(stage_augments) - 1 for stage_augments in augmentations]
```
stiepan, line 481: This is the fix

stiepan:
!build

szalpal (approved):

jantonguirao (approved):

PR #4897, opened 2023-06-06
Fix coverity issues

stiepan:
## Category:
**Other** (*e.g. Documentation, Tests, Configuration*)

## Description:
* Error checking and dynamic casting issues
* Adjusted the npy header parser: 
   * check afeter dyn_casting of file handle
   * make sure to allocate the extra byte for 10 bytes token for the `\0` trick
   * make sure to include the 10 bytes offset in the size of the read for the header
   * adjust the comment about the possible header size

### Tests:

- [x] Existing tests apply
- [ ] New tests added
  - [ ] Python tests
  - [ ] GTests
  - [ ] Benchmark
  - [ ] Other
- [ ] N/A

### Documentation
- [ ] Existing documentation applies
- [ ] Documentation updated
  - [ ] Docstring
  - [ ] Doxygen
  - [ ] RST
  - [ ] Jupyter
  - [ ] Other
- [x] N/A

#### Requirements
- [ ] Implements new requirements
- [ ] Affects existing requirements
- [x] N/A

**REQ IDs**: N/A

**JIRA TASK**: DALI-3472

PR #4903, opened 2023-06-12
Fix parsing numpy header

stiepan:
## Category:

## Description:

In https://github.com/NVIDIA/DALI/pull/4897, I moved the calculation of the header start pointer (`header = token_mem.get() + token_len;`) down in the function, but setting null-termination character stayed higher, which means that the `\0` is set `token_len` bytes too early. 

### Tests:

- [ ] Existing tests apply
- [x] New tests added
  - [x] Python tests
  - [ ] GTests
  - [ ] Benchmark
  - [ ] Other
- [ ] N/A

### Documentation
- [ ] Documentation updated
  - [ ] Docstring
  - [ ] Doxygen
  - [ ] RST
  - [ ] Jupyter
  - [ ] Other
- [ ] N/A

#### Requirements
- [x] N/A

**REQ IDs**: N/A

**JIRA TASK**: N/A

### DIFF
--- a/dali/test/python/reader/test_numpy.py
+++ b/dali/test/python/reader/test_numpy.py
@@ -192,6 +192,42 @@ def _get_type_and_shape_params():
                         num_threads, batch_size
 
 
+@params(
+    (False, ),
+    (True, ),
+)
+def test_header_parse(use_o_direct):
+    # Test different ndims to see how well we handle headers of different lengths and padding.
+    # The NPY token (header meta-data) + header is padded to the size aligned up to 64 bytes.
+    # In particular the `np.full((1,) * 21, 1., dtype=float32)` and
+    # `np.full((1,) * 22, 1., dtype=float32)` are the boundary between 128 and 192 bytes header.
+    # This make a good case for testing the bounds we use for extracting the header.
+    # The 32 is the max dimensionality handled by the numpy
+    ndims = list(range(33))
+    with tempfile.TemporaryDirectory(prefix=gds_data_root) as test_data_root:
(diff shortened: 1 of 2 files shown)


JanuszL (approved):

stiepan:
!build

JanuszL (approved):

dali/test/python/reader/test_numpy.py
```diff
@@ -192,6 +192,42 @@ def _get_type_and_shape_params():
...
+)
+def test_header_parse(use_o_direct):
+    # Test different ndims to see how well we handle headers of different lengths and padding.
+    # The NPY token (header meta-data) + header is padded to the size aligned up to 64 bytes.
+    # In particular the `np.full((1,) * 21, 1., dtype=float32)` and
+    # `np.full((1,) * 22, 1., dtype=float32)` are the boundary between 128 and 192 bytes header.
+    # This make a good case for testing the bounds we use for extracting the header.
+    # The 32 is the max dimensionality handled by the numpy
+    ndims = list(range(33))
+    with tempfile.TemporaryDirectory(prefix=gds_data_root) as test_data_root:
```
mzient, line 207: Do we really need to test _all_ of them? Wouldn't a small subset like `1, 2, 3, 7, 16, 32` do?

stiepan: The most interesting case is 21, as that has minimal padding after the header, which makes it sensitive to small offset errors. But the test is very simple and lightweight so I went for all the cases.

PR #4929, opened 2023-06-28
Add JAX multinode multigpu tests

awolant:
## Category:

**New feature**

Adds tests for distributed JAX workflows that are using JAX sharding mechanism. For now we test the manual way of building distributed arrays from DALI outputs. 
Adds test both for one GPU per process with 8 processes and 4 GPUs per process with 2 processes.
Tests both `PositionalSharding` and `NamedSharding`.
Runs inside NVIDIA JAX container.

### Tests:

- [ ] Existing tests apply
- [x] New tests added
  - [ ] Python tests
  - [ ] GTests
  - [ ] Benchmark
  - [ ] Other
- [ ] N/A

### Documentation
- [ ] Documentation updated
  - [ ] Docstring
  - [ ] Doxygen
  - [ ] RST
  - [ ] Jupyter
  - [ ] Other
- [x] N/A

#### Requirements
- [ ] N/A

**REQ IDs**: N/A

**JIRA TASK**: DALI-3502


qa/TL3_JAX_multiprocess/jax_client.py
```diff
@@ -0,0 +1,28 @@
...
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser()
+    parser.add_argument("--id", type=int, required=True)
+    parser.add_argument("--size", type=int, required=True)
+    args = parser.parse_args()
+    
+    assert args.id != 0, "Client can not have process id == 0"
+    
+    run_multiprocess_workflow(process_id=args.id, cluster_size=args.size)
```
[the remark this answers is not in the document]
awolant: It took definition from different file.

qa/TL3_JAX_multiprocess/jax_server.py
```diff
@@ -0,0 +1,134 @@
...
+    log.info("Passed positional sharding workflow test")
+
+
+def test_named_sharding_workflow(process_id):
+    mesh = Mesh(jax.devices(), axis_names=('device'))
+    sharding = NamedSharding(mesh, PartitionSpec('device'))
+
+    run_distributed_sharing_test(sharding=sharding, process_id=process_id)
+
+    log.info("Passed named sharding workflow test")
```
awolant, line 111: This is the important part previously unreviewed.

qa/TL3_JAX_multiprocess/test_jax.sh
```diff
@@ -6,4 +6,18 @@ function CLEAN_AND_EXIT {
...
 
+echo "Test one GPU per process"
+CUDA_VISIBLE_DEVICES="1" python jax_client.py --id 1 --size 8 &
+CUDA_VISIBLE_DEVICES="2" python jax_client.py --id 2 --size 8 &
+CUDA_VISIBLE_DEVICES="3" python jax_client.py --id 3 --size 8 &
+CUDA_VISIBLE_DEVICES="4" python jax_client.py --id 4 --size 8 &
+CUDA_VISIBLE_DEVICES="5" python jax_client.py --id 5 --size 8 &
+CUDA_VISIBLE_DEVICES="6" python jax_client.py --id 6 --size 8 &
+CUDA_VISIBLE_DEVICES="7" python jax_client.py --id 7 --size 8 &
+CUDA_VISIBLE_DEVICES="0" python jax_server.py --size 8
```
awolant, line 17: This test is one GPU per one process.  I added this for completeness. Since we already have everything set up and this test is cheap to run I decided to do it as well.

qa/TL3_JAX_multiprocess/test_jax.sh
```diff
@@ -6,4 +6,18 @@ function CLEAN_AND_EXIT {
...
 python -c "import jax; print(jax.devices()); assert jax.device_count() > 0"
 
+echo "Test one GPU per process"
+CUDA_VISIBLE_DEVICES="1" python jax_client.py --id 1 --size 8 &
+CUDA_VISIBLE_DEVICES="2" python jax_client.py --id 2 --size 8 &
+CUDA_VISIBLE_DEVICES="3" python jax_client.py --id 3 --size 8 &
+CUDA_VISIBLE_DEVICES="4" python jax_client.py --id 4 --size 8 &
+CUDA_VISIBLE_DEVICES="5" python jax_client.py --id 5 --size 8 &
+CUDA_VISIBLE_DEVICES="6" python jax_client.py --id 6 --size 8 &
+CUDA_VISIBLE_DEVICES="7" python jax_client.py --id 7 --size 8 &
```
klecki, line 16: I think that if we are running things in background that may in some pathological cases leads to hangs or false positives in the test job. 
It would probably be good to capture all the started processes and wait for their completion or terminate if this doesn't happen in reasonable time.

awolant: Main job is not in the background and it will succeed only if all the clients will connect successfully, otherwise it will fail.
I am all for making it more robust so can you elaborate a bit more about the setup that you have in mind what would work better?

klecki: I was thinking at least to wait for all jobs to correctly terminate like here: https://stackoverflow.com/a/73620420

awolant: I read a bit and I come up with something that I think is better. I am using timeout to exit the test even when there is some hang in there and I am waiting for all the jobs.

qa/TL3_JAX_multiprocess/test_jax.sh
```diff
@@ -6,4 +6,22 @@ function CLEAN_AND_EXIT {
...
+CUDA_VISIBLE_DEVICES="7" timeout -k 60s 60s python jax_client.py --id 7 --size 8 &
+CUDA_VISIBLE_DEVICES="0" timeout -k 60s 60s python jax_server.py --size 8 &
+
+wait $(jobs -p)
+
+echo "Test multiple GPUs per process"
+CUDA_VISIBLE_DEVICES="4,5,6,7" timeout -k 60s 60s python jax_client.py --id 1 --size 2 &
+CUDA_VISIBLE_DEVICES="0,1,2,3" timeout -k 60s 60s python jax_server.py --size 2
+
+wait $(jobs -p)
```
klecki, line 25: This is probably my fault, but I think we will check for the status of the wait instead of the jax_server invokation. Should we check for correct execution of all the clients and servers by checking their exit codes?

klecki (approved):

awolant:
!build

PR #4947, opened 2023-07-18
Propagate layout in fn.reductions

stiepan:
## Category:
**Bug fix**/**New feature**

## Description:
Adds missing layout propagation to the family of `fn.reductions`. The layout propagation rules are simple:
* If the input has no layout, the output does not have it too. 
* If the `keep_dims` is True, the input layout is just copied to the output. 
* Otherwise, the reduced extents are removed from the layout.

## Additional information:

Here's a real-life example of an augmentation that can benefit from the layout propagation, otherwise the `center = .. ` line would not work.

```
def contrast(data, parameter=1.8):
    mean = fn.reductions.mean(data, axis_names="HW", keep_dims=True)
    rgb_weights = types.Constant(np.array([0.299, 0.587, 0.114], dtype=np.float32))
    center = fn.reductions.sum(mean * rgb_weights, axis_names="C", keep_dims=True)
    return fn.cast_like(center + (data - center) * parameter, data)
```

### Tests:

- [ ] Existing tests apply
- [x] New tests added
  - [x] Python tests
  - [ ] GTests
  - [ ] Benchmark
  - [ ] Other
- [ ] N/A

### Documentation
- [ ] Documentation updated
  - [ ] Docstring
  - [ ] Doxygen
  - [ ] RST
  - [ ] Jupyter
  - [ ] Other
- [ ] N/A

#### Requirements
- [x] N/A

**REQ IDs**: N/A

**JIRA TASK**: DALI-3554

### DIFF
--- a/dali/operators/generic/reduce/layout_util.h
+++ b/dali/operators/generic/reduce/layout_util.h
@@ -16,6 +16,7 @@
 #define DALI_OPERATORS_GENERIC_REDUCE_LAYOUT_UTIL_H__
 
 #include "dali/core/tensor_layout.h"
+#include "dali/core/util.h"
 
 namespace dali {
 namespace reduce_util {
@@ -24,30 +25,16 @@ namespace reduce_util {
  * @brief Removes the specified axes from the layout
  *
  * @param layout The layout to remove axes from. Max supported layout size is 64.
- * @param axes list of unique axes from [0...layout.size() - 1] range describing
+ * @param axes list of axes from [0...layout.size() - 1] range describing
  *             what axes to remove from the layout
  */
 template <typename Axes>
 inline TensorLayout ReduceLayout(const TensorLayout &layout, Axes &&axes) {
-  int in_ndim = layout.size();
-  assert(in_ndim >= 0 && in_ndim <= 64);
-  assert(axes.size() <= in_ndim);
-  uint64_t mask = 0;
-  for (auto a : axes) {
-    assert(0 <= a && a < in_ndim);
-    uint64_t a_mask = 1_u64 << a;
-    assert(!(mask & a_mask));  // axes must be unique for the correct out layout dim
-    mask |= a_mask;
-  }
-  int out_ndim = in_ndim - axes.size();
+  uint64_t mask = to_bit_mask(axes);
   TensorLayout out_layout;
-  if (out_ndim <= 0) {
-    return out_layout;
-  }
-  out_layout.resize(out_ndim);
-  for (int in_idx = 0, out_idx = 0; in_idx < in_ndim; in_idx++) {
-    if (!(mask & (1_u64 << in_idx))) {
-      out_layout[out_idx++] = layout[in_idx];
+  for (int idx = 0; idx < layout.size(); idx++) {
+    if (!(mask & (1_u64 << idx))) {
+      out_layout += layout[idx];
     }
   }
   return out_layout;
--- a/dali/operators/generic/reduce/layout_util.h
+++ b/dali/operators/generic/reduce/layout_util.h
@@ -1,0 +1,71 @@
+// Copyright (c) 2023, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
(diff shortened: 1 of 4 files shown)


stiepan:
!build

dali/operators/generic/reduce/layout_util.h
```diff
@@ -0,0 +1,71 @@
...
+template <typename Axes>
+inline TensorLayout ReduceLayout(const TensorLayout &layout, Axes &&axes) {
+  int in_ndim = layout.size();
+  assert(in_ndim >= 0 && in_ndim <= 64);
+  assert(axes.size() <= in_ndim);
+  uint64_t mask = 0;
+  for (auto a : axes) {
+    assert(0 <= a && a < in_ndim);
+    uint64_t a_mask = 1_u64 << a;
+    assert(!(mask & a_mask));  // axes must be unique for the correct out layout dim
```
mzient, line 39: don't think it's required or enforced by the operator. We can specify an axis as many times as we want - it's just that the subsequent occurrences are no-op.

dali/operators/generic/reduce/layout_util.h
```diff
@@ -0,0 +1,71 @@
...
+ *             what axes to remove from the layout
+ */
+template <typename Axes>
+inline TensorLayout ReduceLayout(const TensorLayout &layout, Axes &&axes) {
+  int in_ndim = layout.size();
+  assert(in_ndim >= 0 && in_ndim <= 64);
+  assert(axes.size() <= in_ndim);
+  uint64_t mask = 0;
+  for (auto a : axes) {
+    assert(0 <= a && a < in_ndim);
```
mzient, line 37: Do we have some pre-treatment here? Do reduction operators accept numpy-like negative axes? Should they?

dali/operators/generic/reduce/layout_util.h
```diff
@@ -0,0 +1,71 @@
...
+ *
+ * @param layout The layout to remove axes from. Max supported layout size is 64.
+ * @param axes list of unique axes from [0...layout.size() - 1] range describing
+ *             what axes to remove from the layout
+ */
+template <typename Axes>
+inline TensorLayout ReduceLayout(const TensorLayout &layout, Axes &&axes) {
+  int in_ndim = layout.size();
+  assert(in_ndim >= 0 && in_ndim <= 64);
+  assert(axes.size() <= in_ndim);
```
mzient, line 34: Repetitions wouldn't hurt.
Also, `assert` is not a proper tool for validating user-supplied data. Is there another validation step outside, that enforces these?

stiepan: I assume all those assertions hold for axes, assertions here are like debuggable documentation.

dali/operators/generic/reduce/layout_util.h
```diff
@@ -0,0 +1,71 @@
...
+  uint64_t mask = 0;
+  for (auto a : axes) {
+    assert(0 <= a && a < in_ndim);
+    uint64_t a_mask = 1_u64 << a;
+    assert(!(mask & a_mask));  // axes must be unique for the correct out layout dim
+    mask |= a_mask;
+  }
```
mzient, line 41: Instead of defensively double-checking the axes, you can simply do `uint64_t mask = to_bit_mask(axes);`

dali/operators/generic/reduce/layout_util.h
```diff
@@ -0,0 +1,71 @@
...
+  TensorLayout out_layout;
+  if (out_ndim <= 0) {
+    return out_layout;
+  }
+  out_layout.resize(out_ndim);
+  for (int in_idx = 0, out_idx = 0; in_idx < in_ndim; in_idx++) {
+    if (!(mask & (1_u64 << in_idx))) {
+      out_layout[out_idx++] = layout[in_idx];
+    }
+  }
```
mzient, line 52: Simpler:
```suggestion
  TensorLayout out_layout;
  for (int idx = 0; idx < layout.size(); idx++) {
    if (!(mask & (1_u64 << idx))) {
      out_layout[ += layout[idx];
    }
  }
```

stiepan: Nice. I missed that `+=` op. It lets me drop the unique axes assumption.

dali/operators/generic/reduce/layout_util.h
stiepan: done

mzient (approved):

klecki (approved):

dali/operators/generic/reduce/layout_util.h
```diff
@@ -0,0 +1,58 @@
...
+  for (int idx = 0; idx < layout.size(); idx++) {
+    if (!(mask & (1_u64 << idx))) {
+      out_layout += layout[idx];
+    }
+  }
+  return out_layout;
+}
+
+template <typename Input, typename Output, typename Axes>
+inline void PropagateLayout(const Input &input, Output &output, Axes &&axes, bool keep_dims) {
```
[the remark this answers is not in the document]
stiepan: done

stiepan:
!build

PR #4951, opened 2023-07-20
Fix layout broadcasting arithm ops

stiepan:
## Category:

## Description:
Without this change, it is possible to produce tensors whose layout does not match their dimensionality.

The arithm ops layout propagation works by picking just any non-empty layout and making sure that any other operands have either empty or equal layout. This leads to two issues:
* layout of wrong length: if more ndim-ed tensor has no layout and the less ndim-ed tensor does, the output ends up with layout of incorrect lengt. Take `a` of shape (1, 2, 3, 4) and empty layout, and  `b` of shape  (4,) and layout "C". The output ends up to have 4dims and layout "C".
* errors when the layout match but are not equal, for example FHWC and C.

The fix is to 1. check if the picked layout is not too short and if so, discard it, 2. allow for different layouts as long as one if suffix of the other.

## Additional information:
It would be nice to follow-up the layout propagation with more involved mechanism that would:
1. not discard partial information, i.e. when bigger dim-ed tensor does not have layout, but the samller dim-ed has, we should retain the information about layout suffix. Possibly with "*" wildcards in the layout (that would have consitently special treatment in equality). 
2. generalize the matching from common suffix to one of the layouts being subsequence of the other.
### Tests:

- [ ] Existing tests apply
- [x] New tests added
  - [x] Python tests
  - [ ] GTests
  - [ ] Benchmark
  - [ ] Other
- [ ] N/A

### Documentation
- [ ] Documentation updated
  - [ ] Docstring
  - [ ] Doxygen
  - [ ] RST
  - [ ] Jupyter
  - [ ] Other
- [ ] N/A

#### Requirements
- [x] N/A

**REQ IDs**: N/A

**JIRA TASK**: DALI-3556

### DIFF
--- a/dali/operators/math/expressions/arithmetic.h
+++ b/dali/operators/math/expressions/arithmetic.h
@@ -87,12 +87,12 @@ inline TileCover GetOneTilePerSample(const TensorListShape<> &shape) {
 /**
  * @brief Checks if the child is a (possibly improper) suffix of the parent.
  */
-inline bool EndsWith(const TensorLayout &parent, const TensorLayout &child) {
-  if (parent.size() < child.size()) {
+inline bool EndsWith(const TensorLayout &whole, const TensorLayout &suffix) {
+  if (whole.size() < suffix.size()) {
     return false;
   }
-  for (int i = 0; i < child.size(); i++) {
-    if (parent[parent.size() - 1 - i] != child[child.size() - 1 - i]) {
+  for (int i = 0; i < suffix.size(); i++) {
+    if (whole[whole.size() - 1 - i] != suffix[suffix.size() - 1 - i]) {
       return false;
     }
   }
--- a/dali/operators/math/expressions/arithmetic.h
+++ b/dali/operators/math/expressions/arithmetic.h
@@ -1,4 +1,4 @@
-// Copyright (c) 2019-2022, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+// Copyright (c) 2019-2023, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -85,6 +85,20 @@ inline TileCover GetOneTilePerSample(const TensorListShape<> &shape) {
 }
 
 /**
+ * @brief Checks if the child is a (possibly improper) suffix of the parent.
+ * The child must not be longer than the parent.
+ */
+inline bool HasSuffix(const TensorLayout &parent, const TensorLayout &child) {
+  assert(parent.size() >= child.size());
+  for (int i = 0; i < child.size(); i++) {
+    if (parent[parent.size() - 1 - i] != child[child.size() - 1 - i]) {
+      return false;
+    }
+  }
+  return true;
+}
+
+/**
  * @brief Recurse over expression tree and return the only matching layout
  */
 template <typename Backend>
@@ -103,18 +117,19 @@ DLL_PUBLIC TensorLayout GetCommonLayout(ExprNode &expr, const Workspace &ws) {
   auto result_layout = GetCommonLayout<Backend>(func[0], ws);
   for (int i = 1; i < expr.GetSubexpressionCount(); i++) {
     auto next_layout = GetCommonLayout<Backend>(func[i], ws);
-    if (result_layout.empty()) {
+    if (result_layout.size() >= next_layout.size()) {
+      if (HasSuffix(result_layout, next_layout)) {
+        continue;
+      }
+    } else if (HasSuffix(next_layout, result_layout)) {
       result_layout = next_layout;
       continue;
     }
-    if (next_layout.empty()) {
-      continue;
-    }
-    DALI_ENFORCE(
-        result_layout == next_layout,
-        make_string("Layouts of subexpressions ", i - 1, " and ", i, " for arithmetic operation",
-                    func.GetFuncName(), " do not match. Expected ", result_layout, " got ",
-                    next_layout, "."));
+    DALI_FAIL(make_string("Layouts of subexpressions ", i - 1, " and ", i,
+                          " for arithmetic operation ", func.GetFuncName(),
+                          " do not match. They must be equal or one must be a "
+                          "suffix of the other. Got ",
+                          result_layout, " and ", next_layout, "."));
   }
   return result_layout;
 }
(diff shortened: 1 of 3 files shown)


dali/test/python/operator_1/test_arithmetic_ops.py
```diff
@@ -863,6 +864,118 @@
...
+        expected_layout = expected_layout or ""
+        assert o.layout() == expected_layout, f"got `{o.layout()}`, expected `{expected_layout}`"
+        out_shape = o.shape()
+        assert len(out_shape) == batch_size, f"got `{len(out_shape)}`, expected `{batch_size}`"
+        for sample_shape in out_shape:
+            assert sample_shape == expected_shape, \
+                f"got `{sample_shape}`, expected `{expected_shape}`"
+    else:
+        with assert_raises(Exception, glob="They must be equal or one must be a suffix"):
+            o, = p.run()
```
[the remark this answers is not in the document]
stiepan: done

stiepan:
!build

stiepan:
!build

dali/operators/math/expressions/arithmetic.h
```diff
@@ -84,6 +84,20 @@ inline TileCover GetOneTilePerSample(const TensorListShape<> &shape) {
   return std::make_tuple(std::move(descs), std::move(ranges));
 }
 
+/**
+ * @brief Checks if the child is a (possibly improper) suffix of the parent.
+ * The child must not be longer than the parent.
+ */
+inline bool HasSuffix(const TensorLayout &parent, const TensorLayout &child) {
```
mzient, line 91: More common name: EndsWith
```suggestion
inline bool EndsWith(const TensorLayout &parent, const TensorLayout &child) {
```

dali/operators/math/expressions/arithmetic.h
```diff
@@ -84,6 +84,20 @@ inline TileCover GetOneTilePerSample(const TensorListShape<> &shape) {
...
+  assert(parent.size() >= child.size());
```
mzient, line 92: Unnecessary and actually harmful. Obviously a longer string is not a suffix of a shorter one.
```suggestion
  if (parent.size() < child.size())
    return false;
```

dali/operators/math/expressions/arithmetic.h
stiepan: done

dali/operators/math/expressions/arithmetic.h
stiepan: done

stiepan:
!build

dali/operators/math/expressions/arithmetic.h
```diff
@@ -330,9 +346,22 @@ class ArithmeticGenericOp : public Operator<Backend> {
...
     auto curr_batch_size = ws.GetInputBatchSize(0);
 
+    PropagateShapes<Backend>(result_shape_, *expr_, ws, curr_batch_size);
+
     if (!types_layout_inferred_) {
       result_type_id_ = PropagateTypes<Backend>(*expr_, ws);
       result_layout_ = GetCommonLayout<Backend>(*expr_, ws);
+      // sample_ndim is assumed to be constant between iterations.
+      if (result_layout_.size() != result_shape_.sample_dim()) {
+        // If the largest dimentional tensors do not have layout set but some
```
mzient, line 356:
```suggestion
        // If the tensors with the most dimensions do not have a layout set but some
```

dali/operators/math/expressions/arithmetic_test.cc
```diff
@@ -134,7 +134,7 @@ TEST(ArithmeticOpsTest, TreePropagationLayoutError) {
     ptr->Resize({{1}, {2}}, DALI_INT32);
   }
   in[0]->SetLayout(TensorLayout());
-  in[1]->SetLayout(TensorLayout("HW"));
+  in[1]->SetLayout(TensorLayout("DH"));
```
mzient, line 137: Actually, we should test both cases - the old one should _not_ throw, but instead extend the layout to `"DHW"`.

stiepan: We have tests for that in python though.

dali/test/python/operator_1/test_arithmetic_ops.py
```diff
@@ -863,6 +864,118 @@ def get_sh(arg_idx):
...
+        placement.append("gpu" if placement[-1] == "cpu" else "cpu")
+        return tuple(placement)
+
+    def get_input_types(num_inputs, integral_only):
+        types = (np.int32, np.uint8)
+        if not integral_only:
+            types += (np.float32, )
+        return tuple(rng.choice(types, size=(num_inputs, )))
+
+    bin_layouts = [
```
mzient, line 884: An explanation would be nice.
```suggestion
    # The input layouts and the expected output layout.
    # A number N denotes an ND tensor without a layout.
    # `Exception` means that applying an operator with the arguments with given layouts should raise an error.
    bin_layouts = [
```

dali/test/python/operator_1/test_arithmetic_ops.py
```diff
@@ -863,6 +864,118 @@ def get_sh(arg_idx):
...
+        (2, bin_layouts, bin_ops),
+        (3, ternary_layouts, ternary_operations),
+    ]:
+        for i, (args_desc, out_desc) in enumerate(layouts):
+            assert (len(args_desc) == num_inputs)
+            op, op_name = op_lists[i % len(op_lists)][:2]
+            op = op if not isinstance(op, tuple) else op[0]
+            input_devs = get_input_dev(num_inputs)
+            in_types = get_input_types(num_inputs, op_name in ("&|^"))
+            args_desc = tuple(tensor_desc(arg) for arg in args_desc)
```
mzient, line 942: :+1: for not creating a useless Cartesian product

dali/operators/math/expressions/arithmetic.h
```diff
@@ -84,6 +84,21 @@ inline TileCover GetOneTilePerSample(const TensorListShape<> &shape) {
   return std::make_tuple(std::move(descs), std::move(ranges));
 }
 
+/**
+ * @brief Checks if the child is a (possibly improper) suffix of the parent.
+ */
+inline bool EndsWith(const TensorLayout &parent, const TensorLayout &child) {
```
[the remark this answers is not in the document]
stiepan: done

dali/test/python/operator_1/test_arithmetic_ops.py
stiepan: done

dali/operators/math/expressions/arithmetic.h
stiepan: done

stiepan:
!build

mzient (approved):

szalpal (approved):

PR #4954, opened 2023-07-25
Support checkpointing in FileReader

staniewzki:
## Category:

## Description:

This PR adds support for checkpointing to `FileReader`. 

In the prefetching thread, a state snapshot is produced every epoch by the `Loader`,  and stored in a queue.

Currently, only pad_last_batch=true is supported, because otherwise, the epoch might change mid-batch, which complicates checkpointing.

Note: most of the changes in file_label_loader.h diff come from fixing the indentation there.

### Tests:

- [ ] Existing tests apply
- [ ] New tests added
  - [ ] Python tests
  - [ ] GTests
  - [ ] Benchmark
  - [ ] Other
- [ ] N/A

### Documentation
- [x] Documentation updated
  - [ ] Docstring
  - [x] Doxygen
  - [ ] RST
  - [ ] Jupyter
  - [ ] Other
- [ ] N/A

#### Requirements
- [ ] N/A

**REQ IDs**: N/A

**JIRA TASK**: DALI-3527


dali/operators/reader/loader/loader.h
```diff
@@ -81,6 +81,11 @@ class Loader {
     std::seed_seq seq({seed_});
     e_ = std::default_random_engine(seq);
     virtual_shard_id_ = shard_id_;
+
+    // TODO(mstaniewski): add default value in operator schemas
+    if (!options.TryGetArgument(checkpointing_, "checkpointing")) {
+      checkpointing_ = false;
+    }
```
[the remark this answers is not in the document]
staniewzki: Where can I set the default value then?

dali/pipeline/operator/operator.h
```diff
@@ -156,8 +156,10 @@ class DLL_PUBLIC OperatorBase {
 
   /**
    * @brief Saves operator state into a checkpoint.
+   *
+   * Is called exactly once per epoch.
   */
-  virtual void SaveState(OpCheckpoint &cpt, std::optional<cudaStream_t> stream) const {
+  virtual void SaveState(OpCheckpoint &cpt, std::optional<cudaStream_t> stream) {
```
stiepan, line 162: Actually, can we keep it `const` and not make this assumption? Not that it particularily hurt us, but I just find it a bit surprising for the SaveState method to change state of the operator.
Is that because of the removing the state from the loader's queue? How about removing the stale state from the queue when the state is becomes stale, i.e. in the Run method?

dali/operators/reader/loader/loader.h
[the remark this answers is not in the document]
stiepan: I would check op_schema.cc, the L60 looks promising:
`AddInternalArg("num_threads", "Number of CPU threads in a thread pool", -1);`

dali/pipeline/operator/operator.h
staniewzki: The problem I faced at first was that I can't lock a mutex with the `const`, and I don't really know how to deal with it, as the queue is updated from another thread.

dali/operators/reader/loader/file_label_loader.h
```diff
@@ -195,6 +241,12 @@ class DLL_PUBLIC FileLabelLoader : public Loader<CPUBackend, ImageLabelWrapper>
     current_epoch_++;
 
     if (shuffle_after_epoch_) {
+      if (checkpointing_) {
+        // With checkpointing is enabled, dataset order must be easy to restore.
```
stiepan, line 245: nitpick

```suggestion
        // With checkpointing enabled, dataset order must be easy to restore.
```

dali/operators/reader/loader/loader.h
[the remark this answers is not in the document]
stiepan: I don't think the InternalArgs are exposed in the op docs (hopefully not!). But you are right that you can as well get back to it when working on exposing the checkpointing in the pipeline.

dali/operators/reader/loader/file_label_loader.h
```diff
@@ -46,87 +54,118 @@ class DLL_PUBLIC FileLabelLoader : public Loader<CPUBackend, ImageLabelWrapper>
...
+    DALI_ENFORCE(!(shuffle_after_epoch_  && stick_to_shard_),
+                  "shuffle_after_epoch and stick_to_shard cannot be both true");
+    DALI_ENFORCE(!(shuffle_after_epoch_ && shuffle_),
+                  "shuffle_after_epoch and random_shuffle cannot be both true");
+    /*
+     * Imply `stick_to_shard` from  `shuffle_after_epoch`
+     */
+    if (shuffle_after_epoch_) {
+      stick_to_shard_ = true;
+    }
```
staniewzki, line 127: Lengthy diff, but this is just an indentation fix.

dali/operators/reader/reader_op_test.cc
```diff
@@ -473,4 +475,275 @@ TEST(ReaderTestSimple, CheckNumSamples) {
...
+
+  OpSpec MakeOpSpec() {
+      // Currently, only pad_last_batch=true is supported.
+      return OpSpec("FileReader")
+            .AddOutput("data_out", "cpu")
+            .AddOutput("labels", "cpu")
+            .AddArg("files", filepaths_)
+            .AddArg("checkpointing", true)
+            .AddArg("pad_last_batch", true);
+  }
```
stiepan, line 508: Maybe one more thing to check here - the `prefetch_queue_depth` could be different in or two cases? Although, the default is 1 which intuitevely is the most difficult case. :D

dali/operators/reader/loader/file_label_loader.cc
```diff
@@ -73,7 +73,36 @@ void FileLabelLoader::ReadSample(ImageLabelWrapper &image_label) {
   image_label.image.SetMeta(meta);
 }
 
+FileLabelLoaderState FileLabelLoader::PopClonedState() {
+  DALI_ENFORCE(checkpointing_, "Cannot export loader state when checkpointing is not enabled. ");
+  std::lock_guard<std::mutex> lock(state_queue_mutex_);
+  auto result = state_queue_[state_queue_front_];
+  state_queue_front_ = (state_queue_front_ + 1) % state_queue_.size();
+  return result;
+}
```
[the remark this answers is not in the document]
staniewzki: 1. Good call, 
2. I was trying not to use the words checkpoint and state interchangebly, as there is already `class Checkpoint` and naming a method here `PopCheckpoint` has another meaning. I have adjusted the names to make them more clear hopefully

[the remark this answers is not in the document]
stiepan: 2. Pop/PushStateSnapshot sound good!

dali/operators/reader/reader_op.h
```diff
@@ -46,8 +46,10 @@ namespace dali {
  *                    samples.
  * @tparam ParseTarget Type passed into Parser for parsing, usually it is the same
  *                     as the LoadTarget.
+ * @tparam ParseTarget A marker for checkpointing support.
```
stiepan, line 49:
```suggestion
 * @tparam supports_checkpointing A marker for checkpointing support.
```

dali/operators/reader/reader_op_test.cc
```diff
@@ -473,4 +475,279 @@ TEST(ReaderTestSimple, CheckNumSamples) {
...
+    std::vector<std::pair<string, string>> outputs = {{"data_out", "cpu"}};
+    pipe.Build(outputs);
+  }
+
+  std::vector<uint8_t> RunEpoch(Pipeline &pipe, int batch_size,
+                                int num_shards = 1, bool stick_to_shard = false) {
+    std::vector<uint8_t> result;
+    int samples_per_shard = (filepaths_.size() + num_shards - 1) / num_shards;
+    int batches_per_shard = (samples_per_shard + batch_size - 1) / batch_size;
+    int iterations = (stick_to_shard ? 1 : num_shards) * batches_per_shard;
```
stiepan, line 520: Does the number of iterations in epoch depend on the stick_to_shard?

staniewzki: Good catch! I misunderstood how the shards are working and indeed it was a bug (in both checkpointing and tests)

stiepan (approved):

dali/operators/reader/reader_op_test.cc
stiepan: I guess the `Reset()` workings are a bit misleading in that regard.

banasraf (approved):

staniewzki:
!build

PR #4957, opened 2023-07-27
Fix abs diff computation in check_batch test utility

stiepan:
## Category:

## Description:
The `check_batch` utility reports wrong abs differences for unsigned types if the actual difference is bigger than the half of the range of the type. (Take uint8 and 1 vs 254 - the abs diff is 253 while the chekc_batch would report 3).

### Tests:

- [ ] Existing tests apply
- [ ] New tests added
  - [ ] Python tests
  - [ ] GTests
  - [ ] Benchmark
  - [ ] Other
- [ ] N/A

### Documentation
- [ ] Documentation updated
  - [ ] Docstring
  - [ ] Doxygen
  - [ ] RST
  - [ ] Jupyter
  - [ ] Other
- [ ] N/A

#### Requirements
- [ ] N/A

**REQ IDs**: N/A

**JIRA TASK**: N/A

### DIFF
--- a/dali/test/python/test_utils.py
+++ b/dali/test/python/test_utils.py
@@ -129,9 +129,7 @@ def get_gpu_num():
     return len(out_list)
 
 
-def get_absdiff(left, right):
-
-    assert left.dtype == right.dtype, f"dtypes mismatch {left.dtype} {right.dtype}"
+def _get_absdiff(left, right):
 
     def make_signed(dtype):
         if not np.issubdtype(dtype, np.signedinteger):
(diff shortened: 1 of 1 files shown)


stiepan:
!build

stiepan:
!build

dali/test/python/test_utils.py
```diff
@@ -128,6 +129,51 @@ def get_gpu_num():
...
+
+    # np.abs of diff doesn't handle overflow for unsigned types
+    absdiff = np.maximum(left, right) - np.minimum(left, right)
+    # max - min can overflow for signed types, wrap them up
+    absdiff = absdiff.astype(make_signed(absdiff.dtype))
+    return absdiff
+
+
+def _check_absdiff():
+    for i in range(-128, 127):
```
stiepan, line 152: In principle, overflow on signed int is UB (that we rely on anyway without this PR). The following one-time check aims to verify it works as expected.

awolant (approved):

dali/test/python/test_utils.py
```diff
@@ -128,6 +129,51 @@ def get_gpu_num():
     return len(out_list)
 
 
+def _get_absdiff(left, right):
+
+    def make_signed(dtype):
```
klecki, line 134: Isn't it make_unsigned?

dali/test/python/test_utils.py
klecki: Maybe we should explain it in a comment in the code rather than in the PR?

klecki (approved):

dali/test/python/test_utils.py
stiepan: Yes it is.:)

dali/test/python/test_utils.py
stiepan: done

stiepan:
!build

PR #4965, opened 2023-08-02
Enable python operators in async pipelines

banasraf:
## Category:

**New feature**

## Description:

I enables support for Python operators in pipelines with `exec_async=True` and `exec_pipelined=True`.

The most important fix that prevent deadlock that was happening there, is releasing the GIL when waiting for the pipeline run results: 
```c++
{
    py::gil_scoped_release interpreter_unlock{};
    p->Outputs(&ws);
}
```

The eliminated the main problem: when a thread was blocked on .Outputs() call, the python interpreter had no way of making it release the GIL. At the same time the executor threads tried to acquire the GIL to run the Python op and that caused the deadlock.

There was also a bunch of other fixes required:
- DLTensorFunction desctructor needed to be written manually to acquire the GIL before releasing the python objects
- operator input and output objects were added as fields because they need to outlive all the computation scheduled to GPU, so they're deleted only at the beginning of the next run
- Before pipeline object is desctructed, the executor needs to be shutdown with the GIL __released__. That ensures that when we wait for all  the scheduled prefetching to finish (which might include running python ops) we don't block the GIL.

### Tests:

- [x] Existing tests apply
- [ ] New tests added
  - [ ] Python tests
  - [ ] GTests
  - [ ] Benchmark
  - [ ] Other
- [ ] N/A

 test_dltensor_function, test_python_function_operator, test_gpu_python_function_operator
 
I modified them to run with async pipelines.

### Documentation
- [x] Documentation updated
  - [ ] Docstring
  - [ ] Doxygen
  - [ ] RST
  - [ ] Jupyter
  - [ ] Other
- [ ] N/A

I leave updating the notebooks for another PR

#### Requirements
- [x] N/A

**REQ IDs**: N/A

**JIRA TASK**: N/A


dali/python/backend_impl.cc
```diff
@@ -2004,7 +2006,12 @@ PYBIND11_MODULE(backend_impl, m) {
           DALI_ENFORCE(meta,
               "Operator " + op_name + "  not found or does not expose valid metadata.");
           return ReaderMetaToDict(meta);
-        });
+        })
+    // On the pipeline destruction we need to release the GIL to shutdown the executor
+    // This way, Python operators that might be still running do not deadlock
+    .def("Shutdown", [](Pipeline *p) {
```
JanuszL, line 2012: I wonder if `Shutdown` shouldn't be an internal method?

JanuszL: I think we can do it a bit more automatically using [this](https://github.com/pybind/pybind11/issues/1446) approach.
The `init` of the `Pipeline` Python class returns a smart pointer. We can add a custom deleter to it that would release the lock for the time of `p->Shutdown();` and then delete the pipeline. I think it is way less error-prone than exposing part of the teardown sequence to Python.

JanuszL (commented):
Great change. I'm very happy to see it

dali/python/backend_impl.cc
banasraf: it should be, although i want to call it from the `__del__` method of Python pipeline object. I could rename it to _Shutdown, though. 

And it is not exposed through the Pyton pipeline wrapper

banasraf: Ok, that's sounds like a good approach

jantonguirao (approved):

dali/test/python/operator_2/test_python_function.py
```diff
@@ -315,13 +305,6 @@ def test_python_operator_invalid_function():
     invalid_pipe.run()
 
 
-@raises(TypeError, "do not support multiple input sets")
```
JanuszL, line 318: So it is supported now, right?

JanuszL (approved):

dali/python/backend_impl.cc
```diff
@@ -1588,6 +1588,30 @@ void FeedPipeline(Pipeline *p, const string &name, py::list list, AccessOrder or
...
+                      prefetch_queue_depth, async_execution, bytes_per_sample_hint, set_affinity,
+                      max_num_stream, default_cuda_stream_priority) {}
+
+  PyPipeline(string serialized_pipe, int batch_size, int num_threads, int device_id,
+             bool pipelined_execution, int prefetch_queue_depth, bool async_execution,
+             size_t bytes_per_sample_hint, bool set_affinity, int max_num_stream,
+             int default_cuda_stream_priority):
+             Pipeline(serialized_pipe, batch_size, num_threads, device_id, pipelined_execution,
+                      prefetch_queue_depth, async_execution, bytes_per_sample_hint, set_affinity,
+                      max_num_stream, default_cuda_stream_priority) {}
```
mzient, line 1607:
```suggestion
  using Pipeline::Pipeline;
```

dali/python/backend_impl.cc
```diff
@@ -1588,6 +1588,30 @@ void FeedPipeline(Pipeline *p, const string &name, py::list list, AccessOrder or
   p->SetExternalInput(name, tv, order, sync, use_copy_kernel);
 }
 
+struct PyPipeline: public Pipeline {
```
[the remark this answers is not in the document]
banasraf: PyBind doesn't accept it ;)

when we define init method for py:class_<A> it accepted std::unique_ptr<A> but didn't accpept unique_ptr with a deleter.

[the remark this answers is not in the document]
banasraf: I could define py::class_<Pipeline, std::unique_ptr<Pipeline, Deleter>>. It would accept such a ptr in init but,  I would have to change the way all the methods are exposed 
from 
```
.def("device_id", &Pipeline::device_id)
```
to
```
.def("device_id", [](Pipeline *p) { return p->device_id(); })
```
And that would be a lot of work

dali/test/python/operator_2/test_python_function.py
banasraf: Nah, just removed too much tests. Reverted it

JanuszL (approved):

banasraf:
!build

banasraf:
!build

PR #4977, opened 2023-08-03
Add checkpointing support to stateless operators used in EfficientNet

staniewzki:
## Category:

## Description:

- Added `StatelessOperator` class, that provides trivial checkpointing implementation
- Added a template parameter to `SequenceOperator` to specify the operator base class to derive from
- Used the `StatelessOperator` to support checkpointing in stateless operators used in EfficientNet:
   - `fn.resize`
   - `fn.rotate`
   - `fn.flip`
   - `fn.crop_mirror_normalize`
   - `fn.wrap_affine`
   - `fn.saturation`
   - `fn.cast_like`
   - `fn.reductions.*`
   - `fn.experimental.equalize`
   - `fn.transforms.crop`
   - `fn.transforms.rotation`
   - `fn.transforms.shear`
    - `fn.transforms.scale`
   - `fn.transforms.translation`
- Added python tests, to make sure the operators marked as stateless are indeed stateless (in addition to manual verification)

### Tests:

- [ ] Existing tests apply
- [x] New tests added
  - [x] Python tests
  - [ ] GTests
  - [ ] Benchmark
  - [ ] Other
- [ ] N/A

### Documentation
- [ ] Documentation updated
  - [ ] Docstring
  - [ ] Doxygen
  - [ ] RST
  - [ ] Jupyter
  - [ ] Other
- [ ] N/A

#### Requirements
- [ ] N/A

**REQ IDs**: N/A

**JIRA TASK**: N/A


dali/test/python/test_dali_cpu_only.py
```diff
@@ -235,7 +235,7 @@ def test_cast_like_cpu():
     pipe.set_outputs(out)
     pipe.build()
     for _ in range(3):
-        pipe.run()
+        print(pipe.run()[0].as_array())
```
stiepan, line 238: Do we need this print?:)

staniewzki: Opps, already fixed.

dali/test/python/test_dali_stateless_operators.py
```diff
@@ -0,0 +1,159 @@
...
+    def __init__(self):
+        rng = np.random.default_rng(1234)
+        self.batch = [rng.integers(255, size=test_data_shape, dtype=np.uint8)
+                      for _ in range(batch_size)]
+
+    def __call__(self):
+        return self.batch
+
+
+def check_single_input(op, batch=True, exec_async=True, exec_pipelined=True, **kwargs):
```
stiepan, line 55: Do you need the `batch`, `exec_async` and `exec_pipelined` args?

dali/test/python/test_dali_stateless_operators.py
```diff
@@ -0,0 +1,159 @@
...
+            else:
+                pipe.set_outputs(processed)
+        pipe.build()
+        return pipe
+
+    check_is_pipeline_stateless(pipeline_factory)
+
+
+def test_stateful():
+    assert_raises(AssertionError, check_single_input, fn.random.coin_flip)
```
stiepan, line 88: +1 for checking the test actually tests :)

While on that, plese prefer using `assert_raises` from nose_utils that accepts glob/regex to match agains the error message to make sure we fail for the anticipated reason.

dali/test/python/test_dali_stateless_operators.py
```diff
@@ -0,0 +1,159 @@
...
+
+    def __call__(self):
+        return self.batch
+
+
+def check_single_input(op, batch=True, exec_async=True, exec_pipelined=True, **kwargs):
+    def pipeline_factory():
+        pipe = Pipeline(batch_size=batch_size, num_threads=4, device_id=None, exec_async=exec_async,
+                        exec_pipelined=exec_pipelined)
+        with pipe:
```
stiepan, line 59: Nitpick: It may be personal preference, but I'd drop with pipe/set_outputs in favour of `@pipeline_def` decorator on top of pipeline_factory. The `set_outputs` will become just return statement (with `tuple(processed)`, `processed`).

In principle, the decorator is newer and advocated way of creating the pipelines, there are even some features that won't work with old style API.

stiepan (commented):
Looks good overall.

Apart from minor comments to the tests I think we'll need arithmetic expression operators too for EfficientNet example.

dali/test/python/test_dali_stateless_operators.py
staniewzki: No, the args were specified in code I was using as reference, and I thought that it might be useful in future. If you think it is more elegant to remove them, I don't see any problem with that.

dali/test/python/test_dali_stateless_operators.py
staniewzki: Cool, it looks much cleaner now.

dali/test/python/test_dali_stateless_operators.py
staniewzki: Done, my preferences changed.

dali/test/python/test_dali_stateless_operators.py
stiepan, line 1: Two remarks to the tests:

1. all of the cases seem to test a cpu variant only. Please make sure to test both cpu and gpu variants. The easiest way would be to call `.gpu()` on external source outputs before passing them to the tested op. You'll need `device_id: 0` rather than `None` as pipeline parameter.

2. Please check out `DALI/qa/TL0_python-self-test-core/test_body.sh`, I think you may need to add path to this file there in order for the test to be run in the CI.

dali/test/python/test_dali_stateless_operators.py
stiepan: Sure, I am fine either way.

staniewzki:
!build

banasraf (approved):

PR #4985, opened 2023-08-08
Trim CV-CUDA to expose only median blur to reduce the binary size

JanuszL:
## Category:

## Description:
- trims CV-CUDA to expose only median blur to reduce the binary size (as CV-CUDA adds 100 MB to the build)

- NA

### Tests:

- [x] Existing tests apply
  - test_median_blur.py
- [ ] New tests added
  - [ ] Python tests
  - [ ] GTests
  - [ ] Benchmark
  - [ ] Other
- [ ] N/A

### Documentation
- [ ] Documentation updated
  - [ ] Docstring
  - [ ] Doxygen
  - [ ] RST
  - [ ] Jupyter
  - [ ] Other
- [x] N/A

#### Requirements
- [x] N/A

**REQ IDs**: N/A

**JIRA TASK**: N/A


klecki (approved):

klecki:
I would be happy if we could come up with a workflow that just tracks the native CV-CUDA repo rather than using our fork.

stiepan (approved):

JanuszL:
> I would be happy if we could come up with a workflow that just tracks the native CV-CUDA repo rather than using our fork.

Me too. For now, we are still at the very early stage of integration and we require more rapid updates than the CV-CUDA release cycle provides.

PR #4993, opened 2023-08-11
Add optimized variant of CMN for HWC to HWC pad FP16 case

klecki:
## Description:
Introduce an optimized kernel for the HWC -> HWC, pad case for FP16.
With current DALI allocation patterns, all outputs are aligned to multiple of 4,
allowing to implement the output loop as writing pair of two lower channels, pair of two higher channels
for each pixel using vectorized fp16 instructions, resulting in 4 byte accesses to output gmem.

This yields up to 3x speedups compared to the previous version used by DALI for this particular case.
There is a safety fallback in case if the memory is not actually alinged.

CMN 

### Tests:

- [x] Existing tests apply
This is already covered by existing tests, current implementation gets more specialized version
for one particular case.
- [ ] New tests added
  - [ ] Python tests
  - [ ] GTests
  - [ ] Benchmark
  - [ ] Other
- [ ] N/A

### Documentation
- [ ] Documentation updated
  - [ ] Docstring
  - [ ] Doxygen
  - [ ] RST
  - [ ] Jupyter
  - [ ] Other
- [ ] N/A

#### Requirements
- [ ] N/A

**REQ IDs**: N/A

**JIRA TASK**: N/A

### DIFF
--- a/dali/kernels/slice/slice_hwc2chw_normalize_gpu.cu
+++ b/dali/kernels/slice/slice_hwc2chw_normalize_gpu.cu
@@ -731,21 +731,6 @@ __global__ void Hwc2HwcNormalizePadFp16(const Hwc2HwcChwSampleDesc<Out, In> *sam
   int sample_idx = FindSampleIdx(first_blocks, num_samples);
   const auto sample = samples[sample_idx];
 
-  int64_t start_x = (blockIdx.x - sample.first_block) * kBlockSize;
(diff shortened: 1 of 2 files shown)


klecki:
!build

klecki:
!build

klecki:
!build

awolant (approved):
Very nice comments and docs. This is quite hard code - additional info makes it significantly easier to understand.

szalpal (approved):

PR #4999, opened 2023-08-17
Add checkpointing support to ImageRandomCrop

staniewzki:
## Category:

## Description:

This PR adds checkpointing implementation to `fn.decoders.image_random_crop` operator.

As the internal state was kept as through `std::function` capturing a `std::shared_ptr`, there was no way to make a deep copy. To handle that, I replaced the `std::function` with a custom wrapper, that implements `operator()`.

I have also added a utility header for checkpointing tests, to make the operator testing easier.

### Tests:

- [ ] Existing tests apply
- [x] New tests added
  - [ ] Python tests
  - [x] GTests
  - [ ] Benchmark
  - [ ] Other
- [ ] N/A

### Documentation
- [ ] Documentation updated
  - [ ] Docstring
  - [ ] Doxygen
  - [ ] RST
  - [ ] Jupyter
  - [ ] Other
- [ ] N/A

#### Requirements
- [ ] N/A

**REQ IDs**: N/A

**JIRA TASK**: DALI-3586

### DIFF
--- a/dali/operators/decoder/nvjpeg/fused/nvjpeg_decoder_random_crop_test.cc
+++ b/dali/operators/decoder/nvjpeg/fused/nvjpeg_decoder_random_crop_test.cc
@@ -14,6 +14,7 @@
 
 #include "dali/operators/decoder/decoder_test.h"
 #include "dali/operators/image/crop/random_crop_attr.h"
+#include "dali/test/dali_test_checkpointing.h"
 
 namespace dali {
 
@@ -62,6 +63,29 @@ TYPED_TEST(ImageDecoderRandomCropTest_GPU, TiffDecode) {
 
 TYPED_TEST(ImageDecoderRandomCropTest_GPU, Jpeg2kDecode) {
   this->Run(t_jpeg2kImgType);
+}
+
+class ImageRandomCropCheckpointingTest_GPU : public CheckpointingTest {};
+
+TEST_F(ImageRandomCropCheckpointingTest_GPU, Simple) {
+  PipelineWrapper pipe(8, {{"decoded", "gpu"}});
+
+  auto filepath = testing::dali_extra_path() + "/db/single/jpeg/134/site-1534685_1280.jpg";
+  pipe.AddOperator(
+    OpSpec("FileReader")
+      .AddOutput("file", "cpu")
+      .AddOutput("label", "cpu")
+      .AddArg("files", std::vector{filepath}));
+
+  pipe.AddOperator(
+    OpSpec("decoders__ImageRandomCrop")
+      .AddInput("file", "cpu")
+      .AddOutput("decoded", "gpu")
+      .AddArg("device", "mixed")
+      .AddArg("checkpointing", true));
+
+  pipe.Build();
+  this->RunTest<uint8_t>(std::move(pipe), 2);
 }
 
 }  // namespace dali
--- a/dali/operators/decoder/nvjpeg/fused/nvjpeg_decoder_random_crop_test.cc
+++ b/dali/operators/decoder/nvjpeg/fused/nvjpeg_decoder_random_crop_test.cc
@@ -75,6 +75,8 @@ TEST_F(ImageRandomCropCheckpointingTest_GPU, Simple) {
     OpSpec("FileReader")
       .AddOutput("file", "cpu")
       .AddOutput("label", "cpu")
+      .AddArg("checkpointing", true)
+      .AddArg("pad_last_batch", true)
       .AddArg("files", std::vector{filepath}));
 
   pipe.AddOperator(
--- a/dali/operators/image/crop/random_crop_attr.h
+++ b/dali/operators/image/crop/random_crop_attr.h
@@ -28,24 +28,6 @@
 
 namespace dali {
 
-class RandomCropGeneratorWrap {
- public:
-  template<typename... Args>
-  explicit RandomCropGeneratorWrap(Args&& ...args)
-    : random_crop_generator_(std::forward<Args>(args)...) {}
-
-  inline CropWindow operator()(const TensorShape<>& shape, const TensorLayout& shape_layout) const {
-    return random_crop_generator_->GenerateCropWindow(shape);
-  }
-
-  inline const std::mt19937 &GetRNG() const { return random_crop_generator_->GetRNG(); }
-  inline std::mt19937 &GetRNG() { return random_crop_generator_->GetRNG(); }
-  inline void SetRNG(std::mt19937 rng) { random_crop_generator_->SetRNG(rng); }
-
- private:
-  std::shared_ptr<RandomCropGenerator> random_crop_generator_;
-};
-
 /**
  * @brief Crop parameter and input size handling.
  *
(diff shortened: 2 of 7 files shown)


staniewzki:
!build

dali/util/random_crop_generator.h
```diff
@@ -34,6 +34,8 @@ class DLL_PUBLIC RandomCropGenerator {
     int64_t seed = time(0),
     int num_attempts_ = 10);
 
+  DLL_PUBLIC inline std::mt19937 GetRNG() const { return rand_gen_; }
```
mzient, line 37:
```suggestion
  DLL_PUBLIC inline std::mt19937 &GetRNG() const { return rand_gen_; }
```
?

dali/util/random_crop_generator.h
```diff
@@ -34,6 +34,8 @@ class DLL_PUBLIC RandomCropGenerator {
...
+  DLL_PUBLIC inline void SetRNG(std::mt19937 rng) { rand_gen_ = rng; }
```
mzient, line 38:
```suggestion
  DLL_PUBLIC inline void SetRNG(std::mt19937 rng) { rand_gen_ = std::move(rng); }
```

dali/operators/image/crop/random_crop_attr.h
```diff
@@ -53,24 +70,35 @@ class RandomCropAttr {
...
   }
 
+  std::vector<std::mt19937> RNGSnapshot() {
+    std::vector<std::mt19937> rngs;
+    for (const auto &gen : crop_window_generators_)
+      rngs.push_back(gen.GetRNG());
+    return rngs;
+  }
+
+  void RestoreRNGState(std::vector<std::mt19937> rngs) {
```
mzient, line 93:
```suggestion
  void RestoreRNGState(const std::vector<std::mt19937> &rngs) {
```

dali/operators/image/crop/random_crop_attr.h
```diff
@@ -28,6 +28,23 @@
...
+ public:
+  template<typename... Args>
+  explicit RandomCropGeneratorWrap(Args&& ...args)
+    : random_crop_generator_(std::forward<Args>(args)...) {}
+
+  inline CropWindow operator()(const TensorShape<>& shape, const TensorLayout& shape_layout) const {
+    return random_crop_generator_->GenerateCropWindow(shape);
+  }
+
+  inline std::mt19937 GetRNG() const { return random_crop_generator_->GetRNG(); }
```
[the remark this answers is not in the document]
staniewzki: A note to make sure we are on the same page: this code does not compile, I have added two getters for const/mutable references.

dali/test/dali_test_checkpointing.h
```diff
@@ -0,0 +1,132 @@
...
+    PipelineWrapper clone(batch_size_, outputs_);
+    for (const auto &[spec, id] : ops_)
+      clone.AddOperator(spec, id);
+    clone.Build();
+
+    for (const auto &[spec, id] : ops_) {
+      // Only save/restore state for operators with checkpointing enabled.
+      bool checkpointing;
+      if (!spec.TryGetArgument(checkpointing, "checkpointing") || !checkpointing)
+        continue;
```
stiepan, line 69: Is this needed?

dali/test/dali_test_checkpointing.h
```diff
@@ -0,0 +1,132 @@
...
+
+      OpCheckpoint cpt(spec);
+      GetOperator(id)->SaveState(cpt, std::nullopt);
```
stiepan, line 72: What if one uses the wrapper with pipeline that contains stateful GPU op?

staniewzki: It's a problem then. I think that we would need to either:
1. try to extract streams from pipeline's executor, which is not possible with the current API I believe,
2. create another stream, and synchronize device when necessary, which doesn't seem like a great option for detecting race conditions

dali/test/dali_test_checkpointing.h
staniewzki: Then it removes the possibility to get input from an stateless external source in the tested pipeline.

I don't use it in tests, and it is easy to mess up testing this way, so I can remove it.

dali/test/dali_test_checkpointing.h
stiepan: I racked my brain on the word "executor", but... it's irrelevant to this PR, is it? There's no executor here, you've written one in `RunIteration`. You are in control of execution. You can get a stream, set it to the workspace and use it here to schedule the copy and then synchornize the host with it.

dali/operators/decoder/nvjpeg/fused/nvjpeg_decoder_random_crop_test.cc
```diff
@@ -64,4 +65,29 @@ TYPED_TEST(ImageDecoderRandomCropTest_GPU, Jpeg2kDecode) {
...
+
+TEST_F(ImageRandomCropCheckpointingTest_GPU, Simple) {
+  PipelineWrapper pipe(8, {{"decoded", "gpu"}});
+
+  auto filepath = testing::dali_extra_path() + "/db/single/jpeg/134/site-1534685_1280.jpg";
+  pipe.AddOperator(
+    OpSpec("FileReader")
+      .AddOutput("file", "cpu")
+      .AddOutput("label", "cpu")
+      .AddArg("checkpointing", true)
```
stiepan, line 78: It's no-op, right?

stiepan (approved):

dali/operators/image/crop/random_crop_attr.h
```diff
@@ -28,6 +28,24 @@
...
 namespace dali {
 
+class RandomCropGeneratorWrap {
+ public:
+  template<typename... Args>
+  explicit RandomCropGeneratorWrap(Args&& ...args)
+    : random_crop_generator_(std::forward<Args>(args)...) {}
+
+  inline CropWindow operator()(const TensorShape<>& shape, const TensorLayout& shape_layout) const {
+    return random_crop_generator_->GenerateCropWindow(shape);
```
mzient, line 38:
```suggestion
    return random_crop_generator_->GenerateCropWindow(shape, layout);
``` ?

File: dali/operators/image/crop/random_crop_attr.h [hunk 9]
```diff
@@ -28,6 +28,24 @@
 
 namespace dali {
 
+class RandomCropGeneratorWrap {
```
mzient, line 31: What is this class for? It seems that it adds no functionality except maybe the call operator (which could be easily replaced with a lambda).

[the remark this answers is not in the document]
staniewzki: Yes, essentially, I wanted to keep it similar to the lambda that was there, but to with the option to get the RNG. 

The underlying generator can be accessed as well if we kept a `std::vector` of them in the RandomCropAttr, which might be a better solution. Do you like it better?

Refactoring the interface seems outside of the scope of this task, but I am open to suggestions.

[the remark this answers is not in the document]
mzient: I don't know if that's really out of scope. It's not a public interface and we clearly have a mismatch between the two interfaces (the function type and the RandomCropGenerator).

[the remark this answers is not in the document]
stiepan: I agree with Michał (outstanding pun) that it's a bit out of scope. Can you add add a jira task that would describe the expected refactor, @mzient? Then, we can make it a follow-up - as you said it's not public API so doing this now later makes little difference to users, while it can delay work on the checkpointing in general.

dali/operators/image/crop/random_crop_attr.h
```diff
@@ -53,24 +53,36 @@ class RandomCropAttr {
...
     seq.generate(seeds.begin(), seeds.end());
 
-    crop_window_generators_.resize(max_batch_size);
-
+    random_crop_generators_.reserve(max_batch_size);
     for (int i = 0; i < max_batch_size; i++) {
-      std::shared_ptr<RandomCropGenerator> random_crop_generator(
+      random_crop_generators_.emplace_back(
         new RandomCropGenerator(
           {aspect_ratio[0], aspect_ratio[1]}, {area[0], area[1]}, seeds[i], num_attempts));
```
mzient, line 60: It's generally a bad practice to `emplace_back` a dynamically allocated pointer to a collection of smart pointers, because if the allocation of a new collection element fails (e.g. due to `std::bad_alloc` being thrown), the pointer is leaked.

Safer and more efficient:
```suggestion
      random_crop_generators_.push_back(
        std::make_shared<RandomCropGenerator>({aspect_ratio[0], aspect_ratio[1]}, {area[0], area[1]}, seeds[i], num_attempts));
```

staniewzki:
!build

PR #5008, opened 2023-08-21
Add pipeline checkpointing to the Executor

staniewzki:
## Category:

## Description:

This PR adds implementation of pipeline checkpointing to the Executor.

Checkpoint is added to the IterationData, and in the iterations that correspond to end of an epoch, operators states are saved in that checkpoint.

There was an issue with cyclic dependiencies. Because of that, definitions from `op_checkpoint.h` were moved to `op_checkpoint.cc`, and there are few other slight workarounds in the code.

### Tests:

- [ ] Existing tests apply
- [x] New tests added
  - [ ] Python tests
  - [x] GTests
  - [ ] Benchmark
  - [ ] Other
- [ ] N/A

### Documentation
- [x] Documentation updated
  - [ ] Docstring
  - [x] Doxygen
  - [ ] RST
  - [ ] Jupyter
  - [ ] Other
- [ ] N/A

#### Requirements
- [ ] N/A

**REQ IDs**: N/A

**JIRA TASK**: N/A

### DIFF
--- a/dali/pipeline/executor/executor.cc
+++ b/dali/pipeline/executor/executor.cc
@@ -386,7 +386,7 @@ void Executor<WorkspacePolicy, QueuePolicy>::RunHelper(OpNode &op_node, Workspac
   SmallVector<int, 16> empty_layout_in_idxs;
 
   if (checkpointing_ && iteration_id % checkpointing_epoch_size_ == 0) {
-    auto &cpt = checkpoint_queue_[iteration_id % checkpoint_queue_.size()].GetOpCheckpoint(op_node.id);
+    auto &cpt = GetCurrentCheckpoint(iteration_id).GetOpCheckpoint(op_node.id);
     op_node.op->SaveState(cpt, ws.has_stream() ? std::optional{ws.stream()} : std::nullopt);
   }
 
@@ -628,6 +628,8 @@ void Executor<WorkspacePolicy, QueuePolicy>::InitIterationData() {
   iteration_data_.resize(iteration_data_size);
   for (auto& id : iteration_data_) {
     id.operator_traces = std::make_shared<operator_trace_map_t>();
+    if (checkpointing_)
+      id.checkpoint.Build(*graph_);
   }
 }
 
@@ -657,12 +659,6 @@ void Executor<WorkspacePolicy, QueuePolicy>::InitCheckpointing() {
   if (std::is_same_v<QueuePolicy, SeparateQueuePolicy>)
     DALI_FAIL("Checkpointing is not supported with `separated` pipeline exection mode enabled. ")
 
-  // TODO(mstaniewski): verify
-  int queue_depth = stage_queue_depths_[OpType::CPU] + 1;
-  checkpoint_queue_.resize(queue_depth);
-  for (auto &cpt : checkpoint_queue_)
-    cpt.Build(*graph_);
-
   checkpointing_epoch_size_ = -1;
   for (const auto &node : graph_->GetOpNodes()) {
     auto meta = node.op->GetReaderMeta();
@@ -676,7 +672,7 @@ void Executor<WorkspacePolicy, QueuePolicy>::InitCheckpointing() {
       DALI_FAIL(make_string(
         "When the checkpointing is enabled, all readers must have the same epoch size. ",
         "The readers and have different epoch sizes ",
-        "(", checkpointing_epoch_size, " and ", local_epoch_size, " respectively). "));
+        "(", checkpointing_epoch_size_, " and ", local_epoch_size, " respectively). "));
   }
 
   /* If there is no operator with ReaderMeta, set the epoch size to 1. */
@@ -687,7 +683,7 @@ void Executor<WorkspacePolicy, QueuePolicy>::InitCheckpointing() {
 template<typename WorkspacePolicy, typename QueuePolicy>
 Checkpoint &
 Executor<WorkspacePolicy, QueuePolicy>::GetCurrentCheckpoint(size_t iteration_id) {
-  return checkpoint_queue_[iteration_id % checkpoint_queue_.size()];
+  return GetCurrentIterationData(iteration_id).checkpoint;
 }
 
 template<typename WorkspacePolicy, typename QueuePolicy>
--- a/dali/pipeline/executor/executor.cc
+++ b/dali/pipeline/executor/executor.cc
@@ -667,7 +667,7 @@ void Executor<WorkspacePolicy, QueuePolicy>::InitCheckpointing() {
     DALI_FAIL("Checkpointing is not supported with `separated` pipeline exection mode enabled. ")
 
   checkpointing_epoch_size_ = -1;
-  const std::string *reader_name = nullptr;
+  std::string reader_name;
   for (const auto &node : graph_->GetOpNodes()) {
     auto meta = node.op->GetReaderMeta();
     if (meta.epoch_size_padded <= 0)
@@ -676,11 +676,11 @@ void Executor<WorkspacePolicy, QueuePolicy>::InitCheckpointing() {
     int local_epoch_size = (meta.epoch_size_padded + max_batch_size_ - 1) / max_batch_size_;
     if (checkpointing_epoch_size_ == -1) {
       checkpointing_epoch_size_ = local_epoch_size;
-      reader_name = &node.spec.name();
+      reader_name = node.spec.name();
     } else if (checkpointing_epoch_size_ != local_epoch_size) {
       DALI_FAIL(make_string(
         "When the checkpointing is enabled, all readers must have the same epoch size. ",
-        "The readers ", *reader_name, " and ", node.spec.name(), " have different epoch sizes ",
+        "The readers ", reader_name, " and ", node.spec.name(), " have different epoch sizes ",
         "(", checkpointing_epoch_size_, " and ", local_epoch_size, " respectively). "));
     }
   }
--- a/dali/pipeline/executor/executor.cc
+++ b/dali/pipeline/executor/executor.cc
@@ -385,6 +385,11 @@ void Executor<WorkspacePolicy, QueuePolicy>::RunHelper(OpNode &op_node, Workspac
   const auto &schema = spec.GetSchema();
   SmallVector<int, 16> empty_layout_in_idxs;
 
+  if (checkpointing_ && iteration_id % checkpointing_epoch_size_ == 0) {
+    auto &cpt = checkpoint_queue_[iteration_id % checkpoint_queue_.size()].GetOpCheckpoint(op_node.id);
+    op_node.op->SaveState(cpt, ws.has_stream() ? std::optional{ws.stream()} : std::nullopt);
+  }
+
   ws.InjectOperatorTraces(GetCurrentIterationData(iteration_id).operator_traces);
   ws.ClearOperatorTraces();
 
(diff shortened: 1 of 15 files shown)


dali/pipeline/executor/executor.cc
```diff
@@ -644,7 +649,45 @@ bool Executor<WorkspacePolicy, QueuePolicy>::HasConditionals() const {
...
+
+    int local_epoch_size = (meta.epoch_size_padded + max_batch_size_ - 1) / max_batch_size_;
+    if (checkpointing_epoch_size_ == -1)
+      checkpointing_epoch_size_ = local_epoch_size;
+    else if (checkpointing_epoch_size_ != local_epoch_size)
+      DALI_FAIL(make_string(
+        "Incompatibile epoch sizes of operators in a pipeline with checkpointing enabled. ",
+        "Expected: ", checkpointing_epoch_size_, " found: ", local_epoch_size, ". "));
+  }
+}
```
stiepan, line 677: 1. How about moving the part with epoch to a separate method with a nice TODO comment about it being a limitation/simplification for now, to be lifted later on?
2. What would happen if none of the operators had the ReaderMeta (e.g. a pipeline with just the random number generators)
3. Can you please check how the external_source works here? Does it have the meta? Will it fail on an attempt to get the checkpoint?
4. Nitpick, I think the error message is a bit unclear. How about: "When the checkpointing is enabled, all readers must have the same epoch size. The readers <get the op name> and <get the second op name> have different epoch sizes (<> and <> respecitively)."

dali/pipeline/executor/executor.cc
```diff
@@ -644,7 +649,45 @@ bool Executor<WorkspacePolicy, QueuePolicy>::HasConditionals() const {
   return has_conditionals_;
 }
 
+template<typename WorkspacePolicy, typename QueuePolicy>
+void Executor<WorkspacePolicy, QueuePolicy>::InitCheckpointing() {
+  if (!checkpointing_)
+    return;
+
```
stiepan, line 656: Presumingly, we also need a check for the execution policy (we won't support the separated queues).

dali/pipeline/executor/executor.cc
```diff
@@ -644,7 +649,45 @@ bool Executor<WorkspacePolicy, QueuePolicy>::HasConditionals() const {
...
+  // TODO(mstaniewski): verify
+  int queue_depth = stage_queue_depths_[OpType::CPU] + 1;
```
stiepan, line 658: I wonder if we should just extnd the `IterationData` struct, it already computes the queue_depth (the same way I think), has the queue, keeps the indicies of producers and consumers.

dali/pipeline/executor/executor.cc
staniewzki: I am pretty sure that external_source does not have ReaderMeta. I think that it is impossible to make it checkpointable, unless you assume that the input is a pure function, in which case it is stateless, and we have the checkpointing for free.

Am I missing something?

stiepan: Ad 1. Just to have anything that touches the notion of epoch separated and defined in one place. So that we can easily find where they are used and remove the usaged and defnitions when the restriction is ready to be lifted.
Ad 2. It could be either that or an error, I wonder which one is better.
Ad 3. Side note, the external source is special in that regard it allows you to feed smaller batches (than the max_batch_size/batch_size declared in the pipeline_def). What I am asking is not to make it work with checkponting now, but if it fails gracefully. :) I guess if we ever ask it for the checkpoint, the external source op just fails with the unsupported error as any other op. Maybe it's worth to double check though. Especially, as if we go with another behaviour in the 2. point, i.e. we actually never call the SaveCheckpoint, we could end up with reporting to the user that checkpointing is not supported when they try to save the checkpoint (possibly after some heavy data loading and training) rather than just from the start.

staniewzki: 2. Why would an error be better? A pipeline where the only input is a random generator sound little bit sketchy, but checkpointing it is a valid use case in the end, isn't it?
3. ExternalSource in a pipeline with checkpointint enabled is going to fail right before the first iteration, when the first checkpoint is created, so there shouldn't be any problems. Even though, it seems like a good test case for future python tests.

stiepan: Ad 3. Cool!
Ad 2. You're right, there's no point in making the provided example not work. If the epoch size is -1 we are either able to checkpoint every iteration or none (i.e. external_source). I was worried about the latter case, but we should end up with a proper error in the first iteration.

dali/pipeline/operator/checkpointing/checkpoint.h
```diff
@@ -17,10 +17,13 @@
 
 #include <vector>
 
-#include "dali/pipeline/graph/op_graph.h"
+#include "dali/pipeline/operator/checkpointing/op_checkpoint.h"
+// #include "dali/pipeline/graph/op_graph.h"
```
stiepan, line 21: :)

dali/pipeline/executor/executor.cc
```diff
@@ -644,7 +658,47 @@ bool Executor<WorkspacePolicy, QueuePolicy>::HasConditionals() const {
...
+    if (meta.epoch_size_padded <= 0)
+      continue;
+
+    int local_epoch_size = (meta.epoch_size_padded + max_batch_size_ - 1) / max_batch_size_;
+    if (checkpointing_epoch_size_ == -1)
+      checkpointing_epoch_size_ = local_epoch_size;
+    else if (checkpointing_epoch_size_ != local_epoch_size)
+      DALI_FAIL(make_string(
+        "When the checkpointing is enabled, all readers must have the same epoch size. ",
+        "The readers and have different epoch sizes ",
```
stiepan, line 681: Something is missing in the message (`The readers and have`). In the first remark I meant for the message to mention names of both readers. You either need to include them in the message or reformulate the message a bit.

dali/pipeline/executor/executor.h
```diff
@@ -142,6 +145,16 @@ class DLL_PUBLIC Executor : public ExecutorBase, public QueuePolicy {
 
   DISABLE_COPY_MOVE_ASSIGN(Executor);
 
+  /**
+   * Returns the checkpoint for the given iteration ID.
+  */
+  DLL_PUBLIC Checkpoint& GetCurrentCheckpoint(size_t iteration_id);
```
stiepan, line 151: I think we should not expose the iteration_id in the public API, should we? We can even have GetCurrentCheckpoint in public API and GetCheckpoint(int iteration_id) in the protected one. Similarily as ShareOutputs gets the trace based on the internally kept output_idx.

dali/pipeline/executor/executor.cc
```diff
@@ -524,6 +530,12 @@ void Executor<WorkspacePolicy, QueuePolicy>::RunHelper(OpNode &op_node, Workspac
       in.SetLayout({});
     }
   }
+
+  // If it is the end of an epoch, create a checkpoint
+  if (checkpointing_ && (iteration_id + 1) % checkpointing_epoch_size_ == 0) {
```
stiepan, line 535: If `checkpointing_epoch_size_ == 1` we end up doing two checkpoints in the `iteration_id == 0`, right?

dali/pipeline/executor/executor.cc
```diff
@@ -644,7 +658,50 @@ bool Executor<WorkspacePolicy, QueuePolicy>::HasConditionals() const {
...
+template<typename WorkspacePolicy, typename QueuePolicy>
+void Executor<WorkspacePolicy, QueuePolicy>::InitCheckpointing() {
+  if (!checkpointing_)
+    return;
+
+  if (std::is_same_v<QueuePolicy, SeparateQueuePolicy>)
+    DALI_FAIL("Checkpointing is not supported with `separated` pipeline exection mode enabled. ")
+
+  checkpointing_epoch_size_ = -1;
+  const std::string *reader_name = nullptr;
```
stiepan, line 670:
```suggestion
  const std::string reader_name;
```

dali/test/operators/dummy_op.cu
```diff
@@ -14,8 +14,91 @@
...
+  auto &input = ws.Input<CPUBackend>(0);
+  auto &output = ws.Output<GPUBackend>(0);
+  int samples = input.num_samples();
+  output.set_type(DALI_UINT8);
+  output.Resize(uniform_list_shape(samples, {1}));
+
+  std::vector<uint8_t> buffer(samples);
+  for (int i = 0; i < samples; i++) {
+    buffer[i] = input.tensor<uint8_t>(i)[0] + state_++;
+    cudaMemcpyAsync(output.mutable_tensor<uint8_t>(i), &buffer[i], sizeof(uint8_t),
```
stiepan, line 46: `CUDA_CALL`?

klecki (commented):
Posting some initial comments, I didn't read the initialization and the restore yet - I need to understand better how do we cope with output_iteration_id_ vs cpu/mixed/gpu_iteration_id that can be shifted compared to one another.

dali/pipeline/executor/executor.cc
staniewzki: No. The here the checkpoint created is corresponding to the state between epochs `(iterations_id, iteration_id + 1)`. The checkpoint created before `iteration_id == 0` is something else.

I will try to improve the comment here to reflect that.

dali/test/operators/dummy_op.cc
```diff
@@ -18,6 +18,68 @@
...
+  state_ = cpt.CheckpointState<uint8_t>();
+}
+
+bool TestStatefulSource::SetupImpl(std::vector<OutputDesc> &output_desc, const Workspace &ws) {
+  return false;
+}
+
+void TestStatefulSource::RunImpl(Workspace &ws) {
+  auto &output = ws.Output<CPUBackend>(0);
+  int samples = output.num_samples();
```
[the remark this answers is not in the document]
staniewzki: From my testing, the num samples here is equal to max_batch_size.
What is the proper way? Set it in SetupImpl?

dali/pipeline/executor/executor.cc
```diff
@@ -385,6 +385,12 @@ void Executor<WorkspacePolicy, QueuePolicy>::RunHelper(OpNode &op_node, Workspac
   const auto &schema = spec.GetSchema();
   SmallVector<int, 16> empty_layout_in_idxs;
 
+  // If it is the first iteration, create initial checkpoint.
+  if (checkpointing_ && iteration_id == 0) {
```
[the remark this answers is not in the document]
staniewzki: 1. Why? Is it better than just accessing `checkpointing_` variable?
2. Good question. It might be better to just throw an exception in this case.

dali/pipeline/executor/executor.h
```diff
@@ -142,6 +147,16 @@ class DLL_PUBLIC Executor : public ExecutorBase, public QueuePolicy {
...
 
+  /**
+   * Returns the current checkpoint.
+  */
+  DLL_PUBLIC Checkpoint& GetCurrentCheckpoint();
+
+  /**
+   * Restores states of operators.
+  */
+  DLL_PUBLIC void RestoreStateFromCheckpoint(const Checkpoint &cpt);
```
klecki, line 158: We should consider guarding against using the Restore when the executor was started/is running. It should only be allowed before any operator was executed, right?

dali/pipeline/executor/executor.cc
```diff
@@ -524,6 +530,12 @@ void Executor<WorkspacePolicy, QueuePolicy>::RunHelper(OpNode &op_node, Workspac
       in.SetLayout({});
     }
   }
+
+  // If it is the end of an epoch, create a checkpoint
+  if (checkpointing_ && (iteration_id + 1) % checkpointing_epoch_size_ == 0) {
+    auto &cpt = GetCurrentIterationData(iteration_id + 1).checkpoint.GetOpCheckpoint(op_node.id);
+    op_node.op->SaveState(cpt, ws.has_stream() ? std::optional{ws.stream()} : std::nullopt);
+  }
```
klecki, line 538: As a side note, you can consider saving some information in the checkpoint here, that for this iteration you actually saved a checkpoint. When the user calls `GetCurrentCheckpoint` we can verify that we actually have something to get, and the checkpoint was obtained in the correct place/time.

dali/pipeline/executor/executor.h
```diff
@@ -142,6 +147,16 @@ class DLL_PUBLIC Executor : public ExecutorBase, public QueuePolicy {
...
   DISABLE_COPY_MOVE_ASSIGN(Executor);
 
+  /**
+   * Returns the current checkpoint.
+  */
+  DLL_PUBLIC Checkpoint& GetCurrentCheckpoint();
+
+  /**
+   * Restores states of operators.
+  */
```
klecki, line 157: Please expand the documentation for those functions.
I think that the "current checkpoint" is a bit ambiguous. It should specify the prerequisites, that is the `GetCurrentCheckpoint` can be used only after the last output of this epoch was consumed. 

For `RestoreStateFromCheckpoint` it should mention, that we can only resume when the executor was not running. 
Some mention of the post-conditions/state after calling those functions would be welcomed.

dali/pipeline/executor/executor.cc
```diff
@@ -524,6 +530,12 @@ void Executor<WorkspacePolicy, QueuePolicy>::RunHelper(OpNode &op_node, Workspac
       in.SetLayout({});
     }
   }
+
+  // If it is the end of an epoch, create a checkpoint
+  if (checkpointing_ && (iteration_id + 1) % checkpointing_epoch_size_ == 0) {
+    auto &cpt = GetCurrentIterationData(iteration_id + 1).checkpoint.GetOpCheckpoint(op_node.id);
```
klecki, line 536: Hmm, thinking about it, I would put a comment explaining the fact, that if we write into the next iteration id, consume its outputs and than call the "GetCurrentCheckpoint" we will get the data from here (basically explaining that +1). It might make things easier in the future.

klecki (approved):

staniewzki:
!build

dali/pipeline/executor/executor.cc
```diff
@@ -644,7 +667,53 @@ bool Executor<WorkspacePolicy, QueuePolicy>::HasConditionals() const {
...
+  if (checkpointing_epoch_size_ == -1)
+    checkpointing_epoch_size_ = 1;
+}
+
+template<typename WorkspacePolicy, typename QueuePolicy>
+Checkpoint &Executor<WorkspacePolicy, QueuePolicy>::GetCurrentCheckpoint() {
+  auto &cpt = GetCurrentIterationData(output_iteration_id_).checkpoint;
+  // Sanity check
+  DALI_ENFORCE(cpt.GetIterationId() == output_iteration_id_,
+               "Requested checkpoint does not correspond to the matching iteration. ");
```
stiepan, line 707: I think it may be a bit vague from the user perspective. The mismatch here simply means we did not checkpoint the given iteration and are looking at some old/default checkpoint, right? In that case, how about something like:

```suggestion
               "The pipeline cannot be checkpointed at the given iteration. Currently, checkpointing is supported at the end of the epoch only (after the last batch from the epoch was consumed).");
```

dali/pipeline/executor/executor.cc
```diff
@@ -644,7 +667,53 @@ bool Executor<WorkspacePolicy, QueuePolicy>::HasConditionals() const {
...
+  }
+
+  /* If there is no operator with ReaderMeta, set the epoch size to 1. */
+  if (checkpointing_epoch_size_ == -1)
+    checkpointing_epoch_size_ = 1;
+}
+
+template<typename WorkspacePolicy, typename QueuePolicy>
+Checkpoint &Executor<WorkspacePolicy, QueuePolicy>::GetCurrentCheckpoint() {
+  auto &cpt = GetCurrentIterationData(output_iteration_id_).checkpoint;
```
stiepan, line 704: As for validation, should we gaurd it with a check that checkpoiting is enabled? Or will pipeline do that?

stiepan (approved):
Looks good! I left two comments with remarks about error messages. As this is not user-facing yet, I think you may as well address them in the pipeline PR. Or now, up to.

PR #5009, opened 2023-08-21
Fix processing cost calculator: the order of filter support

stiepan:
## Category:

## Description:
In the setup, the resize operator estimates the cost of running the resize with different orders of the resampled axes. It uses output, input sizes and the size of filters used along particular axes. However, it seems that the filters support size is unnecesarily reversed when passed to the best order calculator, which may result in sub-optimal performance in some cases.

The performance of the dali resize operator, esp. for downsampling with antialias on and filters with big radious.

While the shapes and arguments come in a (D)HW (limited to spatial dims) layout, the SampleDesc keeps the extents in the WH(D) order. The filters are setup in the `SetFilters` method called a few lines above the fix. They are kept in the SampleDesc in the WH(D) order, same as the computed input (roi) and output shapes here. 

### Tests:

- [ ] Existing tests apply
- [ ] New tests added
  - [ ] Python tests
  - [ ] GTests
  - [ ] Benchmark
  - [ ] Other
- [x] N/A

### Documentation
- [ ] Documentation updated
  - [ ] Docstring
  - [ ] Doxygen
  - [ ] RST
  - [ ] Jupyter
  - [ ] Other
- [x] N/A

#### Requirements
- [ ] N/A

**REQ IDs**: N/A

**JIRA TASK**: N/A

### DIFF
--- a/dali/kernels/imgproc/resample/resampling_setup.cc
+++ b/dali/kernels/imgproc/resample/resampling_setup.cc
@@ -1,4 +1,4 @@
-// Copyright (c) 2019-2022, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+// Copyright (c) 2019-2023, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


banasraf (approved):

stiepan:
!build

mzient (approved):

PR #5010, opened 2023-08-21
Added Devcontainer

5had3z:
## Category:

New Feature

## Description:

Devcontainer is a fancy docker-compose that intergrates tightly with VSCode also enables the automatic installing of commonly used extensions. This enables contributors to simply run the command "Dev Containers: Rebuild and Reopen in Container" to get working in a bare-metal like manner with all dependencies compiled and installed. From there you can compile, run and debug DALI in the single environment.

The only steps to do for a new developer are:
 - git clone recursive
 - "Rebuild and Reopen in Container"
 - run init_repo.sh in cvcuda
 - select GCC-11.4 from cmake extension + click build button
 - cd build && pip instsall dali/python
 - run tests and debug etc.

New dockerfile in .devcontainer folder based on standard cuda development container. Dependencies included have sufficied for me to develop with and run operator_2 test suite #5007.

Extensions added to devcontainer include C++, CMake, Nsight, Python. By default the devcontainer will try to run with a GPU.

DALI_deps is built, installed then removed to save space on final image size. There is potentially a better way to automatically insert the DALI_DEPS_VERSION_SHA to the Dockerfile than an Arg so the correct commit is used at build time. DALI_extra on the otherhand is needed for testing so that is cloned and retained, so the target SHA doesn't really need to be inserted during image build.

I didn't attempt setting up the environment for working with the tensorflow plugin part of the library.

A few extra packages included from the test-toolchain ppa like gcc-13 and tbb since I may probe a C++23 port later, but this can be removed (doesn't harm anything for now).

Need to discuss how to update compilation.rst, I personally prefer this workflow over docker/build.sh.

### Tests:

The only test perse is the user experience to be able to quickly get a development environment up from scratch with this system.

- [ ] Existing tests apply
- [ ] New tests added
  - [ ] Python tests
  - [ ] GTests
  - [ ] Benchmark
  - [ ] Other
- [x] N/A

### Documentation
- [ ] Documentation updated
  - [ ] Docstring
  - [ ] Doxygen
  - [ ] RST
  - [ ] Jupyter
  - [ ] Other
- [ ] N/A

#### Requirements
- [ ] N/A

**REQ IDs**: N/A

**JIRA TASK**: N/A

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,0 +1,60 @@
+# Base CUDAToolkit image to develop DALI, comment/uncomment 11.x or 12.x
+FROM nvcr.io/nvidia/cuda:12.1.0-cudnn8-devel-ubuntu22.04
+# FROM nvcr.io/nvidia/cuda:11.8.0-cudnn8-devel-ubuntu22.04
+
+# Add required packages for DALI_deps
+RUN --mount=type=cache,target=/var/cache/apt apt-get update && \
+    DEBIAN_FRONTEND=noninteractive apt-get install -y \
+    git g++ wget unzip ninja-build autoconf libtool yasm vim \
+    software-properties-common unzip python3 python-is-python3
+
+# Add nvJPEG2k and nvCOMP since it isn't included in CUDA Toolkit, 
+# CUDA_VERSION_MAJOR should reflect the base toolkit container used above.
+RUN NVJPEG2K_VERSION=0.7.0.31-1 && \
+    NVCOMP_VERSION=2.6.1 && \
+    CUDA_VERSION_MAJOR=12 && \
+    apt-key adv --fetch-keys https://developer.download.nvidia.com/compute/cuda/repos/ubuntu2204/x86_64/3bf863cc.pub && \
+    add-apt-repository "deb https://developer.download.nvidia.com/compute/cuda/repos/ubuntu2204/x86_64/ /" && \
+    apt-get update && \
+    DEBIAN_FRONTEND=noninteractive apt-get install -y libnvjpeg2k0=${NVJPEG2K_VERSION} libnvjpeg2k-dev=${NVJPEG2K_VERSION} && \
+    cp /usr/include/libnvjpeg2k/${CUDA_VERSION_MAJOR}/* /usr/local/cuda/include/ && \
+    cp /usr/lib/x86_64-linux-gnu/libnvjpeg2k/${CUDA_VERSION_MAJOR}/* /usr/local/cuda/lib64/ && \
+    rm -rf /var/lib/apt/lists/* && \
+    mkdir nvcomp && \
+    cd nvcomp && \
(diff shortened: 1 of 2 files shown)


klecki:
Thanks for the contribution.
This looks cool, please allow me a few days to evaluate and test the devcontainer, I never used it before.

I think after we evaluate this, we will try to accept this, and than proceed with adjustments/generalization as next steps. One thing we need to figure out is how to plug this into CI, so we can verify that the container allows for correct compilation of latest DALI and isn't getting out of date.

When we have that streamlined + possibly add a handling for the TF plugin & custom ops, we will adjust the documentation advertising this as an option for development.

As a side note, I am thinking if we can utilize any of the dockerfiles that are used for the CI/build.sh and their versioning here, if you didn't see it, here is an example of the latest one: https://github.com/NVIDIA/DALI/blob/main/docker/Dockerfile.cuda122.x86_64.deps. To be clear, I am not asking for changes in this PR or generalization of the dockerfile, but if you have any suggestions in that area I would be happy to discuss what may work best for a developer using this feature.

One more note regarding the plugin list, we have a config file for yapf formatter: https://github.com/NVIDIA/DALI/blob/main/.style.yapf and our current formatting would probably generate less formatting changes using `yapf` instead of `black`.

JanuszL:
Hi @5had3z,

Thank you for pointing out the gap and providing the initial implementation.
We discussed this internally and we are afraid that this approach may be hard to maintain in the long run. That is why we started working on something that can reuse the parts that are changing most frequently in our setup and have it tested in our CI. The initial PR we want to extend further has been posted https://github.com/NVIDIA/DALI/pull/5017. Feel free to watch it and share your comments when it becomes more mature.

5had3z:
@JanuszL 
I'm wondering if the dockerfile system can be simplified a bit with cuda minor version compatibility? Could all images be just based on the latest minor version, so there's only one 11 and 12 image (and use ngc base as I have done here)? Correct me if I'm wrong with thinking how it works. Also, I'm not sure why 9 and 10 are still hanging around since they're not supported anymore.

If toolchain and dependency things don't change too often, you could build the dev image in CI itself and host the developer image in NGC, so the devcontainer.json becomes
```
{
    "image": "nvcr.io/nvidia/dali-dev:cu122",
    "customizations": {
            ....
```
Maybe save a bit of energy so many people can just download instead of download + compile.

I'll close this here so any future conversation can happen in #5017.

JanuszL:
Hi @5had3z,

> I'm wondering if the dockerfile system can be simplified a bit with cuda minor version compatibility? Could all images be just based on the latest minor version, so there's only one 11 and 12 image (and use ngc base as I have done here)

We already follow [that](https://github.com/NVIDIA/DALI/blob/main/docs/compilation.rst#building-python-wheel).

> Also, I'm not sure why 9 and 10 are still hanging around since they're not supported anymore.

We keep them for eager hackers who may want to build DALI with their own setup. It might work although there is no guarantee as we do not test this.

> If toolchain and dependency things don't change too often, you could build the dev image in CI itself and host the developer image in NGC, so the devcontainer.json becomes

Not that often usually means once every month or two. It is technically possible although this adds significant engineering effort to test and validate provided images. On top of that, we need to patch all CVEs discovered in such images. For now, we would prefer to let everyone build this image on their own.

PR #5012, opened 2023-08-21
Support video in predefined automatic augmentations

stiepan:
## Category:

## Description:
Adds basic support for video to AA/RA/TA auto augments.

## Additional information:
In principle, the required changes were minimal at this point - just assuring that the right axes are used in the auto_contrast and constrast predefined augmentations. The remaining changes are: 1. adjusted tests for predefined augmentations, 2. added basic tests to AA/RA/TA on video input, 3. adjusted and clarified the documentation.

auto_aug module

### Tests:

- [ ] Existing tests apply
- [x] New tests added
  - [x] Python tests
  - [ ] GTests
  - [ ] Benchmark
  - [ ] Other
- [ ] N/A

### Documentation
- [x] Documentation updated
  - [x] Docstring
  - [ ] Doxygen
  - [ ] RST
  - [ ] Jupyter
  - [ ] Other
- [ ] N/A

#### Requirements
- [ ] N/A

**REQ IDs**: N/A

**JIRA TASK**: DALI-3594

### DIFF
--- a/dali/python/nvidia/dali/auto_aug/augmentations.py
+++ b/dali/python/nvidia/dali/auto_aug/augmentations.py
@@ -120,6 +120,7 @@ def contrast(data, parameter):
     It follows PIL implementation of Contrast enhancement which uses a channel-weighted
     mean as a contrast center.
     """
+    # assumes FHWC or HWC layout
     mean = fn.reductions.mean(data, axis_names="HW", keep_dims=True)
     rgb_weights = types.Constant(np.array([0.299, 0.587, 0.114], dtype=np.float32))
     center = fn.reductions.sum(mean * rgb_weights, axis_names="C", keep_dims=True)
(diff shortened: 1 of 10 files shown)


dali/python/nvidia/dali/auto_aug/augmentations.py
```diff
@@ -120,9 +120,10 @@ def contrast(data, parameter):
     It follows PIL implementation of Contrast enhancement which uses a channel-weighted
     mean as a contrast center.
     """
-    mean = fn.reductions.mean(data, axes=[0, 1])
+    # assumes FHWC or HWC layout
+    mean = fn.reductions.mean(data, axis_names="HW", keep_dims=True)
```
klecki, line 124: I wonder if we should normalize over the each frame individually or over the whole sequence. The latter may result in a better temporal stability of the produced output, but I am not sure what kind of sequences we would need to work with, for this problem to actually appear.

stiepan: That's very good point. To sum up the off-line discussion, it seems to be done on per-frame basis in other frameworks. If needed we could include the other behaviour as an option, but for now I'd go with what is already in the wild to avoid surprises. For the record, this consideration applies also to auto_contrast and equalize that rely on some global statistics as well.

klecki (approved):

awolant (approved):

PR #5016, opened 2023-08-23
Add opt-in support for case insensitive webdataset

JanuszL:
- the current implementation of webdataset is case-sensitive when
  it comes to file extensions. This PR adds an option to make
  it case insensitive when it comes to provided by the user extensions
  as well as the file extensions in the container

## Category:

## Description:
- the current implementation of webdataset is case-sensitive when
  it comes to file extensions. This PR adds an option to make
  it case insensitive when it comes to provided by the user extensions
  as well as the file extensions in the container

- webdataset reader

### Tests:

- [ ] Existing tests apply
- [x] New tests added
  - [x] Python tests
    - test_webdataset_requirements.test_case_sensitive_arg_format
    - test_webdataset_requirements.test_case_sensitive_container_format
    - test_webdataset_requirements.test_case_insensitive_container_format
    - test_webdataset_requirements.test_case_insensitive_arg_format
  - [ ] GTests
  - [ ] Benchmark
  - [ ] Other
- [ ] N/A

### Documentation
- [x] Documentation updated
  - [x] Docstring
  - [ ] Doxygen
  - [ ] RST
  - [ ] Jupyter
  - [ ] Other
- [ ] N/A

#### Requirements
- [x] N/A

**REQ IDs**: RDWDS.26

**JIRA TASK**: N/A


dali/operators/reader/webdataset_reader_op.cc
```diff
@@ -123,6 +123,14 @@ with a semicolon (';') and may contain dots.
...
 Example: "left.png;right.jpg")code",
             DALI_STRING_VEC)
+    .AddOptionalArg("case_insensitive_extensions",
+      R"code(Determines whether the extensions provided via the `ext` should be case insensitive.
+
+Allows mixing case sizes in the `ext` argument as well as in the webdataset container. For example
+when turned on: jpg, JPG, jPG should work.
+
+If the extension characters cannot be represented as ASCI the result of this option is undefined.
+)code", false)
```
stiepan, line 133: Isn't the default `case_insensitive_extensions=false` a double negation? We say that, "we are not case insensitive by default". Wouldn't the "we are case senstivie by default" message be clearer?

As as +1 for the positive variant of the name is that we already have `case_sensitive_filter` param in file reader.

dali/test/python/reader/test_webdataset_requirements.py
```diff
@@ -274,6 +274,150 @@ def test_pax_format():
...
+def test_case_sensitive_container_format():
+    global test_batch_size
+    num_samples = 1000
+    tar_file_path = os.path.join(get_dali_extra_path(), "db/webdataset/MNIST/devel-0.tar")
+    case_insensitive_tar_file_path = os.path.join(get_dali_extra_path(),
+                                                  "db/webdataset/case_insensitive/devel-0.tar")
+    index_file = generate_temp_index_file(tar_file_path)
+
+    num_shards = 100
+    with assert_raises(AssertionError):
```
stiepan, line 286: nitpick: Can you add the glob/regex for the expected error message?

dali/test/python/reader/test_webdataset_requirements.py
```diff
@@ -274,6 +274,150 @@ def test_pax_format():
         )
 
 
+def test_case_sensitive_container_format():
+    global test_batch_size
```
stiepan, line 278: I can see this is used in other test cases in this file, but is it needed? It seems to only unnecessary circumvent Python protection against accidental modification of variables in global scope.

dali/test/python/reader/test_webdataset_requirements.py
JanuszL: Done

dali/operators/reader/webdataset_reader_op.cc
JanuszL: Done

dali/test/python/reader/test_webdataset_requirements.py
JanuszL: Apparently not. Removed.

dali/operators/reader/webdataset_reader_op.cc
```diff
@@ -123,6 +123,14 @@ with a semicolon (';') and may contain dots.
 
 Example: "left.png;right.jpg")code",
             DALI_STRING_VEC)
+    .AddOptionalArg("case_sensitive_extensions",
+      R"code(Determines whether the extensions provided via the `ext` should be case sensitive.
+
+Allows mixing case sizes in the `ext` argument as well as in the webdataset container. For example
+when turned on: jpg, JPG, jPG should work.
+
+If the extension characters cannot be represented as ASCI the result of this option is undefined.
```
stiepan, line 132: Oh, but now we need to warn against turing it off.

JanuszL: Missed that. Fixed.

stiepan: I can see the update in the example description, but ther's also the warning 

```suggestion
If the extension characters cannot be represented as ASCI the result of turing this option off is undefined.
```

stiepan (approved):
Please remeber to adjust the warning in the docs as well, apart from that LGTM.

stiepan (approved):

dali/operators/reader/webdataset_reader_op.cc
JanuszL: Done

banasraf (approved):

PR #5017, opened 2023-08-24
Provide a dev container for DALI

JanuszL:
### Tests:

- [ ] Existing tests apply
- [ ] New tests added
  - [ ] Python tests
  - [ ] GTests
  - [ ] Benchmark
  - [ ] Other
- [ ] N/A

### Documentation
- [ ] Documentation updated
  - [ ] Docstring
  - [ ] Doxygen
  - [ ] RST
  - [ ] Jupyter
  - [ ] Other
- [ ] N/A

#### Requirements
- [ ] N/A

**REQ IDs**: N/A

**JIRA TASK**: N/A


docker/Dockerfile.ubuntu
```diff
@@ -0,0 +1,94 @@
+ARG DEPS_IMAGE_NAME
```
klecki, line 1: I think that we should consider skipping this container, unless we want to have an option to tinker with different targets in dev container.
Still, this is mostly an artifact of how we use it in CI, and most of those options can be controlled directly with CMake invocation inside the dev container.

The next step would be a layer that adds the tools for testing/debugging instead of this one.

JanuszL: For now, I would leave it as it is and use it to create test env. When it is done, we can merge it with Dockerfile.deps.ubuntu (unless there is a part that needs to have cuda installed already).

docker/Dockerfile.ubuntu
```diff
@@ -0,0 +1,153 @@
...
+# clean builder without source code inside
+FROM ${DEPS_IMAGE_NAME} as builder
+
+WORKDIR /opt/dali
+
+# to be adjusted later, first it to run from the build dir, the second is for nose test
+# so the user can `python3 -m nose_wrapper --verbose`
+ENV PYTHONPATH=/opt/dali/test_comp/dali/python/:/opt/dali/qa
```
JanuszL, line 9: Thank to that you can just
```
cd dali/test/python/
python3 -m nose2 --verbose -s operator_1 test_constant
```
or
```
cd dali/test/python/
python3 -m nose_wrapper --verbose
```
Any suggestion how to improve that?

PR #5018, opened 2023-08-24
Refactor ops into a submodule

klecki:
## Description:
This is first part of refactoring, with the following steps:
* ops.py was moved into ops/__init__.py
* Utilities related to keeping the names of registered operators were moved to a dedicated _registry.py file.
* Utilities for parsing and reporting name based on schema name were moved to a dedicated _names.py file.
* Docstring generation was moved to dedicated _docs.py file.
* Some docstrings were added.

The visibility of the symbols placed in the new files was kept the same as in the original code to not cause cascading changes in DALI.

The next steps will involve:
* Extending the registration of the custom operator to allow moving the code into separate files (like PythonOp or TFRecordReader).
* Adjusting the processing of inputs and arguments, and filling the OpSpec, so the code can better maintained between the two APIs.

nvidia/dali/ops.py -> nvidia/dali/ops/...

Did we miss any symbols?
Can the adjusted imports cause any problems? - probably a question for CI.
Check if the utilities that rely on the internals, like generation of the sphinx docs, still work as intended.

### Tests:

- [x] Existing tests apply
This changes a placement of the ops module code, but everything should work the same, so all current tests apply, documentation should be generated in the same way as well.
- [ ] New tests added
  - [ ] Python tests
  - [ ] GTests
  - [ ] Benchmark
  - [ ] Other
- [ ] N/A

### Documentation
- [x] Documentation updated
  - [x] Docstring
  - [ ] Doxygen
  - [ ] RST
  - [ ] Jupyter
  - [ ] Other
- [ ] N/A

#### Requirements
- [ ] N/A

**REQ IDs**: N/A

**JIRA TASK**: N/A

### DIFF
--- a/dali/python/nvidia/dali/ops/__init__.py
+++ b/dali/python/nvidia/dali/ops/__init__.py
@@ -25,15 +25,19 @@ from nvidia.dali import fn as _functional
 from nvidia.dali import internal as _internal
 from nvidia.dali.data_node import DataNode as _DataNode
 from nvidia.dali.pipeline import Pipeline as _Pipeline
-from nvidia.dali.types import \
-        _type_name_convert_to_string, _type_convert_value, _default_converter, \
-        _vector_element_type, _bool_types, _int_like_types, _float_types, \
-        DALIDataType, \
-        CUDAStream as _CUDAStream, \
-        ScalarConstant as _ScalarConstant, \
-        Constant as _Constant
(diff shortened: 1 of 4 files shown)


klecki:
!build

klecki:
!build

awolant (approved):

szalpal (approved):

PR #5024, opened 2023-08-29
Add ability to serialize/deserialize Checkpoint

staniewzki:
## Category:

## Description:

This PR adds new protobuf message and a way to translate it into a Checkpoint object and the other way around. 

Opaque `std::any` in OpCheckpoint was replaced with `std::variant` for strong typing in the serialization process.

### Tests:

- [ ] Existing tests apply
- [ ] New tests added
  - [ ] Python tests
  - [ ] GTests
  - [ ] Benchmark
  - [ ] Other
- [ ] N/A

### Documentation
- [ ] Documentation updated
  - [ ] Docstring
  - [ ] Doxygen
  - [ ] RST
  - [ ] Jupyter
  - [ ] Other
- [ ] N/A

#### Requirements
- [ ] N/A

**REQ IDs**: N/A

**JIRA TASK**: N/A

### DIFF
--- a/dali/operators/random/rng_base.h
+++ b/dali/operators/random/rng_base.h
@@ -44,7 +44,9 @@ class RNGBase : public Operator<Backend> {
   void RestoreState(const OpCheckpoint &cpt) override {
     if constexpr (std::is_same_v<Backend, CPUBackend>) {
       auto snapshot = cpt.CheckpointState<RNGSnapshotCPU64>();
-      DALI_ENFORCE(static_cast<int>(snapshot.rng.size()) == max_batch_size_);
+      DALI_ENFORCE(static_cast<int>(snapshot.rng.size()) == max_batch_size_,
+                   "Provided checkpoint doesn't match the expected batch size. "
+                   "Perhaps the batch size setting changed? ");
       rng_ = BatchRNG<std::mt19937_64>::FromVector(snapshot.rng);
     } else {
       static_assert(std::is_same_v<Backend, GPUBackend>);
--- a/dali/operators/random/rng_base.h
+++ b/dali/operators/random/rng_base.h
@@ -34,7 +34,7 @@ class RNGBase : public Operator<Backend> {
  public:
   void SaveState(OpCheckpoint &cpt, std::optional<cudaStream_t> stream) override {
     if constexpr (std::is_same_v<Backend, CPUBackend>) {
-      cpt.MutableCheckpointState() = rng_;
+      cpt.MutableCheckpointState() = RNGSnapshotCPU64{rng_.ToVector()};
     } else {
       static_assert(std::is_same_v<Backend, GPUBackend>);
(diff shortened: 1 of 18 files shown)


dali/pipeline/proto/dali.proto
```diff
@@ -74,3 +74,34 @@ message PipelineDef {
...
+    message DummySnapshot {
+      repeated uint32 dummy_state = 1;
+    }
+
+    optional string operator_name = 1;
+    oneof data {
+      RNGSnapshotCPU rng_cpu = 2;
+      RNGSnapshotCPU64 rng_cpu64 = 3;
+      LoaderStateSnapshot loader = 4;
+      DummySnapshot dummy = 5;
```
stiepan, line 102: If we cannot reasonably avoid this one, maybe let us make it first in the oneof sequence? So that with more types to come this one is not hidden in the middle?

dali/pipeline/operator/checkpointing/checkpointing_data.h
```diff
@@ -0,0 +1,71 @@
...
+
+/**
+ * @brief Universal checkpoint data storage type
+*/
+using CheckpointingData = std::variant<
+  RNGSnapshotCPU,
+  RNGSnapshotCPU64,
+  LoaderStateSnapshot,
+  DummySnapshot
+>;
```
stiepan, line 67: Now, all the stateless operators have RNGSnapshotCPU implictly.

dali/pipeline/operator/checkpointing/checkpoint.cc
stiepan, line 1: We need to somewhere validate the incoming checkpoints, for example the max batch size (frankly that's only example that comes to my mind at the moment). I think it is dependent on the operator logic so the particular RestoreState method should do on per-op basis. In any case, the validation is necessary, otherwise we can easily crash if we assign a vector of rng states that does not match the length expected by the operator

dali/pipeline/operator/checkpointing/op_checkpoint.h
stiepan, line 1: I guess this may be the point when we need to consider the versioning - so that the version number can get serialized.

stiepan (approved):

dali/operators/random/rng_base.h
```diff
@@ -43,7 +43,9 @@ class RNGBase : public Operator<Backend> {
 
   void RestoreState(const OpCheckpoint &cpt) override {
     if constexpr (std::is_same_v<Backend, CPUBackend>) {
-      rng_ = cpt.CheckpointState<BatchRNG<std::mt19937_64>>();
+      auto snapshot = cpt.CheckpointState<RNGSnapshotCPU64>();
+      DALI_ENFORCE(static_cast<int>(snapshot.rng.size()) == max_batch_size_);
```
stiepan, line 47: Could you add here (and in the random crop) a nice a error message about the incorrect checkpoint that is likely due to different batch size between the checkpointed pipeline and the current pipeline?

PR #5028, opened 2023-08-30
Refactor custom operator wrappers into separate files from ops.py

klecki:
## Description:
Second part of ops module refactoring.

Created a dedicated ops/_operators module, where code for additional custom wrappers
of DALI operator was moved. Those operator are later reexposed and registered in the
original nvidia.dali.ops (and fn) modules.

Custom wrappers that were moved:
* Compose "operator" (only class API).
* Helper code for math/arithmetic operator
* Python Function operator family
* TFRecord Reader - the code for the conditional loading was adjusted to not rely on exceptions.

nvidia.dali.ops

Any symbol missing?

### Tests:

- [x] Existing tests apply
This change should not impact the functionality, just rearranges the code, so all tests apply.
- [ ] New tests added
  - [ ] Python tests
  - [ ] GTests
  - [ ] Benchmark
  - [ ] Other
- [ ] N/A

### Documentation
- [x] Documentation updated
  - [x] Docstring
  - [ ] Doxygen
  - [ ] RST
  - [ ] Jupyter
  - [ ] Other
- [x] N/A

#### Requirements
- [x] N/A

**REQ IDs**: N/A

**JIRA TASK**: N/A


klecki:
!build

jantonguirao (approved):

klecki:
!build

dali/python/nvidia/dali/ops/_operators/__init__.py
stiepan, line 1: Likely a silly question, but why nothing happens in here apart from the docstring?

stiepan, line 1: Edit of the previous comment, I first expected you'd collect those utils from submodules here and re-export them here, that's why I was surprised to see only the docstring here. But that's fine I guess.

klecki: I put it here for the import machinery to work, but I just import directly from `_operators/something`. Otherwise I would just import it twice, I don't think we need this `ops/_operators` to do much. Sadly, we use the utils here and there, so I can't just import the op classes and be done with it.

stiepan (approved):

PR #5038, opened 2023-09-06
Replace GPU dltensor per-sample copying kernel with a batched one

stiepan:
## Category:

## Description:
Replace per-sample copying kernel for GPU DL Tensors with a batched one for better performance.

This is done in copy_with_stride - the backend templated `CopyWithStride` that copied a single sample is replaced with the `CopyDlTensorCpu` which just keeps the old implementation and the `CopyDlTensorBatchGpu` that handles gpu tensors in batches.

## Additional information:

The current approach is to copy each tensor separately: either with cudaMemCopyAsync if the tensor does not provide `.strides` (according to dlpack it must be compact then) or run a kernel that addresses each byte separately: from flat output (compact, as that is DALI tensor) index to coordinates and then to input flat index. The latter case was a major bottleneck - in case of CV-CUDA ColorTwist called in `fn.dl_tensor_python_function` vs the native `fn.color_twist`, the e2e throughput drops 4 times.

This PR 
1. keeps per-sample cudaMemCopyAsync copy path for tensors with no strides and extends it to tensors that do have strides but they are compact (i.e. match output DALI tensors strides).
2. the remaining tensors from the input are grouped into a batch and a batched copy-kernel is launched.

The copy-kernel aims to speed-up the copy:
1. instead of always using byte as input and output type, the underlaying element type is used (so that tensors of types bigger than byte do faster reads and writes).
2. relying on the fact that the output tensor is compact, the writes are vectorized - each thread reads four items seprately but stores them in the vectorized type and writes in a singe go. If the input tensor happens to be close to compact, the caches should help here as well. This is the case for CV-CUDA outputs, as they striding boils down to padding row size. Note, this requires separate handling of the possibly unaligned start and end of the output tensor.
3. The translation of output_index to input_index stops early if the inner most strides match. In case of CV-CUDA images it means we need to do the div, mod arithmetic only for the first stride, which speeds up the copy significantly.
4. The common number of dimensions for tensor (0-5) are handled statically for smaller register pressure.

In total, a generic use case should benefit from a single kernel launch and vectorized writes, while the cases akin to CV-CUDA row padded tensors should see major boost. If needed, one could optimize other cases in the future, such as transposed views.

The `copy_with_stride` utility was used only in the family of python_functions.

### Tests:

- [ ] Existing tests apply
- [x] New tests added
  - [x] Python tests
  - [x] GTests
  - [ ] Benchmark
  - [ ] Other
- [ ] N/A

### Documentation
- [ ] Documentation updated
  - [ ] Docstring
  - [ ] Doxygen
  - [ ] RST
  - [ ] Jupyter
  - [ ] Other
- [x] N/A

#### Requirements
- [x] N/A

**REQ IDs**: N/A

**JIRA TASK**: DALI-3609

### DIFF
--- a/dali/operators/python_function/dltensor_function.cc
+++ b/dali/operators/python_function/dltensor_function.cc
@@ -1,4 +1,4 @@
-// Copyright (c) 2019-2022, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+// Copyright (c) 2019-2023, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -148,7 +148,7 @@ void CopyOutputData(TensorList<CPUBackend> &output, std::vector<DLMTensorPtr> &d
   auto out_shape = output.shape();
   for (int i = 0; i < batch_size; ++i) {
     thread_pool.AddWork([&, i](int) {
-      CopyDlTensor<CPUBackend>(output.raw_mutable_tensor(i), dl_tensors[i]);
+      CopyDlTensorCpu(output.raw_mutable_tensor(i), dl_tensors[i]);
     }, out_shape.tensor_size(i));
   }
   thread_pool.RunAll();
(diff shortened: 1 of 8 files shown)


stiepan:
!build

stiepan:
!build

awolant (approved):

stiepan:
Added a bunch of python tests (for sliced, permuted, flipped, and strided) views on the input data.

stiepan:
!build

szalpal (approved):

PR #5039, opened 2023-09-06
Refactor ops and fn APIs

klecki:
## Description: 
This part of refactoring moves the most important parts of handling the argument
parsing, deprecation, constant promotion, conditionals, etc to the _OperatorInstance.
In case of ops API, there needs to be a pre-step that merges the state
from `__init__` and `__call__` and is specific that API.
For the `fn` API, apart from promoting constants first (due to the perf reasons),
we can handle the creation of `_OperatorInstance` and filling the `spec` in one go.

`name` is no longer included in argument inputs but considered scalar argument.

It tries to keep most of the implementation in one place, but also allows for detaching
the ops and fn APIs from one another.
The question is, how much of this change we want.

Additionally, some details around the conditionals are streamlined, allowing
to enable them in multiple input sets.

Next steps:
Python function, TFRecord reader and external source implement their Operator class
without inheriting from the one generated in op factory, so they repeat some code.
We probably can refactor it and unify them with the generic code path even more.
Make the special arguments (name, device), less special so they are added to the
spec through generic code-path.

fn, ops

Should we do the thing that is put in the fn part as a comment?

### Tests:

- [x] Existing tests apply
- [x] New tests added
  - [x] Python tests
  - [ ] GTests
  - [ ] Benchmark
  - [ ] Other
- [ ] N/A

### Documentation
- [x] Documentation updated
  - [x] Docstring
  - [ ] Doxygen
  - [ ] RST
  - [ ] Jupyter
  - [ ] Other
- [ ] N/A

#### Requirements
- [ ] N/A

**REQ IDs**: N/A

**JIRA TASK**: N/A

### DIFF
--- a/dali/python/nvidia/dali/ops/__init__.py
+++ b/dali/python/nvidia/dali/ops/__init__.py
@@ -356,7 +356,6 @@ class _OperatorInstance(object):
     def check_args(self):
         self._op.schema.CheckArgs(self._spec)
 
-
     def _process_instance_name(self, arguments):
         """Extract the name from the arguments or generate the default one.
         The "name" is removed from arguments.
@@ -371,7 +370,6 @@ class _OperatorInstance(object):
             self._name = name
         else:
             self._name = '__' + type(self._op).__name__ + "_" + str(self._counter.id)
-
 
     def _generate_outputs(self):
         pipeline = _Pipeline.current()
@@ -546,7 +544,6 @@ def python_op_factory(name, schema_name=None):
                     f"Operator {type(self).__name__} expects "
                     f"from {self._schema.MinNumInput()} to {self._schema.MaxNumInput()} inputs, "
                     f"but received {len(inputs)}.")
-
 
     Operator.__name__ = str(name)
     Operator.schema_name = schema_name or Operator.__name__
--- a/dali/python/nvidia/dali/ops/__init__.py
+++ b/dali/python/nvidia/dali/ops/__init__.py
@@ -321,13 +321,28 @@ class _OperatorInstance(object):
     Each "_process_[arguments/inputs]" step adds them to the OpSpec specifying the operator for
     the C++ backend.
 
-    Parameters
-    ----------
-    object : _type_
-        _description_
+    * some validation is done in class Operator.__init__ due to legacy reasons.
+    This is the reason for `_processed_arguments` constructor parameter.
     """
 
-    def __init__(self, inputs, arguments, arg_inputs, op):
+    def __init__(self, inputs, arg_inputs, arguments, _processed_arguments, op):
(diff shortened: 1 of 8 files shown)


klecki:
!build

klecki:
!build

klecki:
!build

dali/python/nvidia/dali/ops/__init__.py
```diff
@@ -206,75 +311,67 @@ def _add_spec_args(schema, spec, kwargs):
...
+    the C++ backend.
 
-    def __init__(self, inputs, op, **kwargs):
+    Parameters
+    ----------
+    object : _type_
+        _description_
+    """
+
+    def __init__(self, inputs, arguments, arg_inputs, op):
```
klecki, line 330: Consider propagating the arguments, that were already added to the spec in __init__ for full visibility.

szalpal (approved):

klecki:
!build

klecki:
!build

dali/python/nvidia/dali/ops/__init__.py
```diff
@@ -189,6 +189,116 @@ def _handle_deprecations(schema, kwargs, op_name):
     return kwargs
 
 
+def _handle_op_deprecation(schema, op_name):
+    if schema.IsDeprecated():
+        # TODO(klecki): how to know if this is fn or ops?
```
stiepan, line 194: So for the class-based API we're going to suggest `fn.*` to now? Was it like that before?

dali/python/nvidia/dali/ops/__init__.py
```diff
@@ -189,6 +189,116 @@ def _handle_deprecations(schema, kwargs, op_name):
...
+        with warnings.catch_warnings():
+            warnings.simplefilter("default")
+            warnings.warn(msg, DeprecationWarning, stacklevel=2)
+
+
+def _resolve_double_definitions(current, previous, keep_old=True):
+    """Unify the arguments (or argument inputs) between __call__ and __init__ - ops API only.
+
+    If the same argument was passed in both invocation it should result in error, with exception,
+    of passing `None` in `__call__`.
```
stiepan, line 211:
```suggestion
    If the same argument was passed in both invocation it should result in error, with exception
    of passing `None` in `__call__`.
```

dali/python/nvidia/dali/ops/__init__.py
klecki: Yup, this was just moved around.
I think the best way would be to add some `_api_type` string and dispatch on that in errors (also propagate it to the backend as hidden arg possibly. But this is a case for follow-up.

stiepan (approved):

PR #5055, opened 2023-09-22
Fix the support of videos with different bith dept in the video reader

JanuszL:
- the video reader incorrectly assumes that all videos have the same bit depth.
  This PR makes sure that this is checked and that the decoder instance
  is recreated when it changes.

## Category:

## Description:
- the video reader incorrectly assumes that all videos have the same bit depth.
  This PR makes sure that this is checked and that the decoder instance
  is recreated when it changes.

- video reader

- NA

### Tests:

- [ ] Existing tests apply
- [x] New tests added
  - [x] Python tests
  - test_video_pipeline.py:test_10bit_vid_reconfigure
  - [ ] GTests
  - [ ] Benchmark
  - [ ] Other
- [ ] N/A

### Documentation
- [ ] Documentation updated
  - [ ] Docstring
  - [ ] Doxygen
  - [ ] RST
  - [ ] Jupyter
  - [ ] Other
- [x] N/A

#### Requirements
- [x] N/A

**REQ IDs**: RVID.26

**JIRA TASK**: DALI-3625


JanuszL:
https://github.com/NVIDIA/DALI_extra/pull/125

szalpal (approved):

dali/operators/reader/nvdecoder/cuvideodecoder.cc
```diff
@@ -161,12 +161,16 @@ void CUVideoDecoder::reconfigure(unsigned int height, unsigned int width) {
 }
 
 int CUVideoDecoder::initialize(CUVIDEOFORMAT* format) {
-    if (decoder_) {
+    while (decoder_) {
```
klecki, line 164: Why `while` if there is `return` at the end of this loop?

klecki: Oh, I missed the break. IMO, the logic is a bit convoluted, this is a bit `goto` like :P 
Can we somehow make it into
```
if (some condition)
   FAIL
if (other condition) 
   do something;
   return 1;
if (third condition)
   clear the state & continue with config?
```

JanuszL:
!build

klecki (approved):

PR #5063, opened 2023-09-27
Add implicit scope to batch_permutation

stiepan:
## Category:
**Bug fix** /**New feature** (*non-breaking change which adds functionality*)

## Description:

The PR makes the following code work.
```
batch =  fn.decoders.image(batch, device="mixed")
if fn.random.coin_flip():
        permuted_batch = fn.permute_batch(batch, indices=fn.batch_permutation())
```
Without it, the `fn.permute_batch` will fail due `indicies` exceeding the valid indicies for the `batch` in the inner scope.

Even though we call `fn.batch_permutation()` inside the if body, the operator will run as if it was called in the outer scope - that is it will produce a permutation for a full batch size. Then, that permutation will be split according to the `fn.random.coin_flip()` predicate.

Unfortunately, randomly splitting permutation in two smaller lists usually does not produce a valid permutations for smaller lists, i.e. while `[3, 0, 2, 1]` is a valid permutation of four elements, splitting it in two `[3, 0]`, `[2, 1]` does not produce valid permutations of two elements.

The hoisting of `fn.batch_permutation` into an outer scope happens becuase the fn.batch_permutation does not accept any tensor arguments - those args serve as a source of the "local scope" batch size. This PR fixes the problem by introducing an implicit "_scope" argument that is automatically injected with a dummy batch whose size matches the local batch size.

* OpSchema - by introduction of HiddenOptional arguments.
* fn.batch_permutation - it now has new parent schema with the `_scope` arg.
* ops.py - code that automatically injects a referential scope to those operators that have the `_scope` argument.

A possible follow-up would be to use the same mechanism for any operator that can potentially have no other tensor arugments. For example random generators - in their case, hositing to global scope and then splitting the output does not impact the correctness, but may incur some performence penalty.

Is it tested properly? Is the hidden `_scope` arg hidden in the docs? Anywhere else where you'd rather not see a *hidden* argument that I should check?

### Tests:

- [ ] Existing tests apply
- [x] New tests added
  - [x] Python tests
  - [ ] GTests
  - [ ] Benchmark
  - [ ] Other
- [ ] N/A

### Documentation
- [ ] Documentation updated
  - [ ] Docstring
  - [ ] Doxygen
  - [ ] RST
  - [ ] Jupyter
  - [ ] Other
- [ ] N/A

#### Requirements
- [ ] N/A

**REQ IDs**: N/A

**JIRA TASK**: DALI-3495

### DIFF
--- a/dali/pipeline/operator/builtin/conditional/scope_argument.cc
+++ b/dali/pipeline/operator/builtin/conditional/scope_argument.cc
@@ -1,0 +1,21 @@
+// Copyright (c) 2023, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "dali/pipeline/operator/operator.h"
+
+DALI_SCHEMA(ImplicitScopeAttr)
+    .AddOptionalArg<int>("_scope", R"code(Any batch to infer the current batch size from.
+
+If the operator has no other tensor inputs, this argument can serve as a source of a current
+batch size.)code", 0, true);
--- a/dali/pipeline/operator/op_schema.h
+++ b/dali/pipeline/operator/op_schema.h
@@ -1,4 +1,4 @@
-// Copyright (c) 2017-2022, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+// Copyright (c) 2017-2023, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -46,6 +46,10 @@ struct DefaultedArgumentDef {
   std::string doc;
   DALIDataType dtype;
   Value *default_value;
+  // As opposed to purely internal argument, the hidden argument
+  // can be specified as any other argument on per-operator basis, through Python API, etc.
+  // It is just hidden from the docs.
+  bool hidden;
 };
 
 struct DeprecatedArgDef {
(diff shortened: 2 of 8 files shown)


stiepan:
!build

dali/pipeline/operator/builtin/conditional/scope_argument.cc
```diff
@@ -0,0 +1,21 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "dali/pipeline/operator/operator.h"
+
+DALI_SCHEMA(ImplicitScopeAttr)
+    .AddHiddenOptionalArg<int>("_scope", R"code(Any CPU batch to infer the current batch size from.
```
mzient, line 18:
```suggestion
    .AddHiddenOptionalArg<int>("_scope", R"code(Any batch to infer the current batch size from.
```
That's just current limitation that we don't need to repeat here and may lift in the future.

awolant (approved):

dali/pipeline/operator/builtin/conditional/scope_argument.cc
stiepan: done

dali/pipeline/operator/op_schema.h
```diff
@@ -64,6 +68,20 @@ enum class InputDevice : uint8_t {
   GPU = 2,
 };
 
+inline void EnforceArgNameFormat(const std::string &name, bool hide_argument) {
```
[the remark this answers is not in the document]
stiepan: Good point, I've changed it for an arg to be hidden iff its name starts with "_".

stiepan:
!build

dali/pipeline/operator/op_schema.h
```diff
@@ -64,6 +68,10 @@ enum class InputDevice : uint8_t {
   GPU = 2,
 };
 
+inline bool HideArgument(const std::string &name) {
+  return name.size() && name[0] == '_';
+}
```
mzient, line 73: Can you move it inside the Schema class as a static function?
Also, the imperative name doesn't look like a boolean query.
Alternatives: ShouldHideArgument, ArgumentHidden, IsArgumentHidden or perhaps just `IsHidden(const string &arg_name)`

stiepan: good point

mzient (approved):

stiepan:
!build

klecki (approved):

PR #5065, opened 2023-09-28
Extend HW image decoder bench script to support multiple GPUs

JanuszL:
- adds an ability to run multiple data processing pipelines on
  multiple GPUs inside HW image decoder bench script

## Category:

## Description:
- adds an ability to run multiple data processing pipelines on
  multiple GPUs inside HW image decoder bench script

- tools/hw_decoder_bench.py

- NA

### Tests:

- [ ] Existing tests apply
- [ ] New tests added
  - [ ] Python tests
  - [ ] GTests
  - [ ] Benchmark
  - [ ] Other
- [x] N/A

### Documentation
- [x] Documentation updated
  - [x] Docstring
  - [ ] Doxygen
  - [ ] RST
  - [ ] Jupyter
  - [ ] Other
- [ ] N/A

#### Requirements
- [x] N/A

**REQ IDs**: N/A

**JIRA TASK**: N/A

### DIFF
--- a/tools/hw_decoder_bench.py
+++ b/tools/hw_decoder_bench.py
@@ -21,6 +21,8 @@ from nvidia.dali.pipeline import pipeline_def
 parser = argparse.ArgumentParser(description='DALI HW decoder benchmark')
 parser.add_argument('-b', dest='batch_size', help='batch size', default=1, type=int)
 parser.add_argument('-d', dest='device_id', help='device id', default=0, type=int)
+parser.add_argument('-n', dest='gpu_num',
+                    help='Number of GPUs used starting from device_id', default=1, type=int)
 parser.add_argument('-g', dest='device', choices=['gpu', 'cpu'],
                     help='device to use', default='gpu',
                     type=str)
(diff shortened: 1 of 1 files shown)


JanuszL:
!build

tools/hw_decoder_bench.py
```diff
@@ -21,6 +21,8 @@
 parser = argparse.ArgumentParser(description='DALI HW decoder benchmark')
 parser.add_argument('-b', dest='batch_size', help='batch size', default=1, type=int)
 parser.add_argument('-d', dest='device_id', help='device id', default=0, type=int)
+parser.add_argument('-n', dest='gpu_num',
+                    help='Number of GPUs used starting from device_id', default=1, type=int)
```
szalpal, line 25: I recall some issue that `device_ids` do not have to be consecutive (I might be wrong though). How about providing a list of devices instead? Something along the lines:
```suggestion
parser.add_argument('-d', dest='device_id', help='device_id', default=0, nargs='*')
```
And the usage would be:
```bash
python hw_decoder_bench.py -d 1 3 -b ...
python hw_decoder_bench.py -d 0 -b ...
```
```python
    for di in range(args.device_id):
        pipes.append(DecoderPipeline(device_id=di))
```

szalpal (approved):
Added a tiny suggestion, LGTM regardless.

tools/hw_decoder_bench.py
JanuszL: The numbers are consecutive, but the order may differ from the PCI order.
I rather imagine this script to be used to test on heterogenous systems and the user will just decide how many GPUs is willing to use.

awolant (approved):

PR #5081, opened 2023-10-05
Add user-friendly message about missing numpy

szkarpinski:
## Category:

## Description:

`numpy` is required to print `TensorList` and `Tensor`, but it's not a dependency of DALI. This means that if someone has no `numpy`, they get import error when trying to print a `Tensor`/`TensorList` (for example when following our _Getting started_). 

This PR adds a message explaining what happened.

The solution used is exactly the same as in other `numpy`-dependent places, such as External Source:
https://github.com/NVIDIA/DALI/blob/3e1123d58a72bf559096f8a16adebd2cbc47b78b/dali/python/nvidia/dali/_utils/external_source_impl.py#L22-L33

### Tests:

- [x] Existing tests apply
- [ ] New tests added
  - [ ] Python tests
  - [ ] GTests
  - [ ] Benchmark
  - [ ] Other
- [ ] N/A

### Documentation
- [ ] Documentation updated
  - [ ] Docstring
  - [ ] Doxygen
  - [ ] RST
  - [ ] Jupyter
  - [ ] Other
- [ ] N/A

#### Requirements
- [x] N/A

**REQ IDs**: N/A

**JIRA TASK**: N/A

### DIFF
--- a/dali/python/nvidia/dali/tensors.py
+++ b/dali/python/nvidia/dali/tensors.py
@@ -29,9 +29,23 @@ def _join_string(data, crop, edgeitems, sep=', '):
     return sep.join(data)
 
 
+np = None
+
+
+def import_numpy():
+    global np
+    if np is None:
+        try:
+            import numpy as np
+        except ImportError:
+            raise RuntimeError('Could not import numpy. Numpy is required for '
+                               'Tensor and TensorList printing. '
+                               'Please make sure you have numpy installed.')
+
+
 def _tensor_to_string(self):
     """ Returns string representation of Tensor."""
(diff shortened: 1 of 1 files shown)


JanuszL (approved):

mzient (approved):
I'm not entirely convinced that changing the error type is a good thing, but whatever.

JanuszL:
> I'm not entirely convinced that changing the error type is a good thing, but whatever.

From my point of view it shows that is a conscious decision from our side to not enforce numpy dependency as it is used only by some functionalities.

szkarpinski:
!build

PR #5083, opened 2023-10-06
Fix coverity issues 10/23

banasraf:
## Category:

**Other**

## Description:

It fixes issues detected by coverity

### Tests:

- [x] Existing tests apply
- [ ] New tests added
  - [ ] Python tests
  - [ ] GTests
  - [ ] Benchmark
  - [ ] Other
- [ ] N/A

### Documentation
- [ ] Documentation updated
  - [ ] Docstring
  - [ ] Doxygen
  - [ ] RST
  - [ ] Jupyter
  - [ ] Other
- [ ] N/A

#### Requirements
- [ ] N/A

**REQ IDs**: N/A

**JIRA TASK**: N/A


dali/pipeline/util/copy_with_stride.cu
```diff
@@ -351,7 +351,7 @@ void CopyDlTensorBatchGpu(TensorList<GPUBackend> &output, std::vector<DLMTensorP
   }
   int element_size, ndim;
   ValidateBatch(element_size, ndim, dl_tensors, batch_size);
-  SmallVector<strided_copy::StridedCopyDesc, 128> sample_descs;
+  SmallVector<strided_copy::StridedCopyDesc, 32> sample_descs;
```
banasraf, line 354: Stack allocation was too large

dali/pipeline/util/copy_with_stride.cu
```diff
@@ -181,7 +181,7 @@ DALI_DEVICE DALI_FORCEINLINE void AlignedCopy(const StridedCopyDesc &sample,
                                               MismatchedNdimT mismatched_ndim) {
   using T = typename ElementTypeDesc::type;
   using VecT = typename ElementTypeDesc::vec_type;
-  constexpr int vec_len = ElementTypeDesc::vec_len;
+constexpr int64_t vec_len = ElementTypeDesc::vec_len;
```
banasraf, line 184: It's later used in expressions which could overflow with 32bit types

awolant (approved):

dali/operators/python_function/dltensor_function.h
```diff
@@ -258,7 +258,10 @@ class DLTensorPythonFunctionImpl : public Operator<Backend> {
   }
 
   ~DLTensorPythonFunctionImpl() {
-    auto interpreter_lock = py::gil_scoped_acquire();
+    try {
+      auto interpreter_lock = py::gil_scoped_acquire();
+    } catch (...) {
+    }
```
mzient, line 264: Does it make sense to just swallow this exception? What are the chances that we'll ruin something? Perhaps letting the process terminate is exactly what we should do.

dali/pipeline/util/copy_with_stride.cu
mzient: indent

dali/pipeline/util/copy_with_stride.cu
mzient: Well, it worked, so I wouldn't say with any amount of certainty that it was "too large". For our target it seems to have been OK.... Your call.

dali/operators/python_function/dltensor_function.h
mzient, line 264: This code is broken now. This "lock" is like a lock guard. It goes out of scope and the lock is released; then we do everything outside of the lock.

Also, I don't think that letting this error just disappear is a good idea.

mzient (changes requested):
Broken GIL handling in ~DLTensorPythonFunctionImpl

dali/operators/python_function/dltensor_function.h
awolant: Wouldn't this clear up anyway since this is in the `~DLTensorPythonFunctionImpl` ?
I considered suggesting that this should kill the whole thing but I am not sure that it can occur?

dali/pipeline/util/copy_with_stride.cu
awolant: Strided copy tests have had some random test failures in the CI for a while. Maybe there are still some issue there.

dali/operators/python_function/dltensor_function.h
banasraf: Yeah, this should only throw in some extreme cases. If we cannot access GIL there's probably something more wrong with the process so maybe it's just better to let it terminate

banasraf:
!build

dali/pipeline/util/copy_with_stride.cu
mzient: If we ran out of stack it wouldn't be "some issues". It would be a hard fault (SIGSEGV? I don't think linux has a separate stack overflow error).

PR #5088, opened 2023-10-10
Add Fast-Forward to Loader base

szkarpinski:
## Category:

## Description:

This PR gives `Loader` a capability to fast-forward by _n_ steps, i.e. change its internal state to a state looking as if the `Loader` has returned _n_ more samples. This functionality will be later used for restoring mid-epoch checkpoints.

From the high level, fast-forwarding is done in two steps:

- _(Step 1)_ Simulate the `Loader` state for _n_ steps, but without touching the data: keep track of samples present in the buffer, but don't load them.

- _(Step 2)_ Check which samples need to be loaded and read them using `ReadSample`.

As _(Step 1)_ doesn't touch the data it should be relatively cheap. Cost of _(Step 2)_ depends on how fast the particular loader can load samples from given indices (i.e. how fast can it seek).

To implement _(Step 1)_ the following changes were made:

- **`IndexedLoadTargetSharedPtr`**, which is a pair of a sample index and tensor pointer, was introduced. Both `sample_buffer_` and `last_sample_ptr_tmp_` were changed to contain those instead of just pointers. This way we can keep track of samples
- `ReadOne` was extended with **_dry run_ mode**, controlled by `dry_run` argument. In a _dry_ mode `ReadOne` doesn't call `ReadSample` to read the sample, but instead just puts `nullptr` as the data pointer. This means that after running `ReadOne` in _dry_ mode for a few times the state of the loader is almost the same as it would be in normal mode, except for some samples having `nullptr` as their pointer. Those are the ones to be loaded in _(Step 2)_.

To implement _(Step 2)_ the following changes were made:
- `Loader` base class was extended with **`Skip(n)` method**, which skips reading _n_ samples. Efficient implementation of that method is trivial for some loaders (for example for file loader this just means adding _n_ to the pointer to current file) and harder for others (WDS, video).
- An inefficient **generic implementation of `Skip`** is provided by the baseclass: it simply reads samples and discards them instantly. It should be overriden by all loaders, but I put it here to provide some working implementation for all loaders.
- **`ReadMissingSamples` method** was added, which scans for samples having null pointer and reads them. As we only have `Skip` and thus are unable to seek backward freely, this method first lists all samples to be read and then reads them in a single forward pass through the data.
- To support fast-forward, loaders are required to provide **`Rewind` method**, which will reset the reader to the first sample. That's what `Reset` was probably originally meant to do, but now multiple loaders are also doing things like `current_epoch++` in their `Reset`s. `Rewind` is expected not to have any such side effects. Together with `Skip`, `Rewind` allows to move between samples to be read.

The following minor technical changes were made:
- Originally, the samples in the buffer were kept as `unique_ptr`s and transformed to `shared_ptr`s prior to returning. Now `ReadMissingSamples` might need to load the same sample into the buffer multiple times, so I decided to **keep samples in the buffer as shared pointers** as well, so now the `unique_ptr -> shared_ptr` transformation happens earlier.
- `ReadMissingSamples` uses `Reset`, `Skip` and `ReadSample` to make the loader read samples other that indicated by its current state (`read_sample_counter_` in particular). By design, `ReadSample` shouldn't depend on reading state of the loader, but just return the next sample, because the sample buffer is rather an implementation detail of the Loader. To make sure `ReadSample` won't depend on it, I **make `read_sample_counter_` and `initial_buffer_filled_` private**. 
- Also, **`virtual_shard_id_` should be made private**, but it's used in one place, in `Reset` of `FileLabelLoader` and it was recently introduced there (https://github.com/NVIDIA/DALI/issues/4954). I hope to eliminate this use when implementing mid-epoch checkpointing there. I added a _todo_ to make `virtual_shard_id_` private as well.

Loader baseclass

The only change that could potentially break existing stuff is keeping shared pointers instead of unique pointers in the sample buffer. It is also possible that after fast-forward a few samples in the sample buffer will be pointers to the same tensor. Is there some problem with it that I don't see?

### Tests:

- [ ] Existing tests apply
- [x] New tests added
  - [ ] Python tests
  - [x] GTests
  - [ ] Benchmark
  - [ ] Other
- [ ] N/A

### Documentation
- [ ] Documentation updated
  - [ ] Docstring
  - [ ] Doxygen
  - [ ] RST
  - [ ] Jupyter
  - [ ] Other
- [ ] N/A

#### Requirements
- [x] N/A

**REQ IDs**: N/A

**JIRA TASK**: DALI-3652

### DIFF
--- a/dali/operators/reader/loader/loader.h
+++ b/dali/operators/reader/loader/loader.h
@@ -329,17 +329,9 @@ class Loader {
 
   /**
    * @brief Advances loader position in the data source by skipping n samples.
-   * @param n Number of samples to skip.
+   * @warning This generic implementation is very inefficient and should be overriden.
   */
-  void Advance(uint64_t n) {
-    AdvanceImpl(n);
-  }
-
-  /**
-   * @brief Advances loader position in the data source by skipping n samples.
-   * @warning This generic implementation is inefficient and should be overriden.
-  */
-  virtual void AdvanceImpl(uint64_t n) {
+  virtual void Skip(uint64_t n) {
     LoadTargetUniquePtr tensor_ptr;
     {
       std::lock_guard<std::mutex> lock(empty_tensors_mutex_);
@@ -485,7 +477,7 @@ class Loader {
         continue;
       }
 
-      Advance(target->idx - at);
+      Skip(target->idx - at);
       at = target->idx;
 
       LoadTargetSharedPtr tensor_ptr = {new LoadTarget,
@@ -501,7 +493,7 @@ class Loader {
     }
 
     Reset(true);
-    Advance(read_sample_counter_);
+    Skip(read_sample_counter_);
   }
 
   std::vector<IndexedLoadTargetSharedPtr> sample_buffer_;
--- a/dali/operators/reader/loader/loader.h
+++ b/dali/operators/reader/loader/loader.h
@@ -231,11 +231,12 @@ class Loader {
       for (int i = 0; i < initial_buffer_fill_; ++i) {
         LoadTargetSharedPtr tensor_ptr = nullptr;
         if (!dry_run) {
-          tensor_ptr = LoadTargetSharedPtr(new LoadTarget,
-                           [this](LoadTarget* sample){
-                                      LoadTargetUniquePtr recycle_ptr(sample);
-                                      RecycleTensor(std::move(recycle_ptr));
-                                  });
+          tensor_ptr = LoadTargetSharedPtr(
+            new LoadTarget,
+            [this](LoadTarget* sample){
+              LoadTargetUniquePtr recycle_ptr(sample);
+              RecycleTensor(std::move(recycle_ptr));
+            });
           PrepareEmpty(*tensor_ptr);
           ReadSample(*tensor_ptr);
         }
@@ -287,12 +288,15 @@ class Loader {
       // being called by multiple consumer threads
       {
         std::lock_guard<std::mutex> lock(empty_tensors_mutex_);
-        DALI_ENFORCE(empty_tensors_.size() > 0, "No empty tensors - did you forget to return them?");
-        tensor_ptr = {empty_tensors_.back().release(),
-                      [this](LoadTarget* sample){
-                        LoadTargetUniquePtr recycle_ptr(sample);
-                        RecycleTensor(std::move(recycle_ptr));
-                      }};
+        DALI_ENFORCE(empty_tensors_.size() > 0,
+                     "No empty tensors - did you forget to return them?");
+        tensor_ptr = {
+          empty_tensors_.back().release(),
+          [this](LoadTarget* sample){
+            LoadTargetUniquePtr recycle_ptr(sample);
+            RecycleTensor(std::move(recycle_ptr));
+          }
+        };
         empty_tensors_.pop_back();
       }
       ReadSample(*tensor_ptr);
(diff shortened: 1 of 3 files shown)


dali/operators/reader/loader/loader.h
```diff
@@ -469,17 +573,25 @@ class Loader {
...
     Index start;
     Index end;
   };
 
   std::deque<ShardBoundaries> shards_;
+
+ private:
+  bool initial_buffer_filled_ = false;
+  // Counts how many samples the reader have read already from this epoch
+  Index read_sample_counter_;
```
mzient, line 594:
```suggestion
  Index read_sample_counter_ = 0;
```
...and get rid of the initializer in the constructor?

dali/operators/reader/loader/loader.h
```diff
@@ -469,17 +573,25 @@ class Loader {
   // The epoch number the next returned sample belongs to,
   // tracked only if checkpointing is enabled
   int consumer_epoch_;
```
mzient, line 575:
```suggestion
  int consumer_epoch_ = 0;
```

dali/operators/reader/loader/loader.h
```diff
@@ -451,8 +557,6 @@ class Loader {
   std::once_flag fetch_cache_;
   std::shared_ptr<ImageCache> cache_;
 
-  // Counts how many samples the reader have read already from this epoch
-  Index read_sample_counter_;
   // Counts how many samples the reader have read returned in the current epoch (including padding)
   Index returned_sample_counter_;
```
mzient, line 561: = 0

dali/operators/reader/loader/loader.h
```diff
@@ -81,12 +87,13 @@ class Loader {
...
       loading_flag_(false),
-      read_sample_counter_(0),
       returned_sample_counter_(0),
       pad_last_batch_(options.GetArgument<bool>("pad_last_batch")),
       dont_use_mmap_(options.GetArgument<bool>("dont_use_mmap")),
       checkpointing_(options.GetArgument<bool>("checkpointing")),
-      consumer_epoch_(0) {
+      consumer_epoch_(0),
+      max_batch_size_(options.GetArgument<int>("max_batch_size")),
+      read_sample_counter_(0) {
```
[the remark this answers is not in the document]
szkarpinski: Done

banasraf (approved):

dali/operators/reader/loader/loader_test.cc
awolant, line 1: Maybe it would be good to add a test without mocking. Have two loaders, fast-forward one using this added functionality, the other just manually run n-times and then compare if the state is the same.

awolant (approved):

dali/operators/reader/loader/loader_test.cc
szkarpinski: At the moment `Skip` is not yet implemented for any "real" reader, so testing it on this mock `DummyCountingReader` is the only choice for now. I'll be adding fast-forwarding tests with real loaders as I implement `Skip` in them.

szkarpinski:
!build

dali/operators/reader/loader/loader_test.cc
awolant: Maybe I missed something but wouldn't this generic implementation in base class be enough to write a working test?

szkarpinski: It would, but that generic implementation is just running the `ReadSample` n times. So IMO checking if this generic `Skip` is the same as running `ReadSample` n times doesn't add much value, because it's defined exactly as running `ReadSample` n times. 

I think doing the same with a mock loader with `Skip` implemented is a better choice: we then check if fast-forwarding (which is now different than just reading n samples) is the same as reading n samples (this happens in ` testFastForward`, where I first compute the _reference_ by just reading the data, then reset the loader and read the data again, fast-forwarding from time to time and comparing the values with the reference).

PR #5096, opened 2023-10-12
Generate type hints for fn and ops APIs

klecki:
## Description:
Implement generation of function signatures for `fn` and `ops` APIs with
type hint annotations.
Add generation of .pyi stub files containing those signatures.
The signatures are generated based on DALI_SCHEMA information,
so the build artifacts before wheel packaging are imported using PYTHONPATH
as a build step.
This is done as a new build target that is available only for native builds.
Docstring generation is adjusted to rely on the schema name rather than class.
fn.py is moved into fn/__init__.py as the directory structure is needed for the stub files.

The stub files replicate the existing structure of submodules for DALI operators.
The mechanism of generation uses the inspect.Signature and inspect.Parameter classes
for increased validation.
Due to some specifics of how annotation are printed, we need to introduce
special placeholder classes for DataNode and DALI enums, so they are 
correctly recognized by intellisense and also have terse representation.

TODO/next steps:
* MIS handling - now it is not considered, do it as an overload followup
* Constants as inputs/arguments - ArrayLike annotation similar to the NumPy - also a followup. Maybe DataNode can be converted to such DataNodeOrArrayLike annotation? 
* Extend the annotations to other DALI parts, like FW iterators.

fn, ops.

Can this somehow impact the runtime Python?
Is any part of the stub incorrect.
Do we need to import DataNode and DALIDataType into fn and ops modules?
Does it work on your IDE?
Does it interfere with sphinx docs?

### Tests:
The best way is to inspect the behaviour under different IDEs and adjust if needed.

- [ ] Existing tests apply
- [x] New tests added
  - [x] Python tests - added some simple mypy tests.
  - [ ] GTests
  - [ ] Benchmark
  - [ ] Other
- [ ] N/A

### Documentation
- [ ] Documentation updated
  - [ ] Docstring
  - [ ] Doxygen
  - [ ] RST
  - [ ] Jupyter
  - [ ] Other
- [x] N/A

#### Requirements
- [x] N/A

**REQ IDs**: N/A

**JIRA TASK**: N/A


klecki:
!build

klecki:
!build

klecki:
!build

klecki:
!build

klecki:
!build

klecki:
!build

dali/python/nvidia/dali/ops/_signatures.py
```diff
@@ -0,0 +1,393 @@
...
+
+from nvidia.dali import backend as _b
+from nvidia.dali import types as _types
+from nvidia.dali.ops import _registry, _names, _docs
+from nvidia.dali.fn import _to_snake_case
+from nvidia.dali import types
+
+
+def _create_annotation_placeholder(typename):
+    # Inspect and typing know better how to format annotations. They allow of short versions
```
mzient, line 32: > allow of short versions

English parse error. What was it supposed to mean?

dali/python/nvidia/dali/ops/_signatures.py
```diff
@@ -0,0 +1,393 @@
...
+# Those operators have custom Python code, so the annotations are done in-place for the
+# wrappers rather than via automatic generation.
+_manual_definitions = [
+    "ExternalSource",
+    "TFRecordReader",
+    "readers__TFRecord",
+    "PythonFunction",
+    "DLTensorPythonFunction",
+    "NumbaFunction",
+]
```
mzient, line 89: Don't we have some other way to detect those? I think there's some mechanism in the (former) ops.py that generated (or not) those functions.
How about adding a field `generated` or similar and only processing those here?

dali/python/nvidia/dali/ops/_signatures.py
```diff
@@ -0,0 +1,393 @@
...
+    else:
+        return Parameter(f"__input_{idx}", kind=Parameter.POSITIONAL_ONLY, default=default,
+                         annotation=annotation)
+
+
+def _get_positional_input_params(schema):
+    """Get the list of positional only inputs to the operator.
+    """
+    param_list = []
+    if schema.MaxNumInput() > _MAX_INPUT_SPELLED_OUT:
```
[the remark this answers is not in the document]
klecki: Adding the documented names, but I don't think it matters if we have more than 5 optional or regular arguments. I can maybe make the threshold 10 and leave it at it?

dali/python/nvidia/dali/ops/_signatures.py
klecki: Done

dali/python/nvidia/dali/types.pyi
```diff
@@ -0,0 +1,46 @@
...
+    YCbCr = ...
+    ANY_DATA = ...
+
+class DALIInterpType(Enum):
+    INTERP_NN = ...
+    INTERP_LINEAR = ...
+    INTERP_CUBIC = ...
+    INTERP_LANCZOS3 = ...
+    INTERP_TRIANGULAR = ...
+    INTERP_GAUSSIAN = ...
```
mzient, line 46: Is that written manually? Do we have at least some way to ensure that it doesn't go stale?

klecki: It is, and I don't have a good idea how to check it, considering the fact, that we already have it spelled out in this or that form in 5 places.
The only thing for my defense is that we didn't change it for quite a long time, and this definition is not binding. 
The problem with those enums is, that they are only visible when the actual pybind module is loaded, so any static tool that looks just at the Python code, doesn't see the original definitions. This was the only way, that I found, not to pollute the `fn` module (and submodules) with the `class DALIInterpType: ...` placeholders, that were visible in the intellisense. Otherwise we would need to rewrite those types in Python, maybe this should be the way.

dali/python/nvidia/dali/ops/_operators/python_function.py
```diff
@@ -99,6 +99,7 @@ def _dlpack_from_array(array):
 
 class PythonFunction(PythonFunctionBase):
     schema_name = "PythonFunction"
+    _generated = False
```
mzient, line 102: I presume this is here to override the default `True` from the base class?

klecki: It's here for completeness with the `_generated = True` from the generated classes. I still do `getattr(op, "_generated", False)` just in case, but I am not fond of missing fields.

klecki:
!build

dali/python/nvidia/dali/ops/_operators/python_function.py
mzient: Meh. I'll not sweat it, but I disagree on adding this kind of field "for completeness". It adds noise and creates yet another place where the code can go stale.

dali/python/nvidia/dali/ops/_signatures.py
```diff
@@ -0,0 +1,432 @@
...
+        # We don't want <class 'typename'> when inspect calls inspect.formatannotation directly.
+        def __repr__(cls):
+            return typename
+
+    class _AnnotationPlaceholder(metaclass=_AnnotationPlaceholderMeta):
+        # typing uses those instead of repr
+        __name__ = typename
+        __qualname__ = typename
+        # only if you are part of `typing.` your path is hidden
+        __module__ = "typing"
```
mzient, line 50: What do you mean by "path" here?

klecki: fully qualified type name, so `module.submodule.class`.

dali/python/nvidia/dali/ops/_operators/python_function.py
klecki: Tbh, there is no actual base class for all ops, at least as far as I understand Python 🙃 
And the hand written ones don't inherit from the one in op_factory, there is a TODO to fix it, as it would eliminate even more common code, but I decided not to spend another week on reworking that and introduce the actual hints first.

banasraf (approved):

klecki:
!build

dali/python/nvidia/dali/ops/_signatures.py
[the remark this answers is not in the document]
mzient: My point was that if we have some minimum number of arguments, then it's likely that they have some specific meaning. If we have a tail of 2-3 arguments, likely the same assumption holds. Conversely, if we have a bunch of optional arguments at the end, they're likely just the same thing over and over (like in `cat`,`stack` or `transform.combine`). It doesn't matter all that much for now - it's just that now we know what can (possibly) go wrong and it's easily fixable. Once that code is written, we're unlikely to go back to it.

dali/python/CMakeLists.txt
```diff
@@ -43,3 +43,33 @@ copy_post_build(dali_python "${PROJECT_SOURCE_DIR}/Acknowledgements.txt" "${PROJ
...
+    COMMAND PYTHONPATH=${PYTHONPATH} ${PROJECT_SOURCE_DIR}/tools/python_stub_generator.py --wheel_path ${PROJECT_BINARY_DIR}/dali/python/nvidia/dali
+  )
+
+  # Format the bindings, so they are somewhat readable.
+  if (${FORMAT_PYTHON_STUB_FILES})
+    add_custom_target(dali_python_format_stubs ALL DEPENDS dali_python dali_python_generate_stubs)
+    add_custom_command(
+      TARGET dali_python_format_stubs
+      WORKING_DIRECTORY ${PROJECT_BINARY_DIR}
+      COMMAND find dali/python/nvidia/dali -name '*.pyi' -print0 | xargs -0 yapf --in-place --verbose --parallel --style "${PROJECT_SOURCE_DIR}/.style.yapf"
```
mzient, line 72: This takes _ages_. I recommend using `black` instead.
```suggestion
      COMMAND COMMAND find dali/python/nvidia/dali -name '*.pyi' | xargs black -l 120 -t py38
```

klecki: If we agreed to use `black` uniformly, than yes, but we already have formatting rules for `yapf`.

klecki:
!build

dali/python/CMakeLists.txt
mzient: These responses were posted deliberately as means of psychological torture on a person who holds beliefs of a minority. Shall I report it? I'm being harassed.

PR #5103, opened 2023-10-19
Add mid-epoch checkpointing to Loader

szkarpinski:
## Category:

## Description:

This PR allows `Loader` to checkpoint at any time (including mid-epoch).

If checkpointing is enabled, loader takes a snapshot of its state at the beginning of every epoch. This snapshot is kept in `current_snapshot_` together with its `age`, indicating how many samples were returned since the snapshot was taken.

When restoring a checkpoint, first the state from the snapshot is restored and then the loader is fast-forwarded (https://github.com/NVIDIA/DALI/pull/5088) by `age` steps.

Checkpointing in loader `Loader` baseclass was modified to allow mid-epoch checkpointing. 

In addition to that, the following additional changes were made:
- Sample indices (stored together with actual buffered samples to allow reading them lazily in `ReadMissingSamples`, see https://github.com/NVIDIA/DALI/pull/5088) were changed from being sample's index _in the epoch_ (`read_sample_counter`) to being absolute sample's index (`total_read_sample_counter`). That's because it's possible that there are samples from multiple epochs in the buffer and `ReadMissingSamples` must know which epoch is each sample from, because samples may be shuffled between epochs.
- ~In `reader_op.h`, when restoring the checkpoint, we now stop the prefetch thread to prevent races and clear prefetch queue to return tensors to the loader. This worked before because of the assumption that `restore` will be called only on freshly-created pipeline. I'd like to lift this assumption soon.~ This can wait until the follow-up PR when I deal with this completely. Undone in https://github.com/NVIDIA/DALI/pull/5103/commits/0a1a24b3b6b03d599173ce8aa0b284fd57bc113c
- The changes in `FileLabelLoader` are just implementing `Rewind` to keep existing code working.

### Follow-ups
- `FileReader` will get mid-epoch checkpointing support in a separate PR, coming soon
- To optimize checkpoint restore, I'd like to store snapshots every _n_ samples and not just every epoch.
- This solution should allow us to lift `pad_last_batch=True` requirement from `FileReader`.

### Tests:
Several fixes were made in `DummyCountingLoader ` to make it epoch-aware. Also, to simulate possible differences between epochs, an optional parameter `mark_epoch` (default `0`) was added to  `DummyCountingLoader `, which adds `mark_epoch * current_epoch` to each sample.

- [ ] Existing tests apply
- [x] New tests added
  - [ ] Python tests
  - [x] GTests
  - [ ] Benchmark
  - [ ] Other
- [ ] N/A

### Documentation
- [ ] Documentation updated
  - [ ] Docstring
  - [ ] Doxygen
  - [ ] RST
  - [ ] Jupyter
  - [ ] Other
- [ ] N/A

#### Requirements
- [x] N/A

**REQ IDs**: N/A

**JIRA TASK**: DALI-3652

### DIFF
--- a/dali/operators/reader/loader/file_label_loader.h
+++ b/dali/operators/reader/loader/file_label_loader.h
@@ -196,13 +196,16 @@ class DLL_PUBLIC FileLabelLoaderBase : public Loader<CPUBackend, ImageLabelWrapp
     Reset(true);
   }
 
-  void Reset(bool wrap_to_shard) override {
+  void Rewind(bool wrap_to_shard) override {
     if (wrap_to_shard) {
       current_index_ = start_index(virtual_shard_id_, num_shards_, SizeImpl());
     } else {
       current_index_ = 0;
     }
+  }
 
+  void Reset(bool wrap_to_shard) override {
+    Rewind(wrap_to_shard);
     current_epoch_++;
 
     if (shuffle_after_epoch_) {
--- a/dali/operators/reader/loader/loader.h
+++ b/dali/operators/reader/loader/loader.h
@@ -189,7 +189,7 @@ class Loader {
 
     // Re-run reset
     Reset(true);
-    
+
     SaveStateSnapshot(current_snapshot_);
 
     FastForward(state.age);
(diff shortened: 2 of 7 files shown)


banasraf (approved):

awolant (approved):

szkarpinski:
!build

PR #5113, opened 2023-10-24
Allow mid-epoch checkpointing in FileReader

szkarpinski:
## Category:

## Description:
This PR adds an ability to checkpoint at any iteration to `FileReader`.

- `Reader` and now saves a snapshot of loader after every iteration
- `Executor` now saves a checkpoint after every iteration
- `FileLabelLoader` got a proper implementation of `Skip`

### Tests:

- [ ] Existing tests apply
- [x] New tests added
  - [x] Python tests
  - [x] GTests
  - [ ] Benchmark
  - [ ] Other
- [ ] N/A

### Documentation
- [x] Documentation updated
  - [x] Docstring
  - [ ] Doxygen
  - [ ] RST
  - [ ] Jupyter
  - [ ] Other
- [ ] N/A

#### Requirements
- [x] N/A

**REQ IDs**: N/A

**JIRA TASK**: DALI-3652


dali/pipeline/executor/executor.cc
```diff
@@ -525,13 +525,14 @@ void Executor<WorkspacePolicy, QueuePolicy>::RunHelper(OpNode &op_node, Workspac
...
   }
 
-  // If it is the end of an epoch, create a checkpoint.
+  // Create a checkpoint.
   // The checkpoint corresponds to the state between the iteration `iteration_id`
   // and `iteration_id + 1`.
-  // After consuming all the outputs from the epoch, GetCurrentCheckpoint is going to
+  // After consuming all the outputs from the iteration, GetCurrentCheckpoint is going to
   // return the created checkpoint.
-  if (checkpointing_ && (iteration_id + 1) % checkpointing_epoch_size_ == 0)
```
stiepan, line 533: It seems we could get rid of the `checkpointing_epoch_size_` completely at this point, right? Along with the check that all the readers in the pipeline agree on the epoch size.

dali/test/python/checkpointing/test_dali_checkpointing.py
```diff
@@ -145,11 +145,14 @@ def pipeline_factory():
...
-        (2, 11, 2, 5, False, True, False)
+        (2, 11, 2, 5, False, True, False),
+        (5, 3, 0, 1, True, False, False, 4),
+        (2, 10, 0, 2, True, False, False, 5),
+        (4, 256, 2, 4, False, False, True, 6),
 )
 def test_file_reader(
         num_epochs, batch_size, shard_id, num_shards,
-        random_shuffle, shuffle_after_epoch, stick_to_shard):
+        random_shuffle, shuffle_after_epoch, stick_to_shard, iters_into_epoch=None):
```
stiepan, line 155: Do you use it this test?

dali/test/python/checkpointing/test_dali_checkpointing.py
stiepan, line 1: However weird that case may be, I wonder if we should now have a test with two readers that go over different number of samples. We had a check in the executor that enforces the equal size of epochs for different readers when checkpointing is enabled. The check is no longer needed - some simple test could help catch this kind of vestiges of this assumption.

dali/pipeline/executor/executor.cc
szkarpinski: That's right, removed.

dali/test/python/checkpointing/test_dali_checkpointing.py
szkarpinski: Oops! Fixed

dali/test/python/checkpointing/test_dali_checkpointing.py
szkarpinski: Good idea, I added `test_multiple_readers` to check that

stiepan, line 1: One more thing, do you test `pad_last_batch=False`? It's supported now, right?

szkarpinski: I planned to add support as a follow-up, but now I see I already removed `pad_last_batch=True` check in https://github.com/NVIDIA/DALI/pull/5103 :P I'll add the tests here

szkarpinski: I found a bug with `pad_last_batch=True` + `shuffle_after_epoch=False`: when fast-forwarding, the `Loader` was calling `Reset` too often resulting in wrong epoch number being used for shuffling. I fixed that in https://github.com/NVIDIA/DALI/pull/5113/commits/0dd8d7a6c46d770af9a6ee94de2acddd14d22a88 and added tests

stiepan, line 1: Leaving a comment from our offline discussion just not to miss it in the next read:)

1. We need tests with small `initial_fill` (or make sure to iterate enough times when comparing pipelines to 1. exhaust the shuffling buffer and 2. go over all shards when applicable; for now we checked only the 2. part). 
2. The lack of 1.1. is super insidious, as we may never stumble upon the "FastForwarded" samples when consuming from the shuffling buffer in the tests. That's likely what happens with Rewinding to a wrong spot after the sequence of dry runs of ReadOne.

szkarpinski:
After long discussions with @stiepan and fighting with endless corner-cases, I decided to rework how Loader fast-forwarding is done.

### What was wrong?
The fast-forwarding was done in two steps. First, `ReadOne(..., dry=true)` was run, simulating the buffer contents, working with pseudo-samples (i.e. samples with just indices and no content). Then, `ReadMissingSamples` was called, which used a combination of `Rewind`, `Skip` and `ReadSample` to read the samples that were missing.

This made loader's state incoherent during fast-forward: some counters were already changed by running `ReadOne(..., dry=true)` i the first phase, but we were trying to trick `ReadSample` into reading samples "from the past". This turned out to be very hard to do correctly for some strange cases of sharding. Handling those correctly required adding even more magical code and it was getting really hard to manage.

The main reasons for state incoherence were:
- `ReadMissingSamples` which by design was reading samples "from the past"
- `Skip` which was meant to only change the "reading" part of the loader
- `ReadOne(..., dry=true)` which was meant to only change the "returning" part of the loader.

### How is it done now?
The main goal of the changes was to keep loader's state coherent all the time and to simplify stuff. 

The following changes were made to the utility functions:
- `Skip(n)` is now just `Skip()` (skips one sample only), simply meant to behave like `ReadSample(...)`, but without actually reading the sample.
- `ReadMissingSamples` is gone
- `FastForward` is gone, inlined into `RestoreStateFromSnapshot`

On the high level, restoring checkpoint is now done in the following way:
1. Restore state from checkpoint (using `RestoreEpochState`)
2. Run `ReadOne` in dry mode to simulate buffer contents
3. See what was the oldest sample in the buffer
4. Restore state from checkpoint again
5. Run `ReadOne` in dry mode until we reach that oldest sample and continue in normal mode from there

szkarpinski:
Another bigger change happened in https://github.com/NVIDIA/DALI/pull/5113/commits/ef2af8e36323b31778fd8b6a96c07f14456f1ae1. We weren't happy with the idea of reading all samples from the oldest needed sample because expected performance was poor. We decided to use a predicate instead of `dry` argument. Now `ReadOne` gets a predicate `filter` to decide which samples need to be read and which don't. This allows us to read exactly the samples that we need

stiepan (approved):
Looks good, I happy to see the readers are actually heading towards fully fledged checkpointing!

I left a comment with an idea of a few more test cases (I think those cases are well suportted now, but it's always good to have that actually checked now and with the future changes).

dali/operators/reader/loader/loader.h
```diff
@@ -342,24 +351,14 @@ class Loader {
...
 
   /**
-   * @brief Advances loader position in the data source by skipping n samples.
-   * @warning This generic implementation is very inefficient and should be overriden.
+   * @brief Advances loader position in the data source by skipping a sample.
+   * @warning This generic implementation is very inefficient (it simply reads and discards
+   *          the sample) and should be overriden.
   */
-  virtual void Skip(uint64_t n) {
+  virtual void Skip() {
```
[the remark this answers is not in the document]
szkarpinski: That's a good point, thanks! maybe not having a generic implementation at all is a better option. I'll think about it and I'll remove it in some follow-up PR unless I find a good reason for it to stay.

dali/test/python/checkpointing/test_dali_checkpointing.py
```diff
@@ -133,30 +133,41 @@ def pipeline_factory():
...
+        (5, 3, 0, 1, True, False, False, True, 4),
+        (2, 10, 0, 2, True, False, False, True, 5),
+        (4, 256, 2, 4, False, False, True, True, 6),
+        (3, 64, 3, 4, False, False, True, False),
+        (5, 10, 0, 2, True, False, False, False),
+        (1, 3, 0, 1, True, False, False, False),
+        (10, 3, 0, 1, True, False, False, False, 1),
+        (10, 10, 0, 2, True, False, False, False, 2),
+        (10, 256, 2, 4, False, False, True, False, 3),
+        (10, 10, 1, 2, False, False, False, False),
```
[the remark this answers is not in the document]
szkarpinski: Added the tests, thank you!

stiepan (approved):

szkarpinski:
!build

awolant (approved):

PR #5115, opened 2023-10-24
Add default arg values to JAX decorator

awolant:
## Category:

**New feature** 

## Description:

Adds default values for `device_id` and `num_threads` for decorated JAX iterator functions. Thanks to this change it is easier to apply `sharding` to scale up the computation in the most common pattern.

JAX iterator decorator.

### Tests:

- [ ] Existing tests apply
- [ ] New tests added
  - [x] Python tests
  - [ ] GTests
  - [ ] Benchmark
  - [ ] Other
- [ ] N/A

### Documentation
- [x] Documentation updated
  - [ ] Docstring
  - [ ] Doxygen
  - [ ] RST
  - [x] Jupyter
  - [ ] Other
- [ ] N/A

#### Requirements
- [ ] N/A

**REQ IDs**: N/A

**JIRA TASK**: DALI-3671


awolant:
!build

dali/python/nvidia/dali/plugin/jax/iterator.py
```diff
@@ -246,7 +246,15 @@ def data_iterator_decorator(func):
         def create_iterator(*args, **wrapper_kwargs):
             pipeline_def_fn = pipeline_def(func)
 
+            if 'num_threads' not in wrapper_kwargs:
+                wrapper_kwargs['num_threads'] = 4
```
mzient, line 250: This choice seems quite arbitrary. I'd recommend getting the number of processors (or perhaps the number of processors per GPU?).

awolant: I did that because that's what we do it a lot in our codebase and especially in the tutorials. Do we have some benchmarks of what it does to the performance?

JanuszL: But you need to account for the number of CPUs that may need to be used for the JAX itself, the speed of the network, and things that may run in parallel.
I'm afraid that without an elaborate autotune we can't provide an accurate value here.

mzient: There's a lot of middle ground between elaborate autotune and a fixed number.
Still, I'd at least like to have a big, fat comment saying that it's subject to change.

banasraf (approved):

dali/python/nvidia/dali/plugin/jax/iterator.py
awolant: Done, added docs, comments and tutorial section on `nunm_threads`

awolant:
!build

PR #5118, opened 2023-10-24
Add type hints for ExternalSource, Python Functions, TFRecord Reader, math module

klecki:
## Description:
Add annotations to operators with custom Python wrappers:
* External source
* TFRecord reader
* Python function family
* nvidia.dali.math module
* ops.Compose

Add stub file for nvidia.dali.tfrecord module.

Add return type annotations to the Pipeline.run function.
The TensorList types again are not fully visible due to being generated 
by backend at runtime, next step should provide a dedicated stub file
or alternative implementation.

The stubs are based on the output of mypy `stubgen` and the 
`nvidia.dali.ops._signatures._gen_[fn/ops]_signature`.

The stub generation is reworked, first grouping the operators into 
4 categories, so the generated stubs contain imports first. 
We utilize the fact that the operators with custom Python wrappers
have dedicated implementation modules now, and reexpose them
in the stubs, allowing the type hints to be picked up.

The external source has non-trivial defaults, expressed mostly via `None`
and cross-dependent on other parameters, so creating a meaningful
annotations there is hard.
There are two overloads provided for the `external_source` function,
allowing to disambiguate between the single and multiple outputs
(when `num_outputs` parameter was used, and the return type is 
always a tuple). Such distinction can't be easily made in the ops API.

TODO:
* Numba and Pytorch function are left for a followup as they require 
   a lot of code movement to accommodate for the module structure.

Several operators, Pipeline.run, stub generation.

If the annotations are ok.

### Tests:

- [x] Existing tests apply
 Whole DALI must work
- [ ] New tests added
  - [x] Python tests
   Trivial mypy test with external source.
  - [ ] GTests
  - [ ] Benchmark
  - [ ] Other
- [ ] N/A

### Documentation
- [x] Documentation updated
  - [ ] Docstring
  - [ ] Doxygen
  - [ ] RST
  - [ ] Jupyter
  - [ ] Other
- [ ] N/A

#### Requirements
- [ ] N/A

**REQ IDs**: N/A

**JIRA TASK**: N/A


klecki:
!build

dali/python/nvidia/dali/external_source.pyi
```diff
@@ -0,0 +1,115 @@
...
+
+from nvidia.dali.data_node import DataNode
+from nvidia.dali.types import DALIDataType, DALIImageType, DALIInterpType
+
+
+class ExternalSource:
+
+    def __init__(
+        self,
+        source: Optional[Union[Callable[..., Any], Iterable[Any], Callable[..., Sequence[Any]], Iterable[Sequence[Any]]]] = None,
```
stiepan, line 26: A random thought, but should we go into details like `Sequence[Any]` of the source return type when we already list Any as acceptable type?

dali/python/nvidia/dali/external_source.py
stiepan, line 1: Do we have (or even need) some distinction on when we prefer stubs over inline type annotations?

dali/python/nvidia/dali/external_source.pyi
klecki: This is a really good point, my thinking is basically `Any` ~ `Tensor` and `Sequence[Any]` ~ batch of tensors.
The problem is I don't have any (heh) good approximation of what Tensor can be, there is a proposal to have something like this in the newer Pythons https://github.com/python/typeshed/blob/2a0fc1b582ef84f7a82c0beb39fa617de2539d3d/stdlib/_typeshed/__init__.pyi#L194
but is still doesn't encapsulate GPU tensors.
I want to express the outermost kind of type, and we can iteratively get better, I am not yet fully sure what is going on here: https://github.com/numpy/numpy/blob/main/numpy/_typing/_array_like.py

dali/python/nvidia/dali/external_source.py
klecki: I think we have new enough Python, that we can mix in the annotations without issues. 
They do hide the lookup of the original implementation: if you want to do a partial annotations for given class, you need to mix it in the .py file, otherwise you would need to list all members in the stub, that's why sometimes I put it only in the file.

dali/python/nvidia/dali/external_source.pyi
stiepan: I assumed that's the intention and it feels like a good idea. My concern is whether it's best to mention the Sequence now or make it more detailed once we have some TensorLike type. For example, replacing Any with the TensorLike, probably won't be enough. I think we also need Tuple[Sequence[TensorLike]] and Tuple[TensorLike] for the sample mode.

dali/python/nvidia/dali/ops/_signatures.py
```diff
@@ -370,63 +371,130 @@ def _get_op(api_module, full_qualified_name: List[str]):
...
         if api == "fn":
-            module_to_file[module_path].write(_gen_fn_signature(schema, schema_name, fn_name))
+            stub_manager.get(module_nesting).write(_gen_fn_signature(schema, schema_name, op_name))
         else:
-            module_to_file[module_path].write(_gen_ops_signature(schema, schema_name, op_name))
+            stub_manager.get(module_nesting).write(_gen_ops_signature(schema, schema_name, op_name))
 
-    for _, f in module_to_file.items():
-        f.close()
+    # TODO(klecki): use scope manager
```
stiepan, line 499: I think you can just wrap the `StubFileManager(nvidia_dali_path, api)` into `from contextlib import closing` like `with closing(StubFileManager(nvidia_dali_path, api)) as stub_manager`.

dali/python/nvidia/dali/external_source.pyi
klecki: Oh, sorry, I forgot my intention here. So the Sequence comes from the multiple outputs, looking at the `external_source` overloads, but I didn't want to go too much into the overloads of the class API. 
Looks like we can't put `num_outputs` in the call, and the `source` is kw-only there, will fix.

The type for source is the union of two overloads I did for the fn API, basically `Any` can be either TensorLike or TensorBatchLike. `TensorLike` can be used in sample mode, and `Union[TensorLike, TensorBatchLike]` can be used in batch mode. And than we have Sequence of those in case of `num_outputs != None`. It is a lot of options.

dali/python/nvidia/dali/tfrecord.pyi
```diff
@@ -0,0 +1,45 @@
...
+@overload
+def VarLenFeature(__dtype: int, __default_value: object, /) -> Feature:
+    """Equivalent of TensorFlow's VarLenFeature"""
+    ...
+
+
+@overload
+def VarLenFeature(__partial_shape: List[int], __dtype: int, __default_value: object, /) -> Feature:
+    """Equivalent of TensorFlow's VarLenFeature"""
+    ...
```
stiepan, line 45: Do you actually use those?

dali/python/nvidia/dali/ops/_operators/python_function.pyi
```diff
@@ -0,0 +1,168 @@
...
+    by DALI, which can be obtained by calling the ``current_dali_stream()`` function. In this case,
+    the ``synchronize_stream`` flag can be set to False.
+    """
+
+    def __init__(
+        self,
+        /,
+        *,
+        function: Optional[Callable[..., Union[Any, Sequence[Any]]]] = None,
+        batch_processing: Optional[bool] = False,
```
stiepan, line 86: Isn't the default for DLTensorPythonFunction, `batch_processing=True`?

dali/python/nvidia/dali/tfrecord.pyi
klecki: This is the case of the pybind bindings, without this file, we don't see this types in the IDE, I did the whole module for completeness. 
For operator annotation we need just the `Feature` (so it doesn't show as `Any` in VSCode), but I think it is nice to have `nvidia.dali.tfrecord.Fi<space>` -> `nvidia.dali.tfrecord.FixedLenFeature` work.

dali/python/nvidia/dali/ops/_operators/python_function.pyi
klecki: The doc says: `batch_processing (bool, optional, default = False) –`, the source uses `True`. Nice.

dali/python/nvidia/dali/ops/_signatures.py
klecki: done

dali/python/nvidia/dali/ops/_operators/python_function.pyi
klecki: Fixed in the schema and annotation.

klecki:
!build

klecki:
!build

stiepan (approved):

dali/python/nvidia/dali/ops/_operators/python_function.pyi
```diff
@@ -83,8 +82,8 @@ class DLTensorPythonFunction:
         /,
         *,
         function: Optional[Callable[..., Union[Any, Sequence[Any]]]] = None,
-        batch_processing: Optional[bool] = False,
-        num_outputs: Optional[int] = 1,
+        batch_processing: bool = False,
```
stiepan, line 85: Shouldn't this be True too?

dali/python/nvidia/dali/ops/_operators/python_function.pyi
```diff
@@ -100,14 +99,13 @@ class DLTensorPythonFunction:
         /,
         *input: DataNode,
         function: Optional[Callable[..., Union[Any, Sequence[Any]]]] = None,
-        batch_processing: Optional[bool] = False,
-        num_outputs: Optional[int] = 1,
+        batch_processing: bool = False,
```
stiepan, line 102: Shouldn't this be True too?

dali/python/nvidia/dali/ops/_operators/python_function.pyi
klecki: probably it should

dali/python/nvidia/dali/ops/_operators/python_function.pyi
klecki: I removed those arguments, as I don't think they will work correctly.
We need to consider revisiting the custom implementations for ops, as they do not follow the "pass any argument anywhere" flow.

klecki:
!build

stiepan (approved):

banasraf (approved):

PR #5129, opened 2023-10-26
Add type annotations for numba and pytorch plugins

klecki:
## Description:
Type annotations are added for numba and pytorch plugins, covering
the operators (Numba function and Torch Python function) available
in the plugin module. 

Additionally, type annotations are added to Pytorch data iterators for DALI.

The code is rearranged so the dictionaries reflect the actual module
hierarchy and allow for inserting the interface .pyi files, rather than
generating the modules at runtime (especially the `fn` ones).

Plugins for Numba and Pytorch (annotations only)

### Tests:

- [ ] Existing tests apply
- [ ] New tests added
  - [ ] Python tests
  - [ ] GTests
  - [ ] Benchmark
  - [ ] Other
- [ ] N/A

### Documentation
- [ ] Documentation updated
  - [ ] Docstring
  - [ ] Doxygen
  - [ ] RST
  - [ ] Jupyter
  - [ ] Other
- [ ] N/A

#### Requirements
- [ ] N/A

**REQ IDs**: N/A

**JIRA TASK**: N/A

### DIFF
--- a/dali/python/nvidia/dali/plugin/numba/experimental/__init__.py
+++ b/dali/python/nvidia/dali/plugin/numba/experimental/__init__.py
@@ -371,6 +371,7 @@ class NumbaFunction(metaclass=ops._DaliOperatorMeta):
             NumbaFunction._check_minimal_numba_version()
             NumbaFunction._check_cuda_compatibility()
 
+        # TODO(klecki): Normalize the types into lists first, than apply the checks
         assert len(in_types) == len(ins_ndim), ("Number of input types "
                                                 "and input dimensions should match.")
         assert len(out_types) == len(outs_ndim), ("Number of output types "
--- a/dali/python/nvidia/dali/plugin/numba/experimental/__init__.pyi
+++ b/dali/python/nvidia/dali/plugin/numba/experimental/__init__.pyi
@@ -1,0 +1,67 @@
+# Copyright (c) 2023, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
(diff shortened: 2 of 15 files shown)


dali/python/nvidia/dali/plugin/pytorch/__init__.pyi
```diff
@@ -0,0 +1,126 @@
...
+        fill_last_batch: Optional[bool] = None,
+        dynamic_shape: Optional[bool] = ...,
+        last_batch_padded: bool = False,
+        last_batch_policy: LastBatchPolicy = LastBatchPolicy.FILL,
+        prepare_first_batch: bool = True,
+    ) -> None:
+        ...
+
+    def __next__(self) -> List[Dict[str, torch.Tensor]]:
+        ...
```
mzient, line 55: I don't think that retyping everything here is the right thing to do - rather the contrary; it's prone to go stale in many subtle ways. Why can't we just annotate the proper code instead?

klecki: The technical reason was to keep the TorchPythonFunction without changes, so I put the annotated signatures (that differs from the runtime ones) in the stub file, and thanks to that I need to put all the annotations in the stub, otherwise they become hidden (the annotation file has precedence over which symbols are visible, you can't mix and match this way).

I can move the TorchPythonFunction to a submodule, and move annotations for the iterators back to the file.

mzient: Why do we need a submodule for that?

klecki: So I can have TorchPythonFunction annotated with stub file, as this is needed to expand the definition of `**kwargs`.

klecki: almost done

klecki:
I extracted some adjustments that I found needed for Python function into a separate PR: https://github.com/NVIDIA/DALI/pull/5149

mzient (approved):

klecki:
!build

klecki:
!build

klecki:
!build

PR #5139, opened 2023-11-02
[WiP] Add checkpointing support to Video Reader

szkarpinski:
### Tests:

- [ ] Existing tests apply
- [ ] New tests added
  - [ ] Python tests
  - [ ] GTests
  - [ ] Benchmark
  - [ ] Other
- [ ] N/A

### Documentation
- [ ] Documentation updated
  - [ ] Docstring
  - [ ] Doxygen
  - [ ] RST
  - [ ] Jupyter
  - [ ] Other
- [ ] N/A

#### Requirements
- [ ] N/A

**REQ IDs**: N/A

**JIRA TASK**: N/A


dali/test/python/checkpointing/test_dali_checkpointing.py
```diff
@@ -287,6 +288,90 @@ def pipeline():
...
+
+@dataclass
+class VideoConfig:
+    sequence_length : int
+    stride : int
+    step : int
+
+
+@cartesian_params(
+    (0, 1, 3),
```
stiepan, line 307: How about a case that resumes from a non zero epoch and a non zero iteration?

stiepan (approved):
Looks like most of the heavylifting is done by the "index building" of the video reader and the current checkpointing implementation blends nicely into that!

dali/test/python/checkpointing/test_dali_checkpointing.py
```diff
@@ -287,6 +288,90 @@ def pipeline():
...
+    ),
+    (
+        VideoConfig(sequence_length=3, stride=1, step=-1),
+        VideoConfig(sequence_length=3, stride=1, step=5),
+    ),
+)
+def test_video_reader(num_epochs, batch_size, iters_into_epoch,
+                      config: BaseDecoderConfig, video: VideoConfig):
+
+    files = 2 * [os.path.join(get_dali_extra_path(), 'db/video/multiple_framerate/10/10fps.mp4')]
```
stiepan, line 328: I wonder if that is enough data? I.e. that the multiple shards cases don't get too trivial? Can we use all the files from `multiple_framerate` at once @awolant?

dali/test/python/checkpointing/test_dali_checkpointing.py
szkarpinski: Isn't there one? I choose `num_epochs` from `(0, 1, 3)` and `iters_into_epoch` from `(0, 2)`, so there will be tests with `num_epochs=1, iters_into_epoch=2` and `num_epochs=3, iters_into_epoch=2`

dali/test/python/checkpointing/test_dali_checkpointing.py
szkarpinski: I checked that at some point but it might be good to add some assert in the test to check that there are at least 2 samples in each shard and that the requested number of iterations in epoch is possible

szkarpinski: It turns out that some testcases indeed were trivial, thank you! I added assets and changed the video files used to fix that

dali/test/python/checkpointing/test_dali_checkpointing.py
stiepan: Oh, of course, sorry for the confusion. I somehow misread the suite as if the first three tuples where three different cases for `num_epochs, batch_size, iters_into_epoch` times the decoder times the video setups.

awolant (approved):

szkarpinski:
!build

szkarpinski:
!build

PR #5145, opened 2023-11-06
[WiP] Experiments with spellchecking

szkarpinski:
## Category:

### What it does?
Warnings for typos are printed by `make html`

### What are the dependencies?
This uses `sphinx.spelling` plugin, which is based on `pyenchant`. There are some problems with this approach (see below) and maybe it'll be easier to just write own plugin.

### What's missing?
`pyenchant`'s tokenizer by default splits words on `.` and `_`, resulting in tons of false positives. `sphinx.spelling` doesn't allow to change that, so a bit of hacking in either `pyenchant` or `sphinx.spelling` is needed :(

### Tests:

- [ ] Existing tests apply
- [ ] New tests added
  - [ ] Python tests
  - [ ] GTests
  - [ ] Benchmark
  - [ ] Other
- [ ] N/A

### Documentation
- [ ] Documentation updated
  - [ ] Docstring
  - [ ] Doxygen
  - [ ] RST
  - [ ] Jupyter
  - [ ] Other
- [ ] N/A

#### Requirements
- [ ] N/A

**REQ IDs**: N/A

**JIRA TASK**: N/A


JanuszL:
What happens in case of any spelling mistake? Does it error out?

JanuszL:
> > What happens in case of any spelling mistake? Does it error out?
> 
> @JanuszL There's just a warning printed and the mistake is recorder in `*.spelling` file

I think we can start with that (as this is not an error). And we can try to fix them in the meantime.

szkarpinski:
Closed, replaced with https://github.com/NVIDIA/DALI/pull/5193

PR #5148, opened 2023-11-07
Implement checkpointing for random GPU operators

szkarpinski:
## Category:

## Description:

In this PR I add checkpointing support to `fn.random` operators on the GPU.

`curand_states` (from `randomizer.cuh`) now has `get_states` and `set_states` methods, which allow to export internal state or restore it.

`RNGBase` (from `rng_base.h`) can now save/restore states also on the GPU. The checkpoint of GPU random operator is a pointer to states (as returned by `get_states`). An ability to serialize/deserialize such checkpoints was added.

#### Checkpoints are generally kept on the device
Note that the states returned by `get_states` are still on the GPU and `set_states` also expects its input to be on the GPU. `RNGBase`'s Save and Restore also keep the checkpoint as a pointer to the GPU memory. The checkpoint is copied to the host at the very last moment during `Serialize`. This limits the device->host transfers significantly as we only copy the checkpoints that the user intends to save, while the checkpoints which we just trace and then delete never leave the device.

#### Serialization of `curandState`
During serialization, we treat `curandState` as a sequence of bytes, not caring about its structure. I consider it an implementation detail of curand.

### Tests:

- [ ] Existing tests apply
- [x] New tests added
  - [x] Python tests
  - [ ] GTests
  - [ ] Benchmark
  - [ ] Other
- [ ] N/A

### Documentation
- [ ] Documentation updated
  - [ ] Docstring
  - [ ] Doxygen
  - [ ] RST
  - [ ] Jupyter
  - [ ] Other
- [ ] N/A

#### Requirements
- [x] N/A

**REQ IDs**: N/A

**JIRA TASK**: DALI-3532

### DIFF
--- a/dali/operators/random/rng_base.h
+++ b/dali/operators/random/rng_base.h
@@ -40,7 +40,8 @@ class RNGBase : public Operator<Backend> {
       cpt.MutableCheckpointState() = rng_;
     } else {
       static_assert(std::is_same_v<Backend, GPUBackend>);
-      DALI_FAIL("Checkpointing is not implemented for GPU random operators. ");
+      cpt.SetOrder(order);
+      cpt.MutableCheckpointState() = backend_data_.randomizer_.get_states(order);
     }
   }
 
@@ -53,18 +54,38 @@ class RNGBase : public Operator<Backend> {
       rng_ = rng;
     } else {
       static_assert(std::is_same_v<Backend, GPUBackend>);
-      DALI_FAIL("Checkpointing is not implemented for GPU random operators. ");
+      const auto &states_cpu = cpt.CheckpointState<std::shared_ptr<curandState>>();
+      backend_data_.randomizer_.set_states(states_cpu.get());
     }
   }
 
   std::string SerializeCheckpoint(const OpCheckpoint &cpt) const override {
-    const auto &state = cpt.CheckpointState<BatchRNG<std::mt19937_64>>();
-    return SnapshotSerializer().Serialize(state.ToVector());
+    if constexpr (std::is_same_v<Backend, CPUBackend>) {
+      const auto &state = cpt.CheckpointState<BatchRNG<std::mt19937_64>>();
+      return SnapshotSerializer().Serialize(state.ToVector());
+    } else {
+      static_assert(std::is_same_v<Backend, GPUBackend>);
+      const auto &ptr_gpu = cpt.CheckpointState<std::shared_ptr<curandState>>();
+      size_t n = backend_data_.randomizer_.length();
+      std::vector<curandState> states(n);
+      cudaMemcpy(states.data(), ptr_gpu.get(), n * sizeof(curandState), cudaMemcpyDeviceToHost);
+      return SnapshotSerializer().Serialize(states);
+    }
+
   }
 
   void DeserializeCheckpoint(OpCheckpoint &cpt, const std::string &data) const override {
-    auto deserialized = SnapshotSerializer().Deserialize<std::vector<std::mt19937_64>>(data);
-    cpt.MutableCheckpointState() = BatchRNG<std::mt19937_64>::FromVector(deserialized);
+    if constexpr (std::is_same_v<Backend, CPUBackend>) {
+      auto deserialized = SnapshotSerializer().Deserialize<std::vector<std::mt19937_64>>(data);
+      cpt.MutableCheckpointState() = BatchRNG<std::mt19937_64>::FromVector(deserialized);
+    } else {
+      static_assert(std::is_same_v<Backend, GPUBackend>);
+      auto deserialized = SnapshotSerializer().Deserialize<std::vector<curandState>>(data);
+      auto states = mm::alloc_raw_shared<curandState, mm::memory_kind::device>(deserialized.size());
+      cudaMemcpy(states.get(), deserialized.data(), sizeof(curandState) * deserialized.size(),
+                 cudaMemcpyHostToDevice);
+      cpt.MutableCheckpointState() = states;
+    }
   }
 
  protected:
--- a/dali/operators/random/rng_base.h
+++ b/dali/operators/random/rng_base.h
@@ -70,7 +70,7 @@ class RNGBase : public Operator<Backend> {
     } else {
       static_assert(std::is_same_v<Backend, GPUBackend>);
       const auto &states_gpu = cpt.CheckpointState<curand_states>();
-      size_t n = backend_data_.randomizer_.length();
+      size_t n = states_gpu.length();
       std::vector<curandState> states(n);
       cudaMemcpy(states.data(), states_gpu.states(), n * sizeof(curandState),
                  cudaMemcpyDeviceToHost);
--- a/dali/operators/random/rng_base.h
+++ b/dali/operators/random/rng_base.h
@@ -55,8 +55,8 @@ class RNGBase : public Operator<Backend> {
       rng_ = rng;
     } else {
       static_assert(std::is_same_v<Backend, GPUBackend>);
-      const auto &states_cpu = cpt.CheckpointState<std::shared_ptr<curandState>>();
-      backend_data_.randomizer_.set_states(states_cpu.get());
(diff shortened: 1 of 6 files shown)


szkarpinski:
Found some performance, converting back to draft

szkarpinski:
There are 1024 curand states kept per sample (48B each), serializing them one by one was not the best idea. Now https://github.com/NVIDIA/DALI/pull/5148/commits/a4c521e4a8954739e96af9b8f71315188bb382d3 I serialize array of states to one big bytestring and it reduces the overhead by ~4x. Now the overhead is comparable to that of CPU random operators, so I'd leave it as it is and try to optimize them both in a follow-up.

dali/operators/random/rng_base.h
```diff
@@ -53,18 +55,37 @@ class RNGBase : public Operator<Backend> {
...
   std::string SerializeCheckpoint(const OpCheckpoint &cpt) const override {
-    const auto &state = cpt.CheckpointState<BatchRNG<std::mt19937_64>>();
-    return SnapshotSerializer().Serialize(state.ToVector());
+    if constexpr (std::is_same_v<Backend, CPUBackend>) {
+      const auto &state = cpt.CheckpointState<BatchRNG<std::mt19937_64>>();
+      return SnapshotSerializer().Serialize(state.ToVector());
+    } else {
+      static_assert(std::is_same_v<Backend, GPUBackend>);
+      const auto &ptr_gpu = cpt.CheckpointState<std::shared_ptr<curandState>>();
+      size_t n = backend_data_.randomizer_.length();
```
klecki, line 70: Can the `n` change at any point? I saw some checks for batch size in CPU code, should the `n` be the part of the checkpoint itself?

dali/operators/random/rng_base.h
```diff
@@ -53,18 +55,37 @@ class RNGBase : public Operator<Backend> {
       rng_ = rng;
     } else {
       static_assert(std::is_same_v<Backend, GPUBackend>);
-      DALI_FAIL("Checkpointing is not implemented for GPU random operators. ");
+      const auto &states_cpu = cpt.CheckpointState<std::shared_ptr<curandState>>();
```
klecki, line 58: Isn't it `states_gpu` :P? 
I see that the Serialize and DeserializeCheckpoint do the copies between device and host.

klecki (commented):
Looks ok, one small question.

I also have more general question regarding synchronous operations. Are there any plans to make the checkpointing to run on some stream? Now we do the synchronous copies in few places. (I don't see this as a bad thing, the change is simpler thanks to that, but it may have some impact in cases where the checkpoints are created more often).

dali/operators/random/rng_base.h
szkarpinski: Ooops! The design has changed and the old name remained :P Thanks!

dali/operators/util/randomizer.cuh
```diff
@@ -37,6 +38,21 @@ struct curand_states {
...
   }
 
+  DALI_HOST inline size_t length() const {
+    return len_;
+  }
+
+  DALI_HOST inline std::shared_ptr<curandState> get_states(AccessOrder order) const {
+    auto states = mm::alloc_raw_shared<curandState, mm::memory_kind::device>(len_);
+    cudaMemcpyAsync(states.get(), states_, sizeof(curandState) * len_,
+                    cudaMemcpyDeviceToDevice, order.stream());
```
szkarpinski, line 48: > I also have more general question regarding synchronous operations. Are there any plans to make the checkpointing to run on some stream? Now we do the synchronous copies in few places. (I don't see this as a bad thing, the change is simpler thanks to that, but it may have some impact in cases where the checkpoints are created more often).

@klecki `.checkpoint()` returns a serialized checkpoint to the user as a normal python string, so by design it requires synchronization with the CPU. Hypothetically we could make it asynchronous by returning a promise of a checkpoint or letting the user provide a callback, but I guess it's a bit of overkill and don't plan to do it in the foreseeable future. So unless @stiepan foresees some use case for that I'd keep serialization & checkpoint restoring synchronous.

The only place where I definetely want the copy to be asynchronous is in `get_states`, which will be called after every iteration. This one will be asynchronous, right?

dali/operators/random/rng_base.h
szkarpinski: I suppose that the only way this can happen is when the user tries to restore a pipeline from a checkpoint created with a pipeline with different batch size. 

The issue is not only about batch size, but about all settings of the pipeline and operators. Checking it here doesn't make much sense to me. I'll think about a way to check such things strictly somewhere at the higher level (maybe we should have something some kind of "pipeline hash" and check if it matches?). 

Thank you for reminding me of that - while it's kind of ok to have such behavior in the "experimental" checkpointing, the final implementation should protect the users from harming themselves

dali/operators/util/randomizer.cuh
stiepan: I'd refrain from introducing asynchronous API. Or adding an asynchronous "CPU" stage after the gpu copy for serialization. 

There's one thing that the initial desing misses though and we could revisit it. It seems (maybe that's what @klecki has in mind too) that now when user actually wants a checkpoint, the usage of orders will trigger synchronization with the GPU stage stream. Which means we will wait for all the prefetched GPU batches to complete when, in fact, we only need the back of the GPU queue to be ready. If user wants, for some reason, to make the checkpoints every itration all the time, it may kill GPU prefetching.

dali/operators/random/rng_base.h
szkarpinski: After offline discussion with @stiepan we decided that there aren't that many ways of the user harming themselves, so just adding an assert here is quite a good idea. I'll add one in a moment

dali/operators/util/randomizer.cuh
szkarpinski: After offline discussion: I'll optimize the waiting to be more granular in a follow-up PR

dali/operators/random/rng_base.h
szkarpinski: Adding that assert required surprisingly many changes (see https://github.com/NVIDIA/DALI/pull/5148/commits/af4ccdca32fd762d1b9629138d97b5ac39616d11), because I was not keeping the number of states in the checkpoint: it was just a pointer to the states and the length was deduced from current randomizer's length.

I changed the checkpoint type from `shared_ptr<curandState>` to `curand_states`, which is exactly what I wanted: a shared ptr plus length. This allows me to keep track of the length cleanly

stiepan (approved):
Looks nice! Please remeber to add the more fine-grained synchronization to the roadmap.

dali/operators/random/rng_base.h
klecki: This change solves my reservation from the first comment, where the state/checkpoint was a pointer to memory, but the length of the memory was tied to the actual state of the operator.

dali/operators/random/rng_base.h
```diff
@@ -53,18 +55,41 @@ class RNGBase : public Operator<Backend> {
...
   std::string SerializeCheckpoint(const OpCheckpoint &cpt) const override {
-    const auto &state = cpt.CheckpointState<BatchRNG<std::mt19937_64>>();
-    return SnapshotSerializer().Serialize(state.ToVector());
+    if constexpr (std::is_same_v<Backend, CPUBackend>) {
+      const auto &state = cpt.CheckpointState<BatchRNG<std::mt19937_64>>();
+      return SnapshotSerializer().Serialize(state.ToVector());
+    } else {
+      static_assert(std::is_same_v<Backend, GPUBackend>);
+      const auto &states_gpu = cpt.CheckpointState<curand_states>();
+      size_t n = backend_data_.randomizer_.length();
```
klecki, line 73:
```suggestion
      size_t n = states_gpu.length();
```
Don't we want this^?

klecki (approved):
One more place with state vs curand_state, otherwise ok.

szkarpinski:
!build

PR #5154, opened 2023-11-08
Set __signature__ for fn API

klecki:
Use the code for interface generation. We already generate Signature objects as intermediate step of building
interface files (.pyi), so that object can be used directly for the fn API functions.

Instead of adding a lot of conditionals, the type annotations are filtered at the end from the resulting signature.
(The Parameter objects are immutable, so we need to recreate them without the annotation part).

Setting `__signature__` doesn't change the semantics but can serve as a documentation in tools relying on runtime information like Python help() and Jupyter.

### Tests:

- [ ] Existing tests apply
- [ ] New tests added
  - [ ] Python tests
  - [ ] GTests
  - [ ] Benchmark
  - [ ] Other
- [ ] N/A
I have no idea what to test here?

### Documentation
- [x] Documentation updated
  - [x] the `__signature__` is kind of a documentation
  - [ ] Docstring
  - [ ] Doxygen
  - [ ] RST
  - [ ] Jupyter
  - [ ] Other
- [ ] N/A

#### Requirements
- [x] N/A

**REQ IDs**: N/A

**JIRA TASK**: N/A

### DIFF
--- a/dali/python/nvidia/dali/fn/__init__.py
+++ b/dali/python/nvidia/dali/fn/__init__.py
@@ -89,9 +89,16 @@ def _wrap_op_fn(op_class, wrapper_name, wrapper_doc):
         else:
             return op_wrapper(*inputs, **kwargs)
 
+    from nvidia.dali.ops import _signatures
+    from nvidia.dali import backend as _b
     fn_wrapper.__name__ = wrapper_name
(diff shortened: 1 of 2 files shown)


mzient (approved):

klecki:
!build

szalpal (approved):

PR #5158, opened 2023-11-09
Fix the check for dynamically generated number of outputs

klecki:
## Description:
The documentation for HasOutputFunction suggests that it allows to check if operator calculates the number of outputs dynamically or if it is set statically.

We have two entities:
OutputFn - calculates the number of outputs dynamically based on OpSpec AdditionalOutputsFn - calculates dynamically based on OpSpec additional outputs.

If any of those is present, number of outputs is not know statically.

HasOutputFn allowed to detect only the first case, now it will detect any of the two.
This check is only used in Python when generating signatures.

This additionally fixes incorrect return type annotation for some operators that indicate dynamic number of outputs using AdditionalOutputsFn.

It appears that it could return negative number of additional outputs, so I can't use the NumOutputs as a lower bound.

### Tests:

- [ ] Existing tests apply
- [ ] New tests added
  - [ ] Python tests
  - [ ] GTests
  - [ ] Benchmark
  - [ ] Other
- [x] N/A
I inspected the the signatures of operators that have the AdditionalOutputsFn, now the annotation
suggests dynamic number of outputs.

### Documentation
- [x] Documentation updated
  - [x] Generated signatures
  - [ ] Docstring
  - [ ] Doxygen
  - [ ] RST
  - [ ] Jupyter
  - [ ] Other
- [ ] N/A

#### Requirements
- [ ] N/A

**REQ IDs**: N/A

**JIRA TASK**: N/A

### DIFF
--- a/dali/pipeline/operator/op_schema.cc
+++ b/dali/pipeline/operator/op_schema.cc
@@ -593,7 +593,7 @@ DLL_PUBLIC const DeprecatedArgDef &OpSchema::DeprecatedArgMeta(const std::string
 
 
 bool OpSchema::HasOutputFn() const {
-  return static_cast<bool>(output_fn_);
+  return static_cast<bool>(output_fn_) || static_cast<bool>(additional_outputs_fn_);
 }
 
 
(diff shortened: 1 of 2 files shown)


klecki:
!build

stiepan (approved):

klecki:
!build

szalpal (approved):

PR #5159, opened 2023-11-09
Generate overloads for Multiple Input Sets in Python interface files

klecki:
obraz

## Description:
Expand the API coverage by providing two overloads for each API function.
* Primary one accepts only `DataNodes` and has as constrained type as possible (`DataNode` for 1 output - `Union` return types are much less useful in IDE and with mypy)
* Secondary that allows to pass a `List[DataNode]` for inputs, indicating Multiple Input Sets handling. Such lists can be passed as one of several input, and the other inputs that are `DataNodes` will be bcast to match that.

Python overload resolution appears to be sequential - first matching function is selected, so when using regular `DataNode` inputs, we pick the overload with nicer return types.
Anything indicating MIS will fall into the other overload. 

Next step is to include TensorLike/SampleLike constants in the signatures.

Signature generation

### Tests:

- [ ] Existing tests apply
- [x] New tests added
  - [x] Python tests
  - [ ] GTests
  - [ ] Benchmark
  - [ ] Other
- [ ] N/A

### Documentation
- [ ] Documentation updated
  - [ ] Docstring
  - [ ] Doxygen
  - [ ] RST
  - [ ] Jupyter
  - [ ] Other
- [ ] N/A

#### Requirements
- [ ] N/A

**REQ IDs**: N/A

**JIRA TASK**: N/A

### DIFF
--- a/dali/python/nvidia/dali/ops/_signatures.py
+++ b/dali/python/nvidia/dali/ops/_signatures.py
@@ -75,7 +75,6 @@ _enum_mapping = {
     types.DALIInterpType: _DALIInterpType
 }
 
-
 _MAX_INPUT_SPELLED_OUT = 10
 
 
@@ -140,7 +139,6 @@ def _get_annotation_input_regular(schema):
     return _DataNode
 
 
-
 def _get_annotation_return_regular(schema):
     """Produce the return annotation for DALI operator suitable for primary, non-MIS overload.
     Note the flattening, single output is not packed in Sequence.
@@ -202,7 +200,6 @@ def _get_annotation_return_mis(schema):
     return return_annotation
 
 
-
 def _get_positional_input_params(schema, input_annotation_gen=lambda x, y: _DataNode):
     """Get the list of positional only inputs to the operator.
 
@@ -214,10 +211,12 @@ def _get_positional_input_params(schema, input_annotation_gen=lambda x, y: _Data
     """
     param_list = []
     if not schema.HasInputDox() and schema.MaxNumInput() > _MAX_INPUT_SPELLED_OUT:
-        param_list.append(Parameter("input", Parameter.VAR_POSITIONAL, annotation=input_annotation_gen(schema)))
+        param_list.append(
+            Parameter("input", Parameter.VAR_POSITIONAL, annotation=input_annotation_gen(schema)))
     else:
         for i in range(schema.MaxNumInput()):
-            param_list.append(_get_positional_input_param(schema, i, annotation=input_annotation_gen(schema)))
+            param_list.append(
+                _get_positional_input_param(schema, i, annotation=input_annotation_gen(schema)))
     return param_list
 
 
@@ -314,7 +313,8 @@ def _call_signature(schema, include_inputs=True, include_kwargs=True, include_se
         param_list.append(Parameter("self", kind=Parameter.POSITIONAL_ONLY))
 
     if include_inputs:
-        param_list.extend(_get_positional_input_params(schema, input_annotation_gen=input_annotation_gen))
+        param_list.extend(
+            _get_positional_input_params(schema, input_annotation_gen=input_annotation_gen))
 
     if include_kwargs:
         param_list.extend(_get_keyword_params(schema, all_args_optional=all_args_optional))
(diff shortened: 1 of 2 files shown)


szalpal (approved):

dali/python/nvidia/dali/ops/_signatures.py
```diff
@@ -131,15 +131,87 @@ def _get_positional_input_param(schema, idx):
...
+    """Annotation for function that handles Multiple input sets overload.
+    Note that DALI does a lot of flattening, so single-element sequences are transformed to
+    just the element.
+    """
+    if schema.HasOutputFn():
+        # Dynamic number of outputs, not known at "compile time"
+        # We can return single or multiple outputs in regular case (same as primary overload),
+        # a list of single outputs or a list of multiple outputs for MIS or None.
+        return_annotation = Union[_DataNode, Sequence[_DataNode], List[_DataNode],
+                                  List[Sequence[_DataNode]], None]
```
mzient, line 182: MIS overload never returns a single DataNode.

dali/python/nvidia/dali/ops/_signatures.py
```diff
@@ -131,15 +131,87 @@ def _get_positional_input_param(schema, idx):
...
+        elif num_regular_output == 1:
+            # This allows for type hints with single-return operators to work correctly with MIS
+            # as the return type matches the overload for their input type.
+            return_annotation = Union[_DataNode, List[_DataNode]]
+        else:
+            # Here we could utilize the fact, that the tuple has known length, but we can't
+            # as DALI operators return a list
+            # Also, we don't advertise the actual List type, hence the Sequence, but we say
+            # that the outermost return type of MIS is a List.
+            return_annotation = Union[Sequence[_DataNode], List[Sequence[_DataNode]]]
```
mzient, line 197: MIS overload never returns a single DataNode.

dali/python/nvidia/dali/ops/_signatures.py
```diff
@@ -131,15 +131,87 @@ def _get_positional_input_param(schema, idx):
...
+                                  List[Sequence[_DataNode]], None]
+    else:
+        # Call it with a dummy spec, as we don't have Output function
+        num_regular_output = schema.CalculateOutputs(_b.OpSpec(""))
+        if num_regular_output == 0:
+            return_annotation = None
+        elif num_regular_output == 1:
+            # This allows for type hints with single-return operators to work correctly with MIS
+            # as the return type matches the overload for their input type.
+            return_annotation = Union[_DataNode, List[_DataNode]]
```
mzient, line 191: MIS overload never returns a single DataNode.

klecki: `fn.copy([input])` returns a DataNode as opposed to `fn.copy([input, input])` which returns `List[DataNode]`.

klecki:
!build

klecki:
!build

dali/python/nvidia/dali/ops/_signatures.py
```diff
@@ -131,15 +131,92 @@ def _get_positional_input_param(schema, idx):
...
+
+
+def _get_annotation_input_mis(schema):
+    """Return the annotation for multiple input sets, used for the secondary operator overload.
+    A function is used as a global variable can be confused with type alias.
+    Handles special case for operators with one input.
+    """
+    if schema.MinNumInput() == 1 and schema.MaxNumInput() == 1:
+        return List[_DataNode]
+    return Union[_DataNode, List[_DataNode]]
```
stiepan, line 171: I am confused a little here (likely not seeing the context) - we do we need union with `_DataNode` when the num inputs is not exactly one, but can get rid of it otherwise?

klecki: So, we have primary overload of the form:
```
def operator(input0: DataNode, input1: DataNode, ...) -> ...: ...
```
and the secondary one:
```
def operator(input0: Union[DataNode, List[DataNode]], input1: Union[DataNode, List[DataNode]], ...) -> ...: ...
```
The first catches the case where all inputs are DataNodes, the second will catch cases where at least one of the input is a list (the rest of the inputs can still be DataNodes and can be broadcast to match the length of the list).

If we have just one input, we have two overloads:
```
def operator(input0: DataNode, /, ...) -> ...: ...
def operator(input0: List[DataNode], /, ...) -> ...: ...
```
we don't heed the union in the second one, as the only DataNode inputs always matches the first overload.

stiepan: Oh, I see, I was missing the broadcasting case in my head. Thank you!

stiepan (approved):

PR #5165, opened 2023-11-14
Support checkpointing in `tfrecord` and `mxnet` readers

szkarpinski:
## Category:

## Description:

This PR adds checkpointing support to `mxnet` and `tfrecord` readers.

They are both based on `IndexedFileLoader` (`TFRecordReader` uses it directly, `MXNetReader` uses `RecordIOLoader` which is a subclass of it).

## Additional information:

Adding checkpointing support is a simple task, similar for every loader and reader.

The following changes were required to enable checkpointing in `IndexedFileLoader`:
- Make it inherit from `Loader<..., supports_checkpointing=true>`
- Implement `Skip()` method. `Skip` should behave like `ReadSample` in terms of side effects, but should skip a sample instead of reading it. This method is used to implement fast-forwarding in `Loader` baseclass.
- Change `Reset` to use `virtual_shard_id_` (which is the shard currently processed)
  instead of `shard_id_` (which is the initial shard requested by the user). See [2] for more details.

The following changes were required to enable checkpointing in the readers:
- Make it inherit from `DataReader<..., supports_checkpointing=true>`. See [1] for more details.
- Store the very first snapshot in the constructor with `this->SetInitialSnapshot()`.
  Subsequent snapshots are saved by `DataReader` baseclass.

#### [1] Changing `DataReader` template parameters
Inheriting from `DataReader<..., supports_checkpointing=true>` might look strange in the diff, because the `DataReader` is defined as:
```cpp
template <typename Backend, typename LoadTarget,
          typename ParseTarget = LoadTarget, bool supports_checkpointing = false>
```
and is used mostly as:
```cpp
DataReader<Backend, Target>
```
so to enable checkpointing one needs to add two parameters:
```cpp
DataReader<Backend, Target, Target, true>
```

#### [2] `virtual_shard_id_`
`virtual_shard_id_` and `shard_id_`  might differ when `stick_to_shard=False`.

This change doesn't impact existing code, because `Reset` is normally called after each full pass over the data, so then those two are equal. It might happen that a checkpoint is saved when the reader was is processing a different shard that `shard_id_`,  so to restore from such checkpoint we need to be able to reset to the current shard (`virtual_shard_id_`).

The same change was made in `FileReader` in #4954.

- `IndexedFileLoader` and `RecordIOLoader`
- `TFRecordReader` and `MXNetReader`

### Tests:

A minor change in the `check_reader_checkpointing` helper was required to support single-output readers.

- [ ] Existing tests apply
- [x] New tests added
  - [x] Python tests
  - [ ] GTests
  - [ ] Benchmark
  - [ ] Other
- [ ] N/A

### Documentation
- [ ] Documentation updated
  - [ ] Docstring
  - [ ] Doxygen
  - [ ] RST
  - [ ] Jupyter
  - [ ] Other
- [ ] N/A

#### Requirements
- [x] N/A

**REQ IDs**: N/A

**JIRA TASK**: DALI-3691

### DIFF
--- a/dali/operators/reader/loader/indexed_file_loader.h
+++ b/dali/operators/reader/loader/indexed_file_loader.h
@@ -1,4 +1,4 @@
-// Copyright (c) 2017-2022, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+// Copyright (c) 2017-2023, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -32,7 +32,7 @@
 
 namespace dali {
 
-class IndexedFileLoader : public Loader<CPUBackend, Tensor<CPUBackend>> {
+class IndexedFileLoader : public Loader<CPUBackend, Tensor<CPUBackend>, true> {
  public:
   explicit IndexedFileLoader(const OpSpec& spec)
     : Loader(spec),
@@ -172,6 +172,10 @@ class IndexedFileLoader : public Loader<CPUBackend, Tensor<CPUBackend>> {
     return;
   }
 
+  void Skip() override {
+    MoveToNextShard(current_index_++);
+  }
+
   ~IndexedFileLoader() override {
     current_file_.reset();
   }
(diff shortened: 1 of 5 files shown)


mzient (approved):

awolant (approved):

szkarpinski:
!build

PR #5168, opened 2023-11-15
Adjust configuration for autoformatting with black

klecki:
## Description:
Temporarily disable the linter in this commit, before the whole repository is adjusted to new formatting.

Add black to third party deps listing to include it in the deps update procedure.

Use latest version but require only the one for given year according to the compabilty policy.

Remove obsolete configs related to previous linter or exceptions related to the formatting of some parts of DALI.

Linting, formatting, Python

### Tests:

- [ ] Existing tests apply
- [ ] New tests added
  - [ ] Python tests
  - [ ] GTests
  - [ ] Benchmark
  - [ ] Other
- [x] N/A

### Documentation
- [x] Documentation updated
  - [ ] Docstring
  - [ ] Doxygen
  - [ ] RST
  - [ ] Jupyter
  - [x] Other
- [ ] N/A

#### Requirements
- [ ] N/A

**REQ IDs**: N/A

**JIRA TASK**: N/A


klecki:
!build

.flake8.ag
```diff
@@ -1,13 +1,14 @@
...
+# per-file-ignore do not work well with absolute path (or more complex paths),
+# so just disable them by a special config
 [flake8]
-indent-size = 2
 color = auto
 count = True
 doctests = False
-max-doc-length = 100
+max-doc-length = 120
 max-line-length = 100
```
stiepan, line 8: In the end, does it change anything (given the fact that max-line-length stays the same)?

stiepan (approved):

docker/Dockerfile
```diff
@@ -20,7 +20,7 @@ ENV PYVER=${PYVER} \
 RUN ln -s /opt/python/cp${PYV}* /opt/python/v
 
 # install Python bindings and patch it to use the clang we have here
-RUN pip install future setuptools wheel clang==14.0 flake8 astunparse gast dm-tree black && \
+RUN pip install future setuptools wheel clang==14.0 flake8 astunparse gast dm-tree black==23.11.0 && \
```
szalpal, line 23: Adding this here stipulates, that this particular line shall be changed with every deps update, is that correct?

cmake/lint.cmake
```diff
@@ -31,15 +31,26 @@ set(AUTOGRAPH_LINT_PATHS
...
           flake8 --config=${PROJECT_SOURCE_DIR}/.flake8 ${PYTHON_LINT_PATHS}
         COMMAND
           flake8 --config=${PROJECT_SOURCE_DIR}/.flake8.ag ${AUTOGRAPH_LINT_PATHS}
         COMMENT
           "Performing Python linter check"
 )
+add_dependencies(lint-python-flake lint-python-black)
+
+add_custom_target(lint-python)
+add_dependencies(lint-python lint-python-flake)
```
szalpal, line 53: Why adding flake dependency from black? Would it be better to keep them separate?
```suggestion
add_custom_target(lint-python)
add_dependencies(lint-python lint-python-flake lint-python-black)
```

cmake/lint.cmake
```diff
@@ -31,15 +31,26 @@ set(AUTOGRAPH_LINT_PATHS
...
+          "Performing black Python formatting check"
+)
 
-add_custom_target(lint-python
+
+add_custom_target(lint-python-flake
         COMMAND
           flake8 --config=${PROJECT_SOURCE_DIR}/.flake8 ${PYTHON_LINT_PATHS}
         COMMAND
           flake8 --config=${PROJECT_SOURCE_DIR}/.flake8.ag ${AUTOGRAPH_LINT_PATHS}
```
szalpal, line 46: Since we are unifying the autograph, shouldn't this be removed?

.flake8
```diff
@@ -5,5 +5,7 @@ doctests = False
 max-doc-length = 120
 max-line-length = 100
 statistics = True
+# E203 - whitespace before ':' - PEP8 non-compliant warning in slicing expressions
+extend-ignore = E203
```
szalpal, line 9: Please adjust also remaining checks according to the black-onboarding document

.flake8.ag
```diff
@@ -1,13 +1,14 @@
...
 statistics = True
+# E203 - whitespace before ':' - PEP8 non-compliant warning in slicing expressions
 # E741 - ambiguous variable name 'l'
 # E731 - do not assign a lambda expression, use a def
 # F401 - 'autograph.operators....' imported but unused - used to bring types into another module
-extend-ignore = E741,E731,F401
+extend-ignore = E203,E741,E731,F401
```
szalpal, line 14: Should the checks I've referred to in previous comment be adjusted here also?

docker/Dockerfile
klecki: yes

cmake/lint.cmake
klecki: My intention is to check the linter only after the code is already formatted. Otherwise it is a moving target.

cmake/lint.cmake
klecki: There are 3 warnings that are very common in the ag implementation and tests:
```
# E741 - ambiguous variable name 'l'
# E731 - do not assign a lambda expression, use a def
# F401 - 'autograph.operators....' imported but unused - used to bring types into another module
```
I didn't want to rework couple hundred lines and wanted to keep it as is. When I tried to add per-file-excludes, I didn't manage to write a pattern that would detect `autograph` generically in the path name. I would need to list all absolute paths by hand, and this changes based on the build location. The relative paths don't work well. 
For some reason only one exclude out of several worked. 
At this point I just kept the separate config, and run it with few additional excludes.
If you have better idea on how to merge those configs I would be happy to see it.

.flake8
klecki: It didn't trigger on our code-base so I kinda missed it, added.

.flake8.ag
klecki: Yes, added

docker/Dockerfile
szalpal: Hopefully we'll not forget about this ;)

klecki: That's why I put it in the third_party readme.

szalpal (approved):

klecki:
!build

PR #5172, opened 2023-11-16
Fix sharding in Caffe reader

szkarpinski:
## Category:

### The problem
This PR fixes a problem with sharding in Caffe reader. The observed problem was that shards with `shard_id > 0` were containing samples they shouldn't, in particular there was an overlap between subsequent shards.

### The root cause
Our LMDB wrapper (`IndexedLMDB`) assumed that the cursor is pointing at index `0` when created, which turned out to be false - actually it seems that the cursor position is initialized to `-1`.  

I found no mention in the docs of initial cursor position being defined or undefined,  but in their [getting started](http://www.lmdb.tech/doc/starting.html) they state that:

> For example, to list all key/value pairs in a database, use operation `MDB_FIRST` for the first call to `mdb_cursor_get()`, and `MDB_NEXT` on subsequent calls, until the end is hit.

which suggests that the cursor first needs to be positioned at initial position with `MDB_FIRST`.

### The solution
I now reset the cursor with `MDB_FIRST` in `Open`.

### Why did it work with one shard?
When in the first shard, we reset to `index = 0`, causing `SeekByIndex` to use absolute `MDB_FIRST` instead of relative `MDB_NEXT`.
https://github.com/NVIDIA/DALI/blob/b350581a3eca6491fd772903666aa64fe0b23336/dali/operators/reader/loader/lmdb.h#L80-L90

- LMDB loader, Caffe(2) readers

### Tests:

- [ ] Existing tests apply
- [x] New tests added
  - [ ] Python tests
  - [ ] GTests
  - [ ] Benchmark
  - [ ] Other
- [ ] N/A

### Documentation
- [ ] Documentation updated
  - [ ] Docstring
  - [ ] Doxygen
  - [ ] RST
  - [ ] Jupyter
  - [ ] Other
- [ ] N/A

#### Requirements
- [x] N/A

**REQ IDs**: N/A

**JIRA TASK**: N/A

### DIFF
--- a/dali/operators/reader/loader/lmdb.h
+++ b/dali/operators/reader/loader/lmdb.h
@@ -1,4 +1,4 @@
-// Copyright (c) 2017-2022, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+// Copyright (c) 2017-2023, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -60,6 +60,8 @@ class IndexedLMDB {
     mdb_size_ = stat.ms_entries;
     LOG_LINE << "lmdb " << num_ << " " << db_path_
              << " has " << mdb_size_ << " entries" << std::endl;
+    MDB_val tmp_key, tmp_value;
+    mdb_cursor_get(mdb_cursor_, &tmp_key, &tmp_value, MDB_FIRST);
     mdb_index_ = 0;
   }
   size_t GetSize() const { return mdb_size_; }
--- a/dali/operators/reader/loader/lmdb.h
+++ b/dali/operators/reader/loader/lmdb.h
@@ -60,8 +60,11 @@ class IndexedLMDB {
     mdb_size_ = stat.ms_entries;
     LOG_LINE << "lmdb " << num_ << " " << db_path_
              << " has " << mdb_size_ << " entries" << std::endl;
+
+    // We must reset the cursor to the first entry, because mdb_cursor_open doesn't place it there.
     MDB_val tmp_key, tmp_value;
     mdb_cursor_get(mdb_cursor_, &tmp_key, &tmp_value, MDB_FIRST);
+
     mdb_index_ = 0;
   }
   size_t GetSize() const { return mdb_size_; }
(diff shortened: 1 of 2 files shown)


dali/test/python/reader/test_caffe.py
```diff
@@ -99,6 +99,50 @@ def test_caffe_reader_alias():
...
+    def get_data(shard_id, num_shards, stick_to_shard):
+        p = pipeline(shard_id, num_shards, stick_to_shard)
+        p.build()
+        size = p.reader_meta()['Reader']['epoch_size_padded']
+
+        # This should return some unique number for each sample
+        def sample_id(sample):
+            return sample.as_array().sum()
+
+        return {sample_id(p.run()[0]) for _ in range(size * num_shards)}
```
JanuszL, line 124:
```suggestion
        return {sample_id(p.run()[0]) for _ in range(size / num_shards)}
```

JanuszL (approved):

dali/test/python/reader/test_caffe.py
szkarpinski: Fixed in https://github.com/NVIDIA/DALI/pull/5172/commits/851755778a22d97fee12d985de9eba501f255eaa

JanuszL: 👍

JanuszL (approved):

dali/operators/reader/loader/lmdb.h
```diff
@@ -60,6 +60,8 @@ class IndexedLMDB {
     mdb_size_ = stat.ms_entries;
     LOG_LINE << "lmdb " << num_ << " " << db_path_
              << " has " << mdb_size_ << " entries" << std::endl;
+    MDB_val tmp_key, tmp_value;
+    mdb_cursor_get(mdb_cursor_, &tmp_key, &tmp_value, MDB_FIRST);
```
klecki, line 64: Can you add the comment here why it is needed?

klecki (approved):

dali/operators/reader/loader/lmdb.h
szkarpinski: Added, thank you

szkarpinski:
!build

klecki (approved):

PR #5179, opened 2023-11-20
Introduce TensorLike to signatures utilizing Array Interface

klecki:
## Description:
The generated signatures now include the TensorLike inputs when there is a possibility
to pass a DataNode.

There are two new type aliases introduced in the `_typing` module. 
* `TensorLikeIn` is intended to describe anything that can be passed as a constant to input. This include arrays/tensors of accepted types, sequences of Python scalar numbers and the scalar numbers themselves (int and float). We don't have type constraints until the operator runs and validates the inputs, so we need to list all possibilities.
* `TensorLikeArg` is intended for argument inputs. As we already describe Sequences and scalar there (with a specific types as well), we keep this alias constrained to ArrayInterface.

There are two now protocol classes: https://mypy.readthedocs.io/en/stable/protocols.html#simple-user-defined-protocols

They describe classes implementing NumPy Array Interface (it has a special method `__array__`) and Numba CUDA Array Interface (it has a property `__cuda_array_interface__`).

In the case of constants we actually handle only the NumPy Array Interface - and even this is an approximation.
Our implementation handles only PyTorch and MxNet tensors and NumPy arrays, using the `.flatten().to_list()` sequence.
Still, those classes do implement the `__array__()` method (even if only to throw an exception when they are GPU-placed),
and this is the closest and most generic approximation I can come up with. 

The IDEs are also able to look up the docstrings for those aliases and definition when navigating the interface file.

VSCode example (it keeps the aliases):
obraz
PyCharm (it unpacks the aliases):
obraz

### Tests:

- [ ] Existing tests apply
- [x] New tests added
  - [x] Python tests
  - [ ] GTests
  - [ ] Benchmark
  - [ ] Other
- [ ] N/A

### Documentation
- [x] Documentation updated
  - [x] Docstring
  - [ ] Doxygen
  - [ ] RST
  - [ ] Jupyter
  - [ ] Other
- [ ] N/A

#### Requirements
- [ ] N/A

**REQ IDs**: N/A

**JIRA TASK**: N/A


klecki:
!build

dali/test/python/type_annotations/test_typing_pipelines.py
```diff
@@ -154,6 +154,34 @@ def rn50_ops_pipe():
...
+        expect_data_node(const_int, const_float, const_list, const_tuple, const_torch, const_np)
+        return const_int, const_float, const_list, const_tuple, const_torch, const_np
+
+    pipe = const_copy_pipe()
+    expect_pipeline(pipe)
+    pipe.build()
+    const_int, const_float, const_list, const_tuple, const_torch, const_np = pipe.run()
+    expect_tensor_list(const_int, const_float, const_list, const_tuple, const_torch, const_np)
+    assert np.array_equal(np.array(const_int.as_tensor()), np.full((10, 1), 1))
+    assert np.array_equal(np.array(const_float.as_tensor()), np.full((10, 1), 2.0))
```
mzient, line 178: Scalar handling has been fixed! This will fail.
```suggestion
    assert np.array_equal(np.array(const_int.as_tensor()), np.full((10,), 1))
    assert np.array_equal(np.array(const_float.as_tensor()), np.full((10,), 2.0))
```

klecki: done

klecki:
!build

mzient (approved):

klecki:
!build

PR #5184, opened 2023-11-21
Support checkpointing in Nemo reader

szkarpinski:
## Category:

## Description:

This PR adds checkpointing support to `fn.readers.nemo_asr`.

## Additional information:

**Note:** Nemo reader supports `shuffle_after_epoch`, so two additional steps (marked with **[!]** in the description below) are needed.

The following changes were required to enable checkpointing in the loader:
- Make it inherit from `Loader<..., supports_checkpointing=true>`
- Implement `Skip()` method. `Skip` should behave like `ReadSample` in terms of side effects, but should skip a sample instead of reading it. This method is used to implement fast-forwarding in `Loader` baseclass.
- Change `Reset` to use `virtual_shard_id_` (which is the shard currently processed)
  instead of `shard_id_` (which is the initial shard requested by the user). See [2] for more details.
- **[!]** Restore `current_epoch_ ` from the state in `RestoreStateImpl`. This loader needs to keep track of the epoch, because it's used as a seed for shuffling.
- **[!]** Change the way shuffling after epoch works. See [3] for more details.

The following changes were required to enable checkpointing in the reader:
- Make it inherit from `DataReader<..., supports_checkpointing=true>`. See [1] for more details.
- Store the very first snapshot in the constructor with `this->SetInitialSnapshot()`.
  Subsequent snapshots are saved by `DataReader` baseclass.

#### [1] Changing `DataReader` template parameters
Inheriting from `DataReader<..., supports_checkpointing=true>` might look strange in the diff, because the `DataReader` is defined as:
```cpp
template <typename Backend, typename LoadTarget,
          typename ParseTarget = LoadTarget, bool supports_checkpointing = false>
```
and is used mostly as:
```cpp
DataReader<Backend, Target>
```
so to enable checkpointing one needs to add two parameters:
```cpp
DataReader<Backend, Target, Target, true>
```

#### [2] `virtual_shard_id_`
`virtual_shard_id_` and `shard_id_`  might differ when `stick_to_shard=False`.

This change doesn't impact existing code, because `Reset` is normally called after each full pass over the data, so then those two are equal. It might happen that a checkpoint is saved when the reader was is processing a different shard that `shard_id_`,  so to restore from such checkpoint we need to be able to reset to the current shard (`virtual_shard_id_`).

The same change was made in `FileReader` in #4954.

#### [3] Shuffling after epoch
Originally, the list of indices was shuffled multiple times, once after each epoch. Each time shuffling was applied to already shuffled list. This meant that the list of indices after three epochs was:
```
shuffle(shuffle(shuffle([1, 2, ..., n], seed=S+0), seed=S+1), seed=S+2)
```
To make restoring from checkpoint easier, we now restore the list of indices to 1..n before shuffling it.
So now after three epochs it's just:
```
shuffle([1, 2, ..., n], seed=S+2)
```
It's still as random as it was, but much easier to restore.

- Nemo reader and loader

### Tests:
A minor change in the check_reader_checkpointing helper was required to support single-output readers.

- [ ] Existing tests apply
- [x] New tests added
  - [x] Python tests
  - [ ] GTests
  - [ ] Benchmark
  - [ ] Other
- [ ] N/A

### Documentation
- [ ] Documentation updated
  - [ ] Docstring
  - [ ] Doxygen
  - [ ] RST
  - [ ] Jupyter
  - [ ] Other
- [ ] N/A

#### Requirements
- [x] N/A

**REQ IDs**: N/A

**JIRA TASK**: DALI-3705

### DIFF
--- a/dali/operators/reader/loader/nemo_asr_loader.cc
+++ b/dali/operators/reader/loader/nemo_asr_loader.cc
@@ -1,4 +1,4 @@
-// Copyright (c) 2020-2022, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+// Copyright (c) 2020-2023, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -94,10 +94,16 @@ void NemoAsrLoader::PrepareMetadataImpl() {
 }
 
 void NemoAsrLoader::Reset(bool wrap_to_shard) {
-  current_index_ = wrap_to_shard ? start_index(shard_id_, num_shards_, SizeImpl()) : 0;
+  current_index_ = wrap_to_shard ? start_index(virtual_shard_id_, num_shards_, SizeImpl()) : 0;
   current_epoch_++;
 
   if (shuffle_after_epoch_) {
+    if (IsCheckpointingEnabled()) {
+      // With checkpointing enabled dataset order must be easy to restore.
+      // Shuffling is run with different seed every epoch, so this doesn't
+      // reduce the randomness.
+      std::iota(shuffled_indices_.begin(), shuffled_indices_.end(), 0);
+    }
     std::mt19937 g(kDaliDataloaderSeed + current_epoch_);
     std::shuffle(shuffled_indices_.begin(), shuffled_indices_.end(), g);
   }
@@ -185,7 +191,15 @@ void NemoAsrLoader::ReadSample(AsrSample& sample) {
   ));  // NOLINT
 }
 
+void NemoAsrLoader::Skip() {
+  MoveToNextShard(++current_index_);
+}
+
 Index NemoAsrLoader::SizeImpl() {
   return entries_.size();
+}
+
+void NemoAsrLoader::RestoreStateImpl(const LoaderStateSnapshot &state) {
+  current_epoch_ = state.current_epoch;
 }
 }  // namespace dali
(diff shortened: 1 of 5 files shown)


dali/operators/reader/loader/nemo_asr_loader.cc
```diff
@@ -94,10 +94,16 @@ void NemoAsrLoader::PrepareMetadataImpl() {
...
 void NemoAsrLoader::Reset(bool wrap_to_shard) {
-  current_index_ = wrap_to_shard ? start_index(shard_id_, num_shards_, SizeImpl()) : 0;
+  current_index_ = wrap_to_shard ? start_index(virtual_shard_id_, num_shards_, SizeImpl()) : 0;
   current_epoch_++;
 
   if (shuffle_after_epoch_) {
+    if (IsCheckpointingEnabled()) {
+      // With checkpointing enabled dataset order must be easy to restore.
+      // Shuffling is run with different seed every epoch, so this doesn't
+      // reduce the randomness.
```
mzient, line 104: It actually will, kind of.

szkarpinski: User-provided seed is not used here, so IMO we can use sequential seeds safely here

awolant: We probably should add a small note to release notes that enabling checkpointing for this reader changes the order or something like this as this might be important for the end user.

awolant (approved):

szkarpinski:
!build

PR #5188, opened 2023-11-21
Remove redundant overload signature for operators without inputs

klecki:
## Description:
We provide two overload signatures for operators, but they are disambiguated based on the operator inputs. When the operator has no inputs only one signature is needed.

Notable cases: readers.

Generated interface files with stubs

### Tests:
Just check the output and IDE integration manually.

- [ ] Existing tests apply
- [ ] New tests added
  - [ ] Python tests
  - [ ] GTests
  - [ ] Benchmark
  - [ ] Other
- [x] N/A

### Documentation
- [x] Documentation updated
  - [x] Signatures for no-input ops.
  - [ ] Docstring
  - [ ] Doxygen
  - [ ] RST
  - [ ] Jupyter
  - [ ] Other
- [ ] N/A

#### Requirements
- [ ] N/A

**REQ IDs**: N/A

**JIRA TASK**: N/A

### DIFF
--- a/dali/python/nvidia/dali/ops/_signatures.py
+++ b/dali/python/nvidia/dali/ops/_signatures.py
@@ -341,9 +341,20 @@ def inspect_repr_fixups(signature: str) -> str:
     return signature.replace("NoneType", "None")
 
 
-def _gen_fn_signature(schema, schema_name, fn_name):
-    """Write the stub of the fn API function with the docstring, for given operator.
-    Include two overloads: with regular inputs and secondary accepting MIS.
+def _gen_fn_signature_no_input(schema, schema_name, fn_name):
+    """In case of no input, we don't have overload set, as there is no MIS involved,
+    we write only the default signature, without involving @overload decorator.
+    """
+    return f"""
+def {fn_name}{_call_signature(schema, include_inputs=True, include_kwargs=True)}:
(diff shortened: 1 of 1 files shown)


klecki:
!build

awolant (approved):

dali/python/nvidia/dali/ops/_signatures.py
```diff
@@ -356,7 +367,7 @@ def _gen_fn_signature(schema, schema_name, fn_name):
     pylance, resulting again in the Union[DataNode, List[DataNode]] return type (for single output),
     keeping us in the MIS realm in the simple case.
     """
-    return inspect_repr_fixups(f"""
+    return f"""
 @overload
```
mzient, line 371: Why not have some `if` here and only append the second case when relevant? As-is there's some rather unnecessary repetition (the signature with no inputs is identical to the non-MIS one).

klecki: It was more readable for me to spell out the:
```
@overload
<signature>

@overload
<signature>
```
rather than hide the `@overload` behind yet another condition, considering the fact, that the multiline strings are not the most readable thing on earth due to the significant indentations.

mzient (approved):

PR #5191, opened 2023-11-22
Fix scalar constant dimensionality

mzient:
## Category:
**Other** (*documentation*)

## Description:
The shape of scalar constants was sometimes boosted to `(1,)`, depending on the way the constant was passed to an operator. It should always have an empty shape. The bug was caused by not passing _any_ shape to the constant operator. The operator itself cannot distinguish between a scalar value and a vector of length 1 and, in the absence of shape, always yielded a flat 1D array with the length equal to the length of the argument (`idata` / `fdata`).

Hopefully no code depends on the buggy behavior - we certainly didn't test it either way; the tests were added.

Finally, this PR also hides ops.Constant/fn.constant from the documentation. This operator should not be used directly. Most of the time it should not be used at all, an in the rare cases when it's necessary, `types.Constant` should be used.

Everywhere where we use scalar constant->DataNode promotion.

### Tests:

- [ ] Existing tests apply
- [X] New tests added
  - [X] Python tests
  - [ ] GTests
  - [ ] Benchmark
  - [ ] Other
- [ ] N/A

### Documentation
- [X] Documentation ~updated~ removed
  - [X] Docstring
  - [ ] Doxygen
  - [ ] RST
  - [ ] Jupyter
  - [ ] Other
- [ ] N/A

#### Requirements
- [X] N/A

**REQ IDs**: N/A

**JIRA TASK**: N/A

### DIFF
--- a/docs/examples/use_cases/tensorflow/yolov4/src/dali/ops.py
+++ b/docs/examples/use_cases/tensorflow/yolov4/src/dali/ops.py
@@ -1,4 +1,4 @@
-# Copyright 2021 Kacper Kluk, Piotr Kowalewski. All Rights Reserved.
+# Copyright 2021-2023 Kacper Kluk, Piotr Kowalewski. All Rights Reserved.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
@@ -61,10 +61,7 @@ def select(predicate, if_true, if_false):
     joined = dali.fn.cat(if_true, if_false)
     sh = predicate * true_shape + (1 - predicate) * false_shape
 
-    st = dali.fn.cat(
-            dali.fn.slice(true_shape * (1 - predicate), start=[0], shape=[1], axes=[0]),
-            dali.fn.constant(idata=0, shape=[1])
-        )
+    st = dali.fn.stack(true_shape[0] * (1 - predicate), 0)
 
     return dali.fn.slice(joined, start=st, shape=sh, axes=[0,1], out_of_bounds_policy="trim_to_shape")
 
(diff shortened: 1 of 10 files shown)


awolant (commented):
`fn.constant` that you want to hide from the docs is a part of one of the examples in the docs. Probably, should be removed from there as well:
https://github.com/NVIDIA/DALI/blob/0ba4fc1896a2af503174b09277d3a96d24a69c05/docs/examples/use_cases/tensorflow/yolov4/src/dali/ops.py#L66

mzient:
> `fn.constant` that you want to hide from the docs is a part of one of the examples in the docs. Probably, should be removed from there as well:
> 
> https://github.com/NVIDIA/DALI/blob/0ba4fc1896a2af503174b09277d3a96d24a69c05/docs/examples/use_cases/tensorflow/yolov4/src/dali/ops.py#L66

Definitely! Nice catch - and it really exemplifies that we should hide it, if something like that passed our review process.

docs/examples/use_cases/tensorflow/yolov4/src/dali/ops.py
```diff
@@ -61,10 +61,7 @@ def select(predicate, if_true, if_false):
     joined = dali.fn.cat(if_true, if_false)
     sh = predicate * true_shape + (1 - predicate) * false_shape
 
-    st = dali.fn.cat(
-            dali.fn.slice(true_shape * (1 - predicate), start=[0], shape=[1], axes=[0]),
-            dali.fn.constant(idata=0, shape=[1])
-        )
+    st = dali.fn.stack(true_shape[0] * (1 - predicate), 0)
```
mzient, line 64: I've tested this piece in isolation and it produces the same result as the original code while being much simpler.

awolant (approved):

stiepan (approved):

PR #5114, opened 2023-10-24
Mark more stateless operators

szkarpinski:
## Category:

**New feature** (*non-breaking change which adds functionality*)

## Description:

By default, an operator throws "Checkpointing unsupported" error on an attempt to save/restore checkpoint. Operators without state should inherit from `StatelessOperator` to indicate that checkpoint save/restore are no-ops.

In this PR I mark remaining stateless operators by making them inherit from `StatelessOperator`. For each operator a simple python test is added to verify that it's in fact stateless.

### Key points relevant for the review:
I marked image decoder base classes (`NvJPEGDecoder` and `HostDecoder`) as `Stateless`, because their random-cropping variants override `Save/RestoreState` and have checkpointing implemented.

### Tests:

- [ ] Existing tests apply
- [x] New tests added
  - [x] Python tests
  - [ ] GTests
  - [ ] Benchmark
  - [ ] Other
- [ ] N/A

### Documentation
- [x] Existing documentation applies
- [ ] Documentation updated
  - [ ] Docstring
  - [ ] Doxygen
  - [ ] RST
  - [ ] Jupyter
  - [ ] Other
- [ ] N/A

#### Requirements
- [ ] Implements new requirements
- [ ] Affects existing requirements
- [x] N/A

**REQ IDs**: N/A

**JIRA TASK**: DALI-3675

PR #5195, opened 2023-11-23
Mark all remaining stateless operators

szkarpinski:
## Category:

## Description:

Most of the stateless operators were marked some time ago in #5114 , but some remained unmarked because of one or more of those reasons: a) I missed them b) they were harder to test c) I wasn't sure what they do.

In this PR I mark remaining stateless operators. For each of them, I add a standard basic test to see if they're not stateful.

_Marking as stateless_ means making an operator inherit from `StatelessOperator` baseclass. At some point I'd like to move this information to operator schema as @mzient  suggested.

The following operators are now `StatelessOperator`:
- `AudioDecoder`
- `Inflate`
- `TensorJoin` (`fn.stack`, `fn.cat`)
- `PermuteBatch`
- `SubscriptDimCheck` and `TensorSubscript`
- `SelectMasks`
- `BoxEncoder`
- `NumbaFunction` and `PythonFunction`

This time the tests are much more complex. They aren't very critical (most of the ops are obviously stateless), but still: do the tests make sense?

### Tests:

- [ ] Existing tests apply
- [x] New tests added
  - [x] Python tests
  - [ ] GTests
  - [ ] Benchmark
  - [ ] Other
- [ ] N/A

### Documentation
- [ ] Documentation updated
  - [ ] Docstring
  - [ ] Doxygen
  - [ ] RST
  - [ ] Jupyter
  - [ ] Other
- [ ] N/A

#### Requirements
- [x] N/A

**REQ IDs**: N/A

**JIRA TASK**: N/A


awolant (approved):

mzient:
When checkpointing was first introduced, @staniewzki said that the StatelessOperator base class is an interim solution and when we have checkpointing throughout our code we could get rid of it. We can do it later, of course, but I still find it quite awkward that we have to spell out that an operator doesn't require any special checkpointg effort when that's the default behavior for perhaps 90% of all ops.

mzient (approved):

szkarpinski:
@mzient I agree that it's a bit awkward, but I find it safe - I'm not very familiar with the codebase yet and having to think about every single operator prevents me from missing some stateful ops. But you're right that this should be changed at some point

szkarpinski:
!build

szkarpinski:
!build

PR #5213, opened 2023-11-30
Support checkpointing in External Source

szkarpinski:
### Tests:

- [ ] Existing tests apply
- [ ] New tests added
  - [ ] Python tests
  - [ ] GTests
  - [ ] Benchmark
  - [ ] Other
- [ ] N/A

### Documentation
- [ ] Documentation updated
  - [ ] Docstring
  - [ ] Doxygen
  - [ ] RST
  - [ ] Jupyter
  - [ ] Other
- [ ] N/A

#### Requirements
- [ ] N/A

**REQ IDs**: N/A

**JIRA TASK**: N/A

### DIFF
--- a/dali/pipeline/operator/checkpointing/checkpoint.h
+++ b/dali/pipeline/operator/checkpointing/checkpoint.h
@@ -23,6 +23,10 @@
 namespace dali {
 
 class OpGraph;
+struct DLL_PUBLIC PipelineCheckpoint {
+  int epoch_idx;
+  int iter;
+};
 
 /**
  * @brief Aggregation of operator checkpoints for a whole pipeline.
@@ -69,6 +73,8 @@ class DLL_PUBLIC Checkpoint {
    * @brief Deserializes a protobuf message and builds this object.
   */
   DLL_PUBLIC void DeserializeFromProtobuf(const OpGraph &graph, const std::string &serialized_data);
+
+  PipelineCheckpoint pipeline_cpt_;
 
  private:
   std::vector<OpCheckpoint> cpts_;
--- a/dali/pipeline/proto/dali.proto
+++ b/dali/pipeline/proto/dali.proto
@@ -81,7 +81,13 @@ message Checkpoint {
     optional bytes operator_state = 2;
   }
 
+  message PipelineCheckpoint {
+    optional int64 epoch_idx = 1;
+    optional int64 iteration = 2;
+  }
+
   repeated OpCheckpoint cpts = 1;
+  optional PipelineCheckpoint pipeline_cpt = 2;
 }
 
 // For stateless operators
--- a/dali/python/backend_impl.cc
+++ b/dali/python/backend_impl.cc
@@ -1811,6 +1811,11 @@ PYBIND11_MODULE(backend_impl, m) {
           return node->spec.name();
         });
 
+  py::class_<PipelineCheckpoint>(m, "PipelineCheckpoint")
+    .def(py::init<>())
+    .def_readwrite("epoch_idx", &PipelineCheckpoint::epoch_idx)
+    .def_readwrite("iter", &PipelineCheckpoint::iter);
+
   // Pipeline class
   py::class_<Pipeline, PyPipeline>(m, "Pipeline")
     .def(py::init(
@@ -1897,13 +1902,13 @@ PYBIND11_MODULE(backend_impl, m) {
         },
         "checkpointing"_a = true)
     .def("SerializedCheckpoint",
-        [](Pipeline *p) -> py::bytes {
-          return p->SerializedCheckpoint();
+        [](Pipeline *p, const PipelineCheckpoint &pipeline_cpt) -> py::bytes {
+          return p->SerializedCheckpoint(pipeline_cpt);
           }, py::return_value_policy::take_ownership)
     .def("RestoreFromSerializedCheckpoint",
         [](Pipeline *p, const std::string &serialized_checkpoint) {
-          p->RestoreFromSerializedCheckpoint(serialized_checkpoint);
-        })
+          return p->RestoreFromSerializedCheckpoint(serialized_checkpoint);
+        }, py::return_value_policy::take_ownership)
     .def("executor_statistics",
         [](Pipeline *p) {
           auto ret = p->GetExecutorMeta();
--- a/dali/python/nvidia/dali/pipeline.py
+++ b/dali/python/nvidia/dali/pipeline.py
@@ -269,6 +269,8 @@ class Pipeline(object):
         self._last_iter = False
         self._iter = 0
         self._epoch_idx = 0
+        self._consumer_iter = 0
+        self._consumer_epoch_idx = 0
         self._batches_to_consume = 0
         self._cpu_batches_to_consume = 0
         self._gpu_batches_to_consume = 0
@@ -887,7 +889,17 @@ class Pipeline(object):
 
     def _restore_state_from_checkpoint(self):
         if self._checkpoint is not None:
-            self._pipe.RestoreFromSerializedCheckpoint(self._checkpoint)
+            pipeline_cpt = self._pipe.RestoreFromSerializedCheckpoint(self._checkpoint)
+            self._consumer_epoch_idx = self._epoch_idx = pipeline_cpt.epoch_idx
+            self._consumer_iter = self._iter = pipeline_cpt.iter
+            if self._input_callbacks:
+                for group in self._input_callbacks:
+                    group.current_iter = pipeline_cpt.iter
+                    group.current_sample = pipeline_cpt.iter * self._max_batch_size
+                    if hasattr(group.callback, 'restore'):
+                        idx = group.current_iter if group.batch else group.current_sample
+                        group.callback.restore(pipeline_cpt.epoch_idx, idx)
+
 
     def build(self):
         """Build the pipeline.
@@ -1049,7 +1061,10 @@ class Pipeline(object):
             A list of `TensorList` objects for respective pipeline outputs
         """
         with self._check_api_type_scope(types.PipelineAPIType.SCHEDULED):
+            self._consumer_iter += 1
             if self._batches_to_consume == 0 or self._gpu_batches_to_consume == 0:
+                self._consumer_iter = 0
+                self._consumer_epoch_idx += 1
                 raise StopIteration
             self._batches_to_consume -= 1
             self._gpu_batches_to_consume -= 1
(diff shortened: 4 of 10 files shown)


dali/python/nvidia/dali/pipeline.py
```diff
@@ -1049,7 +1057,10 @@ def outputs(self):
             A list of `TensorList` objects for respective pipeline outputs
         """
         with self._check_api_type_scope(types.PipelineAPIType.SCHEDULED):
+            self._consumer_iter += 1
```
stiepan, line 1060: Note, there's another way (😱) to consume outputs, named `share_outputs`. It's used by the iterators if you wrap the pipeline (and can be called manually too). 

On that note, I think we need tests with an iterator that wraps the pipeline.

dali/pipeline/proto/dali.proto
```diff
@@ -81,7 +81,13 @@ message Checkpoint {
     optional bytes operator_state = 2;
   }
 
+  message PipelineCheckpoint {
```
stiepan, line 84: Just having second thought about the name, maybe it should be more like ExternalContextCheckpoint? (We have pipeline in cpp backend too). I don't have strong feelings here though, what do you think @klecki?

klecki: I agree  with @stiepan. I know this is maybe a bit of nitpicking, but I feel like it's more about ExternalSource or Python state.

dali/python/nvidia/dali/plugin/pytorch/__init__.py
```diff
@@ -224,7 +224,10 @@ def __init__(
...
                 self._ever_consumed = False
             except StopIteration:
-                assert False, (
+                # This case might not be an error if we're iterating over pipeline that is
+                # currently at the end of epoch, for example because it was restored from
+                # checkpoint.
+                assert any(not p._first_iter for p in self._pipes), (
                     "It seems that there is no data in the pipeline. This may happen "
                     "if `last_batch_policy` is set to PARTIAL and the requested batch size is "
                     "greater than the shard size."
```
szkarpinski, line 233: Pytorch iterator has a check for non-empty pipeline. This may fail when a checkpoint was saved after the last iteration. Then, after restoring, we get a pipeline that will throw `StopIteration` at the very beginning, but isn't a reason to worry. Are we ok with this fix?

dali/pipeline/proto/dali.proto
szkarpinski: `ExternalContextCheckpoint` sounds good, will change

dali/python/nvidia/dali/external_source.py
```diff
@@ -360,6 +360,11 @@ class ExternalSource:
     queued. If no stream is provided, DALI will use a default, with a best-effort approach at
     correctness. See the ``cuda_stream`` argument documentation for more information.
 
+    .. note::
+        After restoring from checkpoint, queries to sources that are single-argument callables
+        (accepting :class:`nvidia.dali.types.SampleInfo`, :class:`nvidia.dali.types.BatchInfo`
+        or batch index) will be resumed from the epoch and iteration saved in the checkpoint.
```
klecki, line 366: Do we still allow to checkpoint with other sources, or should we detect them and raise an error, that they are not checkpointable, and only those single-argument callbacks are?

dali/python/nvidia/dali/plugin/pytorch/__init__.py
```diff
@@ -224,7 +224,10 @@ def __init__(
                 # here we should set if to False again
                 self._ever_consumed = False
             except StopIteration:
-                assert False, (
+                # This case might not be an error if we're iterating over pipeline that is
+                # currently at the end of epoch, for example because it was restored from
+                # checkpoint.
+                assert any(not p._first_iter for p in self._pipes), (
```
klecki, line 230: I know that this already exists here, but throwing errors via triggering asserts is not a proper error checking. We should check the condition and throw the exceptions.

One can disable asserts in their code: https://docs.python.org/3/using/cmdline.html#cmdoption-O

dali/python/nvidia/dali/plugin/pytorch/__init__.py
klecki: As for the check itself: should we make it conditional based on some variable that indicates that we are not running from scratch but from a checkpoint? 

We would need to consume the first `StopIteration` so we still have something in `_first_batch`, otherwise, won't this be broken? 

Or we just need to remember that fact, and reraise `StopIteration` from the `__next__`, so the user calls .Reset() and starts from the next batch? <- I think this may work, but is a bit on the spaghetti side. Still, we probably don't want to raise StopIteration from `__init__`.

dali/pipeline/executor/executor_test.cc
stiepan, line 1: That's too much of search and replace, right?

dali/python/nvidia/dali/external_source.py
stiepan: I'd be more for issuing a warning: we don't want to prevent people from working around that limitation with custom initialization of the iterator, e.g. at the given epoch start.

dali/pipeline/operator/checkpointing/checkpoint.h
```diff
@@ -70,6 +74,8 @@ class DLL_PUBLIC Checkpoint {
   */
   DLL_PUBLIC void DeserializeFromProtobuf(const OpGraph &graph, const std::string &serialized_data);
 
+  ExternalContextCheckpoint external_ctx_cpt_;
```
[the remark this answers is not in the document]
klecki: Yes, I mixed C++ with Python convention, sorry.

dali/pipeline/operator/checkpointing/checkpoint.h
```diff
@@ -23,6 +23,10 @@
 namespace dali {
 
 class OpGraph;
+struct DLL_PUBLIC ExternalContextCheckpoint {
+  int epoch_idx;
+  int iter;
+};
```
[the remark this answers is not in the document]
szkarpinski: Added in https://github.com/NVIDIA/DALI/pull/5213/commits/22c037be371460d2ac1a9085f0c0fb4228ed9ed4

dali/python/backend_impl.cc
```diff
@@ -1897,13 +1902,13 @@ PYBIND11_MODULE(backend_impl, m) {
...
-          return p->SerializedCheckpoint();
+        [](Pipeline *p, const ExternalContextCheckpoint &external_ctx_cpt) -> py::bytes {
+          return p->SerializedCheckpoint(external_ctx_cpt);
           }, py::return_value_policy::take_ownership)
     .def("RestoreFromSerializedCheckpoint",
         [](Pipeline *p, const std::string &serialized_checkpoint) {
-          p->RestoreFromSerializedCheckpoint(serialized_checkpoint);
-        })
+          return p->RestoreFromSerializedCheckpoint(serialized_checkpoint);
+        }, py::return_value_policy::take_ownership)
```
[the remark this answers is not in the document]
szkarpinski: Uh, `take_ownership` is wrong here. [Docs ](https://pybind11.readthedocs.io/en/stable/advanced/functions.html) say it's UB for values that were not dynamically allocated. The default should work fine.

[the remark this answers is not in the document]
szkarpinski: Fixed in https://github.com/NVIDIA/DALI/pull/5213/commits/ca89c0fe5ecc60d6c65e7d7cfcd4e417f6b95f03

dali/python/nvidia/dali/plugin/pytorch/__init__.py
szkarpinski: > We would need to consume the first StopIteration so we still have something in _first_batch, otherwise, won't this be broken?

We don't need to fill `_first_batch`. Iterator's `__next__` is opportunistic in that aspect: if there's something in `_first_batch` it's returned, but otherwise the iterator just proceeds the usual way: https://github.com/NVIDIA/DALI/blob/e496263266dce5aa1dc99d4a17208864f277968f/dali/python/nvidia/dali/plugin/pytorch/__init__.py#L238-L241

dali/python/nvidia/dali/plugin/pytorch/__init__.py
szkarpinski: Changed to conditional `RuntimeError` in https://github.com/NVIDIA/DALI/pull/5213/commits/f2c648ca31b7414efc8ca95eb71c82109c0c84d6

dali/python/nvidia/dali/plugin/pytorch/__init__.py
szkarpinski: > As for the check itself: should we make it conditional based on some variable that indicates that we are not running from scratch but from a checkpoint?

Good idea, I added `is_restored_from_checkpoint` to `Pipeline` in https://github.com/NVIDIA/DALI/pull/5213/commits/f2c648ca31b7414efc8ca95eb71c82109c0c84d6

dali/python/nvidia/dali/external_source.py
szkarpinski: Warning added in https://github.com/NVIDIA/DALI/pull/5213/commits/ebcf303d5ff72637eae714f31f8071455db79dfe

dali/python/nvidia/dali/pipeline.py
```diff
@@ -22,6 +22,7 @@
 from nvidia.dali._multiproc.pool import WorkerPool
 from nvidia.dali import pickling as dali_pickle
 from nvidia.dali import _conditionals
+from nvidia.dali._utils.external_source_impl import SourceKind
```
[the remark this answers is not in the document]
szkarpinski: Fixed in https://github.com/NVIDIA/DALI/pull/5213/commits/088f36eb32495e1e0969dac0b18b66fdbda31050

stiepan (approved):

dali/python/nvidia/dali/pipeline.py
```diff
@@ -857,6 +871,28 @@ def _setup_input_callbacks(self):
...
+                        reason = "with parameterless callable as a 'source'"
+
+                    warnings.warn(
+                        "Checkpointing enabled in a pipeline with external source operator, "
+                        f"{reason}. "
+                        "DALI doesn't capture state of such 'source'. When loading the checkpoint, "
+                        "the 'source' must be manaully adjusted by the user to start from the "
+                        "correct point, otherwise it will start from the beginning, "
+                        "potentially leading to mismatch with other data sources."
+                    )
```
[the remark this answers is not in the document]
szkarpinski: moved to `_check_checkpointing_support`

dali/python/nvidia/dali/pipeline.py
```diff
@@ -459,6 +461,16 @@ def gpu_queue_size(self):
...
 
+    @property
+    def is_restored_from_checkpoint(self):
+        """If True, this pipeline was restored from checkpoint."""
+        return self._is_restored_from_checkpoint
+
+    @property
+    def first_iter(self):
+        """True if pipeline is in the first iteration of an epoch."""
+        return self._first_iter
```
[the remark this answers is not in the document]
szkarpinski: Removed

klecki (approved):

szkarpinski:
!build

szkarpinski:
!build

PR #5216, opened 2023-12-04
Support checkpointing in other random operators

szkarpinski:
## Category:

## Description:

This PR adds checkpointing support to random operators not based on `RNGBase`. 

### Splitting `RNGBase`

Originally, `RNGBase` was a baseclass for operators with RNG, which except for its base functionality also took care of shape inference etc. There were many operators that didn't fit into `RNGBase` due to shape-related differences, but shared exactly the same approach to handling RNG. I decided to split `RNGBase` into two parts:

- `OperatorWithRng<Backend>` is an baseclass for an operator with random generator. It handles RNG seeding and checkpointing.

- `RNGBase`, inheriting from `OperatorWithRng`, adds support for shape inference. Its functionality is the same as before.

Also, as there are some operators that don't fit into `OperatorWithRng`, I extracted code for handling RNG checkpoints to `CheckpointingUtils`, defined in `rng_checkpointing_utils.h`.

#### Easy ones

The following operators were using `BatchRNG<...>` and handling it on their own. Now they inherit form `OperatorWithRng`:
- `roi_random_crop`
- `random_mask_pixel`
- `random_object_bbox`
- `ssd_random_crop`

#### Not so easy ones

In `batch_permutation`  only one RNG is needed. I added a second template argument `RngPerSample` (defaulting to `true`) to `OperatorWithRng`. If it's `false`, only one RNG (instead of batch_size RNGs) is created.

The `jitter` operator was a bit harder due to RNG being hidden inside a member (`JitterAugment`). I made the RNG public and checkpointing functions in `Jitter` op can now access it. Checkpointing is handled there manually with `CheckpointingUtils`.

In `random_bbox_crop`, RNG is inside `OpImpl`, so I added `*Impl` equivalents of checkpointing methods to `RandomBBoxCropImplBase` and used them to handle checkpointing.

- Operators: `roi_random_crop`, `random_mask_pixel`, `random_object_bbox`, `ssd_random_crop`, `batch_permutation`, `random_bbox_crop`, `jitter`.
- `RNGBase`, but it should behave exactly the same as before

- Do you see a better way to handle RNG hidden in members in `jitter` and/or `random_bbox_crop`?

PS. Due to small pieces of code being moved around, the diff of `rng_base.h` might be hard to read. You might want to [just read the file](https://github.com/szkarpinski/DALI/blob/rng-base/dali/operators/random/rng_base.h).

### Tests:

- [ ] Existing tests apply
- [x] New tests added
  - [x] Python tests
  - [ ] GTests
  - [ ] Benchmark
  - [ ] Other
- [ ] N/A

### Documentation
- [ ] Documentation updated
  - [ ] Docstring
  - [ ] Doxygen
  - [ ] RST
  - [ ] Jupyter
  - [ ] Other
- [ ] N/A

#### Requirements
- [x] N/A

**REQ IDs**: N/A

**JIRA TASK**: DALI-3717


jantonguirao (approved):

dali/operators/image/remap/displacement_filter.h
```diff
@@ -49,7 +50,7 @@ class DisplacementIdentity {
 template <typename Backend,
           class Displacement = DisplacementIdentity,
           bool per_channel_transform = false>
-class DisplacementFilter : public Operator<Backend> {};
+class DisplacementFilter : public StatelessOperator<Backend> {};
```
mzient, line 53: This code smells.... Some of the derived classes are apparently stateful (jitter?)... I don't like the idea of stating one fact in a base and changing it in a derived class. Perhaps the StatelessOperator should be defined not in terms of backend, but in terms of base class, e.g.:
```C++
template <typename Base>
class StatelessOp<Base> : public Base {
/* add trivial imlementation of checkpointing here */
};

template <typename Backend>
using StatelessOperator = StatelessOp<Operator<Backend>>;
```

With that you could do:
```C++
template <typename Backend>
class Water : public StatelessOp<DisplacementFilter<Backend>>;
```

szkarpinski: @mzient getting rid of StatelessOperator is the first thing on my refactoring todo list. As soon as the checkpointing will work as a whole (hopefully, this week), I'll try to get rid of `StatelessOperator` and either move this to schema or make it the default as you suggest.

dali/operators/image/crop/bbox_crop.h
```diff
@@ -17,24 +17,53 @@
...
+  }
+
+  std::string SerializeCheckpoint(const OpCheckpoint &cpt) const override {
+    return impl_->SerializeCheckpointImpl(cpt);
+  }
+
+  void DeserializeCheckpoint(OpCheckpoint &cpt, const std::string &data) const override {
+    impl_->DeserializeCheckpointImpl(cpt, data);
+  }
+
```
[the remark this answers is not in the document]
szkarpinski: > move BatchRNG to the operator class and pass a pointer/reference to it to RandomBBoxCropImpl at construction
Thanks for the suggestion, I did that in https://github.com/NVIDIA/DALI/pull/5216/commits/f8a144555cfda34ab45c3338df903cf5d8549cfc

dali/operators/image/crop/bbox_crop.h
```diff
@@ -19,12 +19,24 @@
...
-class RandomBBoxCrop : public Operator<Backend> {
+class RandomBBoxCropImplBase : public OpImplBase<Backend> {
+ public:
+  RandomBBoxCropImplBase(const OpSpec *spec, BatchRNG<std::mt19937_64> &rng)
+    : spec_(*spec), rngs_(rng) {}
+
+ protected:
+  const OpSpec &spec_;
+  BatchRNG<std::mt19937_64> &rngs_;
+};
```
mzient, line 36: Do we still need this intermediate class?

szkarpinski: Gone!

dali/operators/random/batch_permutation.h
```diff
@@ -18,13 +18,14 @@
...
+#include "dali/operators/random/rng_base_cpu.h"
 
 namespace dali {
 
-class BatchPermutation : public Operator<CPUBackend> {
+class BatchPermutation : public rng::OperatorWithRng<CPUBackend, false> {
  public:
   explicit BatchPermutation(const OpSpec &spec)
-  : Operator<CPUBackend>(spec), rng_(spec.GetArgument<int64_t>("seed")) {}
+  : rng::OperatorWithRng<CPUBackend, false>(spec) {}
```
mzient, line 28: Well..... I'm not a fan of this particular abuse ;) This operator only needs one RNG. Why bother creating (and serializing) all of them?

mzient (approved):

szkarpinski:
!build

szkarpinski:
!build

PR #5217, opened 2023-12-04
Unify positional input name handling in docs and signatures

klecki:
## Description:
Extract the common name processing and use the same naming format for positional-only input parameter names for both generated docstrings and the generated signatures.

The documentation generated names as `input0`, `input1`, ... (or just `input`). 
In the case of signatures, we used `__input_0` - the double underscore at the start is the convention to indicate argument as positional only. Based on the experiments, it is needed by some IDEs to not suggest setting the name as a keyword parameter.
Also, in case of named inputs, without the `__` there might be clash between them and the keywords otherwise (for example `shape` in slice). 

Formatting is unified to the latter form of `__input_0`, `__input` in case there is only one.

Adjust the way that the signatures are generated, so they list all mandatory inputs.
* signature has a cutoff point for too many inputs: https://github.com/NVIDIA/DALI/blob/main/dali/python/nvidia/dali/ops/_signatures.py#L80
* Previously, if there were any kind of inputs (mandatory or optional, above 10) we just represented all of them as `*input`.
* Documentation lists all the mandatory inputs, and adds a generic: `input[XX..YY]` for optional ones. 

Now, the signatures are generated in the form `def foo(__input_0, __input_1, ..., __input_9, /, *inputs)` - we will list all mandatory inputs, and hide the optional ones behind `VAR_POSITIONAL` only if it would be over threshold, to list all of them.
DALI doesn't have an op with more than 3 mandatory inputs. You can verify this by grepping for `NumInput(...`

This should match closer between the docs and the signatures.
The only downside is the complicated looking signature (but fully Python conformant): 
`def op(positional_only, /, *var_positional, *, keyword=default)`. 

Cleanup some todos on the way.

Docs and signatures generation

Verify if everything renders ok in sphinx and if there are no hidden cases somewhere in the source code.
Specifically check the operators, that have more than 1 input or other non-trivial cases:
* Slice and the slice decoders
* CastLike
* Cat and Stack
* transforms.combine
* numba func, python func,
* some random operators

### Tests:

- [ ] Existing tests apply
- [ ] New tests added
  - [ ] Python tests
  - [ ] GTests
  - [ ] Benchmark
  - [ ] Other
- [x] N/A

### Documentation
- [x] Documentation updated
  - [ ] Docstring
  - [ ] Doxygen
  - [ ] RST
  - [ ] Jupyter
  - [x] Other: Generated docstrings and signatures.
- [ ] N/A

#### Requirements
- [ ] N/A

**REQ IDs**: N/A

**JIRA TASK**: N/A

### DIFF
--- a/dali/python/nvidia/dali/ops/_docs.py
+++ b/dali/python/nvidia/dali/ops/_docs.py
@@ -64,13 +64,13 @@ Args
                 schema.GetSupportedLayouts(i)
             )
             dox = schema.GetInputDox(i)
-            input_name = schema.GetInputName(i)
+            input_name = _names._get_input_name(schema, i)
             ret += _numpydoc_formatter(input_name, input_type_str, dox, optional) + "\n"
     else:
         for i in range(schema.MinNumInput()):
             input_type_str = "TensorList" + _supported_layouts_str(schema.GetSupportedLayouts(i))
             dox = "Input to the operator."
-            input_name = f"input{i}" if schema.MaxNumInput() > 1 else "input"
+            input_name = _names._get_input_name(schema, i)
             ret += _numpydoc_formatter(input_name, input_type_str, dox, False) + "\n"
 
         extra_opt_args = schema.MaxNumInput() - schema.MinNumInput()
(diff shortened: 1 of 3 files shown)


klecki:
!build

jantonguirao (approved):

szkarpinski (approved):

Issue #5155, opened 2023-11-09
How to deal with corrupt files?


Pedrexus:
### Describe the question.

Hello all,

I have been using DALI for a while now, but recently I have been dealing with a problem of having corrupt files:

```python
Error when constructing operator: readers__VideoResize encountered:
[/opt/dali/dali/operators/reader/loader/video_loader.cc:324] Could not find video stream in datasets/db/Kinetics400/train/bungee jumping/oyj6TFAxpiw_000229_000239.mp4
```

I tried finding them with a loop using ffmpeg, but it seems it was not enough, so I would like to ask if there is any way to just ignore them in DALI, identify them with DALI or any other simple work around?

Thank for the hard work on the library.

### Check for duplicates

- [X] I have searched the [open bugs/issues](https://github.com/NVIDIA/DALI/issues) and have found no duplicates for this bug report

JanuszL:
Hi @Pedrexus,

Thank you for reaching out.
Is it a regression at your end? Or a convenient functionality you would like to have in DALI?
Maybe you can provide the video so we can reproduce this on our end and see if we can implement skipping unsupported videos in a reliable way?

Pedrexus:
Hello @JanuszL 

This is a problem of the datasets I am using. There are thousands of corrupted video files, and I would like to conveniently skip these by the data loader if possible.

example dataset: https://github.com/cvdfoundation/kinetics-dataset

example working video: https://github.com/NVIDIA/DALI/assets/19956280/c39e048b-f83c-454b-b2d5-35606df4c52f

example corrupted video: https://github.com/NVIDIA/DALI/assets/19956280/5ecb5019-623c-4eae-974a-98150c8d28ac

Pedrexus:
Thank you very much for the help @JanuszL. This is precisely what I needed. I think this thread could be closed then!

JanuszL:
Let us wait till it is merged and you can test it with the nightly build... just in case.

JanuszL:
It has just been merged. It should be available the next nightly we publish.

Pedrexus:
Besides that, I noticed I also get a "variable frame rate" error in the middle of the run when I lower the `stride` to 1 or 2 spite of setting `skip_vfr_check=True`. However, I don't know which video has this problem. Is there any way to identify the defect video or would this PR also help with this?

Pedrexus:
I see in the nightly version I am able to identify the vrf videos now! Thank you for this feature:

[...]

PR #5218, opened 2023-12-04
Avoids infinite video decoder seek loop

JanuszL:
- when the video decoder is not able to start decoding from a given keyframe
  it seeks a preceding one compared to it. When it happens for the keyframe
  0 we end up in an infinite try and fail loop. This PR makes this case a hard error.
- adds graceful shutdown of the decoder when reading packets or submitting
  them to the decoder fails

## Category:

## Description:
- when the video decoder is not able to start decoding from a given keyframe
  it seeks a preceding one compared to it. When it happens for the keyframe
  0 we end up in an infinite try and fail loop. This PR makes this case a hard error.
- adds graceful shutdown of the decoder when reading packets or submitting
  them to the decoder fails

- video reader

- NA

### Tests:

- [ ] Existing tests apply
- [ ] New tests added
  - [ ] Python tests
  - [ ] GTests
  - [ ] Benchmark
  - [ ] Other
- [x] N/A

### Documentation
- [ ] Documentation updated
  - [ ] Docstring
  - [ ] Doxygen
  - [ ] RST
  - [ ] Jupyter
  - [ ] Other
- [x] N/A

#### Requirements
- [x] N/A

**REQ IDs**: N/A

**JIRA TASK**: N/A

### DIFF
--- a/dali/operators/reader/loader/video_loader.cc
+++ b/dali/operators/reader/loader/video_loader.cc
@@ -591,6 +591,9 @@ void VideoLoader::read_file() {
     if (last_key_frame != -1 &&
         ((key && last_key_frame != frame) || frame > last_key_frame + kStartupFrameTreshold) &&
         dec_status == VidReqStatus::REQ_NOT_STARTED) {
+      DALI_ENFORCE(last_key_frame > 0, "Decoding not started when starting from frame 0, "
+                                       "no point in seeking to preceding keyframe which is "
+                                       "already 0");
       LOG_LINE << "Decoding not started, seek to preceding key frame, "
                 << "current frame " << frame
                 << ", last key frame " << last_key_frame
--- a/dali/operators/reader/loader/video_loader.cc
+++ b/dali/operators/reader/loader/video_loader.cc
@@ -539,7 +539,12 @@ void VideoLoader::read_file() {
 
 
   auto& file = get_or_open_file(req.filename);
-  DALI_ENFORCE(!file.empty(), "Cannot open video file");
+  if (file.empty()) {
+    if (vid_decoder_) {
+      vid_decoder_->finish();
+    }
+    DALI_FAIL("Cannot open video file");
+  }
   auto stream = file.fmt_ctx_->streams[file.vid_stream_idx_];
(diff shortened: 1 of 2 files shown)


JanuszL:
!build

awolant (approved):

JanuszL:
!build

dali/operators/reader/loader/video_loader.cc
```diff
@@ -586,11 +591,21 @@ void VideoLoader::read_file() {
...
+        ((key && last_key_frame != frame && last_key_frame != 0) ||
+          frame > last_key_frame + kStartupFrameThreshold) &&
         dec_status == VidReqStatus::REQ_NOT_STARTED) {
+        if (last_key_frame <= 0) {
+          if (vid_decoder_) {
+            vid_decoder_->finish();
+          }
+          DALI_FAIL("Decoding not started when starting from frame 0, "
+                    "no point in seeking to preceding keyframe which is "
+                    "already 0");
```
szkarpinski, line 606: Is there particular kind of video file or parameter set that triggers this problem? Maybe it's worth mentioning in the message? The current one doesn't seem very helpful for the user as it only refers to the internals of the decoder

JanuszL: The faulty sample was provided by the user in https://github.com/NVIDIA/DALI/issues/5155#issuecomment-1837082791, but I don't know how to create such one on our own and I don't think we can add this one to DALI_extra.

dali/operators/reader/loader/video_loader.cc
```diff
@@ -539,7 +539,12 @@ void VideoLoader::read_file() {
 
 
   auto& file = get_or_open_file(req.filename);
-  DALI_ENFORCE(!file.empty(), "Cannot open video file");
+  if (file.empty()) {
+    if (vid_decoder_) {
+      vid_decoder_->finish();
```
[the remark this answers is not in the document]
szkarpinski: I see, thank you

szkarpinski (approved):

JanuszL:
!build

PR #5230, opened 2023-12-08
Add improved checkpointing docs

szkarpinski:
## Category:

**Other** – docs

## Description:

This PR adds "checkpointing" page to DALI docs.

My first time writing RST page to DALI docs, something might be wrong in a way you don't expect it to be :)

### Tests:

- [ ] Existing tests apply
- [ ] New tests added
  - [ ] Python tests
  - [ ] GTests
  - [ ] Benchmark
  - [ ] Other
- [x] N/A

### Documentation
- [ ] Documentation updated
  - [ ] Docstring
  - [ ] Doxygen
  - [x] RST
  - [ ] Jupyter
  - [ ] Other
- [ ] N/A

#### Requirements
- [ ] N/A

**REQ IDs**: N/A

**JIRA TASK**: DALI-3739

### DIFF
--- a/docs/advanced_topics_checkpointing.rst
+++ b/docs/advanced_topics_checkpointing.rst
@@ -1,0 +1,89 @@
+Checkpointing
+=============
+
+.. currentmodule:: nvidia.dali
+
+Checkpointing is a feature in DALI which allows you to save current state of the pipeline to a file.
+Then, you can restore the pipeline from a saved checkpoint and the new pipeline will produce exactly the same outputs as the old one would.
+It is particularly useful for long-running training jobs which are likely to be interrupted.
+
+A checkpoint of DALI pipeline contains information about states of all random number generators used in the pipeline and about the progress of each reader.
+
+Checkpointing API
+-----------------
+
+Enabling checkpointing
+~~~~~~~~~~~~~~~~~~~~~~
+
+To enable checkpointing, set ``enable_checkpointing=True`` when creating a pipeline.
+With this option enabled, DALI will track state of each operator, allowing you to save it on demand.
+Enabling checkpointing shouldn't have any impact on the performance.
+
+.. code-block:: python
+
+  @pipeline_def(..., enable_checkpointing=True)
+  def pipeline():
+      ...
+
+  p = pipeline()
+  p.build()
+
+
+.. note::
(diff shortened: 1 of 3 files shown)


szkarpinski:
!build

docs/advanced_topics_checkpointing.rst
```diff
@@ -0,0 +1,89 @@
...
+
+  # Write the checkpoint to file:
+  checkpoint = p.checkpoint()
+  open('checkpoint_file.cpt', 'wb')
+
+  # Or simply:
+  checkpoint = p.checkpoint('checkpoint_file.cpt')
+
+.. note::
+    Calling :meth:`Pipeline.checkpoint method` may introduce an observable overhead.
```
awolant, line 55: image

I think this does not look like intended.

szkarpinski: Fixed, thank you

docs/advanced_topics_checkpointing.rst
```diff
@@ -0,0 +1,89 @@
...
+  @pipeline_def(..., enable_checkpointing=True)
+  def pipeline():
+      ...
+
+  p = pipeline()
+  p.build()
+
+
+.. note::
+    Readers with ``shuffle_after_epoch=True`` might produce different (but still equally random) output if checkpointing is enabled.
```
stiepan, line 33: Just a suggestion, but I'd make it a bit more concrete (i.e. shuffle, not produce outputs). At the same time, I would avoid going into the details on randomness quality. The fact that we had such conernes seems to me like a detail of design/implementation steps, but, as lon as we do not degrade it, I would not bother user with that at all.

```suggestion
    Readers with ``shuffle_after_epoch=True`` might shuffle samples differently if checkpointing is enabled.
```

stiepan (approved):

docs/advanced_topics_checkpointing.rst
szkarpinski: I wanted to make sure that the user won't understand this as "shuffling is not random if checkpointing is enabled"

szkarpinski: But I guess "shuffle samples differently" implies that it's still random

szkarpinski:
!build

PR #5236, opened 2023-12-11
Add linter GitHub Action

klecki:
## Description:
GH actions are automatically run on PR changes, giving immediate linter verification that is integrated
with GH PR status.

Use psf/black action with DALI-compatible configuration. Add custom action that installs and runs flake8.

There is dependency added between those actions, so the flake8 linter is only checked when the black formatting is correct.

There is a bit of repetition with this file and `lint.cmake`. I am not attempting on writing a meta-meta
configuration file to handle both. We adjust linting scripts about once a year.

Here is an example of this script where linter error is triggered:
obraz
https://github.com/klecki/DALI/pull/31

CI/GH.

### Tests:

- [ ] Existing tests apply
- [ ] New tests added
  - [ ] Python tests
  - [ ] GTests
  - [ ] Benchmark
  - [ ] Other
- [x] N/A

### Documentation
- [ ] Documentation updated
  - [ ] Docstring
  - [ ] Doxygen
  - [ ] RST
  - [ ] Jupyter
  - [ ] Other
- [x] N/A

#### Requirements
- [x] N/A

**REQ IDs**: N/A

**JIRA TASK**: N/A

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -32,3 +32,11 @@ jobs:
       - run: flake8 --config=.flake8 ${{ env.PYTHON_LINT_PATHS }}
       - run: flake8 --config=.flake8.ag ${{ env.AUTOGRAPH_LINT_PATHS }}
     needs: black
+  cpp:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-python@v4
+        with:
+          python-version: '3.10'
+      - run: python ./tools/lint.py . --nproc=5
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -9,7 +9,6 @@ env:
   PYTHON_LINT_PATHS: "./dali ./docs ./tools ./dali_tf_plugin ./qa"
   AUTOGRAPH_LINT_PATHS: "./dali/python/nvidia/dali/_autograph ./dali/test/python/autograph/"
 
-
 jobs:
   python:
     runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


JanuszL (approved):
Can we improve the efficiency by combining both of these steps into one action?
Can we add our c++ lint too?

klecki:
> Can we improve the efficiency by combining both of these steps into one action?

I like the look of two separate statuses. The jobs are rather quick:
obraz
If you think this is too slow, or we realize that during the fact, than we can adjust.

> Can we add our c++ lint too?

Should be doable. I thought we need a list of files based on CMake, but it just runs recursively on the directory.

klecki:
@JanuszL  I can remove the dependency between black and flake, that would allow them to run in parallel, but I don't think that ~30-40s matters that much, and we save some GH resources that way.

You can see flake and cpp linter being triggered here: https://github.com/klecki/DALI/pull/33

klecki:
> FTR, I've opened a similar PR, which uses pre-commit instead of github actions: https://github.com/NVIDIA/DALI/pull/5190

I am aware about it, I just wanted to try out GitHub actions. Doesn't pre-commit integration also require them to run? So this is just a direct way of running those linters inside GH actions. 
I think similar code would need to be added to the pre-commit to handle all the details that we use (like split between flake configs).

> Pre-commit allows for broader spectrum of code-checking and features, we may consider this approach

This allows arbitrary Python commands to be run.

awolant (approved):

klecki:
I removed the psf/black action and replaced it with pure `run` command after installing black manually. We can revisit when it is allowed within the organization.

PR #5244, opened 2023-12-13
Remove stages from Pipeline API

mzient:
## Category:

### Tests:

- [ ] Existing tests apply
- [ ] New tests added
  - [ ] Python tests
  - [ ] GTests
  - [ ] Benchmark
  - [ ] Other
- [ ] N/A

### Documentation
- [ ] Documentation updated
  - [ ] Docstring
  - [ ] Doxygen
  - [ ] RST
  - [ ] Jupyter
  - [ ] Other
- [ ] N/A

#### Requirements
- [ ] N/A

**REQ IDs**: N/A

**JIRA TASK**: N/A


dali/pipeline/pipeline.h
```diff
@@ -756,8 +771,18 @@ class DLL_PUBLIC Pipeline {
       return true;
     }
 
+    /**
+     * @brief Feeds the recently set inputs to the inputs that have `repeat_last` property
+     *
+     * @param owner       The pipeline
+     * @param fill_queue  If true, the inputs are fed `InputFeedCount(name)` times;
+     *                    otherwise it's fed once.
```
szalpal, line 779: Should it be plural?
```suggestion
     * @param fill_queue  If true, the inputs are fed `InputFeedCount(name)` times;
     *                    otherwise they're fed once.
```

mzient: Yeah, I guess...

dali/pipeline/executor/executor.h
```diff
@@ -164,7 +163,12 @@ class DLL_PUBLIC Executor : public ExecutorBase, public QueuePolicy {
   */
   DLL_PUBLIC void RestoreStateFromCheckpoint(const Checkpoint &cpt) override;
 
+  DLL_PUBLIC int InputFeedCount(const std::string &op_name) override;
+
  protected:
+  DLL_PUBLIC virtual void RunCPU();
+  DLL_PUBLIC virtual void RunMixed();
+  DLL_PUBLIC virtual void RunGPU();
```
mzient, line 171: ...and re-emerge as an implementation detail.

awolant (approved):

dali/python/nvidia/dali/pipeline.py
```diff
@@ -1255,12 +1243,61 @@ def _prefetch(self):
...
         self._schedule_py_workers()
-        if self._exec_separated:
-            self._fill_separated_queues()
-        else:
-            for _ in range(self._prefetch_queue_depth):
-                self._run_once()
+
+        # We probably need some benchmarking before we remove this code path
+        if not self._exec_separated:
+            self._legacy_interleaved_prefetch()
```
stiepan, line 1249: I have a concern/question here. It mainly impacts PES (but it may be relevant for memory usage in regular ES too).

First, do I read it right? For the most typical case (async pipelined), with the new way, we first produce `InputFeedCount = 2 * prefetch_queue_depth (cpu + gpu queues)` batches, fed them into external source (that is happy to accept batches irrespecitve of any prefetch_queues) and only then poke the executor to start?

If so, we increase the number of input batches to keep in external source twice, right? 

In the PES, we need to adjust for that explicitly, as the batches reside in a circular buffer of shared memory and are fed with no_copy into the backend external source op. Implementation wise it's easy, just pass the InputFeedCount to WorkerPool.from_groups call (instead of prefetch_queue_depth). 

The question is, if we want that: it seems to increase required memory and potentially slow down the start. What are the benefits?

stiepan (commented):
I have yet to read a few files, but I am leaving a few comment that crossed my mind so far.

dali/pipeline/operator/builtin/external_source_test.cc
```diff
@@ -1,4 +1,4 @@
-// Copyright (c) 2019-2022, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+// Copyright (c) 2019-2023, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
stiepan, line 1: It's so 2023. :)

dali/python/nvidia/dali/pipeline.py
```diff
@@ -938,7 +939,12 @@ def build(self):
         self._restore_state_from_checkpoint()
         self._built = True
 
-    def _feed_input(self, name, data, layout=None, cuda_stream=None, use_copy_kernel=False):
+    def input_feed_count(self, input_name):
+        return self._pipe.InputFeedCount(input_name)
+
+    def _feed_input(
+        self, name, data, layout=None, cuda_stream=None, use_copy_kernel=False, is_prefetch=False
```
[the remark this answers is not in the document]
mzient: Apparently not - looks like a leftover.

dali/c_api/c_api.cc
```diff
@@ -283,26 +284,36 @@ int daliGetMaxBatchSize(daliPipelineHandle_t pipe_handle) {
...
   auto &pipeline = (*pipe_handle)->pipeline;
-  for (int i = 0; i < gpu_queue_depth; ++i) {
-    pipeline->RunCPU();
-    pipeline->RunGPU();
-  }
-  for (int i = 0; i < cpu_queue_depth; ++i) {
-    pipeline->RunCPU();
+  auto sz = pipeline->GetQueueSizes();
+  if (cpu_queue_depth != sz.cpu_size || gpu_queue_depth != sz.gpu_size) {
+    DALI_WARN("daliPrefetchUniform is deprecated and setting queue_length different than"
```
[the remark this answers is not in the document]
mzient: Yes, it's normally only called once per the lifetime of the pipeline. WARN_ONCE has a downside that it would be only once per the lifetime of the process.

dali/python/nvidia/dali/pipeline.py
stiepan: Maybe it wasn't clear, but, sure, I get we go with the legacy approach for now. The concern was more for the future consideration. Do you plan on getting rid of the legacy mode?

mzient: Plan? No. Consider? Yes.

stiepan (approved):

PR #5245, opened 2023-12-14
Suppress leaks from inside dlopen

szkarpinski:
## Category:

## Description:

We dlopen things but not dlclose them, so some internal dlopen structures get leaked. This was suppressed in https://github.com/NVIDIA/DALI/pull/4521, but some of the functions listed there got inlined, resulting in new false positives.

In this PR I add `dlopen` to the suppressions list to avoid further false-positives

Sanitizer CI runs

Maybe suppressing `dlopen` is too wide?

### Tests:

- [x] Existing tests apply
- [ ] New tests added
  - [ ] Python tests
  - [ ] GTests
  - [ ] Benchmark
  - [ ] Other
- [ ] N/A

### Documentation
- [ ] Documentation updated
  - [ ] Docstring
  - [ ] Doxygen
  - [ ] RST
  - [ ] Jupyter
  - [ ] Other
- [ ] N/A

#### Requirements
- [x] N/A

**REQ IDs**: N/A

**JIRA TASK**: N/A

### DIFF
--- a/qa/leak.sup
+++ b/qa/leak.sup
@@ -24,6 +24,7 @@ leak:_Unwind_ForcedUnwind
 leak:xla
 # as DALI dlopens but not dlcloses libraries some resources are marked as leaked
 # suppress them
+leak:dlopen
 leak:cuFileDriverOpen
 leak:cudaGetDeviceCount
 leak:cuvidGetDecoderCaps


awolant (approved):

szkarpinski:
@mzient pointed out that `dlopen` calls globals constructors, so if there's any leak in there it'll get suppressed. So I guess suppressing `dlopen` was a bad idea.

szkarpinski:
@awolant @banasraf I changed (https://github.com/NVIDIA/DALI/pull/5245/commits/62256ef38bb92d9355be714141567abadcade814) the suppressions to be more precise, `dlopen` was too broad and might suppress true leaks. I'll run sanitizer build in a moment to see if this still helps.

JanuszL (approved):

szkarpinski:
After a long rollercoaster of trying different ways to suppress things I came back to the original solution of just suppressing `dlopen` itself and the CI is finally passing :tada: . @JanuszL @awolant is this OK?

JanuszL:
I think we can proceed that way.

PR #5263, opened 2024-01-03
Expose do_not_convert decorator

klecki:
### Tests:

- [ ] Existing tests apply
- [ ] New tests added
  - [ ] Python tests
  - [ ] GTests
  - [ ] Benchmark
  - [ ] Other
- [ ] N/A

### Documentation
- [ ] Documentation updated
  - [ ] Docstring
  - [ ] Doxygen
  - [ ] RST
  - [ ] Jupyter
  - [ ] Other
- [ ] N/A

#### Requirements
- [ ] N/A

**REQ IDs**: N/A

**JIRA TASK**: N/A


dali/python/nvidia/dali/external_source.py
```diff
@@ -783,6 +784,23 @@ def __call__(
...
+                    "The `source` parameter that was passed to external source was created "
+                    "in a scope that was converted with AutoGraph. To allow the `source` to be "
+                    "correctly used with parallel external source, it must remain unconverted. "
+                    "To prevent conversion, two steps may need to be taken:\n"
+                    "1. The `source` or a factory creating the parameter passed to `source` must "
+                    "be defined at global scope (or at least outside of `pipeline_def` scope).\n"
+                    "2. If the `source` is created by some factory function that is invoked within "
+                    "pipeline definition, that function must be decorated with "
+                    "`@nvidia.dali.pipeline.do_not_convert`, otherwise it will be recursively "
+                    "converted when the pipeline definition is traced.\n"
```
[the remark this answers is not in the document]
klecki: done

dali/python/nvidia/dali/external_source.py
```diff
@@ -783,6 +784,23 @@ def __call__(
...
                     "your source returns batches)".format(what)
                 )
+            if _conditionals._is_autograph_artifact(source_desc.source):
+                raise ValueError(
+                    "The `source` parameter that was passed to external source was created "
+                    "in a scope that was converted with AutoGraph. To allow the `source` to be "
+                    "correctly used with parallel external source, it must remain unconverted. "
+                    "To prevent conversion, two steps may need to be taken:\n"
+                    "1. The `source` or a factory creating the parameter passed to `source` must "
+                    "be defined at global scope (or at least outside of `pipeline_def` scope).\n"
```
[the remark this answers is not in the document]
klecki: done

dali/python/nvidia/dali/pipeline.py
```diff
@@ -1884,6 +1885,98 @@ def create_pipeline(*args, **kwargs):
...
+
+    In conditional mode, DALI uses a fork of
+    `TensorFlow's AutoGraph <https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/autograph/g3doc/reference/index.md>`_
+    to transform the code, enabling us to rewrite and detect the ``if`` statements, so they can be
+    used in processing the DALI pipeline.
+
+    When used with :meth:`external source <nvidia.dali.fn.external_source>` in parallel mode
+    (``parallel=True``), this may interfere with the serialization of the provided ``source``
+    parameter. To prevent this, functions that are used to create the ``source`` parameter,
+    should be decorated with :meth:`@do_not_convert <nvidia.dali.pipeline.do_not_convert>`.
```
mzient, line 1903: All that is true - but on the other hand, this decorator is in `pipeline` module, not `external_source`, so the focus on a particular flavor of external source seems exaggerated. I know that this was a precipitating factor for the whole effort, but it has wider implications and this documentation seems a bit awkward and misplaced when read without context. It also leaves out other potentially problematic areas, like PythonFunction (perf. hit) or numba (do we know if this works?).
```suggestion
    In conditional mode, DALI uses a fork of
    `TensorFlow's AutoGraph <https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/autograph/g3doc/reference/index.md>`_
    to transform the code, enabling us to rewrite and detect the ``if`` statements, so they can be
    used in processing the DALI pipeline. The rewriting applies to all functions defined within or called
    from a pipeline definition function. This, among other effects, makes these functions non-serializable.

    Parallel :meth:`external source <nvidia.dali.fn.external_source>` (``parallel=True``), requires that their
    ``source`` argument is serializable. To prevent rewriting of the ``source``, the functions used to create it
    should be decorated with :meth:`@do_not_convert <nvidia.dali.pipeline.do_not_convert>`.
```

dali/python/nvidia/dali/pipeline.py
```diff
@@ -1884,6 +1885,98 @@ def create_pipeline(*args, **kwargs):
...
+
+        from nvidia.dali import pipeline_def, fn
+
+        @pipeline_def(enable_conditionals=True)
+        def pipe():
+
+            def source_factory(size):
+                def source_fun(sample_info):
+                    return np.full(size, sample_info.iter_idx)
+                return source_fun
```
mzient, line 1927: Can't we just do:
```suggestion
            @do_not_convert
            def source_factory(size):
                def source_fun(sample_info):
                    return np.full(size, sample_info.iter_idx)
                return source_fun
```
?
This would be much less invasive.

klecki: > When a function is converted, all functions defined within its syntactical scope are also converted.

Adding a decorator doesn't stop this process, we must prevent a conversion when a function from external scope is called, as those from this scope, were already converted.

mzient: ...and can't we tweak autograph so it doesn't touch those?

dali/python/nvidia/dali/pipeline.py
klecki: done

klecki:
!build

klecki:
!build

klecki:
!build

stiepan (approved):

awolant (approved):

klecki:
!build

PR #5266, opened 2024-01-04
Add framework attributes to DLFW iterator tests.

mzient:
## Category:

## Description:
Prior to this change it was impossible to run the DLFW iterator tests only for selected frameworks. This PR makes it possible. There is also some refactoring for nose2, which doesn't requires attributes on test cases produced by test generators (`yield func, *params`).

### Tests:

- [ ] Existing tests apply
- [ ] New tests added
  - [ ] Python tests
  - [ ] GTests
  - [ ] Benchmark
  - [ ] Other
- [X] N/A

### Documentation
- [ ] Documentation updated
  - [ ] Docstring
  - [ ] Doxygen
  - [ ] RST
  - [ ] Jupyter
  - [ ] Other
- [X] N/A

#### Requirements
- [X] N/A

**REQ IDs**: N/A

**JIRA TASK**: N/A

### DIFF
--- a/dali/test/python/test_fw_iterators.py
+++ b/dali/test/python/test_fw_iterators.py
@@ -902,7 +902,7 @@ def test_gluon_iterator_pass_reader_name_autoreset():
         )
 
 
-@attr("torch")
+@attr("pytorch")
 def test_pytorch_iterator_last_batch_no_pad_last_batch():
     from nvidia.dali.plugin.pytorch import DALIGenericIterator as PyTorchIterator
 
@@ -941,7 +941,7 @@ def test_pytorch_iterator_last_batch_no_pad_last_batch():
     assert len(set(mirrored_data)) != 1
 
 
-@attr("torch")
+@attr("pytorch")
 def test_pytorch_iterator_last_batch_pad_last_batch():
     from nvidia.dali.plugin.pytorch import DALIGenericIterator as PyTorchIterator
 
@@ -989,7 +989,7 @@ def test_pytorch_iterator_last_batch_pad_last_batch():
     assert len(set(next_mirrored_data)) == 1
 
 
-@attr("torch")
+@attr("pytorch")
 def test_pytorch_iterator_not_fill_last_batch_pad_last_batch():
     from nvidia.dali.plugin.pytorch import DALIGenericIterator as PyTorchIterator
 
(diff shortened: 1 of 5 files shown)


dali/test/python/test_fw_iterators.py
```diff
@@ -441,6 +448,7 @@ def check_iterator_results(
     return (ids, sample_counter, per_gpu_counter, epoch_counter, rounded_shard_size)
 
 
+@attr("mxnet")
 def check_mxnet_iterator_pass_reader_name(
```
mzient, line 452: `yield check_mxnet_iterator_pass_reader_name, ...` caused nose2 to lose the attribute - it has to be added to this function, too - unlike for nosetests, which require it on the generator.

jantonguirao (approved):

qa/TL0_FW_iterators/test_mxnet.sh
```diff
@@ -21,8 +21,7 @@ test_body() {
         done
     fi
     ${python_invoke_test} -m '(?:^|[\b_\./-])[Tt]est.*mxnet*' test_fw_iterators_detection.py
-    ${python_invoke_test} -m '(?:^|[\b_\./-])[Tt]est.*mxnet*' test_fw_iterators.py
-    ${python_invoke_test} -m '(?:^|[\b_\./-])[Tt]est.*gluon*' test_fw_iterators.py
+    ${python_new_invoke_test} -A 'mxnet,gluon' test_fw_iterators
```
stiepan, line 24: I looked into the L0_TL0_FW_iterators--fw-mxnet job and it seems no tests were run for mxnet. Can you double check if it picks them correctly?

JanuszL: There should be a separate job for that L0_TL0_FW_iterators--fw-mxnet

mzient: Nice catch! It seems that comma-separated attributes don't work in nose2 (again, nose2 sucks big time).

JanuszL: How about using expressions - https://docs.nose2.io/en/latest/plugins/attrib.html#select-tests-using-python-expressions ?

JanuszL: According to https://github.com/nose-devs/nose2/blob/main/nose2/plugins/attrib.py if you pass a list of tuple of attributes it should work as well.

mzient: I could, but is it really worth the complication? Using attributes (like in the updated code) works quite fine. My goal here was to simplify manual launches and especially filtering out unwanted frameworks. As I understand, it works like this:
```
-A 'a1,!a2' -A 'a3'
```
translates to `(a1 and not a2 ) or a3` - it makes perfect sense.

PR #5268, opened 2024-01-05
Allow source to be a decorated function or method

klecki:
## Description:
This changes how the external source counts the number of arguments that the source callback uses.
Instead of relying on the `__code__.co_argcount` we use `inspect.signature`.
The reason is that a decorator can (and usually does) propagate the signature, but due to the use of
generic `*args, **kwargs` parameters it will hide the original argcount of the original function that is wrapped.

Such simple decorator is added as a test case, before this change the following code won't work.

This fixes the potential error of DALI assuming that there are mandatory arguments to the function,
that doesn't expect them (or vice-versa) and invoking the callback with (or without) passing the source_info. 
The error would point to the callback invocation and would not indicate the decorators at all.

Now we can sensibly error out on:
* var-args
* var-kwargs
* kwarg-only argument

There is no need to differentiate between positional only and named argument, if we first detect
any disallowed constructs. Note, that sources like `foo(arg, *args)` were already disallowed, but the reported arg count was something like `-1` or `-2` by mistake rather than by design.

Next we can validate the correct number of arguments.

The detection of callbacks was done inside an exception handler, and throwing another exception
from there would chain the exception, so the construction of argument description is moved to line below.

### Tests:

- [ ] Existing tests apply
- [x] New tests added
  - [x] Python tests
  - [ ] GTests
  - [ ] Benchmark
  - [ ] Other
- [ ] N/A

### Documentation
- [ ] Documentation updated
  - [ ] Docstring
  - [ ] Doxygen
  - [ ] RST
  - [ ] Jupyter
  - [ ] Other
- [ ] N/A

#### Requirements
- [x] N/A

**REQ IDs**: N/A

**JIRA TASK**: N/A

### DIFF
--- a/dali/python/nvidia/dali/_utils/external_source_impl.py
+++ b/dali/python/nvidia/dali/_utils/external_source_impl.py
@@ -238,6 +238,14 @@ def _cycle_enabled(cycle):
 
 
 def accepted_arg_count(callable):
+    """Checks the number of accepted arguments by the callable, and validates if it is either
+    0 or 1 positional argument allowed by the external source.
+
+    Raises
+    ------
+    TypeError
+        Indicates that the `source` callable accepts wrong number of type of arguments.
+    """
     if not (inspect.isfunction(callable) or inspect.ismethod(callable)) and hasattr(
         callable, "__call__"
     ):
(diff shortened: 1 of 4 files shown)


mzient (changes requested):
We can fix (and simplify!) autograph do_not_convert instead of building workarounds in DALI, which is just kicking the can down the road.
See #5270

mzient (commented):
We can do it for the sake of different decorators than do_not_convert.

mzient (approved):

klecki:
!build

klecki:
!build

klecki:
!build

dali/python/nvidia/dali/_utils/external_source_impl.py
```diff
@@ -247,7 +255,31 @@ def accepted_arg_count(callable):
...
+            )
+        if p.kind == inspect.Parameter.VAR_KEYWORD:
+            raise TypeError(
+                error_msg + f" Found var-keyword argument `**{p.name}` which is not allowed."
+            )
+        if p.kind == inspect.Parameter.KEYWORD_ONLY:
+            raise TypeError(
+                error_msg + f" Found keyword-only argument `{p.name}` which is not allowed."
+            )
+    result = len(inspect.signature(callable).parameters) - implicit_args
```
stiepan, line 277: I thought `parameters` skips the implicit param already.

stiepan: Btw. the `inspect.signature(callable)` is already assigned to a signature var, right?

klecki:
```
    class MethodOne:
        def __call__(self, a):
            pass

    import inspect

    print(inspect.signature(MethodOne()))
    print(inspect.signature(MethodOne.__call__))
```
prints
```
(a)
(self, a)
```

We extract `__call__`, so it doesn't hide the self anymore.

stiepan (approved):

banasraf (approved):

PR #5273, opened 2024-01-09
Fix synchronization issues in RNG checkpointing utils

szkarpinski:
## Category:

## Description:
This PR fixes two synchronization bugs pointed out by @mzient . Both problems happen when you're restoring from checkpoint in random GPU operators and are due to lack of synchronization after H2D & D2D copies.

PS. Another problem pointed out was that synchronization between `SaveState` and `Serialize` is handled by the pipeline and not by those functions. This is a design problem and I added a TODO to fix it. I'm currently reworking how synchronization works when a checkpoint is saved, so I'll fix it there.

RngCheckpointingUtils

### Tests:

- [x] Existing tests apply
- [ ] New tests added
  - [ ] Python tests
  - [ ] GTests
  - [ ] Benchmark
  - [ ] Other
- [ ] N/A

### Documentation
- [ ] Documentation updated
  - [ ] Docstring
  - [ ] Doxygen
  - [ ] RST
  - [ ] Jupyter
  - [ ] Other
- [ ] N/A

#### Requirements
- [x] N/A

**REQ IDs**: N/A

**JIRA TASK**: DALI-3777

### DIFF
--- a/dali/operators/util/randomizer.cuh
+++ b/dali/operators/util/randomizer.cuh
@@ -59,8 +59,9 @@ struct curand_states {
   }
 
   DALI_HOST inline void set(const curand_states &other) {
-    CUDA_CALL(cudaMemcpy(states_, other.states_, sizeof(curandState) * len_,
-                         cudaMemcpyDeviceToDevice));
+    CUDA_CALL(cudaMemcpyAsync(states_, other.states_, sizeof(curandState) * len_,
+                              cudaMemcpyDeviceToDevice, cudaStreamDefault));
+    CUDA_CALL(cudaStreamSynchronize(cudaStreamDefault));
   }
 
  private:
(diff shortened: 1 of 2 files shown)


dali/operators/util/randomizer.cuh
```diff
@@ -59,8 +59,9 @@ struct curand_states {
   }
 
   DALI_HOST inline void set(const curand_states &other) {
-    CUDA_CALL(cudaMemcpy(states_, other.states_, sizeof(curandState) * len_,
-                         cudaMemcpyDeviceToDevice));
+    CUDA_CALL(cudaMemcpyAsync(states_, other.states_, sizeof(curandState) * len_,
+                              cudaMemcpyDeviceToDevice, cudaStreamDefault));
+    CUDA_CALL(cudaStreamSynchronize(cudaStreamDefault));
```
mzient, line 64: It will do for a quick bug fix, but it would probably be better to use a proper stream.

banasraf (approved):

szkarpinski:
!build

PR #2175, opened 2020-08-03
Adjust documentation after review

JanuszL:
Signed-off-by: Janusz Lisiecki <jlisiecki@nvidia.com>

#### Why we need this PR?
*Pick one, remove the rest*
- It fixes a numerous language and formatting defects in the documentation

#### What happened in this PR?
*Fill relevant points, put NA otherwise. Replace anything inside []*
 - What solution was applied:
     fixes a numerous language and formatting defects in the documentation
 - Affected modules and functionalities:
     whole documentation
 - Key points relevant for the review:
     NA
 - Validation and testing:
     CI build and documentation review of the output
 - Documentation (including examples):
     Documentation has been updated

**JIRA TASK**: *[DALI-1552]*

PR #5285, opened 2024-01-19
Fix capitalized property descriptions

szkarpinski:
## Category:

## Description:

This PR fixes capitalized property descriptions, which since 1.25 look like this:
image
And are sometimes aligned next to each other like this:
image

This is because the _Property_ text itself is (and always has been) of class `.property` and we added some custom styling to it in #2175 (in particular `capitalize` transform to make it _Property_ not _property_ and `display:inline-box`).

It was all fine util the property entries (whole entries, including doc text) started to be rendered as `<... class="py property">` which conflicted with already used class `.property`. This way our styling was applied to whole text.

In this PR I change the selectors to still apply to `property`, but not `py property`.

### Tests:

- [ ] Existing tests apply
- [ ] New tests added
  - [ ] Python tests
  - [ ] GTests
  - [ ] Benchmark
  - [ ] Other
- [x] N/A

### Documentation
- [x] Documentation updated
  - [ ] Docstring
  - [ ] Doxygen
  - [ ] RST
  - [ ] Jupyter
  - [x] Other - **template**
- [ ] N/A

#### Requirements
- [x] N/A

**REQ IDs**: N/A

**JIRA TASK**: DALI-3780

### DIFF
--- a/docs/_templates/layout.html
+++ b/docs/_templates/layout.html
@@ -80,7 +80,7 @@
     border-top: solid 3px rgba(59,93,0,1);
   }
 
-  html.writer-html4 .rst-content dl:not(.docutils) .property, html.writer-html5 .rst-content dl[class]:not(.option-list):not(.field-list):not(.footnote):not(.glossary):not(.simple) .property {
+  html.writer-html4 .rst-content dl:not(.docutils):not(.py) .property, html.writer-html5 .rst-content dl[class]:not(.py):not(.option-list):not(.field-list):not(.footnote):not(.glossary):not(.simple) .property {
     text-transform: capitalize;
     display: inline-block;
     padding-right: 8px;


klecki (approved):

JanuszL (approved):

szkarpinski:
!build

szkarpinski:
The previous fix didn't work - the tag with "property" text inherited `py` class from parent node, so it was also `py property`

Now I fix it by using a simpler selector:
```
.rst-content dl.py [class=property]
```
which selects elements *inside* `dl.py` which have `class=property` set (not inherited).

I also add a separate rule to make sure that the entries are displayed under each other.

szkarpinski:
!build

JanuszL (approved):

klecki (approved):

PR #5294, opened 2024-01-29
Add morphological operators and the nvcvop module

banasraf:
## Category:

**New feature**

## Description:

This PR introduces the integration tools for CVCUDA operators and the morphological operators as a usage of them.

The cvcuda ops can now be implemented as classes deriving from the NVCVOperator class. This provides methods to wrap the DALI op inputs, outputs and arguments as NVCV types.

New operator, new nvcvop module

The nvcvop module

### Tests:

- [ ] Existing tests apply
- [x] New tests added
  - [ ] Python tests
  - [ ] GTests
  - [ ] Benchmark
  - [ ] Other
- [ ] N/A

### Documentation
- [x] Documentation updated
  - [ ] Docstring
  - [ ] Doxygen
  - [ ] RST
  - [ ] Jupyter
  - [ ] Other
- [ ] N/A

#### Requirements
- [ ] N/A

**REQ IDs**: N/A

**JIRA TASK**: DALI-3626

### DIFF
--- a/dali/operators/image/morphology/morphology.cc
+++ b/dali/operators/image/morphology/morphology.cc
@@ -1,0 +1,85 @@
+// Copyright (c) 2024, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "dali/operators/image/morphology/morphology.h"
+
+namespace dali {
+
+bool Morphology::SetupImpl(std::vector<OutputDesc> &output_desc, const Workspace &ws) {
+  const auto &input = ws.Input<GPUBackend>(0);
+  auto sh = input.shape();
+  output_desc.resize(1);
+  output_desc[0] = {sh, input.type()};
+  return true;
+}
+
+void Morphology::RunImpl(Workspace &ws) {
+  const auto &input = ws.Input<GPUBackend>(0);
+  auto &output = ws.Output<GPUBackend>(0);
+  output.SetLayout(input.GetLayout());
+
+  kernels::DynamicScratchpad scratchpad({}, AccessOrder(ws.stream()));
+  auto mask = AcquireTensorArgument<int32_t, NVCV_DATA_TYPE_2S32>(ws, scratchpad, mask_arg_,
+                                                                  TensorShape<1>(2));
+  auto anchor = AcquireTensorArgument<int32_t, NVCV_DATA_TYPE_2S32>(ws, scratchpad, anchor_arg_,
+                                                                    TensorShape<1>(2));
+
+  if (workspace_.capacity() < input.num_samples()) {
+    workspace_ = nvcv::ImageBatchVarShape(std::max(workspace_.capacity() * 2, input.num_samples()));
+  }
+  workspace_.clear();
+  nvcvop::AllocateImagesLike(input, scratchpad, workspace_);
+
+  auto input_images = GetInputBatch(ws, 0);
+  auto output_images = GetOutputBatch(ws, 0);
+  cvcuda::Morphology op{};
+  op(ws.stream(), input_images, output_images, workspace_, morph_type_, mask, anchor, iteration_,
+     border_mode_);
+}
+
+DALI_SCHEMA(Morphology)
+  .AddOptionalArg("mask", "Size of the convolution kernel.",
+                  std::vector<int32_t>({3, 3}), true, true)
+  .AddOptionalArg("anchor",
+                  "Sets the anchor point of the kernel. Default value (-1, -1)"
+                  " uses the kernel center as an anchor point.",
+                  std::vector<int32_t>({-1, -1}), true, true)
+  .AddOptionalArg("iterations",
+                  "Number of times to execute the operation, typically set to 1. "
+                  "Setting to higher than 1 is equivelent of increasing the kernel "
+                  "mask by (mask_width - 1, mask_height -1) for every iteration.",
+                  1, false, false)
+  .AddOptionalArg("border_mode",
+                  "Border mode to be used when accessing elements outside input image.",
+                  "constant")
+  .AllowSequences();
+
+DALI_SCHEMA(experimental__Dilate)
+  .DocStr("Perform a dilation operation on the input image.")
+  .NumInput(1)
+  .NumOutput(1)
+  .AddParent("Morphology");
+
+DALI_SCHEMA(experimental__Erode)
+  .DocStr("Perform an erosion operation on the input image.")
+  .NumInput(1)
+  .NumOutput(1)
+  .AddParent("Morphology");
+
+
+DALI_REGISTER_OPERATOR(experimental__Dilate, Dilate, GPU);
+
+DALI_REGISTER_OPERATOR(experimental__Erode, Erode, GPU);
+
+}  // namespace dali
--- a/dali/operators/image/morphology/morphology.h
+++ b/dali/operators/image/morphology/morphology.h
@@ -1,0 +1,77 @@
+// Copyright (c) 2024, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef DALI_OPERATORS_IMAGE_MORPHOLOGY_MORPHOLOGY_H_
+#define DALI_OPERATORS_IMAGE_MORPHOLOGY_MORPHOLOGY_H_
+
+#include <vector>
+#include <string>
+
+#include <cvcuda/OpMorphology.hpp>
+#include <nvcv/DataType.hpp>
+#include <nvcv/Tensor.hpp>
+
+#include "dali/operators/nvcvop/nvcvop.h"
+#include "dali/kernels/dynamic_scratchpad.h"
+#include "dali/pipeline/operator/arg_helper.h"
+#include "dali/pipeline/operator/checkpointing/stateless_operator.h"
+#include "dali/pipeline/operator/operator.h"
+#include "dali/pipeline/operator/sequence_operator.h"
+
+namespace dali {
+
+class Morphology : public nvcvop::NVCVSequenceOperator<StatelessOperator> {
+ public:
+  Morphology(const OpSpec &spec, NVCVMorphologyType morph_type) :
+    nvcvop::NVCVSequenceOperator<StatelessOperator>(spec),
+    morph_type_(morph_type),
+    border_mode_(nvcvop::GetBorderMode(spec.GetArgument<std::string>("border_mode"))),
+    iteration_(spec.GetArgument<int>("iterations")) {
+      DALI_ENFORCE(iteration_ >= 1, "iterations must be >= 1");
+    }
+
+  bool CanInferOutputs() const override {
+    return true;
+  }
+
+  bool SetupImpl(std::vector<OutputDesc> &output_desc, const Workspace &ws) override;
+
+  bool ShouldExpandChannels(int input_idx) const override {
+    return true;
+  }
+
+  void RunImpl(Workspace &ws) override;
+
+  USE_OPERATOR_MEMBERS();
+  NVCVMorphologyType morph_type_;
+  ArgValue<int32_t, 1> mask_arg_{"mask", spec_};
+  ArgValue<int32_t, 1> anchor_arg_{"anchor", spec_};
+  nvcv::ImageBatchVarShape workspace_{1};
+  NVCVBorderType border_mode_{NVCV_BORDER_CONSTANT};
+  int32_t iteration_ = 1;
+};
+
+class Dilate : public Morphology {
+ public:
+  explicit Dilate(const OpSpec &spec): Morphology(spec, NVCVMorphologyType::NVCV_DILATE) {}
+};
+
+class Erode : public Morphology {
+ public:
+  explicit Erode(const OpSpec &spec): Morphology(spec, NVCVMorphologyType::NVCV_ERODE) {}
+};
+
+}  // namespace dali
+
+#endif  // DALI_OPERATORS_IMAGE_MORPHOLOGY_MORPHOLOGY_H_
--- a/dali/operators/nvcvop/nvcvop.cc
+++ b/dali/operators/nvcvop/nvcvop.cc
@@ -1,0 +1,141 @@
+// Copyright (c) 2024, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include <string>
+
+#include "dali/operators/nvcvop/nvcvop.h"
+
+namespace dali::nvcvop {
+
+NVCVBorderType GetBorderMode(const std::string &border_mode) {
+  if (border_mode == "constant") {
+    return NVCV_BORDER_CONSTANT;
+  } else if (border_mode == "replicate") {
+    return NVCV_BORDER_REPLICATE;
+  } else if (border_mode == "reflect") {
+    return NVCV_BORDER_REFLECT;
+  } else if (border_mode == "reflect_101") {
+    return NVCV_BORDER_REFLECT101;
+  } else if (border_mode == "wrap") {
+    return NVCV_BORDER_WRAP;
+  } else {
+    DALI_FAIL("Unknown border mode: " + border_mode);
+  }
+}
+
+nvcv::DataKind GetDataKind(DALIDataType dtype) {
+  switch (dtype) {
+    case DALI_UINT8:
+    case DALI_UINT16:
+    case DALI_UINT32:
+    case DALI_UINT64:
+      return nvcv::DataKind::UNSIGNED;
+    case DALI_INT8:
+    case DALI_INT16:
+    case DALI_INT32:
+    case DALI_INT64:
+      return nvcv::DataKind::SIGNED;
+    case DALI_FLOAT16:
+    case DALI_FLOAT:
+    case DALI_FLOAT64:
+      return nvcv::DataKind::FLOAT;
+    default:
+      DALI_FAIL("Unsupported data type");
+  }
+}
+
+nvcv::Swizzle GetSimpleSwizzle(int num_channels) {
+  if (num_channels == 1) {
+      return nvcv::Swizzle::S_X000;
+  } else if (num_channels == 2) {
+      return nvcv::Swizzle::S_XY00;
+  } else if (num_channels == 3) {
+      return nvcv::Swizzle::S_XYZ0;
+  } else if (num_channels == 4) {
+      return nvcv::Swizzle::S_XYZW;
+  } else {
+      DALI_FAIL("Unsupported number of channels");
+  }
+}
+
+nvcv::ImageFormat GetImageFormat(DALIDataType dtype, int num_channels) {
+  nvcv::MemLayout mem_layout = nvcv::MemLayout::PITCH_LINEAR;
+  nvcv::DataKind data_kind = GetDataKind(dtype);
+  nvcv::Swizzle swizzle = GetSimpleSwizzle(num_channels);
+  nvcv::PackingParams packing_params{};
+  packing_params.byteOrder = nvcv::ByteOrder::MSB;
+  packing_params.swizzle = swizzle;
+  for (int c = 0; c < num_channels; ++c) {
+    packing_params.bits[c] = TypeTable::GetTypeInfo(dtype).size() * 8;
+  }
+
+  auto packing = nvcv::MakePacking(packing_params);
+  return nvcv::ImageFormat(mem_layout, data_kind, swizzle, packing);
+}
+
+/**
+ * @brief Wrap tensor data buffer into nvcv::Image.
+ *
+ * @param data pointer to beginning of the buffer
+ * @param shape underlying tensor shape
+ * @param format image format
+ */
+nvcv::Image WrapImage(void *data, const TensorShape<> &shape, const nvcv::ImageFormat &format) {
+  DALI_ENFORCE(format.numPlanes() == 1);
+  auto pixel_size = format.planeBitsPerPixel(0) / 8;
+  nvcv::ImageDataStridedCuda::Buffer buf{};
+  buf.numPlanes = 1;
+  buf.planes[0].basePtr = static_cast<NVCVByte*>(data);
+  buf.planes[0].height = shape[0];
+  buf.planes[0].width  = shape[1];
+  buf.planes[0].rowStride = shape[1] * pixel_size;
+  nvcv::ImageDataStridedCuda img_data(format, buf);
+  return nvcv::ImageWrapData(img_data);
+}
+
+nvcv::Image AsImage(SampleView<GPUBackend> sample, const nvcv::ImageFormat &format) {
+  return AsImage(ConstSampleView<GPUBackend>(sample), format);
+}
+
+nvcv::Image AsImage(ConstSampleView<GPUBackend> sample, const nvcv::ImageFormat &format) {
+  auto &shape = sample.shape();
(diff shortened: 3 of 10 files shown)


banasraf:
!build

banasraf:
!build

dali/operators/nvcvop/nvcvop.cc
```diff
@@ -0,0 +1,141 @@
...
+}
+
+nvcv::Image AsImage(ConstSampleView<GPUBackend> sample, const nvcv::ImageFormat &format) {
+  auto &shape = sample.shape();
+  auto data = const_cast<void*>(sample.raw_data());
+  return WrapImage(data, shape, format);
+}
+
+void AllocateImagesLike(const TensorList<GPUBackend> &t_list,
+                        kernels::DynamicScratchpad &scratchpad, nvcv::ImageBatchVarShape &output) {
```
mzient, line 118: Nitpick: DALI uses `memcpy` argument order (`func(out, in)`).

dali/operators/nvcvop/nvcvop.cc
```diff
@@ -0,0 +1,141 @@
...
+  for (int s = 0; s < t_list.num_samples(); ++s) {
+    auto num_channels = (channel_dim >= 0) ? t_list[s].shape()[channel_dim] : 1;
+    auto format = GetImageFormat(t_list.type(), num_channels);
+    auto image = WrapImage(buffer + offset, t_list[s].shape(), format);
+    output.pushBack(image);
+    offset += volume(t_list[s].shape()) * t_list.type_info().size();
+  }
+}
+
+void PushImagesToBatch(const TensorList<GPUBackend> &t_list, nvcv::ImageBatchVarShape &batch) {
```
mzient, line 131: Nitpick: Argument order - output or "this-like" argument should go first.

dali/operators/nvcvop/nvcvop.cc
```diff
@@ -0,0 +1,141 @@
...
+  return WrapImage(data, shape, format);
+}
+
+void AllocateImagesLike(const TensorList<GPUBackend> &t_list,
+                        kernels::DynamicScratchpad &scratchpad, nvcv::ImageBatchVarShape &output) {
+  auto channel_dim = t_list.GetLayout().find('C');
+  uint8_t *buffer = scratchpad.AllocateGPU<uint8_t>(t_list.nbytes());
+  size_t offset = 0;
+  for (int s = 0; s < t_list.num_samples(); ++s) {
+    auto num_channels = (channel_dim >= 0) ? t_list[s].shape()[channel_dim] : 1;
```
mzient, line 123: 1. TensorList's `operator[]` is complex enough to warrant having a temporary local variable instead of re-invoking it several times below.
2. I think that it'd be nice (and saving us trouble in the future) if we added an option to get a `shared_ptr` for a sample in TensorList. The SampleView, which was intended only for data access (as opposed to data ownership) doesn't provide that. With a shared_ptr we could provide a custom data deleter to the resulting NVCV tensor, thus making it safe for the NVCV object to outlive the DALI tensor.

dali/operators/nvcvop/nvcvop.h
```diff
@@ -0,0 +1,183 @@
...
+   * @brief Get an operator argument as an nvcv Tensor.
+   * This method uses the dynamic scratchpad to allocate the (pinned and device) buffers
+   *  used by the output tensor.
+   *
+   * @tparam T data type of the DALI operator argument
+   * @tparam DTYPE expected nvcv Tensor data type
+   * @param arg_shape shape of the DALI operator argument
+   */
+  template <typename T, NVCVDataType DTYPE>
+  nvcv::Tensor AcquireTensorArgument(Workspace &ws, kernels::DynamicScratchpad &scratchpad,
```
mzient, line 102: I guess we could just use an interface here - it doesn't matter if the scratchpad is dynamic or not. What matters (and should be documented!) is that the scratchpad returns its memory in an order that's compatible with `ws.stream()`.
```suggestion
  nvcv::Tensor AcquireTensorArgument(Workspace &ws, kernels::Scratchpad &scratchpad,
```

dali/operators/nvcvop/nvcvop.h
```diff
@@ -0,0 +1,183 @@
...
+    for (int s = 0; s < num_samples; ++s) {
+      std::copy(arg[s].data, arg[s].data + arg_sample_vol,
+                staging_buffer + s * arg_sample_vol);
+    }
+    T *device_buffer = scratchpad.AllocateGPU<T>(num_samples * arg_sample_vol, dtype.alignment());
+    MemCopy(device_buffer, staging_buffer, num_samples * arg_sample_vol * sizeof(T), ws.stream());
+    int ndim = arg_shape.sample_dim();
+    auto inner_dim = arg_shape[ndim - 1];
+    DALI_ENFORCE(inner_dim % dtype.numChannels() == 0, "Invalid argument shape.");
+    std::vector<int64_t> shape_data(ndim);
```
mzient, line 119: Bug!
```suggestion
    std::vector<int64_t> shape_data(ndim + 1);
```

dali/operators/nvcvop/nvcvop.h
```diff
@@ -0,0 +1,183 @@
...
+    shape_data[0] = num_samples;
+    for (int d = 0; d < ndim; ++d) {
+      shape_data[d + 1] = arg_shape[d];
```
mzient, line 122: Out-of-bounds write.

dali/operators/nvcvop/nvcvop.h
```diff
@@ -0,0 +1,183 @@
...
+    }
+    if (inner_dim != dtype.numChannels()) {
+      shape_data.push_back(inner_dim / dtype.numChannels());
+    }
```
mzient, line 126: This overwrites the (incorrectly written) last element of the shape. Is it even used?

dali/operators/nvcvop/nvcvop.h
```diff
@@ -0,0 +1,183 @@
...
+   * This method uses the dynamic scratchpad to allocate the (pinned and device) buffers
+   *  used by the output tensor.
+   *
+   * @tparam T data type of the DALI operator argument
+   * @tparam DTYPE expected nvcv Tensor data type
+   * @param arg_shape shape of the DALI operator argument
+   */
+  template <typename T, NVCVDataType DTYPE>
+  nvcv::Tensor AcquireTensorArgument(Workspace &ws, kernels::DynamicScratchpad &scratchpad,
+                                     ArgValue<T, 1> &arg, const TensorShape<> &arg_shape) {
```
mzient, line 103: How about
```suggestion
                                     ArgValue<T, 1> &arg, const TensorShape<> &arg_shape,
                                     TensorLayout layout = "") {
```

dali/operators/nvcvop/nvcvop.h
```diff
@@ -0,0 +1,183 @@
...
+    if (inner_dim != dtype.numChannels()) {
+      shape_data.push_back(inner_dim / dtype.numChannels());
+    }
+    nvcv::TensorDataStridedCuda::Buffer inBuf;
+    inBuf.basePtr = reinterpret_cast<NVCVByte*>(device_buffer);
+    inBuf.strides[shape_data.size() - 1] = dtype.strideBytes();
+    for (int d = shape_data.size() - 2; d >= 0; --d) {
+      inBuf.strides[d] = shape_data[d + 1] * inBuf.strides[d + 1];
+    }
+    nvcv::TensorShape shape(shape_data.data(), shape_data.size(), nvcv::TensorLayout(""));
```
mzient, line 133: (See suggestion for the function prototype)
```suggestion
    TensorLayout out_layout = layout.empty() ? layout : "N" + layout;
    nvcv::TensorShape shape(shape_data.data(), shape_data.size(), nvcv::TensorLayout(out_layout.c_str));
```

dali/operators/nvcvop/nvcvop.h
```diff
@@ -0,0 +1,183 @@
...
+    nvcv::TensorDataStridedCuda inData(shape, dtype, inBuf);
+    return nvcv::TensorWrapData(inData);
+  }
+
+  /**
+   * @brief Get input image batch.
+   * This method allocates the ImageBatchVarShape and wraps the input data as nvcv images.
```
mzient, line 140: Please add an information that the memory is not owned by the newly created NVCV tensor and instead is managed by the scratchpad object.

dali/operators/nvcvop/nvcvop.h
```diff
@@ -0,0 +1,183 @@
...
+    if (output.capacity() < tl_output.num_samples()) {
+      output = nvcv::ImageBatchVarShape(std::max(output.capacity() * 2, tl_output.num_samples()));
+    }
+    output.clear();
+    PushImagesToBatch(tl_output, output);
+    return output;
+  }
+
+  std::vector<nvcv::ImageBatchVarShape> inputs_;
+  std::vector<nvcv::ImageBatchVarShape> outputs_;
```
mzient, line 175: This is dangerous. The image batch objects (as of now) are ephemeral - they only live as long as the workspace is valid. As such, we should not extend their lifetime beyond the Run function. I propose that we implement a small `RunImpl` function here which clears those vectors at the end.

dali/operators/image/morphology/morphology.h
```diff
@@ -0,0 +1,77 @@
...
+    return true;
+  }
+
+  void RunImpl(Workspace &ws) override;
+
+  USE_OPERATOR_MEMBERS();
+  NVCVMorphologyType morph_type_;
+  ArgValue<int32_t, 1> mask_arg_{"mask", spec_};
+  ArgValue<int32_t, 1> anchor_arg_{"anchor", spec_};
+  nvcv::ImageBatchVarShape workspace_{1};
```
mzient, line 60: Why `{1}`? Can't we just have a null resource?

dali/operators/image/morphology/morphology.cc
```diff
@@ -0,0 +1,92 @@
...
+  auto &output = ws.Output<GPUBackend>(0);
+  output.SetLayout(input.GetLayout());
+
+  kernels::DynamicScratchpad scratchpad({}, AccessOrder(ws.stream()));
+  auto mask = AcquireTensorArgument<int32_t, NVCV_DATA_TYPE_2S32>(ws, scratchpad, mask_arg_,
+                                                                  TensorShape<1>(2));
+  auto anchor = AcquireTensorArgument<int32_t, NVCV_DATA_TYPE_2S32>(ws, scratchpad, anchor_arg_,
+                                                                    TensorShape<1>(2));
+
+  if (workspace_.capacity() < input.num_samples()) {
```
mzient, line 38:
```suggestion
  if (!workspace_ || workspace_.capacity() < input.num_samples()) {
```
...and get rid of the (usually incorrect) initialization in the constructor.

dali/operators/nvcvop/nvcvop.cc
banasraf: 1. Done
2. I added the info about lifetime bounds to the docs

dali/operators/nvcvop/nvcvop.h
banasraf: This shape calculations have to variants. We either remove the inner dimension (e.g, The tensor is {10, 2} of ints, and the dtype is int2 - the resulting tensor is a {10} of int2) or we don't. 

I fixed the bug that was there, moved the shape calculation to a separate function and made it more verbose.

dali/operators/nvcvop/nvcvop.h
banasraf: It's a function for argument intputs. Do they even have any layout?

dali/operators/image/morphology/morphology.cc
banasraf: Actually, I couldn't make it work. It thrown the invalid handle error

dali/operators/nvcvop/nvcvop.h
banasraf: I implemented the Run() method that cleans the batches after the RunImpl

banasraf:
!build

dali/operators/nvcvop/nvcvop.h
mzient: Why was it marked as resolved? The DynamicScratchpad is still there - is it required or we could just use the Scratchpad interface?

dali/operators/nvcvop/nvcvop.h
```diff
@@ -0,0 +1,227 @@
...
+    DALI_ENFORCE(inner_dim % dtype.numChannels() == 0, "Invalid argument shape.");
+    std::vector<int64_t> shape_data;
+    // If number of channels in the data type matches the innermost dimension,
+    // we remove this dimension
+    if (inner_dim == dtype.numChannels()) {
+      shape_data.resize(ndim);
+      shape_data[0] = num_samples;
+      for (int d = 0; d < ndim - 1; ++d) {
+        shape_data[d + 1] = arg_shape[d];
+      }
```
mzient, line 151: I think this logic can lead to all sorts of problems - are we sure that ambiguous cases (like numChannels == 1 && inner_dim == 1) are handled properly?

banasraf: I changed the logic here to be less ambiguous. 
Now there are two cases: dtype has single channel or dtype has multiple channels.
In single-channel case, the resulting shape is equal to the DALI shape.
In multi-channel case, the inner dim of DALI shape **has to be equal** to the number of channels and it's removed.

Generally these two cases come from the fact that cvcuda is not consistent in how it expects the args tensors to be shaped, e.g. window size for median blur is expected to be a {Nx2} tensor of ints and windows size for morphology is expected to be a {N} tensor of int2s.

dali/test/python/test_utils.py
```diff
@@ -424,7 +431,13 @@ def __next__(self):
                     self.rng.randint(min_s, max_s)
                     for min_s, max_s in zip(self.min_shape, self.max_shape)
                 ]
-            if self.dtype == np.float32:
+            if self.val_range:
+                min_val = self.val_range[0]
+                max_val = self.val_range[1]
+                self.test_data.append(
+                    np.array(self.np_rng.random(shape) * max_val, dtype=self.dtype) - min_val
```
stiepan, line 438: I am not sure what's the desired output, but shouldn't it be?
```suggestion
                    np.array(self.np_rng.random(shape) * (max_val - min_val) + min_val, dtype=self.dtype)
```
?

dali/operators/image/morphology/morphology.cc
```diff
@@ -0,0 +1,94 @@
...
+                  "Setting to higher than 1 is equivelent of increasing the kernel "
+                  "mask by (mask_width - 1, mask_height -1) for every iteration.",
+                  1, false, false)
+  .AddOptionalArg("border_mode",
+                  "Border mode to be used when accessing elements outside input image.",
+                  "constant");
+
+DALI_SCHEMA(experimental__Dilate)
+  .AddParent("Morphology")
+  .DocStr("Perform a dilation operation on the input image.")
```
stiepan, line 71: nitpick: 
```suggestion
  .DocStr("Performs a dilation operation on the input image.")
```

dali/operators/image/morphology/morphology.cc
```diff
@@ -0,0 +1,94 @@
...
+  .NumInput(1)
+  .NumOutput(1)
+  .InputDox(0, "input", "TensorList",
+            "Input data. Must be images in HWC or CHW layout, or a sequence of those.")
+  .AllowSequences()
+  .InputLayout({"HW", "HWC", "FHWC", "CHW", "FCHW"});
+
+DALI_SCHEMA(experimental__Erode)
+  .AddParent("Morphology")
+  .DocStr("Perform an erosion operation on the input image.")
```
stiepan, line 81:
```suggestion
  .DocStr("Performs an erosion operation on the input image.")
```

stiepan (approved):
Pretty cool how brief the definition of the operator itself becomes with the utilities factored out.

banasraf:
!build

dali/test/python/test_utils.py
banasraf: It should be, thanks

banasraf:
!build

dali/operators/nvcvop/nvcvop.h
```diff
@@ -0,0 +1,222 @@
...
+   * The output should be used only within the RunImpl method.
+   */
+  const nvcv::ImageBatchVarShape &GetInputBatch(Workspace &ws, size_t input_idx) {
+    if (inputs_.size() < input_idx + 1) {
+      inputs_.resize(input_idx + 1, nvcv::ImageBatchVarShape{1});
+    }
+    const auto &tl_input = ws.Input<GPUBackend>(input_idx);
+    auto &input = inputs_[input_idx];
+    if (input.capacity() < tl_input.num_samples()) {
+      input = nvcv::ImageBatchVarShape(std::max(input.capacity() * 2, tl_input.num_samples()));
```
stiepan, line 171: Just to leave a mark on what we've discussed offline - do we need or consider making the ImageBatchVarShape internal allocations handled by DALI allocators/scratchpad? With some aggresive prefetching scenarios it could(?) trade more memory usage for less synchronization, but I don't know if we have actual need for that.

dali/operators/nvcvop/nvcvop.cc
```diff
@@ -0,0 +1,142 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include <string>
+
+#include "dali/operators/nvcvop/nvcvop.h"
+
+namespace dali::nvcvop {
+
+NVCVBorderType GetBorderMode(const std::string &border_mode) {
```
mzient, line 21: Nitpick - C++17 :)
```suggestion
NVCVBorderType GetBorderMode(std::string_view border_mode) {
```

dali/operators/nvcvop/nvcvop.h
```diff
@@ -0,0 +1,222 @@
...
+      inBuf.strides[d] = shape_data[d + 1] * inBuf.strides[d + 1];
+    }
+    TensorLayout out_layout = layout.empty() ? layout : "N" + layout;
+    nvcv::TensorShape shape(shape_data.data(), shape_data.size(),
+                            nvcv::TensorLayout(out_layout.c_str()));
+    nvcv::TensorDataStridedCuda inData(shape, dtype, inBuf);
+    return nvcv::TensorWrapData(inData);
+  }
+
+  std::vector<int64_t> CalcTensorShape(const TensorShape<> &arg_shape, int num_samples,
```
mzient, line 138:
```suggestion
  TensorShape<> CalcTensorShape(const TensorShape<> &arg_shape, int num_samples,
```
?
Most of the time we won't need dynamic allocation.

dali/operators/nvcvop/nvcvop.h
```diff
@@ -0,0 +1,222 @@
...
+    if (output.capacity() < tl_output.num_samples()) {
+      output = nvcv::ImageBatchVarShape(std::max(output.capacity() * 2, tl_output.num_samples()));
+    }
+    PushImagesToBatch(output, tl_output);
+    return output;
+  }
+
+  void Run(Workspace &ws) override {
+    BaseOp::Run(ws);
+    ClearBatches();
```
mzient, line 198: Slightly better:
```suggestion
    auto atexit = AtScopeExit([]() { ClearBatches(); });
    BaseOp::Run(ws);
```

dali/operators/nvcvop/nvcvop.h
```diff
@@ -0,0 +1,222 @@
...
+
+  /**
+   * @brief Get input image batch.
+   * This method allocates the ImageBatchVarShape and wraps the input data as nvcv images.
+   *
+   * The output should be used only within the RunImpl method.
+   */
+  const nvcv::ImageBatchVarShape &GetInputBatch(Workspace &ws, size_t input_idx) {
+    if (inputs_.size() < input_idx + 1) {
+      inputs_.resize(input_idx + 1, nvcv::ImageBatchVarShape{1});
```
mzient, line 166:
```suggestion
      inputs_.resize(input_idx + 1, nullptr);
```

dali/operators/nvcvop/nvcvop.h
mzient: I think we should wrap our allocators into an NVCV Allocator. Perhaps this can be done in a follow-up, but it definitely _should_ be done.

dali/operators/nvcvop/nvcvop.h
```diff
@@ -0,0 +1,222 @@
...
+    }
+    const auto &tl_input = ws.Input<GPUBackend>(input_idx);
+    auto &input = inputs_[input_idx];
+    if (input.capacity() < tl_input.num_samples()) {
+      input = nvcv::ImageBatchVarShape(std::max(input.capacity() * 2, tl_input.num_samples()));
```
mzient, line 171:
```suggestion
    int curr_cap = input ? input.capacity() : 0;
    if (curr_cap < tl_input.num_samples()) {
      input = nvcv::ImageBatchVarShape(std::max(curr_cap * 2, tl_input.num_samples()));
```

dali/operators/nvcvop/nvcvop.h
```diff
@@ -0,0 +1,222 @@
...
+
+  /**
+   * @brief Get output image batch.
+   * This method allocates the ImageBatchVarShape and wraps the input data as nvcv images.
+   *
+   * The output should be used only within the RunImpl method.
+   */
+  nvcv::ImageBatchVarShape &GetOutputBatch(Workspace &ws, size_t output_idx) {
+    if (outputs_.size() < output_idx + 1) {
+      outputs_.resize(output_idx + 1, nvcv::ImageBatchVarShape{1});
```
mzient, line 185:
```suggestion
      outputs_.resize(output_idx + 1, nullptr);
```

dali/operators/nvcvop/nvcvop.h
```diff
@@ -0,0 +1,222 @@
...
+    }
+    const auto &tl_output = ws.Output<GPUBackend>(output_idx);
+    auto &output = outputs_[output_idx];
+    if (output.capacity() < tl_output.num_samples()) {
+      output = nvcv::ImageBatchVarShape(std::max(output.capacity() * 2, tl_output.num_samples()));
```
mzient, line 190:
```suggestion
    int curr_cap = output ? output.capacity() : 0;
    if (curr_cap < tl_output.num_samples()) {
      output = nvcv::ImageBatchVarShape(std::max(curr_cap * 2, tl_output.num_samples()));
```

banasraf: done

dali/operators/nvcvop/nvcvop.h
banasraf: done

dali/operators/nvcvop/nvcvop.h
banasraf: done

dali/operators/nvcvop/nvcvop.h
banasraf: done

dali/operators/nvcvop/nvcvop.h
banasraf: done

dali/operators/nvcvop/nvcvop.cc
banasraf: done

dali/operators/image/morphology/morphology.cc
```diff
@@ -0,0 +1,94 @@
...
+
+  auto input_images = GetInputBatch(ws, 0);
+  auto output_images = GetOutputBatch(ws, 0);
+  cvcuda::Morphology op{};
+  op(ws.stream(), input_images, output_images, op_workspace_, morph_type_, mask, anchor, iteration_,
+     border_mode_);
+}
+
+DALI_SCHEMA(Morphology)
+  .AddOptionalArg("mask", "Size of the convolution kernel.",
```
mzient, line 54:
```suggestion
  .AddOptionalArg("mask", "Size of the kernel.",
```

dali/operators/image/morphology/morphology.cc
```diff
@@ -0,0 +1,94 @@
...
+                  std::vector<int32_t>({3, 3}), true, true)
+  .AddOptionalArg("anchor",
+                  "Sets the anchor point of the kernel. Default value (-1, -1)"
+                  " uses the kernel center as an anchor point.",
```
mzient, line 58:
```suggestion
                  "Sets the anchor point of the structuring element. Default value (-1, -1)"
                  " uses the element's center as an anchor point.",
```

dali/operators/image/morphology/morphology.cc
```diff
@@ -0,0 +1,94 @@
...
+                  std::vector<int32_t>({-1, -1}), true, true)
+  .AddOptionalArg("iterations",
+                  "Number of times to execute the operation, typically set to 1. "
+                  "Setting to higher than 1 is equivelent of increasing the kernel "
+                  "mask by (mask_width - 1, mask_height -1) for every iteration.",
+                  1, false, false)
```
mzient, line 64:
```suggestion
                  "Setting to a value higher than 1 is equivalent to increasing the "
                  "mask size by (mask_width - 1, mask_height -1) for every additional iteration.",
                  1, false, false)
```

dali/operators/nvcvop/nvcvop.h
```diff
@@ -0,0 +1,225 @@
...
+   *
+   * @tparam T data type of the DALI operator argument
+   * @tparam DTYPE expected nvcv Tensor data type
+   * @param arg_shape shape of the DALI operator argument.
+   */
+  template <typename T, NVCVDataType DTYPE>
+  nvcv::Tensor AcquireTensorArgument(Workspace &ws, kernels::Scratchpad &scratchpad,
+                                     ArgValue<T, 1> &arg, const TensorShape<> &arg_shape,
+                                     TensorLayout layout = "") {
+    auto dtype = nvcv::DataType(DTYPE);
```
mzient, line 112: Why compile-time DTYPE?
```suggestion
  template <typename T>
  nvcv::Tensor AcquireTensorArgument(Workspace &ws, kernels::Scratchpad &scratchpad,
                                     ArgValue<T, 1> &arg, const TensorShape<> &arg_shape,
                                     nvcv::DataType dtype,
                                     TensorLayout layout = "") {
```
Also, we could probably have some default inferred from T.

banasraf: done

dali/operators/image/morphology/morphology.cc
banasraf: done

dali/operators/image/morphology/morphology.cc
banasraf: done

mzient (approved):

PR #5302, opened 2024-02-01
Add origin stack trace capture for DALI operators

klecki:
## Description:
Add capture of origin stack trace for DALI operators.
The `extract_stack` collects all frames from the first frame of pipeline definition to the frame with operator invocation (the outermost user API).
In regular mode no further processing is needed.
For code that was transformed by AutoGraph, the contents of captured frames are remapped back to the user code, filtered out of the _autograph and _conditionals modules (that contain internal DALI implementation). 
Due to how autograph introduces additional frames (for example by implementing if statements with additional function calls), we remove repeated occurrences of the same function, keeping only the last one. AutoGraph's entry point to a function call is used to detect such regions.

Stack traces are collected in OperatorInstance, the outermost layer of the API adds an argument denoting the current stack level, so we can compute how many frames we need to skip related to DALI internals when capturing the stack.

The collected stack is added as hidden arguments to OpSpec. This is backward-compatible for serialized pipelines, as well as allows for disabling and enabling the feature. 

As a default the feature is disabled and only available via hidden API.

C++ test operators are extended to work as a loadable plugin based on dummy operator test. 
It allows to implement tests for the new feature without extending regular DALI operators.

Follow-up to this PR will utilize this information for providing better error messages - pointing to the origin of error in DALI pipeline definition.

New functionality, extensions for Python API and new hidden arguments in all operators.

### Tests:

- [ ] Existing tests apply
- [x] New tests added
  - [x] Python tests
  - [ ] GTests
  - [ ] Benchmark
  - [ ] Other
- [ ] N/A

### Documentation
- [ ] Documentation updated
  - [ ] Docstring
  - [ ] Doxygen
  - [ ] RST
  - [ ] Jupyter
  - [ ] Other
- [ ] N/A

#### Requirements
- [ ] N/A

**REQ IDs**: N/A

**JIRA TASK**: N/A


klecki:
!build

dali/python/nvidia/dali/ops/__init__.py
```diff
@@ -411,6 +415,23 @@ def _process_instance_name(self, arguments):
         else:
             self._name = "__" + type(self._op).__name__ + "_" + str(self._counter.id)
 
+    def _process_trace(self, arguments):
+        if _dali_trace.is_tracing_enabled():
+            if _Pipeline.current():
+                skip_bottom = _Pipeline.current()._definition_stack_frame
+            else:
+                skip_bottom = 0
+            skip_top = 5 if self._op._api == "fn" else 3
```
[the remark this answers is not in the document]
klecki: I was thinking about other solutions: 
* I got the suggestion to use decorator, that would adjust the API functions with stop marker - but I only deal with the frames in the for of `FrameSummary` that contain: filename, the line number, the function name and the code context (as a string) - I cannot look up runtime metadata hera without going through the actual stack frames and this will probably be much heavier.
* Filtering based on the function names also ties with the implementation, I would say even more than just the magic number.
If we change something in the internal implementation, the length of the stack trace will change and the test will fail.
I am also not super happy about the magic numbers, but I saw it in TF as well - when you want to get the stack trace for something in your API, and you put the code in the helper function you change the contents of the stack trace :(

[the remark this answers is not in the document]
stiepan: For going through actuall stack frames, I think the same, it will be likely much heavier and kind of easier to make it wrong (e.g. to introduce some circular references that would bloat up memory usage).
For filtering by name, sure, it's still tying to the implementation, I just feel that being explicit here is a benefit for maintainers. If you think otherwise, how about a comment listing exact functions we expect to remove with those magic numbers? I am mostly conerned with the part that looks back and makes assumption of how many calls have been made before entering the function that has the magic constant defined.

Or maybe, on the explicity part, the front-end operator constructors could just create some stack_context (as simple as an integer - the stack size) and just painstakingly pass it through the utlities to the point where we acquire the stack. When making adjustments to the code in the future, you just need to remember to pass it and it works.

klecki:
Steps based on the offline discussion with @stiepan :
* extend estimation of the bottom of the stack trace by looking up if parent frames are user-defined or from DALI internals
* record the stack trace size when entering the API function and use it to skip the top
* test with pipe and push_current
* disable this feature for debug and eager mode - we rely purely on Python stack trace there
* I will probably remove `_api` for now if it is not needed here and introduce it later.

dali/python/nvidia/dali/ops/__init__.py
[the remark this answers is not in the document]
klecki: Done (I hope).

dali/python/nvidia/dali/_utils/dali_trace.py
```diff
@@ -0,0 +1,165 @@
...
+            is_ag_function_call_start = False
+            current_function_region = origin_frame_entry
+            skip = True
+            origin_stack_summary.append(origin_frame_entry)
+
+        # User code - not filtered out
+        if not skip:
+            # If we are in the same function region, we replace previous entry so we keep only the
+            # last one
+            assert origin_stack_summary
```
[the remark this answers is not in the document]
klecki: Good question. In the first call to AG, we are doing converted call to a pipeline definition, so we will always place something on the stack first.

dali/pipeline/operator/error_reporting.h
```diff
@@ -0,0 +1,59 @@
...
+        line(std::move(line)) {}
+  std::string filename;
+  int lineno;
+  std::string name;
+  std::string line;
+};
+
+/**
+ * @brief Get the origin stack trace for operator constructed with given spec.
+ * The stack trace defines frames between the invocation od pipeline definition and operator call.
```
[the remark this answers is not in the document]
klecki: done

dali/pipeline/operator/error_reporting.cc
```diff
@@ -0,0 +1,53 @@
...
+namespace dali {
+
+std::vector<DaliFrameSummary> GetOperatorOriginInfo(const OpSpec &spec) {
+  auto origin_stack_filename = spec.GetRepeatedArgument<std::string>("_origin_stack_filename");
+  auto origin_stack_lineno = spec.GetRepeatedArgument<int>("_origin_stack_lineno");
+  auto origin_stack_name = spec.GetRepeatedArgument<std::string>("_origin_stack_name");
+  auto origin_stack_line = spec.GetRepeatedArgument<std::string>("_origin_stack_line");
+
+  std::vector<DaliFrameSummary> origin_stack;
+  origin_stack.reserve(origin_stack_filename.size());
```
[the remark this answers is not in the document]
klecki: done

klecki:
!build

klecki:
!build

klecki:
!build

dali/pipeline/operator/error_reporting.cc
```diff
@@ -0,0 +1,57 @@
...
+  std::string error = "Internal error, mismatch in origin stack trace data.";
```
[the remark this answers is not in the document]
klecki: done

dali/pipeline/operator/op_schema.cc
```diff
@@ -78,6 +78,37 @@ to accommodate a batch of samples of this size.)code",
   AddOptionalArg("preserve", R"code(Prevents the operator from being removed from the
 graph even if its outputs are not used.)code",
                  false);
+
+
+  // For simplicity we pass StackSummary as 4 separate arguments so we don't need to extend DALI
+  // with support for special FrameSummary type.
+  // List of FrameSummaries can be reconstructed using utility functions.
+  AddOptionalArg("_origin_stack_filename", R"code(Every operator defined in Python captures and
+processes the StackSummary (a List[FrameSummary], defined in Python traceback module) that describe
```
szkarpinski, line 87:
```suggestion
processes the StackSummary (a List[FrameSummary], defined in Python traceback module) that describes
```

dali/python/nvidia/dali/_autograph/utils/tf_stack.py
```diff
@@ -148,42 +170,54 @@ def get_filtered_filenames(self):
...
+        super().__init__()
+        self._filtered_filenames = set()
+        if not isinstance(module_filter, list):
+            module_filter = [module_filter]
+        for module in module_filter:
+            try:
+                module_file = inspect.getfile(module)
+                init_py = "/__init__.py"
+                if module_file.endswith(init_py):
+                    module_file = module_file[: -len(init_py)]
```
szkarpinski, line 193: That's a bit hacky. How about just splitting on `/`, discarding last item if it's `__init__.py` and then joining?

dali/test/python/operator_1/test_operator_origin_trace.py
```diff
@@ -0,0 +1,331 @@
+# Copyright (c) 2024, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
szkarpinski, line 1: How about adding a simple test to check if DALI internals are correctly filtered out? This piece of code:
```py
if (
   len(call_context) > 2
   and call_context[-2].filename == current_filename
   and call_context[-2].name == "__enter__"
   and call_context[-3].filename == current_filename
    and call_context[-3].name == "_generate_graph"
):
```
is a heuristic and I can imagine it might stop working one day when somebody for example adds a helper function between `_generate_graph` and `__enter__`. A test for that would allow us to notice it faster

klecki: The tests already check against an expected stack trace, captured by hand (or by running the pipeline definition as normal function) so I think this should already be covered.

szkarpinski: That's right, I thought they only check if the stack trace _contains_ the pattern, but now I see they look for exact match. Sorry!

dali/python/nvidia/dali/_utils/dali_trace.py
```diff
@@ -0,0 +1,184 @@
...
+            if _is_matching_function(origin_stack_summary[-1], current_function_region):
+                if _collapse_ag_frames:
+                    origin_stack_summary.pop()
+            else:
+                current_function_region = None
+            origin_stack_summary.append(origin_frame_entry)
+        elif not _filter_ag_frames and (
+            len(origin_stack_summary) == 0 or origin_stack_summary[-1] != origin_frame_entry
+        ):
+            origin_stack_summary.append(origin_frame_entry)
```
stiepan, line 137:
```suggestion
        # User code - not filtered out
        if not skip:
            if is_ag_function_call_start:
                is_ag_function_call_start = False
                current_function_region = origin_frame_entry
                origin_stack_summary.append(origin_frame_entry)
            else:    
                # If we are in the same function region, we replace previous entry so we keep only the
                # last one
                assert origin_stack_summary
                if _is_matching_function(origin_stack_summary[-1], current_function_region):
                    if _collapse_ag_frames:
                        origin_stack_summary.pop()
                else:
                    current_function_region = None
                origin_stack_summary.append(origin_frame_entry)
        elif not _filter_ag_frames:
            origin_stack_summary.append(origin_frame_entry)
```

Sorry for so intrusive suggestions, but the `skip=True` part got me confused a little. I'd rephrase it a little, first for clarity of what the skip means, second, so that you don't need to double check (in `elif not _filter_ag_frames`) if you don't add the same frame twice). The latter part could potentially filter too much (if there existed recursive internal ag calls:)), right?

dali/test/python/operator_1/test_operator_origin_trace.py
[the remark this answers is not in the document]
klecki: Hmm, I added the 
```
try: 
   ...
finally: 
   del frame
```
to the code that handles actual frames. Otherwise we use library functions that return the Stack/FrameInfo objects and they do not contain references to the actual frames/variables so they should be safe.
If anything this would make life a bit more tedious for GC rather than introduce actual bug AFAIK.

dali/pipeline/operator/op_schema.cc
klecki: done

dali/python/nvidia/dali/_autograph/utils/tf_stack.py
klecki: Tbh I don't see much of a difference in how we detect and remove the `__init__.py` that indicates a module directory.

dali/python/nvidia/dali/ops/__init__.py
```diff
@@ -411,6 +415,27 @@ def _process_instance_name(self, arguments):
...
+        current_pipeline = _PipelineDebug.current()
+        is_debug = getattr(current_pipeline, "_debug_on", False)
+        if _dali_trace.is_tracing_enabled() and not is_debug:
+            if _Pipeline.current():
+                skip_bottom = _Pipeline.current()._definition_frame_start
+            else:
+                skip_bottom = 0
+            skip_top = _dali_trace.get_stack_depth() - self._op._definition_frame_end
+            stack_summary = _dali_trace.extract_stack(
+                skip_bottom_frames=skip_bottom, skip_top_frames=skip_top
```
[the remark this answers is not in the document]
klecki: I eliminated one depth calculation by using start and end frame index directly.

dali/python/nvidia/dali/_utils/dali_trace.py
```diff
@@ -0,0 +1,184 @@
...
+    # objects. Frame summary contains filename, lineno, name and line (string representing context).
+    # -1 so we drop extract_stack frame
+    stack_depth = get_stack_depth()
+    limit = stack_depth - skip_bottom_frames
+    stack_summary = traceback.extract_stack(limit=limit)[: -1 - skip_top_frames]
+
+    # If those are empty, AutoGraph transformations were not used, we can return as is
+    frame_map = get_frame_map()
+    frame_filter = get_frame_filter()
+    if not frame_map and not frame_filter:
```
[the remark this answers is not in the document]
klecki: Hmm, without autograph, both are empty.

dali/python/nvidia/dali/_utils/dali_trace.py
klecki: I'm up for those kind of suggestions that can make it clearer.

klecki:
!build

stiepan (approved):

dali/python/nvidia/dali/_utils/dali_trace.py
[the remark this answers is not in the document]
stiepan: Sure, I just though this way it would rely less on that external assumption and more on when this particular function can work. But I guess it's a bit hypothetical discussion.

szkarpinski (approved):

dali/pipeline/operator/error_reporting.h
```diff
@@ -0,0 +1,59 @@
...
+namespace dali {
+
+// TODO(klecki): Throw this one into a namespace?
+
+/**
+ * @brief Direct equivalent of Python's traceback.FrameSummary:
+ * https://docs.python.org/3/library/traceback.html#traceback.FrameSummary
+ *
+ */
+struct DaliFrameSummary {
```
mzient, line 34: For consistency with the remaining 90% of the codebase
```suggestion
struct DALIFrameSummary {
```
or perhaps just drop it - it's in namespace `dali` anyway.

dali/pipeline/operator/error_reporting.h
```diff
@@ -0,0 +1,59 @@
...
+  DaliFrameSummary(std::string &&filename, int lineno, std::string &&name, std::string &&line)
```
mzient, line 35:
```suggestion
  DaliFrameSummary(std::string filename, int lineno, std::string name, std::string line)
```

dali/pipeline/operator/error_reporting.h
```diff
@@ -0,0 +1,59 @@
...
+      : filename(std::move(filename)),
+        lineno(lineno),
+        name(std::move(name)),
+        line(std::move(line)) {}
+  std::string filename;
+  int lineno;
+  std::string name;
```
mzient, line 42: What name is it? Function name? Operator name? Some comment would be nice (and perhaps a better name).

dali/pipeline/operator/op_schema.cc
```diff
@@ -78,6 +78,37 @@ to accommodate a batch of samples of this size.)code",
   AddOptionalArg("preserve", R"code(Prevents the operator from being removed from the
 graph even if its outputs are not used.)code",
                  false);
+
+
+  // For simplicity we pass StackSummary as 4 separate arguments so we don't need to extend DALI
+  // with support for special FrameSummary type.
+  // List of FrameSummaries can be reconstructed using utility functions.
+  AddOptionalArg("_origin_stack_filename", R"code(Every operator defined in Python captures and
+processes the StackSummary (a List[FrameSummary], defined in Python traceback module) that describes
```
mzient, line 87: ...I will complain about the use of the name "summary" here like I did at the point of declaration of the FrameSummary structure.

dali/python/nvidia/dali/_autograph/impl/api.py
```diff
@@ -130,40 +129,47 @@ def _attach_error_metadata(e, f):
...
-#     super().__init__()
-#     self._source_map = converted_fn.ag_source_map
-#     # This may be called repeatedly: once on entry, by the superclass, then by
-#     # each child context manager.
-#     self._cached_map = None
+    def __init__(self, converted_fn):
+        super().__init__()
+        self._source_map = converted_fn.ag_source_map
+        # This may be called repeatedly: once on entry, by the superclass, then by
+        # each child context manager.
```
mzient, line 139: I don't understand this comment nor do I know what "call" it refers to - it sits between two assignments (no calls).

dali/python/nvidia/dali/_autograph/impl/api.py
```diff
@@ -283,6 +289,20 @@ def is_autograph_artifact(entity):
     return hasattr(entity, "autograph_info__")
 
 
+def is_frame_converted_call(frame_info):
```
mzient, line 292:
```suggestion
def is_frame_call_converted(frame_info):
```
?
This order of words is used below.... or is this name supposed to mean something else?

dali/pipeline/operator/error_reporting.h
klecki: I just took most of the naming from the Python struct that is linked in the docstring.
PythonStackFrame will be probably closest thing, even though I would be inclined to keep the name correspondence.

dali/python/nvidia/dali/_autograph/utils/tf_stack.py
```diff
@@ -148,42 +170,54 @@ def get_filtered_filenames(self):
...
+    We detect the top directory of given module and filter all frames from that path or its subpath.
     """
-    thread_key = _get_thread_key()
-    return _tf_stack.extract_stack(
-        _source_mapper_stacks[thread_key][-1].internal_map,
-        _source_filter_stacks[thread_key][-1].internal_set,
-        stacklevel,
-    )
 
+    def __init__(self, module_filter: "Union[List[ModuleType], ModuleType]"):
```
mzient, line 183: Why is the type hint a string?

dali/python/nvidia/dali/_autograph/utils/tf_stack.py
mzient: I know we're not windows-ready, but how does Python represent paths on windows? Shouldn't we look for directory separator (or replace it with `/` before this check)?

dali/pipeline/operator/error_reporting.h
klecki: Done

dali/pipeline/operator/error_reporting.h
klecki: Added comments, I thought that the Python reference is enough, but probably it is better to have it in the source as well.
The names are taken from Python, and I want to keep them for this reason.

dali/pipeline/operator/error_reporting.h
```diff
@@ -0,0 +1,59 @@
...
+  std::string line;
```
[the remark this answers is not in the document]
klecki: Comment added as explained above.

dali/pipeline/operator/error_reporting.h
```diff
@@ -0,0 +1,59 @@
...
+/**
+ * @brief Get the origin stack trace for operator constructed with given spec.
+ * The stack trace defines frames between invocation of the pipeline definition and the operator
+ * call. The returned DaliFrameSummary corresponds to Python traceback.FrameSummary, but the `line`
+ * context may be invalid in some autograph transformed code.
+ */
+DLL_PUBLIC std::vector<DaliFrameSummary> GetOperatorOriginInfo(const OpSpec &spec);
+
+DLL_PUBLIC std::string FormatStack(const std::vector<DaliFrameSummary> &stack_summary,
+                                   bool include_context);
```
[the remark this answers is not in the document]
klecki: We pass the list of FrameSummary cut into 4 separate lists via the op spec arguments. This means we create one per operator. We will retrieve them only for the purpose of building error message. Any processing is thus done outside of the main iteration loop, either when building pipeline or when erroring out.

dali/python/nvidia/dali/_autograph/impl/api.py
klecki: This comment is inherited from TensorFlow.
The whole structure of StackTraceMappers/Filters is used as context managers, and this comment is a bit exaggerated in the expectation of further inheritance from this class.

dali/python/nvidia/dali/_autograph/utils/tf_stack.py
klecki: Done (it's equivalent)

dali/python/nvidia/dali/_autograph/utils/tf_stack.py
klecki: I removed the `/` from the comparison string, the filtering should work with training directory separator as well.

klecki:
!build

mzient (approved):

PR #5311, opened 2024-02-06
Add missing calls to nvmlShutdown

JanuszL:
- in some places, DALI initializes nvml but misses
  to call nvmlShutdown to shut it down gracefully

## Category:

## Description:
- in some places, DALI initializes nvml but misses
  to call nvmlShutdown to shut it down gracefully

- cuda malloc resource
- nvjpeg tests
- gds tests
- optical flow
- video loader_decoder

### Tests:

- [x] Existing tests apply
- [ ] New tests added
  - [ ] Python tests
  - [ ] GTests
  - [ ] Benchmark
  - [ ] Other
- [ ] N/A

### Documentation
- [ ] Documentation updated
  - [ ] Docstring
  - [ ] Doxygen
  - [ ] RST
  - [ ] Jupyter
  - [ ] Other
- [x] N/A

#### Requirements
- [x] N/A

**REQ IDs**: N/A

**JIRA TASK**: N/A

### DIFF
--- a/dali/operators/decoder/nvjpeg/nvjpeg_decoder_decoupled_api_test.cc
+++ b/dali/operators/decoder/nvjpeg/nvjpeg_decoder_decoupled_api_test.cc
@@ -182,13 +182,10 @@ TYPED_TEST(nvjpegDecodeDecoupledAPITest, TestSingleTiffDecode4T) {
   this->TiffTestDecode(4);
 }
 
-#if NVJPEG_VER_MAJOR >= 11
+#if NVJPEG_VER_MAJOR >= 11 && NVML_ENABLED
 void PrintDeviceInfo() {
   unsigned int device_count;
-  if (!nvmlIsInitialized()) {
-    nvml::Init();
-    return;
-  }
+  nvml::Init();
   CUDA_CALL(nvmlDeviceGetCount_v2(&device_count));
   for (unsigned int device_idx = 0; device_idx < device_count; device_idx++) {
     auto info = nvml::GetDeviceInfo(device_idx);
@@ -198,6 +195,7 @@ void PrintDeviceInfo() {
               << " cc_m " << info.cap_minor
               << std::endl;
   }
+  nvml::Shutdown();
 }
 
 /**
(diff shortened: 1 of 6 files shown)


JanuszL:
!build

dali/operators/decoder/nvjpeg/nvjpeg_decoder_decoupled_api_test.cc
```diff
@@ -206,7 +204,9 @@ void PrintDeviceInfo() {
 bool ShouldUseHwDecoder() {
   // HW decoder is disabled for drivers < 455.x, see
   // dali/operators/decoder/nvjpeg/nvjpeg_decoder_decoupled_api.h for details
+  nvml::Init();
   static float driver_version = nvml::GetDriverVersion();
+  nvml::Shutdown();
   static bool device_supports_hw_decoder = nvml::isHWDecoderSupported();
   return device_supports_hw_decoder && driver_version >= 455;
```
mzient, line 211: This is a potential perf hit.
```suggestion
  static const float driver_version = []() {
    nvml::Init();
    float ver = nvml::GetDriverVersion();
    nvml::Shutdown();
    return ver;
  };
  static const bool device_supports_hw_decoder = nvml::isHWDecoderSupported();
  return device_supports_hw_decoder && driver_version >= 455;
```

JanuszL: In the hot path, I would go for something faster. This is just a test. I don't think it will affect it much.

dali/operators/sequence/optical_flow/optical_flow.h
```diff
@@ -78,7 +78,11 @@ class OpticalFlow : public StatelessOperator<Backend> {
 #endif
   }
 
-  ~OpticalFlow();
+  ~OpticalFlow() {
+#if NVML_ENABLED
+    nvml::Shutdown();
```
stiepan, line 83: Isn't that a problem if we have a couple of the ops instances (two pipelines for instance) and one of them goes out of scope while the other remains in use?

JanuszL: I think it should not (based on [the docs](https://docs.nvidia.com/deploy/nvml-api/group__nvmlInitializationAndCleanup.html#group__nvmlInitializationAndCleanup_1gcb6e78434b4d14250ce79c3e7494f06d))
```
This method, should be called once before invoking any other methods in the library. A reference count of the number of initializations is maintained. Shutdown only occurs when the reference count reaches zero.
```

stiepan (approved):

mzient (approved):

JanuszL:
Let me run with sanitizers ON as it passes the basic CI.

PR #5317, opened 2024-02-13
Fix inconsistent calls to nvml::Init and nvml::Shutdown

JanuszL:
- in the worker thread and thread poll the nvml is called
  only for non-CPU pipelines but the shutdown is called unconditionally

## Category:

## Description:
- in the worker thread and thread poll the nvml is called
  only for non-CPU pipelines but the shutdown is called unconditionally

- thread_pool
- worker thread

- NA

### Tests:

- [x] Existing tests apply
- [ ] New tests added
  - [ ] Python tests
  - [ ] GTests
  - [ ] Benchmark
  - [ ] Other
- [ ] N/A

### Documentation
- [ ] Documentation updated
  - [ ] Docstring
  - [ ] Doxygen
  - [ ] RST
  - [ ] Jupyter
  - [ ] Other
- [x] N/A

#### Requirements
- [x] N/A

**REQ IDs**: N/A

**JIRA TASK**: N/A


dali/util/nvml.h
```diff
@@ -241,6 +241,43 @@ inline void Shutdown() {
   CUDA_CALL(nvmlShutdown());
 }
 
+
+class nvmlHandle {
```
mzient, line 245: We have NvmlError and NvmlBadAlloc, so I guess it'd be good to have Nvml... something.
I'm not convinced we should call it NvmlHandle, since it's not a handle nor does it wrap any handle.
Other names to consider:
NvmlScope
NvmlInstance

JanuszL: Good point. Fixed.

dali/core/mm/malloc_resource.cc
```diff
@@ -94,9 +94,8 @@ cuda_malloc_async_memory_resource::cuda_malloc_async_memory_resource(int device_
   dummy_host_stream_ = CUDAStreamPool::instance().Get(device_id_);
 #if NVML_ENABLED
   static const float driverVersion = []() {
-    nvml::Init();
+    auto nvml_handle = nvml::NvmlInstance::CreateNvmlHandle();
```
stiepan, line 97:
```suggestion
    auto nvml_handle = nvml::NvmlInstance::CreateNvmlInstance();
```
Not full search and replace? :)

JanuszL: Missed that. Fixed.

dali/util/nvml.h
```diff
@@ -241,6 +241,43 @@ inline void Shutdown() {
...
+
+  inline NvmlInstance(NvmlInstance &&other) : is_created_(other.is_created_) {
+    other.is_created_ = false;
+  }
+
+  inline NvmlInstance &operator=(NvmlInstance &&other) {
+    is_created_ = other.is_created_;
+    other.is_created_ = false;
+    return *this;
+  }
```
stiepan, line 270: Isn't one Shutdown missing here?

```
NvmlInstance a{true};
a = NvmlInstance{true};
```
We take the ownership of the other's initialization, but forget ours?

JanuszL: Right. Fixed.. I hope.

dali/util/nvml.h
```diff
@@ -241,6 +241,43 @@ inline void Shutdown() {
...
+
+  NvmlInstance &operator=(const NvmlInstance &) = delete;
+
+  inline NvmlInstance(NvmlInstance &&other) : is_created_(other.is_created_) {
+    other.is_created_ = false;
+  }
+
+  inline NvmlInstance &operator=(NvmlInstance &&other) {
+    std::swap(is_created_, other.is_created_);
+    other.~NvmlInstance();
```
stiepan, line 268: In that case, I think the destructor should set the flag to false, otherwise it may Shutdown twice, with the expicit call and when the moved instance goes out of scope. Omitting the explict call to the destructor should be fine, right?

JanuszL: Ture that. Fixed.

mzient (approved):

stiepan (approved):

stiepan (approved):

PR #5331, opened 2024-02-19
[Cutmix] Make fn.multi_paste more flexible, fix validation

stiepan:
## Category:

## Description:
This PR reworks arguments and their parsing for fn.multi_paste to make it more flexible and easier to use (with cutmix augmentation in mind; namely ability to mix different batches and no need to use image shape explicitly if the inputs are uniformly shaped). It fixes a couple of bugs.

Fixes:
* There was no validation of in_ids leading to out-of-bound accesses for incorrect input
* The number of channels was handled incorrectly
  * GPU assumed 3 channels no matter the actual number of channels - leading to incorrect cuda mem accesses.
  * Both backends did not verify if all the pasted regions have the same number of channels (and if the output number of channels matches). The number of output channels was copied from the corresponding input sample. The PR changes that - the number of output channels is inferred from the actually pasted regions and uses the old behaviour only if there are none (to be compatible with the only case it could have worked previously).

New features:
* The in_anchors, region shapes and out_anchors have now relative counterparts - they can be specified as [0, 1] floats relative to input shape, input shape and output shape respectively.
* If all the input shapes are uniform and no output size is provided, the output shape is assumed to be the same.
* To allow mixing images of diffrent batches, the operator can accept multiple inputs - in that case the in_ids must not be specified and the regions are pasted elementwise.

The first two points are aimed to make it easier to use fn.multi_paste without explcitly handling the actual shapes of the samples. 
The multi-input mode should make it easier to use the operator in simple applications where the number of paste regions is uniform - with DALI implict batch you can think in terms of samples rather than indicies in the implicit batch. And it enables cases were you need to mix images from different sources.

Other changes:
* The args are parsed once and saved not to recompute them along the way
* No intersection check is moved to the CPU impl, it's outputs were ignored by the GPU impl anyway.
* Added video support.

cutmix

### Tests:

- [ ] Existing tests apply
- [x] New tests added
  - [x] Python tests
  - [ ] GTests
  - [ ] Benchmark
  - [ ] Other
- [ ] N/A

### Documentation
- [x] Documentation updated
  - [x] Docstring
  - [ ] Doxygen
  - [ ] RST
  - [ ] Jupyter
  - [ ] Other
- [ ] N/A

#### Requirements
- [ ] N/A

**REQ IDs**: N/A

**JIRA TASK**: DALI-3496

### DIFF
--- a/dali/kernels/imgproc/paste/paste_gpu.h
+++ b/dali/kernels/imgproc/paste/paste_gpu.h
@@ -1,4 +1,4 @@
-// Copyright (c) 2021, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+// Copyright (c) 2021-2024, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -38,6 +38,7 @@ template<class InputType, int ndims>
 struct PatchDesc {
   const InputType *in;
   int in_sample_idx;
+  int in_batch_idx;
   ivec<ndims> patch_start, patch_end, in_anchor, in_pitch;
 };
 
@@ -53,7 +54,7 @@ template <class OutputType, class InputType, int ndims>
 void FillPointers(span<SampleDescriptorGPU<OutputType, InputType, ndims - 1>> descs,
                   span<PatchDesc<InputType, ndims - 1>> patches,
                   const OutListGPU<OutputType, ndims> &out,
-                  const InListGPU<InputType, ndims> &in) {
+                  const span<InListGPU<InputType, ndims>> &ins) {
   int batch_size = descs.size();
 
   for (int i = 0; i < batch_size; i++) {
@@ -61,7 +62,7 @@ void FillPointers(span<SampleDescriptorGPU<OutputType, InputType, ndims - 1>> de
   }
 
   for (auto &p : patches) {
-    p.in = p.in_sample_idx < 0 ? nullptr : in.data[p.in_sample_idx];
+    p.in = p.in_sample_idx < 0 ? nullptr : ins[p.in_batch_idx].data[p.in_sample_idx];
   }
 }
 
@@ -73,9 +74,9 @@ void FillPointers(span<SampleDescriptorGPU<OutputType, InputType, ndims - 1>> de
  */
 template <class OutputType, class InputType, int ndims>
 void CreateSampleDescriptors(
-    vector<SampleDescriptorGPU<OutputType, InputType, ndims - 1>> &out_descs,
-    vector<PatchDesc<InputType, ndims - 1>> &out_patches,
-    const TensorListShape<ndims> &in_shape,
+    std::vector<SampleDescriptorGPU<OutputType, InputType, ndims - 1>> &out_descs,
+    std::vector<PatchDesc<InputType, ndims - 1>> &out_patches,
+    const span<TensorListShape<ndims>> &in_shapes,
     span<paste::MultiPasteSampleInput<ndims - 1>> samples) {
   static_assert(ndims == 3, "Only 2D data with channels supported");
 
@@ -174,8 +175,11 @@ void CreateSampleDescriptors(
           patch.patch_end[1] = scaled_x_to_x[x + 1];
           patch.in = nullptr;  // to be filled later
           patch.in_sample_idx = max_paste == -1 ? -1 : sample.inputs[max_paste].in_idx;
+          patch.in_batch_idx = max_paste == -1 ? -1 : sample.inputs[max_paste].batch_idx;
           patch.in_pitch[0] = 0;
-          patch.in_pitch[1] = max_paste == -1 ? -1 : in_shape[sample.inputs[max_paste].in_idx][1];
+          int mp_in_idx = sample.inputs[max_paste].in_idx;
+          int mp_batch_idx = sample.inputs[max_paste].batch_idx;
+          patch.in_pitch[1] = max_paste == -1 ? -1 : in_shapes[mp_batch_idx][mp_in_idx][1];
 
           if (max_paste != -1) {
             patch.in_anchor[0] = sample.inputs[max_paste].in_anchor[0] +
@@ -263,8 +267,8 @@ class PasteGPU {
       KernelContext &context,
       span<paste::MultiPasteSampleInput<spatial_dims>> samples,
       const TensorListShape<ndims> &out_shape,
-      const TensorListShape<ndims> &in_shape) {
-    paste::CreateSampleDescriptors(sample_descriptors_, patch_descriptors_, in_shape, samples);
+      const span<TensorListShape<ndims>> &in_shapes) {
+    paste::CreateSampleDescriptors(sample_descriptors_, patch_descriptors_, in_shapes, samples);
 
     KernelRequirements req;
     ScratchpadEstimator se;
@@ -283,8 +287,8 @@ class PasteGPU {
   void Run(
       KernelContext &context,
       const OutListGPU<OutputType, ndims> &out,
-      const InListGPU<InputType, ndims> &in) {
-    paste::FillPointers(make_span(sample_descriptors_), make_span(patch_descriptors_), out, in);
+      const span<InListGPU<InputType, ndims>> &ins) {
+    paste::FillPointers(make_span(sample_descriptors_), make_span(patch_descriptors_), out, ins);
 
     SampleDesc *samples_gpu;
     PatchDesc  *patches_gpu;
--- a/dali/kernels/imgproc/paste/paste_gpu.h
+++ b/dali/kernels/imgproc/paste/paste_gpu.h
@@ -174,14 +174,17 @@ void CreateSampleDescriptors(
           patch.patch_end[0] = scaled_y_to_y[y + 1];
           patch.patch_end[1] = scaled_x_to_x[x + 1];
           patch.in = nullptr;  // to be filled later
-          patch.in_sample_idx = max_paste == -1 ? -1 : sample.inputs[max_paste].in_idx;
-          patch.in_batch_idx = max_paste == -1 ? -1 : sample.inputs[max_paste].batch_idx;
           patch.in_pitch[0] = 0;
-          int mp_in_idx = sample.inputs[max_paste].in_idx;
-          int mp_batch_idx = sample.inputs[max_paste].batch_idx;
-          patch.in_pitch[1] = max_paste == -1 ? -1 : in_shapes[mp_batch_idx][mp_in_idx][1];
-
-          if (max_paste != -1) {
+          if (max_paste == -1) {
+            patch.in_sample_idx = -1;
+            patch.in_batch_idx = -1;
+            patch.in_pitch[1] = -1;
+          } else {
+            patch.in_sample_idx = sample.inputs[max_paste].in_idx;
+            patch.in_batch_idx = sample.inputs[max_paste].batch_idx;
+            int mp_in_idx = sample.inputs[max_paste].in_idx;
+            int mp_batch_idx = sample.inputs[max_paste].batch_idx;
+            patch.in_pitch[1] = in_shapes[mp_batch_idx][mp_in_idx][1];
             patch.in_anchor[0] = sample.inputs[max_paste].in_anchor[0] +
                               patch.patch_start[0] - sample.inputs[max_paste].out_anchor[0];
             patch.in_anchor[1] = sample.inputs[max_paste].in_anchor[1] +
--- a/dali/operators/image/paste/multipaste.cc
+++ b/dali/operators/image/paste/multipaste.cc
@@ -1,4 +1,4 @@
-// Copyright (c) 2021-2022, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+// Copyright (c) 2021-2024, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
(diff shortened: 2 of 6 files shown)


stiepan:
!build

stiepan:
!build

dali/operators/image/paste/multipaste.cc
```diff
@@ -19,44 +19,83 @@
...
+of the destination region.
+
+Works like `out_anchors` argument, but the values should be floats in `[0, 1]` range,
+describing the LU corner of the pasted region relative to the output shape.)code",
+                           nullptr, true, true)
+    .AddOptionalArg<std::vector<int>>("output_size",
+                                      R"code(A tuple (HW) describing the output shape.
+
+Can be omitted if the operator is run with inputs of uniform shape. In that case,
+the same shape is used as the output shape.)code",
```
klecki, line 95: We can mention `canvas` here, and discuss the output in its terms.

dali/kernels/imgproc/paste/paste_gpu.h
```diff
@@ -174,8 +175,11 @@ void CreateSampleDescriptors(
           patch.patch_end[1] = scaled_x_to_x[x + 1];
           patch.in = nullptr;  // to be filled later
           patch.in_sample_idx = max_paste == -1 ? -1 : sample.inputs[max_paste].in_idx;
+          patch.in_batch_idx = max_paste == -1 ? -1 : sample.inputs[max_paste].batch_idx;
           patch.in_pitch[0] = 0;
-          patch.in_pitch[1] = max_paste == -1 ? -1 : in_shape[sample.inputs[max_paste].in_idx][1];
+          int mp_in_idx = sample.inputs[max_paste].in_idx;
+          int mp_batch_idx = sample.inputs[max_paste].batch_idx;
```
klecki, line 181: Just a line above you are comparing the `max_paste == -1` and here you are indexing into inputs, unless you did offset the pointer by one for this case, this might be wrong.

dali/kernels/imgproc/paste/paste_gpu.h
```diff
@@ -174,8 +175,11 @@ void CreateSampleDescriptors(
...
+          patch.in_pitch[1] = max_paste == -1 ? -1 : in_shapes[mp_batch_idx][mp_in_idx][1];
```
klecki, line 182: Btw, what kind of pitch is that and why do we need to take it from the last pasted sample as far as I understand?

klecki (commented):
Before looking at tests, posting the comments

dali/operators/image/paste/multipaste.h
```diff
@@ -210,94 +428,83 @@ class MultiPasteOp : public StatelessOperator<Backend> {
 
   bool SetupImpl(std::vector<OutputDesc> &output_desc,
                  const Workspace &ws) override {
-    const auto &images = ws.Input<Backend>(0);
-    int ndim = images.sample_dim();
-    DALI_ENFORCE(ndim == 3, "MultiPaste supports only 2D data with channels (HWC)");
+    int ndim;
+    multipaste_utils::GetValidateInputTypeAndDim<Backend>(ndim, input_type_, ws);
+    DALI_ENFORCE(ndim == 3, make_string("Unsupported sample dimensionality, got ", ndim,
+                                        ", expected 3 (2D data with channels - HWC)."));
```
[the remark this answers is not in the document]
stiepan: Exactly.

dali/kernels/imgproc/paste/paste_gpu.h
stiepan: oops, right, it's an artifact of hoisting this up from the inline if expr. good catch!

dali/kernels/imgproc/paste/paste_gpu.h
stiepan: My understanding is that we split the image into non-overlapping recatngular patches, where each patch has a single source and for each sample we take the top-most sample, the pithc is the row stride. At least from a glimpse into the code.

dali/operators/image/paste/multipaste.h
```diff
@@ -87,112 +115,302 @@ class MultiPasteOp : public StatelessOperator<Backend> {
...
+  }
+
+  // Get the num channels for the output sample from the pasted regions
+  // or if there are none for given sample
+  inline int InferChannelsNum(const Workspace &ws, int sample_num) {
+    int num_channels = -1;
+    if (!in_idx_.HasExplicitValue()) {
+      num_channels = ws.Input<Backend>(0)[sample_num].shape()[spatial_ndim_];
+      for (int i = 1; i < ws.NumInput(); i++) {
+        int input_channels = ws.Input<Backend>(i)[sample_num].shape()[spatial_ndim_];
```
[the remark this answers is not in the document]
stiepan: I though about it, but I forgot we have those cool lightweight views with type erasure and assumed it would call for making sure it's all in type-switch.

dali/operators/image/paste/multipaste.h
```diff
@@ -87,112 +115,302 @@ class MultiPasteOp : public StatelessOperator<Backend> {
...
+    const auto validateArgShape = [spatial_ndim = spatial_ndim_](int sample_idx, int paste_count,
+                                                                 const std::string &arg_name,
+                                                                 const TensorShape<> arg_shape) {
+      int num_args = arg_shape[0];
+      int arg_len = arg_shape[1];
+      DALI_ENFORCE(
+          num_args == paste_count && arg_len == spatial_ndim,
+          make_string("Unexpected shape for argument `", arg_name, "` for output sample at index ",
+                      sample_idx, ". It should be a 2D tensor of shape [number of pasted regions]",
+                      "x2 (i.e. ", paste_count, "x", spatial_ndim,
```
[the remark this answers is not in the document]
stiepan: image

dali/operators/image/paste/multipaste.h
```diff
@@ -87,112 +115,302 @@ class MultiPasteOp : public StatelessOperator<Backend> {
...
+      auto sh_view = subtensor(shapes_[sample_idx], paste_idx);
+      for (int d = 0; d < spatial_ndim_; d++) {
+        region_shape[d] = sh_view.data[d];
+      }
+    } else if (shapes_rel_.HasExplicitValue()) {  // scale the source_shape
+      auto shape_rel = subtensor(shapes_rel_[sample_idx], paste_idx);
+      for (int d = 0; d < spatial_ndim_; d++) {
+        float scale = shape_rel.data[d];
+        DALI_ENFORCE(0.f <= scale && scale <= 1.f,
+                     make_string("The `shapes_rel` must be in [0, 1] range, got ", scale, "."));
```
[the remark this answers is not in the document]
stiepan: Added one.

dali/operators/image/paste/multipaste.h
```diff
@@ -87,112 +115,302 @@ class MultiPasteOp : public StatelessOperator<Backend> {
...
+          make_string("The pasted region must be within input sample. Got input anchor: ",
+                      in_anchor[k], ", pasted region shape: ", region_shape[k],
+                      ", input shape: ", region_source_shape[k], ", for the region at index ",
+                      paste_idx, ", pasted into output sample at index ", sample_idx, "."));
+      DALI_ENFORCE(
+          out_anchor[k] >= 0 && out_anchor[k] + region_shape[k] <= out_shape[k],
+          make_string("The pasted region must be within output bounds. Got output anchor: ",
+                      out_anchor[k], ", pasted region shape: ", region_shape[k],
+                      ", output shape: ", out_shape[k], ", for the region at index ", paste_idx,
+                      ", pasted into output sample at index ", sample_idx, "."));
```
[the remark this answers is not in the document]
stiepan: Sure, I error out with the sample shape now.

dali/operators/image/paste/multipaste.h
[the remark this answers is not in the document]
stiepan: Added GetPasteSample.

dali/operators/image/paste/multipaste.h
```diff
@@ -15,30 +15,55 @@
...
+  type = images.type();
+  for (int i = 1; i < ws.NumInput(); i++) {
+    int sample_dim = ws.Input<Backend>(i).sample_dim();
+    auto sample_type = ws.Input<Backend>(i).type();
+    DALI_ENFORCE(ndim == sample_dim,
+                 make_string("All input batches must have the same dimensionality. Got ", ndim,
+                             " and ", sample_dim, " at indicies 0 and ", i, "."));
+    DALI_ENFORCE(type == sample_type,
+                 make_string("All input batches must have the same data type. Got ", images.type(),
+                             " and ", type, " at indicies 0 and ", i, "."));
```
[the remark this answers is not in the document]
stiepan: Fixed.

dali/operators/image/paste/multipaste.cc
```diff
@@ -68,64 +107,67 @@ void MultiPasteCPU::SetupTyped(const Workspace & /*ws*/,
...
+            auto tvin = in_views[sample_input][sample_idx];
+            auto tvout = out_view[i];
+
+            Coords region_shape_view{&region_shapes_data_[i][iter][0], coords_sh_};
+            Coords in_anchor_view{&in_anchors_data_[i][iter][0], coords_sh_};
+            Coords out_anchor_view{&out_anchors_data_[i][iter][0], coords_sh_};
             kernel_manager_.Run<Kernel>(
-                    to_sample, ctx, tvout, tvin,
+                    i, ctx, tvout, tvin,
                     in_anchor_view, region_shape_view, out_anchor_view);
```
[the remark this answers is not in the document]
stiepan: Moved it to a helper.

dali/operators/image/paste/multipaste.cc
stiepan: Sure, I use canvas when referring to output shape now.

dali/operators/image/paste/multipaste.cc
```diff
@@ -19,44 +19,86 @@
...
-If not provided, all anchors are zero.)code", nullptr, true)
-.AddArg("output_size",
-R"code(Shape of the output.)code", DALI_INT_VEC, true)
-.AddOptionalTypeArg("dtype",
-R"code(Output data type. If not set, the input type is used.)code")
-.NumOutput(1);
+In neither `out_anchors` nor `out_anchors_rel` are provided, all anchors are zero,
+making all the pasted region start at the top-left corner of the output canvas.)code",
+                         nullptr, true, true)
+    .AddOptionalArg<float>("out_anchors_rel", R"code(Absolute coordinates of LU corner
```
szkarpinski, line 86:
```suggestion
    .AddOptionalArg<float>("out_anchors_rel", R"code(Relative coordinates of LU corner
```

dali/operators/image/paste/multipaste.cc
```diff
@@ -19,44 +19,86 @@
...
+                         nullptr, true, true)
+    .AddOptionalArg<float>("shapes_rel", R"code(Relative shape of the paste regions.
+
+Works like `shape` argument, but the values should be floats in `[0, 1]` range,
+describing the paste region shape relative to the input shape.)code",
+                           nullptr, true, true)
+    .AddOptionalArg<int>("out_anchors", R"code(Absolute coordinates of LU corner
+of the pasted region in the output canvas.
 
 The anchors are represented as 2D tensors where the first dimension corresponds to the
```
szkarpinski, line 79:
```suggestion
The anchors are represented as 2D tensors where the first dimension is equal to the
```
Before it was: _corresponds to the elements of (...)_, but now it doesn`t _correspond_ to the number of pasted regions, but it's equal to it. Right?

dali/operators/image/paste/multipaste.cc
```diff
@@ -19,44 +19,86 @@
...
+number of pasted regions and the second one is equal to the number of dimensions of the
 data, excluding channels.
 
-If not provided, all anchors are zero.)code", nullptr, true)
-.AddArg("output_size",
-R"code(Shape of the output.)code", DALI_INT_VEC, true)
-.AddOptionalTypeArg("dtype",
-R"code(Output data type. If not set, the input type is used.)code")
-.NumOutput(1);
+In neither `out_anchors` nor `out_anchors_rel` are provided, all anchors are zero,
```
szkarpinski, line 83:
```suggestion
If neither `out_anchors` nor `out_anchors_rel` are provided, all anchors are zero,
```

dali/operators/image/paste/multipaste.cc
```diff
@@ -19,44 +19,86 @@
...
+making all the pasted region start at the top-left corner of the output canvas.)code",
```
szkarpinski, line 84:
```suggestion
making all the pasted regions start at the top-left corner of the output canvas.)code",
```

dali/operators/image/paste/multipaste.cu
```diff
@@ -25,61 +25,70 @@ inline void to_vec(vec<n, T> &out, Source &&src) {
     out[i] = src[i];
 }
 
-void MultiPasteGPU::InitSamples(const TensorListShape<> &out_shape) {
+void MultiPasteGPU::InitSamples(const Workspace &ws, const TensorListShape<> &out_shape) {
   assert(spatial_ndim_ == 2);
+  assert(out_shape.sample_dim() == 3);
```
szkarpinski, line 30: `DALI_ENFORCE`?

dali/operators/image/paste/multipaste.h
```diff
@@ -87,112 +115,295 @@ class MultiPasteOp : public StatelessOperator<Backend> {
...
   }
 
-  void AcquireArguments(const OpSpec &spec, const Workspace &ws) {
-    const auto &images = ws.Input<Backend>(0);
+  void ValidateFactor(float factor, const std::string &arg_name, int out_sample_idx,
+                      int paste_idx) {
+    DALI_ENFORCE(0.f <= factor && factor <= 1.f,
+                 make_string("The `", arg_name, "` must be float in [0, 1] range, got ", factor,
+                             " for region at index ", paste_idx,
+                             " pasted into output sampe at index ", out_sample_idx, "."));
```
szkarpinski, line 134:
```suggestion
                             " pasted into output sample at index ", out_sample_idx, "."));
```

dali/operators/image/paste/multipaste.h
```diff
@@ -87,112 +115,295 @@ class MultiPasteOp : public StatelessOperator<Backend> {
...
+    }
+
+    TensorShape<2> ref_shape = in_shape[0].first(spatial_ndim_);
+    // validate sources of the shape
+    if (output_size_.HasExplicitValue()) {
+      output_size_.Acquire(spec_, ws, nsamples, ArgValue_EnforceUniform);
+      for (int sample_idx = 0; sample_idx < nsamples; sample_idx++) {
+        const auto &sample_out_shape = output_size_[sample_idx];
+        DALI_ENFORCE(sample_out_shape.num_elements() >= spatial_ndim_,
+                     make_string("The `output_size` should be a tuple (HW), got ",
```
szkarpinski, line 191:
```suggestion
                     make_string("The `output_size` should be a tuple (H, W), got ",
```
Nitpick, I just like it more when tuple notation is used for tuple-like

dali/operators/image/paste/multipaste.h
```diff
@@ -87,112 +115,295 @@ class MultiPasteOp : public StatelessOperator<Backend> {
...
+                                 sample_out_shape.num_elements(),
+                                 " elements for a sample at index ", sample_idx, "."));
+        for (int d = 0; d < spatial_ndim_; d++) {
+          DALI_ENFORCE(
+              sample_out_shape.data[d] >= 0,
+              make_string("The `output_size` must be non-negative, got ", sample_out_shape.data[d],
+                          " for a sample at index ", sample_idx, "."));
```
szkarpinski, line 198: Nitpick: maybe also tell which dimension it is?

dali/operators/image/paste/multipaste.cu
stiepan: I went for an assert because we validate the input shape to have ndim == 3 and set the output shape ourseleves requiring the same dimensionality, so it should be met. Or is there a gap?

dali/operators/image/paste/multipaste.h
stiepan: How about printing the whole shape? I think it should be easier to pin-point than decoding in head the dimension idx.

dali/operators/image/paste/multipaste.h
```diff
@@ -480,29 +473,31 @@ class MultiPasteOp : public SequenceOperator<Backend, StatelessOperator> {
...
-      }
+      return ws.Input<Backend>(paste_idx)[out_sample_idx];
+    }
+  }
+
+  inline ivec2 GetInputShape(const Workspace &ws, int out_sample_idx, int paste_idx) const {
+    ivec2 shape;
+    const auto &sample = GetPasteSample(ws, out_sample_idx, paste_idx);
+    const auto& sample_shape = sample.shape();
+    DALI_ENFORCE(sample_shape.size() >= spatial_ndim_);
```
stiepan, line 498: should be an assert, missed that.

dali/operators/image/paste/multipaste.cc
stiepan: Equal is less ambigious, I'll adjust it. But functionally, it's the same before and now (just with the multi-input you can't have variadic number of regions per output).

dali/operators/image/paste/multipaste.h
szkarpinski: Good idea

dali/operators/image/paste/multipaste.cu
szkarpinski: Makes sense, thanks

klecki (approved):

szkarpinski (approved):

stiepan:
!build

dali/operators/image/paste/multipaste.h
```diff
@@ -87,112 +115,302 @@ class MultiPasteOp : public StatelessOperator<Backend> {
...
-                     spatial_ndim_, ", got ", out_anchors_[i].shape[1]));
+        validateArgShape(i, n_paste, "out_anchors", out_anchors_[i].shape);
+      } else if (out_anchors_rel_.HasExplicitValue()) {
+        validateArgShape(i, n_paste, "out_anchors_rel", out_anchors_rel_[i].shape);
       }
 
-      bool found_intersection = false;
+      in_anchors_data_[i].resize(n_paste);
+      region_shapes_data_[i].resize(n_paste);
+      out_anchors_data_[i].resize(n_paste);
```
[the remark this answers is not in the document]
stiepan: I tried replacing it with the TensorListView to have the var sample sizes handled, but if anything, it performed sligthly worse end-to-end.

stiepan:
!build

PR #5338, opened 2024-02-27
Refactor custom ops classes to use python_op_factory as base class

klecki:
## Description:
`python_op_factory` was extended and documented. There is a new, optional `internal_schema_name` parameter, that will be used to retrieve schema and spec for argument handling on Python side, while allowing the original `schema` to be used for the purpose of exposing the documentation.

As both `TFRecordReader` and `PythonFunction` has several variants with different base implementation schemas, the base classes are converted into a class generator function. The new classes are marked as not generated, as the type hints are done manually for them.

This reduces the custom code to minimum, allowing the base class to handle all properties and some common arguments.
The argument handling is moved to base class by updating the kwargs before invoking the base class.

When a `tfrec.Feature` is encountered in the Python argument serialization layer, it is again processed by the `tfrec.Feature` constructor (as we do for other types, normalizing numbers with int() or float()). For this purpose a copy constructor is added and exposed in Pybind. An alternative would be to change the conversion to an identity function - that would keep the error in the spec.AddArg rather than moving it to constructor parameter matching.
Validation code is added to the operator, to check for matching type with better error message.

NumbaFunction was also reworked, inheriting directly from the python_op_factory.

Calls to `_raise_no_current_pipeline` were eliminated - this function doesn't exist!

Breaking change: There is no longer a `PythonFunctionBase` class in `nvidia.dali.ops`.

TFRecord, Python Function and Numba Function operators in Python

Does it has a potential to break something?
Does the documentation render correctly?
What would be a cleaner way to have schema and internal schema? - otherwise we would need to rewrite the operators to not use two schemas - one for presentation and one for implementation.
Should we keep implementation detail of `PythonFunctionBase` alive? 

As a followup a code that prohibits MIS would be a nice generalization for those custom implementations.

### Tests:

- [x] Existing tests apply
   All TFRecord, Python function and Numba Function tests must pass
- [x] New tests added
  - [x] Python tests - new type error checking for TFRecord is tested.
  - [ ] GTests
  - [ ] Benchmark
  - [ ] Other
- [ ] N/A

### Documentation
- [ ] Documentation updated
  - [ ] Docstring
  - [ ] Doxygen
  - [ ] RST
  - [ ] Jupyter
  - [ ] Other
- [ ] N/A

#### Requirements
- [ ] N/A

**REQ IDs**: N/A

**JIRA TASK**: N/A

### DIFF
--- a/dali/python/nvidia/dali/ops/__init__.py
+++ b/dali/python/nvidia/dali/ops/__init__.py
@@ -517,10 +517,25 @@ def _check_arg_input(schema, op_name, name):
         )
 
 
-def python_op_factory(name, schema_name=None):
+def python_op_factory(name, schema_name, internal_schema_name=None):
+    """Generate the ops API class bindings for operator.
+
+    Parameters
+    ----------
+    name : str
+        The name of the operator (without the module) - this will be the name of the class
+    schema_name : str
+        Name of the schema, used for documentation lookups and schema/spec retrieval unless
+        internal_schema_name  is provided
+    internal_schema_name : str, optional
+        If provided, this will be the schema used to process the arguments, by default None
+    """
     class Operator(metaclass=_DaliOperatorMeta):
         def __init__(self, *, device="cpu", **kwargs):
-            schema_name = _schema_name(type(self))
+            if self._internal_schema_name is None:
+                schema_name = _schema_name(type(self))
+            else:
+                schema_name = self._internal_schema_name
             self._spec = _b.OpSpec(schema_name)
             self._schema = _b.GetSchema(schema_name)
 
@@ -611,7 +626,8 @@ def python_op_factory(name, schema_name=None):
             return result
 
     Operator.__name__ = str(name)
-    Operator.schema_name = schema_name or Operator.__name__
+    Operator.schema_name = schema_name
+    Operator._internal_schema_name = internal_schema_name
     Operator._generated = True  # The class was generated using python_op_factory
     Operator.__call__.__doc__ = _docs._docstring_generator_call(Operator.schema_name)
     return Operator
@@ -763,7 +779,6 @@ _internal._adjust_operator_module(ExternalSource, sys.modules[__name__], [])
 
 # Expose the PythonFunction family of classes and generate the fn bindings for them
 from nvidia.dali.ops._operators.python_function import (  # noqa: E402, F401
-    PythonFunctionBase,  # noqa: F401
     PythonFunction,
     DLTensorPythonFunction,
     _dlpack_to_array,  # noqa: F401
--- a/dali/python/nvidia/dali/ops/__init__.py
+++ b/dali/python/nvidia/dali/ops/__init__.py
@@ -662,10 +662,6 @@ def _load_readers_tfrecord():
     if not tfrecord.tfrecord_enabled():
         return
 
-    tfrecord._TFRecordReaderImpl.__call__.__doc__ = _docs._docstring_generator_call(
-        "readers__TFRecord"
-    )
-
     _registry.register_cpu_op("readers__TFRecord")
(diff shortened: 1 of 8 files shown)


klecki:
!build

klecki:
!build

klecki:
!build

klecki:
!build

klecki:
!build

klecki:
!build_nightly

JanuszL:
!build_nightly

dali/python/nvidia/dali/ops/_operators/python_function.py
```diff
@@ -31,63 +30,53 @@ def _setup_cupy():
...
-                    f"Expected inputs of type `DataNode`. "
-                    f"Received input of type '{type(inp).__name__}'. "
-                    f"Python Operators do not support Multiple Input Sets."
+def _get_base_impl(name, impl_name):
+
+    class PythonFunctionBase(ops.python_op_factory(impl_name, name, impl_name)):
+
+        def __init__(self, function, num_outputs=1, **kwargs):
+
+            # The layout need to be handled manually due to implementation detail
```
mzient, line 39:
```suggestion
            # The layouts need to be handled manually due to implementation detail
```

dali/python/nvidia/dali/ops/__init__.py
```diff
@@ -494,10 +494,28 @@ def _check_arg_input(schema, op_name, name):
...
+    ----------
+    name : str
+        The name of the operator (without the module) - this will be the name of the class
+    schema_name : str
+        Name of the schema, used for documentation lookups and schema/spec retrieval unless
+        internal_schema_name  is provided
+    internal_schema_name : str, optional
+        If provided, this will be the schema used to process the arguments, by default None
+        If it is not provided, the class is assumed to be `_generated=True`, otherwise, we mark
+        it as False - the user will provide a custom wrapper.
```
mzient, line 510: Making a (formerly) well-defined property a side effect of some other condition is making the code hard to read and understand. I'd rather have an explicit argument or override this attribute manually for those few operators that need it to be `False`.

dali/python/nvidia/dali/ops/__init__.py
```diff
@@ -581,8 +599,11 @@ def __call__(self, *inputs, **kwargs):
             return result
 
     Operator.__name__ = str(name)
-    Operator.schema_name = schema_name or Operator.__name__
-    Operator._generated = True  # The class was generated using python_op_factory
+    Operator.schema_name = schema_name
+    Operator._internal_schema_name = internal_schema_name
+    # The class was generated using python_op_factory, and we don't expect custom wrapper.
+    # If needed, allow this tag to be overridden by an argument to this function
+    Operator._generated = internal_schema_name is None
```
mzient, line 606: 1. What I've written at the declaration.
2. Now that this attribute is always there (?) we should probably get rid of those `getattr(op, "_generated", None)`

dali/python/nvidia/dali/ops/_operators/python_function.py
klecki: done

dali/python/nvidia/dali/ops/__init__.py
klecki: This still leaves the External Source without the use of python_op_factory, maybe I can look at it as a follow-up, but the ExternalSourceGroup captures the OperatorInstances directly, so this whole abstraction is not usable there at this point in time.

dali/python/nvidia/dali/ops/__init__.py
klecki: Done

klecki:
!build

mzient (approved):

klecki:
!build

PR #5340, opened 2024-02-28
Handle arguments more strictly

TheodoreEhrenborg:
## Category:

## Description:

This code looks like it validates that the `auto_reset` argument is `True`, `False`, `"yes"`, `"no"`, or `None`:
```python
        if not auto_reset or auto_reset is None or auto_reset == "no":
            self._auto_reset = "no"
        elif auto_reset or auto_reset == "yes":
            self._auto_reset = "yes"
        else:
            raise ValueError(f"Unsupported value for `auto_reset` {auto_reset}")
```
But in fact the `else` clause will never trigger, because all falsy values will trigger the first branch, and all truthy values (except `"no"`) will trigger the second branch. So the `else` clause is dead code.

And if the user provides `"False"` or `"No"`, the code will silently set `self._auto_reset` to `"yes"`

This PR limits acceptable values to `True`, `False`, `"yes"`, `"no"`, or `None`

### Tests:

- [ ] Existing tests apply
- [ ] New tests added
  - [ ] Python tests
  - [ ] GTests
  - [ ] Benchmark
  - [ ] Other
- [ ] N/A

### Documentation
- [ ] Documentation updated
  - [ ] Docstring
  - [ ] Doxygen
  - [ ] RST
  - [ ] Jupyter
  - [ ] Other
- [ ] N/A

#### Requirements
- [ ] N/A

**REQ IDs**: N/A

**JIRA TASK**: N/A

### DIFF
--- a/dali/python/nvidia/dali/plugin/base_iterator.py
+++ b/dali/python/nvidia/dali/plugin/base_iterator.py
@@ -88,7 +88,7 @@ class _DaliBaseIterator(object):
                 * ``"no"``, ``False`` or ``None`` - at the end of epoch StopIteration is raised and
                   reset() needs to be called. Calling ``iter()`` on the iterator would reset
                   it as well.
-                * ``"yes"`` or ``"True"``- at the end of epoch StopIteration is raised but reset()
+                * ``"yes"`` or ``True``- at the end of epoch StopIteration is raised but reset()
                   is called internally automatically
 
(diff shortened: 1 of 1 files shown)


mzient (approved):

JanuszL:
Hi @TheodoreEhrenborg,

Nice catch. Thank you for our PR.

PR #5345, opened 2024-02-29
Map backend exceptions into Python exception types

klecki:
## Description:
Introduce DaliError and it subtypes that map directly into Python exception types.
Introduce exception translator in Pybind for that purpose. 
Thanks to that we don't pull pybind or Python into DALI backend code, we just transform 
the exceptions when they cross the boundary from backend to frontend.

Add code unifying the exception translation using std::exception_ptr.
Capture context (with preparation for providing source location in pipeline definition) 
and additional message to be added to the thrown exception.
Additionally handle c++ exception that translate to non-trivial Python exception types in pybind
by default, rather than uniformly translate them to runtime_error.

Simplify a bit of message creation, so we add the "Unknown critical error" in one place only,
and we can capture the exception for processing everywhere with `catch (...)`.

This PR focuses on the translation mechanism, while mostly preserving the current error messages.

Executor/Pipeline exception handling. 
Actual exception types raised in Python will change.

Some parts duplicate other PRs under review.

Breaking change: some error types and error messages might change.

### Tests:

- [x] Existing tests apply
- [x] New tests added
  - [x] Python tests - operator that triggers exceptions.
  - [ ] GTests
  - [ ] Benchmark
  - [ ] Other
- [ ] N/A

### Documentation
- [x] Documentation updated
  - [ ] Docstring
  - [x] Doxygen
  - [ ] RST
  - [ ] Jupyter
  - [ ] Other
- [ ] N/A

#### Requirements
- [ ] N/A

**REQ IDs**: N/A

**JIRA TASK**: N/A

### DIFF
--- a/dali/pipeline/operator/error_reporting.h
+++ b/dali/pipeline/operator/error_reporting.h
@@ -18,6 +18,7 @@
 #include <exception>
 #include <stdexcept>
 #include <string>
+#include <utility>
 #include <vector>
 
 #include "dali/core/api_helper.h"
--- a/dali/pipeline/operator/error_reporting.h
+++ b/dali/pipeline/operator/error_reporting.h
@@ -15,8 +15,14 @@
 #ifndef DALI_PIPELINE_OPERATOR_ERROR_REPORTING_H_
 #define DALI_PIPELINE_OPERATOR_ERROR_REPORTING_H_
 
+<<<<<<< HEAD
 #include <string>
 #include <utility>
+=======
+#include <exception>
+#include <stdexcept>
+#include <string>
+>>>>>>> ffa6178c2 (Make everything a tad more generic)
 #include <vector>
 
 #include "dali/core/api_helper.h"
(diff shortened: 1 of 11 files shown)


klecki:
!build

klecki:
!build

klecki:
!build

dali/pipeline/operator/error_reporting.h
```diff
@@ -0,0 +1,142 @@
...
+
+
+/**
+ * @brief Base error class for exceptions produced in the DALI pipeline. Subtypes should be used
+ * for specific errors that will be mapped into appropriate Python error types of matching names.
+ *
+ * DALI executor automatically extends those kind of errors when they are thrown from operators
+ * and the origin information is present.
+ */
+class DaliError : public std::exception {
```
mzient, line 48: We hardly ever use `Dali` in our source base (less than 10% of occurrences vs DALI - and that excludes macros).

dali/pipeline/operator/error_reporting.h
```diff
@@ -0,0 +1,142 @@
...
+ * IndexError
+ * ValueError
+ * TypeError
+ * StopIteration
+ */
+
+/**
+ * @brief Error class that will be mapped to Python RuntimeError
+ */
+class DaliRuntimeError : public DaliError {
```
mzient, line 103: This is a bit meh. It would be nice if it inherited from `std::runtime_error` - but that leaves us with one of the following:
- no common base
- common base not inheriting from `std::exception`
- virtual inheritance
none of which is pretty.

klecki: Yes, I was thinking about inheriting from the C++ counterparts, but the problems arise.
Not reporting them as std::runtime_error or std::out_of_range may be not ideal for someone using DALI via some sort of C++ API (we don't officially support that, but still).

klecki:
!build

szkarpinski (approved):

awolant (approved):

PR #5356, opened 2024-03-06
External context refactor

szkarpinski:
## Category:

**Refactoring** (*Redesign of existing code that doesn't affect functionality*)

## Description:
In this PR, I refactor `ExternalContext` to be more flexible. 

Instead of enforcing specific data format, just a bytes field for `pipeline_data` is there,  allowing `Pipeline` to populate it in any way. This is a preparation for allowing `ExternalContext` to keep arbitrary `iterator_data` alongside `pipeline_data`, which will improve and simplify iterator checkpointing.

#### What is ExternalContext?
ExternalContext is a structure used to pass data from the frontend to be saved in a checkpoint. 
Currently, it's used to pass Pipeline's epoch and iteration, but it will be extended in the future to store other frontend data (in particular, iterator's counters).

#### General changes
ExternalContext no longer stores `epoch_idx` and `iter` directly. Instead, a `bytes pipeline_data` is provided and the pipeline uses it this field to store its counters. Pipeline is responsible for encoding and decoding this data.

#### Changes in the pipeline
Pipeline now stores **a pickled dict** in `pipeline_data`, with two keys: `"epoch_idx"` and `"iter"`.

### Tests:

- [ ] Existing tests apply
- [ ] New tests added
  - [ ] Python tests
  - [ ] GTests
  - [ ] Benchmark
  - [ ] Other
- [ ] N/A

### Documentation
- [ ] Existing documentation applies
- [ ] Documentation updated
  - [ ] Docstring
  - [ ] Doxygen
  - [ ] RST
  - [ ] Jupyter
  - [ ] Other
- [ ] N/A

#### Requirements
- [ ] Implements new requirements
- [ ] Affects existing requirements
- [ ] N/A

**REQ IDs**: N/A

**JIRA TASK**: N/A

PR #5371, opened 2024-03-13
Improve iterator checkpointing

szkarpinski:
## Category:

## Description:

In this PR I explicitly save iterator data to the checkpoint and then use it to restore. Previously, I attempted to calculate iterator state based on pipeline state, which turned out to be a bad idea.

Iterator checkpointing now works as follows:
1. User calls `it.checkpoints()`
2. (!) The iterator creates a dict with values of its internal counters, pickles it and saves it to `pipe._iterator_data` in each of its pipelines.
3. Iterator calls `pipe.checkpoint()` for each of its pipelines (`pipe._iterator_data` is now saved in a checkpoint as well)
4. Later, user restores the pipelines and creates a new iterator.
5. (!) The iterator notices that the pipelines are restored, so it grabs `pipe._iterator_data` and restores its counters

This approach enables the following new features, which were problematic before:
- Checkpointing for `FILL` + `pad_last_batch=False`, which was not supported
- Reflecting `.reset()` in the checkpoint. Before, the fact that user has called `it.reset()` was not reflected in a checkpoint in any way, resulting in ambiguous restore.

`ExternalContext`, Base iterator, C API

### Tests:

- [ ] Existing tests apply
- [x] New tests added
  - [x] Python tests (`test_reset` to check if `it.reset()` is now checkpointed correctly)
  - [ ] GTests
  - [ ] Benchmark
  - [ ] Other
- [ ] N/A

### Documentation
- [ ] Documentation updated
  - [ ] Docstring
  - [ ] Doxygen
  - [ ] RST
  - [ ] Jupyter
  - [ ] Other
- [ ] N/A

#### Requirements
- [x] N/A

**REQ IDs**: N/A

**JIRA TASK**: DALI-3870

### DIFF
--- a/dali/c_api/c_api_test.cc
+++ b/dali/c_api/c_api_test.cc
@@ -1140,6 +1140,7 @@ TEST(CApiTest, CheckpointingTest) {
     restored_external_context.iterator_data.data,
     mock_external_context.iterator_data.data,
     mock_external_context.iterator_data.size), 0);
+  daliDestroyExternalContextCheckpoint(&mock_external_context);
   daliDestroyExternalContextCheckpoint(&restored_external_context);
 
   // Check the result of the new pipeline
--- a/dali/c_api/c_api_test.cc
+++ b/dali/c_api/c_api_test.cc
@@ -1103,11 +1103,16 @@ TEST(CApiTest, CheckpointingTest) {
 
   // Save the checkpoint
   daliExternalContextCheckpoint mock_external_context{};
-  const std::string data = "Hello world";
 
-  mock_external_context.pipeline_data.data = static_cast<char *>(daliAlloc(data.size()));
-  memcpy(mock_external_context.pipeline_data.data, data.c_str(), data.size());
-  mock_external_context.pipeline_data.size = data.size();
+  const std::string pipeline_data = "Hello world";
+  mock_external_context.pipeline_data.data = static_cast<char *>(daliAlloc(pipeline_data.size()));
+  memcpy(mock_external_context.pipeline_data.data, pipeline_data.c_str(), pipeline_data.size());
+  mock_external_context.pipeline_data.size = pipeline_data.size();
+
+  const std::string iterator_data = "Iterator data!";
+  mock_external_context.iterator_data.data = static_cast<char *>(daliAlloc(iterator_data.size()));
+  memcpy(mock_external_context.iterator_data.data, iterator_data.c_str(), iterator_data.size());
+  mock_external_context.iterator_data.size = iterator_data.size();
 
   char *cpt;
   size_t n;
@@ -1130,7 +1135,11 @@ TEST(CApiTest, CheckpointingTest) {
   EXPECT_EQ(strncmp(
     restored_external_context.pipeline_data.data,
     mock_external_context.pipeline_data.data,
-    restored_external_context.pipeline_data.size), 0);
+    mock_external_context.pipeline_data.size), 0);
+  EXPECT_EQ(strncmp(
+    restored_external_context.iterator_data.data,
+    mock_external_context.iterator_data.data,
+    mock_external_context.iterator_data.size), 0);
   daliDestroyExternalContextCheckpoint(&restored_external_context);
 
   // Check the result of the new pipeline
--- a/dali/python/nvidia/dali/pipeline.py
+++ b/dali/python/nvidia/dali/pipeline.py
@@ -282,6 +282,7 @@ class Pipeline(object):
         self._checkpoint = checkpoint
         self._prefetch_queue_depth = prefetch_queue_depth
         self._is_restored_from_checkpoint = False
+        self._iterator_data = None
         if type(prefetch_queue_depth) is dict:
             self._exec_separated = True
             if not exec_async:
@@ -956,6 +957,7 @@ class Pipeline(object):
                 for group in self._input_callbacks:
                     group.current_iter = pipeline_data["iter"]
                     group.current_sample = pipeline_data["iter"] * self._max_batch_size
+            self._iterator_data = external_ctx_cpt.iterator_data
             self._is_restored_from_checkpoint = True
 
     def build(self):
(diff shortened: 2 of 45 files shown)


dali/c_api/c_api_test.cc
```diff
@@ -1130,7 +1135,12 @@ TEST(CApiTest, CheckpointingTest) {
   EXPECT_EQ(strncmp(
     restored_external_context.pipeline_data.data,
     mock_external_context.pipeline_data.data,
-    restored_external_context.pipeline_data.size), 0);
+    mock_external_context.pipeline_data.size), 0);
+  EXPECT_EQ(strncmp(
+    restored_external_context.iterator_data.data,
+    mock_external_context.iterator_data.data,
+    mock_external_context.iterator_data.size), 0);
+  daliDestroyExternalContextCheckpoint(&mock_external_context);
```
szkarpinski, line 1143: Adding
```cpp
daliDestroyExternalContextCheckpoint(&mock_external_context);
```
fixes a leak which was there before, introduced with #5356

dali/python/nvidia/dali/pipeline.py
```diff
@@ -1568,6 +1570,7 @@ def checkpoint(self, filename=None):
         external_ctx_cpt.pipeline_data = pickle.dumps(
             {"iter": self._consumer_iter, "epoch_idx": self._epoch_idx}
         )
+        external_ctx_cpt.iterator_data = self._iterator_data or ""
```
stiepan, line 1573: It's a nitpick, but I'd rather make the iterator_data explicit.

It seems to me it would be easier to think of the `_iterator_data` the same way as about the `_checkpoint` - something static/readonly throught execution that describes the state we resumed from (and given instance will ever resume once). It's there seprately from _checkpoint as a necessary evil because it is the backend that can parse it out from the _checkpoint and, with the way most iterators are created, we don't want user to pass checkpoint twice.

But the fact that there is special private field in the pipeline to set from the iterator before calling the checkpoint method seems like one more unexpected twist in the (internal but still) API.

We could have `._checkpoint` and `checkpoint` methods with the later calling the first and the first accepting the iterator_data argument that is not present in the `checkpoint` method not to expose it in public API. It makes little difference to what's there now, but I think it's clearer how to use it properly (plus the pipeline._iterator_data does not evolve). It feels to me like less mental burden for the future developers. :)

dali/python/nvidia/dali/plugin/base_iterator.py
```diff
@@ -226,59 +227,60 @@ def __init__(
...
         """
-
-        if self._last_batch_policy == LastBatchPolicy.FILL and self._last_batch_padded is False:
-            raise NotImplementedError(
-                "Currently, checkpointing is not supported for iterators with "
-                + "last_batch_policy=FILL and last_batch_padded=False"
+        if not iterator_data:
+            logging.warning(
+                "Iterator data was not saved in the checkpoint. "
+                + "This iterator might produce unexpected results."
```
stiepan, line 266: nitpick

```suggestion
                "Iterator data was not saved in the checkpoint. "
                "This iterator might produce unexpected results."
```

dali/python/nvidia/dali/plugin/base_iterator.py
```diff
@@ -465,6 +467,9 @@ def checkpoints(self):
         if not self._ever_consumed:
             return self._initial_checkpoints
         else:
+            iterator_data = self._save_state()
+            for p in self._pipes:
+                p._iterator_data = iterator_data
             return [p.checkpoint() for p in self._pipes]
```
stiepan, line 473: Following the comment on readonly `p._iterator_data`

```suggestion
            iterator_data = self._save_state()
            return [p._get_checkpoint(iterator_data=iterator_data) for p in self._pipes]
```

dali/test/python/checkpointing/test_dali_checkpointing_fw_iterators.py
```diff
@@ -178,6 +178,7 @@ class DatasetConfig:
         (
             # (last_batch_policy, pad_last_batch)
             (LastBatchPolicy.FILL, True),
+            (LastBatchPolicy.FILL, False),
```
stiepan, line 181: Cool it's all there now!

stiepan (approved):

dali/python/nvidia/dali/pipeline.py
szkarpinski: That's a great idea, thank you! I changed it in https://github.com/NVIDIA/DALI/pull/5371/commits/4be32487c0ce79790a5327dee27b065ffe950e65 , is it ok now?

dali/python/nvidia/dali/plugin/base_iterator.py
```diff
@@ -226,59 +227,60 @@ def __init__(
...
         if self._enable_checkpointing:
-            if self._last_batch_policy == LastBatchPolicy.FILL and self._last_batch_padded is False:
-                raise NotImplementedError(
-                    "Currently, checkpointing is not supported for iterators with "
-                    + "last_batch_policy=FILL and last_batch_padded=False"
-                )
-
             # Precompute the initial checkpoints, to prevent any problems
             # related to the `prepare_first_batch` flag.
             self._initial_checkpoints = [p.checkpoint() for p in self._pipes]
```
stiepan, line 232: That one needs be adjusted too, not to skip iterator state if there's any, doesn't it?

Would moving it just before `self._restore_state(iterator_data)` and making it `self._initial_checkpoints = [p._get_checkpoint(iterator_data) for p in self._pipes]` work?

banasraf (approved):

dali/python/nvidia/dali/plugin/base_iterator.py
szkarpinski: That's a nice catch and we don't have a test for that. I'll fix this add a test with multiple save+restore

szkarpinski: Fixed in https://github.com/NVIDIA/DALI/pull/5371/commits/d13c0f6bfabfb5d5da97bd7205b36f7030cfc868

szkarpinski: Thank you!

szkarpinski:
!build

PR #5380, opened 2024-03-18
Add random choice CPU operator for 0D samples

klecki:
## Description:
New operator: `fn.random.choice`.
This implementation is limited to 0D sampling - scalar and 1D inputs only are accepted.
The operator is based on `numpy.random.Generator.choice` API,
Probabilities are required to sum close to 1.0. 

`RNGBase` is refactoring to meet new requirements:
* instead of the only input, the first optional input is now treated as shape-like parameter.
* more customization for `dtype` processing. 
* documentation of the customization points is introduced.

The followup PR will introduce baseline implementation for N-D inputs.

### Tests:

- [x] Existing tests apply
The new changes don't break old RNG operators.
- [x] New tests added
  - [ ] Python tests
  - [ ] GTests
  - [ ] Benchmark
  - [ ] Other
- [ ] N/A

### Documentation
- [x] Documentation updated
  - [x] Docstring
  - [x] Doxygen
  - [ ] RST
  - [ ] Jupyter
  - [ ] Other
- [ ] N/A

#### Requirements
- [ ] N/A

**REQ IDs**: N/A

**JIRA TASK**: N/A

### DIFF
--- a/dali/operators/random/choice.h
+++ b/dali/operators/random/choice.h
@@ -131,8 +131,8 @@ class Choice : public rng::RNGBase<Backend, Choice<Backend>, false> {
       ),  // NOLINT
       (
         DALI_FAIL("Data type ", dtype_, " is not supported for 0D inputs. "
-                  "Supported types are: ", ListTypeNames<DALI_CHOICE_0D_TYPES>(), ".");)
-      );
+                  "Supported types are: ", ListTypeNames<DALI_CHOICE_0D_TYPES>(), ".");
+      ));  // NOLINT
     }
     return input.tensor_shape_span(sample_idx)[0];
   }
--- a/dali/operators/random/choice.h
+++ b/dali/operators/random/choice.h
@@ -74,8 +74,7 @@ struct ChoiceSampleDist {
 
 /**
  * @brief Variant for generating indices. In case of 0D input, we treat those indices as the
- * actual sampled values in range [0, element_count). For n-D elements, where n > 1, we used the
- * result as the index of the element to be copied.
+ * actual sampled values in range [0, element_count).
  */
 template <typename T, bool uniform>
 struct ChoiceSampleDist<T, uniform, false> {
@@ -180,6 +179,10 @@ class Choice : public rng::RNGBase<Backend, Choice<Backend>, false> {
     }
   }
 
+  /**
+   * @brief We only allow sampling of the input elements or generating integral types if the
+   * input is scalar.
+   */
   void ValidateDataType(const OpSpec &spec, const Workspace &ws) const {
     if (ws.Input<CPUBackend>(0).sample_dim() != 0) {
       DALI_ENFORCE(ws.Input<CPUBackend>(0).type() == dtype_,
@@ -190,6 +193,9 @@ class Choice : public rng::RNGBase<Backend, Choice<Backend>, false> {
     }
   }
 
+  /**
+   * @brief Concatenate the requested shape with the shape of sampled elements.
+   */
   TensorListShape<> PostprocessShape(const OpSpec &spec, const Workspace &ws) {
     const auto &input = ws.Input<CPUBackend>(0);
     int element_dim = GetElementDim(input);
@@ -250,8 +256,8 @@ class Choice : public rng::RNGBase<Backend, Choice<Backend>, false> {
           BaseImpl::template RunImplTyped<T, Impl<T, true, false>>(ws);
         }
       ), (  // NOLINT
-        DALI_FAIL(make_string("Data type ", dtype_, " is not supported for 0D inputs. "
-                              "Supported types are: ", ListTypeNames<DALI_CHOICE_0D_TYPES>(), "."));
+        DALI_FAIL("Data type ", dtype_, " is not supported for 0D inputs. "
+                  "Supported types are: ", ListTypeNames<DALI_CHOICE_0D_TYPES>(), ".");
       ));  // NOLINT
     } else if (input.sample_dim() == 1) {
       TYPE_SWITCH(dtype_, type2id, T, (DALI_CHOICE_1D_TYPES), (
@@ -261,11 +267,12 @@ class Choice : public rng::RNGBase<Backend, Choice<Backend>, false> {
           BaseImpl::template RunImplTyped<T, Impl<T, true, true>>(ws);
         }
       ), (  // NOLINT
-        DALI_FAIL(make_string("Data type ", dtype_, " is not supported for 1D inputs. "
-                              "Supported types are: ", ListTypeNames<DALI_CHOICE_1D_TYPES>(), "."));
+        DALI_FAIL("Data type ", dtype_, " is not supported for 1D inputs. "
+                  "Supported types are: ", ListTypeNames<DALI_CHOICE_1D_TYPES>(), ".");
       ));  // NOLINT
     } else {
-      ElementCopy(ws);
+      DALI_FAIL("The operator only supports sampling of 0D elements, got: ", input.sample_dim(),
+                "D input.");
     }
     if (!input.GetLayout().empty()) {
       if (input.sample_dim() == output.sample_dim()) {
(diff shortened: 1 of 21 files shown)


klecki:
!build

klecki:
!build

klecki:
!build

dali/operators/random/rng_base.h
```diff
@@ -89,7 +90,13 @@ class OperatorWithRng : public Operator<Backend>{
   OperatorWithRngFields<Backend> backend_data_;
 };
 
-
+/**
+ * @brief CRTP class for implementing random number and noise generators.
+ *
+ * @tparam IsNoiseGen - noise generators by default copy type from input and compute the output
+ * value based on the input value at given coordinate.
+ * compute
```
szkarpinski, line 98:
```suggestion
```

dali/operators/random/choice_cpu.cc
```diff
@@ -0,0 +1,46 @@
...
+
+#include "dali/operators/random/choice.h"
+#include "dali/operators/random/rng_base_cpu.h"
+#include "dali/pipeline/data/backend.h"
+#include "dali/pipeline/operator/arg_helper.h"
+
+namespace dali {
+
+DALI_SCHEMA(random__Choice)
+    .DocStr(R"code(Generates a random sample from a given 1D array.
```
szkarpinski, line 23: Not only: _if a scalar is provided, values from [0, a) are sampled_

dali/operators/random/choice.h
```diff
@@ -0,0 +1,277 @@
...
+#define DALI_CHOICE_1D_TYPES                                                                      \
+  bool, uint8_t, uint16_t, uint32_t, uint64_t, int8_t, int16_t, int32_t, int64_t, float16, float, \
+      double
+
+namespace dali {
+
+/**
+ * @brief Variant for 1D input - we sample index and take the corresponding value from input pointer
+ */
+template <typename T, bool uniform, bool indirect = true>
```
szkarpinski, line 39: Nitpick: It's wasn't obvious for me what `indirect` does. Maybe a tparam doc would help

dali/operators/random/choice.h
```diff
@@ -0,0 +1,277 @@
...
+    auto choice_idx = dist_(st);
+    return elements_[choice_idx];
+  }
+
+  const T *elements_ = nullptr;
+  DistType dist_ = {};
+};
+
+/**
+ * @brief Variant for generating indices. In case of 0D input, we treat those indices as the
```
szkarpinski, line 74: Nitpick: _Variant for generating indices_ is misleading for me. In this specialization we *don't* generate indices, but we generate actual values directly.

dali/operators/random/choice.h
```diff
@@ -0,0 +1,277 @@
...
+  DALIDataType DefaultDataType(const OpSpec &spec, const Workspace &ws) const {
+    if (ws.Input<CPUBackend>(0).sample_dim() == 0) {
+      return DALI_INT32;
+    } else {
+      return ws.Input<CPUBackend>(0).type();
+    }
+  }
+
+  /**
+   * @brief We only allow sampling of the input elements or generating integral types if the
```
szkarpinski, line 172: Nitpick: I'd rather fill @ brief with a brief description of what this function generally does, and add that remark below

dali/operators/random/choice_cpu.cc
```diff
@@ -0,0 +1,46 @@
...
+
+The probability of selecting a sample from the input is determined by the probability distribution
+specified in ``p`` argument.
```
szkarpinski, line 26: Providing `p` is optional, maybe it's worth mentioning here

dali/test/python/operator_2/test_random_choice.py
```diff
@@ -0,0 +1,201 @@
...
+        (
+            (2,),
+            {"p": 0.25},
+            "Sum of probabilities must be 1.0, but got 0.5 for sample: 0.",
+        ),
+        (
+            (-5,),
+            {},
+            "Expected positive number of elements for sampling, got: -5 for sample: 0.",
+        ),
```
szkarpinski, line 164: How about adding a test with 0 elements? It's less obvious than -5

dali/test/python/operator_2/test_random_choice.py
```diff
@@ -0,0 +1,201 @@
+# Copyright (c) 2024, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
szkarpinski, line 1: The following tests are worth adding in my opinion:
- Some probabilities equal to 0
- Only 1 element to choose from
- Randomly choosing a number for some huuge range, like [0, 2^40). This is a bit of misuse, but maybe we should check that DALI won't explode in such case

dali/operators/random/choice.h
klecki: done

dali/operators/random/choice.h
klecki: done. I was thinking about the comment specified in the base class.

dali/operators/random/choice_cpu.cc
klecki: I described the behavior in the argument.

dali/operators/random/choice_cpu.cc
klecki: I adjusted the doc for the first input, so this now holds for all inputs.

dali/operators/random/rng_base.h
klecki: done

dali/test/python/operator_2/test_random_choice.py
klecki: Done. It explodes only a bit.

dali/test/python/operator_2/test_random_choice.py
klecki: done

dali/operators/random/choice.h
```diff
@@ -0,0 +1,275 @@
...
+      double
+
+namespace dali {
+
+/**
+ * @brief Variant for 1D input - we sample index and take the corresponding value from input pointer
+ * @tparam indirect - whether the distribution result is used as the index into the `elements`
+ * array (true) or returned directly (false).
+ */
+template <typename T, bool uniform, bool indirect = true>
```
stiepan, line 41: very much a nitpick: `direct` would read easier than `indirect`, because of the double negation `direct=false` vs `indirect=false`.

dali/test/python/operator_2/test_random_choice.py
stiepan, line 1: How about veryfing the layout of 1d is propagated properly?

stiepan, line 1: Do I read it right that if we pass the probabilities, for the 0d case, the passed scalar value is meaningless? I can see numpy validates it nontheless to match the number of ps.

stiepan (commented):
It's going to be super useful addition to DALI.

Leaving a couple of mostly nitpicks, overall looks good to me.

dali/operators/random/choice.h
```diff
@@ -0,0 +1,277 @@
...
+    } else {
+      assert(false);  // Should not be called
+    }
+  }
+
+  DALI_HOST_DEV ChoiceSampleDist(int64_t element_count) {
+    if constexpr (uniform) {
+      dist_ = DistType(0, element_count - 1);
+    } else {
+      assert(false);  // Should not be called
```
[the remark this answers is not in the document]
klecki: I wanted to finally get away from all the enable_if_t fun, but whatever.

dali/operators/random/choice.h
klecki: As @szalpal reminded us this week, I will gladly not do this, due to the risk of not flipping a boolean and spending 2 hours debugging it.

dali/test/python/operator_2/test_random_choice.py
klecki: If you pass mismatched `a` != `len(p)` you get an error, as I put `a` in the required shape in the `p_dist_.Acquire`

klecki: done

klecki:
!build

szkarpinski (approved):

stiepan (approved):

Issue #5404, opened 2024-04-02
COCO Reader pixelwise_masks Emtpy Output


5had3z:
### Version

1.35.0

### Describe the bug.

When enabling pixelwise_masks (with intention to use for panoptic sementation), the output image is only zeros (with sometimes a stray non-zero pixel, which is also concerning).

I've tried this with both the normal coco dataset, and with examples in DALI_extra, both have the same behaviour.

I'm also not sure what the format of pixelwise_masks is meant to be. The index of the segment?

![output](https://github.com/NVIDIA/DALI/assets/59598353/ed63557d-de41-48cf-a90d-5727c676b254)

### Minimum reproducible example

```shell
from nvidia.dali.pipeline import Pipeline
import nvidia.dali.fn as fn
from nvidia.dali.types import DALIImageType
from matplotlib import pyplot as plt

file_root = REPLACEME
annotations_file = REPLACEME

pipe = Pipeline(batch_size=1, num_threads=4, device_id=0)
with pipe:
    jpegs, bboxes, labels, masks = fn.readers.coco(
        file_root=file_root,
        annotations_file=annotations_file,
        pixelwise_masks=True,
        ratio=True,
    )
    images = fn.decoders.image(jpegs, device="mixed", output_type=DALIImageType.RGB)
    pipe.set_outputs(images, bboxes, labels, masks)
pipe.build()

pipe_out = pipe.run()

images_cpu = pipe_out[0].as_cpu()
bboxes_cpu = pipe_out[1]
labels_cpu = pipe_out[2]
masks = pipe_out[3]

print(labels_cpu)
print(bboxes_cpu)
plt.subplot(121)
plt.imshow(images_cpu.as_array()[0, ...])
plt.subplot(122)
plt.imshow(masks.as_array()[0, ..., 0])
plt.show()
```

### Other/Misc.

COCO Reader [tutorial](https://docs.nvidia.com/deeplearning/dali/user-guide/docs/examples/general/data_loading/coco_reader.html) also needs to be  updated to
`from nvidia.dali.types import DALIImageType` instead of `import nvidia.dali.types as types` with its usage in `fn.decoders.image`.

### Check for duplicates

- [X] I have searched the [open bugs/issues](https://github.com/NVIDIA/DALI/issues) and have found no duplicates for this bug report

5had3z:
[...]

PR #5407, opened 2024-04-02
Fix pixelwise_masks support when the ratio is on in the coco reader

JanuszL:
- when the ratio option is on the pixelwise_masks calculation
  uses relative coordinates instead of the real one which leads
  to creating a 1-pixel map in the corner of the tensor. This PR
  fixes the math.
- adds relevant test

## Category:

## Description:
- when the ratio option is on the pixelwise_masks calculation
  uses relative coordinates instead of the real one which leads
  to creating a 1-pixel map in the corner of the tensor. This PR
  fixes the math.
- adds relevant test
- relates to https://github.com/NVIDIA/DALI/issues/5404

- coco reader

### Tests:

- [ ] Existing tests apply
- [x] New tests added
  - [x] Python tests
    - test_coco.test_coco_pix_mask_ratio
  - [ ] GTests
  - [ ] Benchmark
  - [ ] Other
- [ ] N/A

### Documentation
- [ ] Documentation updated
  - [ ] Docstring
  - [ ] Doxygen
  - [ ] RST
  - [ ] Jupyter
  - [ ] Other
- [x] N/A

#### Requirements
- [x] N/A

**REQ IDs**: N/A

**JIRA TASK**: N/A

### DIFF
--- a/dali/operators/reader/coco_reader_op.cc
+++ b/dali/operators/reader/coco_reader_op.cc
@@ -260,6 +260,7 @@ void COCOReader::RunImpl(SampleWorkspace &ws) {
 }
 
 void COCOReader::PixelwiseMasks(int image_idx, int* mask) {
+  bool ratio = spec_.GetArgument<bool>("ratio");
   auto &loader_impl = LoaderImpl();
   auto pol = loader_impl.polygons(image_idx);
   auto ver = loader_impl.vertices(image_idx);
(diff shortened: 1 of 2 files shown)


5had3z:
I think it would be more efficient to not have normalized points in the first place (since mask and points are exclusive anyway) and change condition [here](https://github.com/NVIDIA/DALI/blob/main/dali/operators/reader/loader/coco_loader.cc#L585) to be something along the lines of `if (ratio && !spec_.GetArgument<bool>("pixelwise_masks"))`

banasraf (approved):

PR #5422, opened 2024-04-10
Improve support for DALI enum types

klecki:
## Description:
Support for DALI enum types (DALIDataType, DALIImageType, DALIInterpType) is added to Constant, Cast, Choice and Copy operators.

Thanks to Constant op support, followin syntax will work now:
```
fn.random.choice([types.DALIInterpType.INTERP_LINEAR, types.DALIInterpType.INTERP_NN])
```
allowing convenient selection of enum parameters.
Constant promotions insert constant nodes automatically.
When types.Constant is used, for true scalars the ScalarConstant placeholder is typically created.
For enum values we always use proper constant node as enums are not usable in artihmetic operations
and this is the main target of ScalarConstant.

Casting support is allowed only between non-fp types and enums.

Explicit error is raised when the enums are used with buffer protcol (conversion to numpy, printing etc) - Python expectes pointer-to-object representation there, while we return them as C-style enums with numeric value under the hood.
As an alternative we can allow to just access the underlying data but I have chosen a bit more restricive approach for now.

The values indicating the enum types were exposed in type hints for DALIDataType.

`__rerp__` of Tensor and TensorList is adjusted to not access the underlying data (it is shown only for `__str__`).
It invokes buffer protocol to show the summary of data contents and in this case it is not possible. As the repr
is used in Python-produced errors, the actual error might have been obscured by trying to access data.

Enum support, type hints, several operators.

Are there any additional operators that we might want to cover?
Equal/not equal comparisons between enum values?

### Tests:

- [ ] Existing tests apply
- [x] New tests added
  - [x] Python tests
  - [ ] GTests
  - [ ] Benchmark
  - [ ] Other
- [ ] N/A

### Documentation
- [x] Documentation updated
  - [ ] Docstring
  - [ ] Doxygen
  - [ ] RST
  - [ ] Jupyter
  - [ ] Other
- [ ] N/A

#### Requirements
- [ ] N/A

**REQ IDs**: N/A

**JIRA TASK**: N/A


dali/kernels/common/cast_gpu.cu
```diff
@@ -19,6 +19,7 @@
 #include "dali/kernels/common/utils.h"
 #include "dali/kernels/kernel.h"
 #include "dali/kernels/dynamic_scratchpad.h"
+#include "dali/pipeline/data/types.h"
```
klecki, line 22: I need to pull the pipeline/data/types.h into kernel implementation here, to be able to instantiate the kernel for the DALIDataType etc.

klecki:
!build

klecki:
!build

banasraf (approved):

klecki:
!build

PR #5447, opened 2024-04-29
Make shared IterationData a first class member of Workspace.

mzient:
Make operator traces thread-safe.

## Category:

## Description:
This change does 2 things:
- it generalizes the mechanism of iteration data beyond operator traces
- it makes operator traces thread-safe

Executor, Workspace, operator traces

N/A

### Tests:
- C API operator trace tests 
- Executor tests

- [X] Existing tests apply
- [ ] New tests added
  - [ ] Python tests
  - [ ] GTests
  - [ ] Benchmark
  - [ ] Other
- [ ] N/A

### Documentation
- [X] Documentation updated
  - [ ] Docstring
  - [X] Doxygen
  - [ ] RST
  - [ ] Jupyter
  - [ ] Other
- [ ] N/A

#### Requirements
- [X] N/A

**REQ IDs**: N/A

**JIRA TASK**: N/A

### DIFF
--- a/dali/pipeline/executor/executor_impl.cc
+++ b/dali/pipeline/executor/executor_impl.cc
@@ -157,6 +157,7 @@ void Executor<WorkspacePolicy, QueuePolicy>::SyncDevice() {
 
 template <typename WorkspacePolicy, typename QueuePolicy>
 void Executor<WorkspacePolicy, QueuePolicy>::RunCPUImpl(size_t iteration_id) {
+  GetCurrentIterationData(iteration_id).iter_data->iteration_index = iteration_id;
   PreRun();
   const char placement_error[] =
       "Cannot run a pipeline with Mixed/GPU ops in CPU-only mode. Please provide "
@@ -408,7 +409,7 @@ void Executor<WorkspacePolicy, QueuePolicy>::RunHelper(OpNode &op_node, Workspac
   const auto &schema = spec.GetSchema();
   SmallVector<int, 16> empty_layout_in_idxs;
 
-  ws.InjectOperatorTraces(GetCurrentIterationData(iteration_id).operator_traces);
+  ws.InjectIterationData(GetCurrentIterationData(iteration_id).iter_data);
   ws.ClearOperatorTraces();
 
   auto ws_order = ws.has_stream() ? AccessOrder(ws.stream()) : AccessOrder::host();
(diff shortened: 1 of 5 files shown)


dali/pipeline/executor/executor_impl.cc
```diff
@@ -157,6 +157,7 @@ void Executor<WorkspacePolicy, QueuePolicy>::SyncDevice() {
 
 template <typename WorkspacePolicy, typename QueuePolicy>
 void Executor<WorkspacePolicy, QueuePolicy>::RunCPUImpl(size_t iteration_id) {
+  GetCurrentIterationData(iteration_id).iter_data->iteration_index = iteration_id;
```
JanuszL, line 160: So it was not use by the CPU stage?

dali/pipeline/workspace/workspace.h
```diff
@@ -603,16 +606,9 @@ class WorkspaceBase : public ArgumentWorkspace {
...
-
-  /**
-   * Get the operator trace map for all operators in the pipeline.
-   *
-   * @see operator_trace_map_t
-   */
-  DLL_PUBLIC const auto &GetOperatorTraceMap() const {
+  DLL_PUBLIC auto &GetOperatorTraces(const std::string &operator_name) const {
+    if (!operator_traces_)
+      operator_traces_ = &iter_data_->operator_traces.Get(operator_name);
```
JanuszL, line 611: So it is to avoid expensive locking?

dali/pipeline/workspace/workspace.h
```diff
@@ -565,8 +565,7 @@ class WorkspaceBase : public ArgumentWorkspace {
    * @see operator_trace_map_t
    */
   DLL_PUBLIC void SetOperatorTrace(const std::string &trace_key, std::string trace_value) {
-    (*operator_traces_)[GetOperatorInstanceName()].insert_or_assign(
-            trace_key, std::move(trace_value));
+    GetOperatorTraces().insert_or_assign(trace_key, std::move(trace_value));
```
JanuszL, line 568: Should we mention here that this function is not thread safe as mentioned `OperatorTraces::Get`?
(other functions touching operator_trace_map_t in this file as well)

dali/pipeline/executor/executor_impl.cc
JanuszL: OK, it is a new thing.

dali/pipeline/workspace/workspace.h
mzient: I don't think we assume that _anything_ in Workspace is thread-safe. Mentioning here that it isn't might create an illusion of thread-safety elsewhere.

dali/pipeline/workspace/workspace.h
mzient: Yes, that's the point.

dali/pipeline/workspace/workspace.h
JanuszL: Understood.

jantonguirao (approved):

PR #5455, opened 2024-05-06
Remove the use of (soon deprecated) aligned_storage.

mzient:
## Category:
**Other** C++ standard update

## Description:
According to C++ 17 standard, type-punning pointers to `std::aligned_storage_t` is UB and in C++23 `std::aligned_*` is deprecated. The recommended way is to use an appropriately aligned array of bytes.

aux_alloc
small_vector
"leaker" in default_resources (that's a hack anyway)
some tests

alignment syntax?

### Tests:
small_vector_test and everything that uses SmallVector
allocator tests indirectly test aux_alloc
- [X] Existing tests apply
- [ ] New tests added
  - [ ] Python tests
  - [X] GTests
  - [ ] Benchmark
  - [ ] Other
- [ ] N/A

### Documentation
- [ ] Documentation updated
  - [ ] Docstring
  - [ ] Doxygen
  - [ ] RST
  - [ ] Jupyter
  - [ ] Other
- [X] N/A

#### Requirements
- [x] N/A

**REQ IDs**: N/A

**JIRA TASK**: DALI-3036

### DIFF
--- a/dali/kernels/test/scratch_test.cc
+++ b/dali/kernels/test/scratch_test.cc
@@ -1,4 +1,4 @@
-// Copyright (c) 2018 - 2021, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+// Copyright (c) 2018-2021, 2024, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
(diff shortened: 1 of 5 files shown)


dali/kernels/test/scratch_test.cc
```diff
@@ -81,31 +81,31 @@ TEST(Scratch, Estimator) {
 }
 
 TEST(Scratch, BumpAllocator) {
-  const size_t size = 1024;
-  std::aligned_storage<size, 64>::type storage;
-  BumpAllocator allocator(reinterpret_cast<char*>(&storage), size);
+  const size_t kSize = 1024;
+  alignas(64) char storage[kSize];
+  BumpAllocator allocator(storage, kSize);
```
mzient, line 86: Here I use `char` instead of `std::byte` because they're equally good wrt aliasing and the BumpAllocator API takes `char *`.

